PR #24, opened 2026-02-26
feat: session-scoped signatures — remove signature from persistent storage

ghermet:
## Summary

- **Signatures are no longer persisted** to IndexedDB/storage. The EIP-712 wallet signature is held in-memory only, scoped to the page session.
- **New public API** on `CredentialsManager`: `lock()`, `unlock()`, `isUnlocked()` for apps to control the session lifecycle.
- **Silent migration** for existing stored credentials that contain a `signature` field — used once on load, then re-persisted without it.
- **New events**: `credentials:locked` and `credentials:unlocked`.

## Motivation

The EIP-712 signature was stored in plaintext in IndexedDB. Since it also serves as the PBKDF2 key material for encrypting the FHE private key, anyone with devtools/XSS access could derive the AES key and decrypt all encrypted balances. Removing it from persistent storage eliminates this attack surface entirely.

## Trade-offs

- One wallet popup per session (page load) to re-sign. Standard UX for dApps.
- Each tab has its own session — acceptable since each tab already has its own SDK instance.

## Test plan

- [x] All 744 existing tests pass
- [x] New tests for: session re-sign flow, new-format storage loading, `lock()`/`unlock()`/`isUnlocked()`, `clear()` clears session, `isExpired()` without session signature
- [x] Legacy migration test (stored data with signature field)
- [x] TypeScript type check passes
- [x] API reports unchanged (no public type changes)


ghermet:
Replaced by a cleaner PR with only session-scoped signature commits cherry-picked onto main.

PR #25, opened 2026-02-26
feat: session-scoped signatures & SDK API improvements [SDK-6]

ghermet:
## Summary

- **Session-scoped signatures**: wallet signatures are no longer persisted to IndexedDB — held in-memory (or pluggable `sessionStorage`) only, scoped to the page session
- **Renamed credential API**: `get`/`getAll` → `allow(...addresses)`, `lock` → `revoke()`, `unlock` → `allow`, `isUnlocked` → `isAllowed`
- **Pluggable session storage**: new `sessionStorage` config on `CredentialsManager` and `ReadonlyToken`, enabling Chrome extensions to use `chrome.storage.session` for cross-context signature sharing
- **ChromeSessionStorage**: ready-made `GenericStorage` implementation for MV3 web extensions (`chrome.storage.session`)
- **Renamed `GenericStringStorage` → `GenericStorage<T>`**: generic type parameter, methods renamed from `getItem`/`setItem`/`removeItem` to `get`/`set`/`delete`, no longer forces JSON serialization; type parameter moved from interface level to individual methods
- **Signer lifecycle subscribe**: `ViemSigner`, `EthersSigner`, and `WagmiSigner` now expose a `.subscribe()` method (EIP-1193 `accountsChanged`/`disconnect`) that auto-revokes session credentials on wallet disconnect or account switch
- **Removed `window.ethereum` auto-detection**: signers no longer reach for the global provider — the caller must supply the EIP-1193 provider explicitly
- **New React hooks**: `useTokenAllow`, `useTokenRevoke`, `useIsTokenAllowed` (replaces `useAuthorizeAll`)
- **New events**: `credentials:revoked` and `credentials:allowed` (replaces `credentials:locked`/`credentials:unlocked`)
- **Chain-scoped credential keys**: credential storage keys now include `chainId`, preventing cross-chain credential leakage
- **CI**: API report check split into its own parallel workflow; aligned with vitest/playwright
- **Docs**: encryption architecture document with Mermaid diagrams, EIP-1193 signer subscribe design docs, updated GitBook guides and READMEs

## Motivation

The EIP-712 signature was stored in plaintext in IndexedDB. Since it also serves as the PBKDF2 key material for encrypting the FHE private key, anyone with devtools/XSS access could derive the AES key and decrypt all encrypted balances. Removing it from persistent storage eliminates this attack surface entirely.

The API rename (`allow`/`revoke`/`isAllowed`) aligns the SDK vocabulary with the domain language used in confidential token contracts and makes the rest-parameter signature more ergonomic than the previous array-based API.

The signer `.subscribe()` addition ensures credentials are automatically revoked when the user switches accounts or disconnects their wallet, closing a session-hygiene gap.

## Breaking changes

| Before | After |
|--------|-------|
| `credentials.get(addr)` | `credentials.allow(addr)` |
| `credentials.getAll([a, b])` | `credentials.allow(a, b)` |
| `credentials.lock()` | `credentials.revoke()` |
| `credentials.unlock(addrs?)` | `credentials.allow(...addrs)` |
| `credentials.isUnlocked()` | `credentials.isAllowed()` |
| `GenericStringStorage` | `GenericStorage<T>` (type param on methods) |
| `.getItem()` / `.setItem()` / `.removeItem()` | `.get()` / `.set()` / `.delete()` |
| `useAuthorizeAll` | `useTokenAllow` |
| `credentials:locked` / `credentials:unlocked` | `credentials:revoked` / `credentials:allowed` |
| `ReadonlyToken.authorizeAll(tokens)` | `ReadonlyToken.allow(...tokens)` |
| `new ViemSigner(walletClient)` auto-detects provider | Must pass EIP-1193 provider explicitly |
| `EIP1193Provider` exported from public API | Removed — use `import { EIP1193Provider } from 'viem'` or equivalent |

## Trade-offs

- One wallet popup per session (page load) to re-sign — standard UX for dApps
- Each tab has its own session unless a shared `sessionStorage` backend (e.g. `ChromeSessionStorage`) is provided
- Removing `window.ethereum` auto-detection requires callers to be explicit, but avoids silent provider-mismatch bugs

## Test plan

- [x] All existing tests pass
- [x] New tests for: session re-sign flow, new-format storage loading, `allow()`/`revoke()`/`isAllowed()`, `clear()` clears session, `isExpired()` without session signature
- [x] Legacy migration test (stored data with signature field)
- [x] Chain-scoped credential key tests
- [x] Signer subscribe/auto-revoke tests
- [x] TypeScript type check passes
- [x] API reports regenerated


ankurdotb:
Thanks for this change, this is moving in the right direction from a security standpoint.

I worked through the edge cases I had in mind and broadly I think this works. A few practical edge cases I still see:

1. **Dynamic token sets may still trigger re-signing (probably not solvable?)**: As wallets discover new tokens over time say from the registry contract, adding a token not included in the original signed contract set can require a fresh signature. I think this may be protocol-bound (EIP-712 payload contract scope), so this may not be fully solvable in SDK alone.
2. **Wallet lock/unlock lifecycle needs explicit app wiring**: Session signatures are revoked only when session storage is cleared or `revoke()` is called. If integrators don’t wire wallet lock/disconnect/account-change events to `revoke()`, users may get inconsistent UX/security expectations. Do you think this is fine, or should we give any explicit guidance?
3. **Chain-switch resilience**: Consider what testing needs to be done if a wallet switches between mainnet and sepolia. The EIP-712 is scoped to the chainId, but is there anything to account for in this sessionStorage?
4. **Event contract mismatch: `credentials:allowed` appears declared but not emitted**. PR summary says new `credentials:allowed`; event type exists, but manager code emits `loading/cached/created/revoked`, not `allowed`.
5. **Multi-context behavior is easy to misconfigure**: Default in-memory session storage remains per-context/tab. In extension environments, if integrators don’t pass shared session storage (`chrome.storage.session`), users may still see repeated prompts after context restarts. Codex seems to think this can be a problem in scenarios like browser wallet extensions if left on default in-memory session storage. In MV3, popup/background/content-script contexts are separate, and service workers restart often.

ghermet:
> Thanks for this change, this is moving in the right direction from a security standpoint.
> 
> I worked through the edge cases I had in mind and broadly I think this works. A few practical edge cases I still see:
> 
> 1. **Dynamic token sets may still trigger re-signing (probably not solvable?)**: As wallets discover new tokens over time say from the registry contract, adding a token not included in the original signed contract set can require a fresh signature. I think this may be protocol-bound (EIP-712 payload contract scope), so this may not be fully solvable in SDK alone.
> 2. **Wallet lock/unlock lifecycle needs explicit app wiring**: Session signatures are revoked only when session storage is cleared or `revoke()` is called. If integrators don’t wire wallet lock/disconnect/account-change events to `revoke()`, users may get inconsistent UX/security expectations. Do you think this is fine, or should we give any explicit guidance?
> 3. **Chain-switch resilience**: Consider what testing needs to be done if a wallet switches between mainnet and sepolia. The EIP-712 is scoped to the chainId, but is there anything to account for in this sessionStorage?
> 4. **Event contract mismatch: `credentials:allowed` appears declared but not emitted**. PR summary says new `credentials:allowed`; event type exists, but manager code emits `loading/cached/created/revoked`, not `allowed`.
> 5. **Multi-context behavior is easy to misconfigure**: Default in-memory session storage remains per-context/tab. In extension environments, if integrators don’t pass shared session storage (`chrome.storage.session`), users may still see repeated prompts after context restarts. Codex seems to think this can be a problem in scenarios like browser wallet extensions if left on default in-memory session storage. In MV3, popup/background/content-script contexts are separate, and service workers restart often.

  ## 1. Dynamic token sets triggering re-signing                                                      
                                                                                                   
  You're right — this is protocol-bound. The EIP-712 message includes contractAddresses and the    
  relayer verifies the decrypt target is in that signed set. When allow() is called with a new     
  contract address not in the existing signed set, #isValid() fails the coverage check and         
  #reSign() fires with the expanded set.                                                           
                                                                                                   
  This is by design and not really solvable without changing the protocol. The signature must cover
   the contracts being decrypted. One mitigation already in place: allow(...contractAddresses)
  accepts multiple addresses, so integrators can batch-allow upfront if they know their token set.
  The re-sign only happens when a genuinely new contract appears.


  ## 2. Wallet lock/unlock lifecycle — explicit guidance needed?

  revoke() only deletes the session signature from sessionStorage — the persistent keypair in
  IndexedDB survives. So if integrators don't call revoke() on disconnect/account-change, the
  session signature stays valid until expiry.

  I'd recommend explicit guidance in the docs. A short "Integration Checklist" section showing:
  wallet.on("disconnect", () => token.credentials.revoke());
  wallet.on("accountsChanged", () => token.credentials.revoke());

  This is standard practice for wallet SDKs. Not wiring it isn't a security hole per se (the
  signature is time-bounded and chain-scoped), but it creates UX confusion if a user switches
  accounts and old credentials are still "allowed."

  ## 3. Chain-switch resilience

  The relayer already handles this well — relayer-web.ts:103-107 detects chainId changes, tears
  down the old worker, and creates a fresh one. The EIP-712 domain includes chainId, so signatures
  are chain-scoped.

  However, the session storage key (#storeKey()) should be checked — if it doesn't include chainId,
   credentials signed for chain A could be returned when queried on chain B. Let me check.

⏺ Searched for 1 pattern, read 1 file (ctrl+o to expand)     
                                                 
⏺ Confirmed: #storeKey() is address-only, no chainId. If a user switches from mainnet to sepolia
  with the same wallet, the same storage key returns credentials signed for the old chain. The     
  EIP-712 signature would be invalid on the new chain, but the SDK would try to use it and likely  
  fail at the relayer.              
                                                                                                   
  This is a real issue. The store key should incorporate chainId:                                  
  async #storeKey(): Promise<string> {                                                             
    const address = (await this.#signer.getAddress()).toLowerCase();                               
    const chainId = await this.#config.getChainId();        
    const hash = await crypto.subtle.digest("SHA-256", new
  TextEncoder().encode(`${address}:${chainId}`));
    // ...
  }

  ## 4. Event contract mismatch — credentials:allowed

  Confirmed: CredentialsAllowed is declared in sdk-events.ts:14 but never emitted. The allow()
  method emits CredentialsLoading at entry, then
  CredentialsCached/CredentialsCreated/CredentialsExpired depending on outcome. There's no terminal
   "allowed" event.

  Options:
  - Remove it if it was speculative and the existing events cover the lifecycle
  - Emit it at the end of allow() as a "credentials are ready" signal (could be useful for UI state
   machines that want a single "done" event regardless of cache/create path)

  ## 5. Multi-context misconfiguration (MV3 extensions)

  You're right. Default memoryStorage is per-JS-context. In MV3:
  - Service worker restarts kill the memory store
  - Popup, background, and content scripts each get separate instances

  ChromeSessionStorage solves this (shared via chrome.storage.session), and the singleton
  chromeSessionStorage export makes it easy. But integrators have to opt in.

  Recommendation: Add a detection helper or a loud console warning:
  if (typeof chrome !== "undefined" && chrome.runtime?.id && storage instanceof MemoryStorage) {
    console.warn("[zama-sdk] Detected extension context with in-memory storage. Consider using
  chromeSessionStorage for cross-context credential sharing.");
  }

  This is low-effort and saves integrators from a confusing debugging session.

ankurdotb:
Potential remaining behavioral gap: auto-revoke may clear the wrong session key on account change/disconnect.

We now auto-wire signer lifecycle events to `sdk.revokeSession()` (`onDisconnect` / `onAccountChange`), which is a good direction.  
But `revokeSession()` derives the key to delete from `signer.getAddress()` + `chainId` *at revoke time*.

Why this matters:
- On **account change**, `getAddress()` is typically already the **new** account.
- So revoke deletes `newAddress:chainId` (often empty), while the previous account’s key can remain in `sessionStorage`.
- If the user switches back to the previous account later, that old session signature may still be considered allowed (no fresh re-sign).
- On **disconnect**, some signer implementations may fail `getAddress()`, which can make revoke best-effort fail and leave the key untouched.

Repro scenario:
1. Account A calls `allow()` (session key `A:chain` stored).  
2. User switches to account B (auto-revoke runs, but deletes `B:chain`).  
3. User switches back to A (`A:chain` may still be present).

Suggested fix:
- Revoke by explicit identity from lifecycle callback (prefer passing previous address, and chain if needed), instead of reading current signer state inside revoke.
- Alternatively track last-known `{address, chainId}` in SDK and revoke that on lifecycle transitions.
- Add a regression test for `A -> B -> A` to assert old account session is cleared on account change.

ankurdotb (changes requested):
I left a comment separately. The push I made was to align the CI triggers for the split out workflow for API Report (include prerelease, similar triggers as vitest and playwright).

ankurdotb:
[AUTOMATED]

One remaining behavioral gap to fix before merge:

`eip1193Subscribe` does not clear `currentAddress` when `accountsChanged([])` happens (disconnect/lock path), which can break auto-revoke over repeated lock/unlock cycles.

Why this matters:
- In `eip1193Subscribe`, `accountsChanged([])` calls `onDisconnect()` and returns, but leaves `currentAddress` unchanged.
- In `ZamaSDK`, `onDisconnect` revokes once and then clears tracked identity (`#lastAddress/#lastChainId`).
- If the wallet unlocks back to the same account, `onAccountChange` will not fire (same address), so tracked identity is not restored.
- On the next disconnect, `#revokeByTrackedIdentity()` has no identity and returns early, leaving the active session signature in `sessionStorage`.

Practical impact for MetaMask-like wallets:
- First lock/unlock cycle behaves as expected.
- Subsequent lock cycles can stop revoking the session signature if account remains unchanged.

Suggested fix:
- In `eip1193Subscribe`, set `currentAddress = undefined` before calling `onDisconnect()` when `accounts.length === 0`.
- Add a regression test for repeated lock/unlock with same account:
  1) allow A
  2) disconnect
  3) reconnect A
  4) disconnect again
  5) assert session key revoked both times.

docs/gitbook/src/guides/sdk/configuration.md
```diff
@@ -269,6 +287,109 @@ const sdk = new ZamaSDK({
 });
 ```
 
+## Session management
+
+FHE credentials are stored encrypted at rest. The wallet signature that unlocks them lives in `sessionStorage` (in-memory by default). This means:
```
ankurdotb: What are the "FHE credentials"...is this the ephemeral key used to encrypt transport? How are they encrypted?

ghermet:
Good catch! This is already fixed in f0a3797 — `currentAddress` is now cleared to `undefined` before calling `onDisconnect()` when `accounts.length === 0`, which is exactly the suggested fix.

Regression test covering repeated lock/unlock cycles with the same account is also in place (`eip1193-subscribe.test.ts`), asserting that `onAccountChange` fires on every reconnect after disconnect.

docs/gitbook/src/guides/sdk/configuration.md
```diff
@@ -269,6 +287,109 @@ const sdk = new ZamaSDK({
...
+await sdk.revoke("0xTokenA", "0xTokenB");
+
+// Or revoke the session without specifying addresses
+await sdk.revokeSession();
+// Next decrypt will require a fresh wallet signature
+```
+
+### Wallet lifecycle integration
+
+When the wallet disconnects or the user switches accounts, the session signature should be cleared so stale credentials don't persist.
```
ankurdotb: Disconnect, yes I agree. Switches accounts, I'm less sure of.

If I switch on the same address from mainnet to Sepolia and back, what happens in that scenario?

docs/gitbook/src/guides/sdk/configuration.md
ghermet: **FHE credentials** are an asymmetric keypair (public key + private key) generated by the relayer for a specific wallet address and chain. They are used to re-encrypt FHE ciphertexts so that only this user's keypair can decrypt them (the relayer uses the public key to re-encrypt, and the SDK uses the private key locally to decrypt the result).

**How they are encrypted at rest:**
- The **private key** is AES-GCM encrypted before being written to persistent storage (`IndexedDB` by default)
- The encryption key is derived from the wallet's EIP-712 signature via PBKDF2 (600k iterations of SHA-256, salted with the wallet address)
- So the wallet signature is effectively the "master key" — without it, the private key cannot be decrypted even if someone reads `IndexedDB`
- The **public key** and **contract addresses** are stored in plaintext (they are not secret)

**The session signature** (stored in ephemeral `sessionStorage`) is the EIP-712 wallet signature that doubles as the decryption key seed. On page reload, the user re-signs once to unlock the credentials for that session.

I'll clarify the wording in the docs — "FHE credentials" is vague. Will update to explicitly describe what's stored and how.

docs/gitbook/src/guides/sdk/configuration.md
ghermet: Good question. Chain switching is actually handled well — the storage key is `SHA-256(address:chainId)`, so credentials for mainnet and Sepolia are stored independently even for the same address.

**Scenario: same address, mainnet → Sepolia → mainnet:**

1. User is on mainnet, calls `allow()` → session signature stored under key `SHA-256("0xAddr:1")`
2. User switches to Sepolia → auto-revoke clears the mainnet session key (`SHA-256("0xAddr:1")`)
3. User calls `allow()` on Sepolia → new session signature stored under `SHA-256("0xAddr:11155111")`
4. User switches back to mainnet → auto-revoke clears Sepolia session key
5. User must re-sign for mainnet (session was revoked in step 2)

The encrypted keypair in persistent storage (`IndexedDB`) survives across chain switches — only the session signature is revoked. So step 5 is a re-sign (one wallet popup), not a full keypair regeneration, assuming the mainnet credentials haven't expired.

On your broader point about account switches: the reason we auto-revoke on account change is that the EIP-712 signature is address-scoped. If the user switches from address A to address B, A's session signature should not remain valid — it would be confusing (and a minor security concern) if B's session could decrypt A's data. The `#lastAddress`/`#lastChainId` tracking ensures we revoke the **previous** identity, not the new one.

That said, I agree it's a different concern than disconnect. Happy to split the docs to explain the two cases separately if that would help.

docs/gitbook/src/guides/sdk/configuration.md
ghermet: Updated all docs to use **"decrypt keys"** instead of "FHE credentials" — much clearer for integrators. First occurrence explains what they are: "a keypair used to decrypt confidential balances."

ankurdotb: I see, that's what I thought it might refer to. Can we maybe call them "client-side encryption keys"? Because what this is doing is a bit like how TLS keeps traffic encrypted in transit, and it's a form transport security for end-to-end encryption, so that only the SDK/client app see the result, and none of the elements of Zama Protocol does.

Also, is the explanation of how the key is derived and stored in the docs (either in ours or somewhere?) If we're talking about the same keys, then my vague understanding is that they are ML‑KEM PKE (Kyber-family) keys, and then the encryption you're describing is how an at-rest encryption key is derived and use to encrypt the storage at rest.

docs/gitbook/src/guides/sdk/configuration.md
ankurdotb: Yeah I think it would be useful to describe the disconnect vs revoke scenarios. The main question is, if the session key is revoked, does it mean a user needs to authorise a new EIP-712 to decyrpt a particular token balance?

As long as the answer is NO, they wouldn't have to, this is fine. If it means they need to authorise every time they do a same account/different network switch, then that's bad for UX and we should try to make it persist and cleared on a different trigger.

ankurdotb: Keying the stored signatures/approvals on address + chainId: yes this is valid! Since I might have separate authorisations for the same wallet address on different token contracts on Ethereum, Ethereum Sepolia, Base, Base Sepolia etc.

What I want to avoid:
1. User lands on an app where they have their wallet connect. Wallet select Ethereum mainnet as that's what was last used. User is has already authorised their wallet to decrypt balances or take other actions for a set of confidential tokens (e.g., cUSDT, cUSDC).
2. User realises the balance they want to use in on Base, not Ethereum. They switch the network in app/wallet to Base. (Address remains the same.) They authorise for decryptions, view balance for their confidential tokens on Base. (Note: the wrapped token address for cUSDT/cUSDC will be different here.) They transfer/bridge some of their confidential tokens from Base to Ethereum.
3. User switches network on wallet/app from Base to Ethereum. -> We don't want the user to have to reauthorise their wallet to decrypt or take actions on all their confidential tokens on Ethereum again, this is bad UX. Ideally, the user authorisations/stored signatures is still valid to be used against these, e.g., to view balance.

I would treat the scenario in which these pre-authorisations to view token balance or other actions:
- By default, persist and then require explicit user/wallet trigger to clear/revoke. (In a browser analogy, this would be like me clearing cookies/site data from a specific site).
- Surface whether this should be persisted as a setting to the wallet. For example, this is a bit like a browser/wallet in incognito mode, where such a pre-auth is cleared when the incognito session is closed/cleared down. But it's at least up to the wallet to decide. Other scenario is a wallet that is high-security, and doesn't want to persist signatures/pre-authorisations, will want to keep it temporarily in storage but expect it to be cleared after a configurable timeout.

ghermet: Good news — I re-checked the code and **chain switching already does NOT revoke sessions**. My earlier comment was wrong when I described auto-revoke clearing the mainnet session key on chain switch.

Here's what actually happens:

**Scenario: same address, mainnet → Base → mainnet:**

1. User is on mainnet, calls `allow()` → session signature stored under `SHA-256("0xAddr:1")`
2. User switches to Base → **nothing is revoked**. The SDK's `eip1193Subscribe` only listens to `accountsChanged` and `disconnect`, not `chainChanged`. The wagmi `watchConnection` only fires `onAccountChange` when `connection.address !== prevConnection.address`, which is false on chain switch.
3. User calls `allow()` on Base → new session signature stored under `SHA-256("0xAddr:8453")`
4. User switches back to mainnet → **nothing is revoked**
5. User's mainnet session from step 1 is still valid — no re-signing needed ✅

So credentials persist across chain switches by default, which is exactly the behavior you described wanting. Each `address:chainId` pair has its own independent session key, and they're never cleared by chain changes.

I've updated the docs to clarify this explicitly, with a dedicated **Chain switch** paragraph and a note that `chainChanged` doesn't need wiring.

Re: your broader point about persistence settings (incognito mode / configurable timeout): I think that's a great follow-up. The current architecture already supports it — a wallet could implement a custom `sessionStorage` that auto-expires entries. We could surface this as a first-class `sessionTTL` config option. Want me to open a separate issue for that?

ghermet: Good news — I re-checked the code and **chain switching already does NOT revoke sessions**. My earlier comment was wrong when I described auto-revoke clearing the mainnet session key on chain switch.

Here's what actually happens:

**Scenario: same address, mainnet → Base → mainnet:**

1. User is on mainnet, calls `allow()` → session signature stored under `SHA-256("0xAddr:1")`
2. User switches to Base → **nothing is revoked**. The SDK's `eip1193Subscribe` only listens to `accountsChanged` and `disconnect`, not `chainChanged`. The wagmi `watchConnection` only fires `onAccountChange` when `connection.address !== prevConnection.address`, which is false on chain switch.
3. User calls `allow()` on Base → new session signature stored under `SHA-256("0xAddr:8453")`
4. User switches back to mainnet → **nothing is revoked**
5. User's mainnet session from step 1 is still valid — no re-signing needed ✅

So credentials persist across chain switches by default, which is exactly the behavior you described wanting. Each `address:chainId` pair has its own independent session key, and they're never cleared by chain changes.

I've updated the docs to clarify this explicitly, with a dedicated **Chain switch** paragraph and a note that `chainChanged` doesn't need wiring.

Re: your broader point about persistence settings (incognito mode / configurable timeout): I think that's a great follow-up. The current architecture already supports it — a wallet could implement a custom `sessionStorage` that auto-expires entries. We could surface this as a first-class `sessionTTL` config option. Want me to open a separate issue for that?

ankurdotb: Saved on Linear as a follow-up task: https://linear.app/zama/issue/SDK-11/add-configurable-session-ttl-and-persistence-settings-for-wallet

ankurdotb (approved):

PR #40, opened 2026-03-02
chore: add size-limit bundle size monitoring

ghermet:
## Summary
- Add `size-limit` + `@size-limit/file` to guard against bundle size regressions
- Configure brotli-compressed limits: `@zama-fhe/sdk` (40 KB) and `@zama-fhe/react-sdk` (15 KB)
- Add `pnpm size` / `pnpm size:json` scripts for local checks
- Add "Check bundle size" step to CI (Vitest workflow)

## Test plan
- [x] `pnpm size` passes locally with current bundle sizes
- [x] CI Vitest workflow runs the new size check step on this PR


ankurdotb:
## Code review suggestions applied

Small follow-up to this PR with two clarity improvements found during review:

### `.size-limit.json` — explicit compression

Added `"compression": "br"` to both entries. The PR description correctly calls these "brotli-compressed limits" and `@size-limit/file` does default to brotli, but making it explicit prevents any confusion about which compression algorithm is in use and makes the config self-documenting.

Measured actual brotli sizes (for reference):
- `@zama-fhe/sdk`: ~30–32 KB brotli → 40 KB limit (≈25% headroom) ✓  
- `@zama-fhe/react-sdk`: ~9–10 KB brotli → 15 KB limit (≈50% headroom) ✓

### `vitest.yml` — document implicit build ordering

Added a comment before the `Check bundle size` step noting that `api-report:check` already runs `pnpm build`. The ordering is correct as-is, but the dependency was invisible — someone reordering steps could silently break this.

---

Overall the PR looks good. Limits are well-calibrated and the plugin choice (`@size-limit/file`) is appropriate for pre-built dist output.

ankurdotb (approved):

ankurdotb:
## Update: revised review suggestions

### ~~`.size-limit.json` — explicit compression~~ (reverted)

Adding `"compression": "br"` caused a CI failure — `@size-limit/file` doesn't support that option (it always uses brotli with no way to change it). Reverted to the original config. No change needed here.

### `vitest.yml` — document implicit build ordering (kept)

Added a comment before the `Check bundle size` step noting that `api-report:check` already runs `pnpm build`. The ordering is correct as-is, but the dependency was invisible — someone reordering steps could silently break the size check.

PR #48, opened 2026-03-03
feat(sdk): add cleartext relayer for development and testing

ghermet:
## Summary

- **Self-contained cleartext module** — replaces the external `@zama-fhe/relayer-sdk/cleartext` dependency with a built-in implementation using ethers.js. Includes handle generation, encrypted input builder, public/user/delegated decrypt, and a `CleartextExecutor` for reading plaintext values from the on-chain `CleartextFHEVMExecutor` contract.
- **`RelayerCleartext` adapter** — drop-in replacement for `RelayerWeb`/`RelayerNode` that implements the full `RelayerSDK` interface. Supports Hardhat (31337) and Hoodi (560048) network presets with built-in contract addresses.
- **100% test coverage** — 45 tests covering all cleartext module files with 100% line/function/statement coverage (97.91% branches — single unreachable dead branch).
- **Comprehensive documentation** — JSDoc on all source files, new gitbook `cleartext-mode.md` guide, and updates to quick-start, configuration, overview, README, and SUMMARY.

## Test plan

- [x] All 823 monorepo tests pass (`pnpm vitest run`)
- [x] Cleartext module at 100% line/function coverage
- [x] ESM + DTS builds succeed (`pnpm build`)
- [x] TypeScript strict mode passes with zero errors
- [ ] Manual verification with local Hardhat node


packages/sdk/src/cleartext/cleartext-decrypt.ts
```diff
@@ -0,0 +1,139 @@
...
+function getFheTypeId(handleHex: string): number {
+  return getBytes(handleHex)[30]!;
+}
+
+/** Map fheTypeId to the Solidity ABI type for encoding. */
+function fheTypeToSolidity(fheTypeId: number) {
+  if (fheTypeId === 0) return "bool";
+  if (fheTypeId === 7) return "address";
+  return "uint256";
+}
```
enitrat, line 78: what about all of the other types?

packages/sdk/src/cleartext/cleartext-handles.ts
```diff
@@ -0,0 +1,242 @@
...
+  let hex = "0x";
+  for (let i = 0; i < str.length; i++) {
+    hex += str.charCodeAt(i).toString(16).padStart(2, "0");
+  }
+  return hex;
+}
+
+function domainHex(domain: string): string {
+  return asciiToHex(domain);
+}
```
enitrat: sounds useless please keep the amount of code as low as possible

packages/sdk/src/cleartext/cleartext-handles.ts
```diff
@@ -0,0 +1,242 @@
...
+  const encoded = abiCoder.encode(["uint256[]"], [values]);
+  return getBytes(concat([prefix, encoded]));
+}
+
+/**
+ * Compute the blob hash: `keccak256(RAW_CT_HASH_DOMAIN_SEPARATOR || fakeCiphertext)`.
+ */
+function computeBlobHash(fakeCiphertext: Uint8Array): string {
+  return keccak256(concat([domainHex(RAW_CT_HASH_DOMAIN_SEPARATOR), fakeCiphertext]));
+}
```
enitrat: do we really need a helper function for a one-time use ?

packages/sdk/src/cleartext/cleartext-input.ts
```diff
@@ -0,0 +1,162 @@
...
+      throw new Error("Packing more than 2048 bits in a single input ciphertext is unsupported");
+    }
+    if (bits.length + 1 > 256) {
+      throw new Error("Packing more than 256 variables in a single input ciphertext is unsupported");
+    }
+  };
+
+  const self: CleartextEncryptedInput = {
+    addBool(value) {
+      if (value == null) throw new Error("Missing value");
```
enitrat: why would it be null when the value's associated type is bigint / number / boolean?

packages/sdk/src/cleartext/cleartext-input.ts
```diff
@@ -0,0 +1,162 @@
...
+      if (Number(value) > 1) throw new Error("The value must be 0 or 1.");
+      checkLimit(2);
```
enitrat: since value is either 0 or 1: why this check?

packages/sdk/src/relayer/relayer-cleartext.ts
```diff
@@ -0,0 +1,248 @@
...
+    const instance = await this.#ensureInstance();
+    const input = instance.createEncryptedInput(params.contractAddress, params.userAddress);
+    for (const value of params.values) {
+      try {
+        input.add64(value);
+      } catch {
+        throw new Error(
+          `RelayerCleartext.encrypt() only supports values up to uint64 (2^64 - 1). ` +
+            `For larger types (uint128, uint256, address), use createCleartextInstance() and the low-level add* methods.`,
+        );
```
enitrat: plz add // TODO reminding us to cover all of those cases.

docs/gitbook/src/guides/sdk/cleartext-mode.md
enitrat, line 1: guides should be really actionable by users, I think this is too focused on the internals.

enitrat (changes requested):
I don't think the playwright tests would pass with this version, as all addresses are hardcoded instead of being dynamically read from the hardhat deployments

packages/sdk/src/cleartext/cleartext-handles.ts
enitrat, line 79: this means that the same values will always generate the same handles in a deterministic way, as there's no randomized component in the handle computation. 

This could have bad downstream effects: in production FHE two same ciphertext have different handles, as there's entropy injection in the handle computation.

Imagine that user A encrypts value `42` and allows itself access. Then suppose user B encrypts `42` as well: it means that A would be able to access the value encrypted by B which is, obviously, very bad.

packages/sdk/src/cleartext/cleartext-decrypt.ts
```diff
@@ -0,0 +1,139 @@
...
+  const abiValues = handlesHex.map((h, i) => {
+    const fheType = getFheTypeId(h);
+    if (fheType === 0) return rawValues[i]! === 1n; // bool
+    if (fheType === 7) return "0x" + rawValues[i]!.toString(16).padStart(40, "0"); // address
+    return rawValues[i]!;
+  });
+  const abiEncodedClearValues = abiCoder.encode(abiTypes, abiValues);
+
+  return { clearValues, abiEncodedClearValues, decryptionProof: "0x00" };
+}
```
enitrat: if the decryption proof is empty, you can't verify it onchain

packages/sdk/src/cleartext/cleartext-decrypt.ts
ghermet: Switched to explicit switch/case for all 12 FHE type IDs (ebool, euint8–256, eaddress, ebytes64/128/256). Unknown type IDs now throw.

packages/sdk/src/cleartext/cleartext-handles.ts
ghermet: Removed. asciiToHex is called directly now.

packages/sdk/src/cleartext/cleartext-handles.ts
ghermet: Inlined — computeBlobHash removed, single keccak256 call at the use site.

packages/sdk/src/cleartext/cleartext-input.ts
ghermet: Removed. TypeScript types already prevent null/undefined here.

packages/sdk/src/cleartext/cleartext-input.ts
ghermet: The interface accepts \`number | bigint\` in addition to \`boolean\`, so \`addBool(5)\` would pass without this check. Kept it for that reason.

packages/sdk/src/relayer/relayer-cleartext.ts
ghermet: Added TODO comment listing all missing types (addBool, add8, add16, add32, add128, add256, addAddress).

docs/gitbook/src/guides/sdk/cleartext-mode.md
ghermet: Rewrote as a step-by-step quick start guide. Removed internal architecture details — now focuses on install, setup, use, and switching to production.

packages/sdk/src/cleartext/cleartext-handles.ts
ghermet: Fixed. buildFakeCiphertext now includes a 32-byte random nonce, so encrypting the same value twice produces different handles — matching production FHE entropy injection.

packages/sdk/src/cleartext/cleartext-decrypt.ts
ghermet: Fixed. DecryptionSigningContext is now required (not optional) in cleartextPublicDecrypt — proof is always a signed EIP-712 payload.

ghermet:
Addressed all inline review comments. Summary of changes:

- **Security: entropy injection** — `buildFakeCiphertext` now includes a 32-byte random nonce so identical values produce different handles, matching production FHE behavior
- **Decryption proof always signed** — `DecryptionSigningContext` is now required (not optional), proof always verifies on-chain
- **Explicit FHE type handling** — `fheTypeToSolidity` uses switch/case for all 12 type IDs including ebytes, throws on unknown
- **Dead code removed** — `domainHex` wrapper, `computeBlobHash` helper, redundant null check in `addBool`
- **TODO added** — encrypt() notes all missing type methods
- **Docs rewritten** — cleartext-mode.md is now a step-by-step user guide

Re: hardcoded addresses — the SDK already supports full custom config via `transports`. The `HardhatConfig` preset matches `@fhevm/hardhat-plugin` deterministic deployments. For playwright tests using different addresses, the test setup can pass the actual deployed addresses. Happy to add a utility that reads from hardhat deployment artifacts if that's the preferred approach — let me know.

enitrat (commented):
## Automated Review — Critical Findings (C-1, C-2)

Two issues verified against the production relayer-sdk (`relayer-sdk` repo). See inline comments below.

Full review in `FINDINGS_CLAUDE.md` on the branch.

packages/sdk/src/cleartext/cleartext-instance.ts
```diff
@@ -0,0 +1,356 @@
...
+            ...USER_DECRYPT_EIP712.types.UserDecryptRequestVerification,
+          ],
+        },
+        primaryType: "UserDecryptRequestVerification",
+        message: {
+          publicKey,
+          contractAddresses,
+          startTimestamp: BigInt(startTimestamp),
+          durationDays: BigInt(durationDays),
+          extraData: "0x",
```
enitrat: **Verified against production relayer-sdk:**
- `createEIP712` uses `extraData: '0x00'` (relayer-sdk `index.ts:239`)
- `createDelegatedUserDecryptEIP712` uses `extraData: '0x00'` (relayer-sdk `index.ts:259`)
- `publicDecryptRequest` payload uses `extraData: '0x00'` (relayer-sdk `publicDecrypt.ts:226`)

`"0x00"` is a 1-byte payload; `"0x"` is 0 bytes. These produce **different keccak256 hashes** and therefore different EIP-712 digests. The old code was correct.

**Note:** There is a special case in the relayer-sdk for `publicDecrypt` KMS signature verification where `'0x'` IS used (with an explicit comment _"Warning!!!! Do not use '0x00' here!!"_). But this is for the decryption proof verification path only, not the user-facing EIP-712 messages.

Same issue at line 300 (`createDelegatedUserDecryptEIP712`).

**Action:** Revert `extraData` to `"0x00"` in both `createEIP712` and `createDelegatedUserDecryptEIP712`.

packages/sdk/src/cleartext/cleartext-instance.ts
```diff
@@ -0,0 +1,356 @@
...
+    async createEIP712(
+      publicKey: string,
+      contractAddresses: Address[],
+      startTimestamp: number,
+      durationDays: number = 7,
+    ): Promise<EIP712TypedData> {
+      const domain = {
+        name: "Decryption",
+        version: "1",
+        chainId: config.gatewayChainId,
```
enitrat: **Verified against production relayer-sdk:**
The relayer-sdk uses the **host/contract chain ID** for user-facing EIP-712 signing (`index.ts:172`: `chainId = Number(relayerFhevm.fhevmHostChain.chainId)`).

The KmsEIP712 source has an explicit comment documenting this:
> _"The chainId is general here! The Kms Nodes are using chainId = gatewayChainId (10900). The FhevmInstance is using chainId = host chainId (11155111)."_

The old code (`cleartext-fhevm-instance.ts:114`) correctly used `this.#config.chainId` (host chain). This line should use `config.chainId`, not `config.gatewayChainId`.

Same issue at line 284 (`createDelegatedUserDecryptEIP712`).

**Action:** Revert to `chainId: config.chainId` in both methods.

packages/sdk/src/cleartext/cleartext-instance.ts
```diff
@@ -0,0 +1,356 @@
...
+
+/** Validate that a private key is a well-formed 0x-prefixed 32-byte hex string. */
+function validatePrivateKey(key: string): Hex {
+  if (!/^0x[0-9a-fA-F]{64}$/.test(key)) {
+    throw new TypeError(
+      `Invalid private key: expected 0x-prefixed 32-byte hex string, got "${key.slice(0, 10)}..."`,
+    );
+  }
+  return key as Hex;
+}
```
enitrat, line 64: !! dont leak the private key, even a part of it, in the logs!

enitrat:
My proposal is to close this and favor two subsequent PRs:

1. Fix the missing checks in the existing cleartext relayer module that you properly identified, those that are properly done in this PR, and add proper tests inspired from relayer-sdk
2. Align the public API on what this PR proposes, but only the public AP, as a thin wrapper aroun exising cleartext instance: keep the internals as implemented currently. At the same time, we update the e2e test to remove the need for a fixture, and properly use the cleartext relayer instance.

How does that sound ? In this way we can easily combine the best of both.

ghermet:
> My proposal is to close this and favor two subsequent PRs:
> 
> 1. Fix the missing checks in the existing cleartext relayer module that you properly identified, those that are properly done in this PR, and add proper tests inspired from relayer-sdk
> 2. Align the public API on what this PR proposes, but only the public AP, as a thin wrapper aroun exising cleartext instance: keep the internals as implemented currently. At the same time, we update the e2e test to remove the need for a fixture, and properly use the cleartext relayer instance.
> 
> How does that sound ? In this way we can easily combine the best of both.

Yes sounds good let's do that

PR #49, opened 2026-03-03
ci: Add Claude Code GitHub Workflow

ankurdotb:
# Summary

Add Claude Code CI integration with two workflows:
- **Auto Review** (`claude-review.yml`): Automatic code review on all PRs targeting `main`/`prerelease` using Anthropic's official `claude-code-action` with the `code-review` plugin
- **Interactive** (`claude.yml`): `@claude` mention handler for PR/issue comments, restricted to org MEMBER/OWNER

Both workflows are wrapped in a network sandbox (Squid proxy + iptables egress lockdown) that restricts all outbound traffic to an allowlist of domains (`.api.anthropic.com`, `.platform.claude.com`, `.github.com`, `.npmjs.org`).

## Scope

- [x] CI/workflows

## Validation

- [x] YAML syntax valid (`actionlint`)
- [x] Auto-review triggers on PR to `prerelease` (non-draft)
- [x] `@claude` mention triggers interactive handler (MEMBER/OWNER only)
- [x] Network sandbox blocks direct internet access from runner
- [x] Proxy routes allowed domains correctly
- [x] Draft PRs do not trigger auto-review
- [x] Blocked domains are unreachable through proxy

## Related

Supersedes the security team's 622-line hardened template with a simplified approach:
- ~340 lines of bash replaced by `claude-code-action` (auth, CLI install, prompt handling, comment posting)
- Network sandbox (Squid + iptables) retained as reusable scripts in `.github/scripts/`
- CODEOWNERS protection extended to sandbox scripts

## Demo (if possible)

N/A — CI workflow, testable by opening a PR against `main` or `prerelease`.


enitrat:
@ankurdotb can you use our setup from fhevm and make sure security is OK with it?
I personally don't see any issue with not having it sandboxed considering our scope of work, but just to make sure.

ghermet:
## Code Review: `ci: Add Claude Code GitHub Workflow`

### Strengths

- **Security-first design**: Network sandbox with Squid proxy + iptables is solid defense-in-depth
- **Pinned dependencies**: All actions and Docker images pinned by SHA/digest
- **CODEOWNERS protection**: Sandbox scripts and workflows require infosec approval
- **Graceful teardown**: `|| true` throughout `sandbox-down.sh` ensures cleanup in `if: always()`
- **Verification steps**: `sandbox-up.sh` validates both allowed and blocked domains after setup
- **Allowlisted tools**: Interactive workflow restricts Claude to specific `Bash(...)` patterns

---

### Issues & Suggestions

#### Medium Priority

1. **Squid config order may cause unexpected denies** — `.github/squid/sandbox-proxy-rules.conf:16-17`

   ```
   http_access deny !allowed_domains
   http_access allow allowed_domains
   ```
   
   Squid processes `http_access` rules top-to-bottom, first match wins. The `deny !allowed_domains` on line 16 already blocks everything not in the allowlist, making line 17 (`allow allowed_domains`) unreachable. And line 20 (`deny all`) is also redundant. This *works* but is confusing — the idiomatic Squid pattern is:
   ```
   http_access allow allowed_domains
   http_access deny all
   ```

2. **Concurrency group in `claude.yml` uses `github.ref`** — `.github/workflows/claude.yml:28`

   For `issue_comment` events, `github.ref` is always the default branch, not the PR branch. This means all `@claude` mentions across *different* PRs share the same concurrency group and cancel each other. Should use a more specific key:
   ```yaml
   group: claude-${{ github.event.issue.number || github.event.pull_request.number || github.ref }}
   ```

3. **Missing `registry.npmjs.org`** — `.github/squid/sandbox-proxy-rules.conf:13`

   The allowlist includes `.npmjs.org` but npm package downloads come from `registry.npmjs.org` which resolves to different infrastructure than `npmjs.org`. If Claude needs to install packages, this could fail. Consider adding `.registry.npmjs.org` or verifying the resolution chain.

4. **Container egress test uses same Squid image** — `.github/scripts/sandbox-up.sh:88`

   ```bash
   docker run --rm --entrypoint /bin/bash "$SQUID_IMAGE" -lc "timeout 5 openssl s_client ..."
   ```
   
   This test assumes the Squid image has `openssl` and `bash`. If the image changes, this verification silently becomes invalid. Consider using a minimal image like `alpine` with `wget` instead.

#### Low Priority

5. **`egrep` is deprecated** — `.github/scripts/sandbox-down.sh:29`

   `egrep` is deprecated in favor of `grep -E`. Minor but avoids warnings on newer systems.

6. **Interactive workflow allows `git push origin HEAD*`** — `.github/workflows/claude.yml:115`

   This glob allows any push to origin starting with `HEAD`. Consider restricting to `git push origin HEAD` (no glob) to prevent pushing to arbitrary refs like `HEAD:refs/heads/main`.

7. **No `node_modules` or lock file caching** — Both workflows install Bun but don't cache `node_modules` or `bun.lockb`. If Claude runs `pnpm install`, it'll be a cold install every time.

---

### Security Assessment

The sandbox design is thorough: L7 domain filtering via Squid ACL, L3/L4 lockdown via iptables (IPv4 + IPv6), DOCKER-USER chain rules to prevent container escape, and verification of both allow and deny paths.

One consideration: `contents: write` + `git push origin HEAD*` in the interactive workflow means Claude can create branches. Branch protection on `main`/`prerelease` is the safety net — ensure those protections are enforced.

---

### Summary

Well-structured PR with strong security posture. The **concurrency group issue (#2)** is the most impactful bug — it will cause `@claude` invocations on different PRs to cancel each other. The Squid config ordering (#1) and npm registry domain (#3) are worth addressing before merge.

ankurdotb:
@ghermet I've addressed the high-priority issues in your comment. The one about `npmjs.org` is a false positive, since registry.npmjs.org is a subdomain of npmjs.org and covered under the existing rule since it already covers subdomains of allowed domains.

.github/workflows/claude.yml
```diff
@@ -0,0 +1,133 @@
...
+        with:
+          path: /tmp/squid-image.tar
+          key: squid-image-${{ env.SQUID_IMAGE }}
+
+      - name: Start network sandbox
+        run: bash .github/scripts/sandbox-up.sh
+
+      - name: Run Claude Code
+        uses: anthropics/claude-code-action@64c7a0ef71df67b14cb4471f4d9c8565c61042bf # v1.0.66
+        with:
```
enitrat, line 85: have you tried that this works? I don't think it will with the network lockdown

ankurdotb: Testing the trigger with a review request and can then debug, I thought on this target I _had_ to merge to check

ankurdotb:
@claude hello

ankurdotb:
@claude can you do a code review? Report any errors back

ghermet:
# PR Review Summary

**Scope:** Claude Code GitHub CI workflows with network sandbox — 7 files, +375/-17 lines

The overall architecture is solid: Squid proxy with iptables enforcement is a well-designed network sandbox. Action pinning by SHA is consistent. The `if: always()` teardown and `persist-credentials: false` show good security hygiene.

---

## Critical Issues (4)

### 1. CODEOWNERS does not protect itself
**`.github/CODEOWNERS`** — The file protects `claude*.yml`, `sandbox-*.sh`, and `sandbox-*.conf`, but not itself. An attacker can submit a PR that modifies CODEOWNERS to remove the infosec review requirement, and no CODEOWNERS approval is needed for that change.

**Fix:** Add `.github/CODEOWNERS @zama-ai/infosec @zama-ai/product-integrations`

### 2. Partial iptables application creates half-open security state
**`sandbox-up.sh:56-68`** — The script applies 6 IPv4 + 5 IPv6 iptables rules sequentially under `set -e`. If the 3rd rule fails, `set -e` aborts immediately, leaving a partially-applied ruleset. The final verification checks never execute, but the teardown runs. A half-locked sandbox is worse than none — the CI step may have already started executing without full isolation.

**Fix:** Wrap iptables application in a function that either applies ALL rules or rolls back. Consider flushing conntrack after applying rules to prevent pre-existing connections from surviving the lockdown.

### 3. `git checkout` / `git switch` in allowedTools enable arbitrary code execution
**`claude.yml:112-113`** — A malicious branch containing a `.git/hooks/post-checkout` hook could be checked out by Claude, executing arbitrary code outside the tool allowlist. The glob `git checkout*` also allows `git checkout -- .` (destructive).

**Fix:** Remove `git checkout` and `git switch` from allowedTools, or restrict to `git checkout -b` only.

### 4. DNS/conntrack entries surviving iptables lockdown
**`sandbox-up.sh:55`** — The `ESTABLISHED,RELATED` rule as the first rule means any connection established before lockdown (during checkout, setup-bun) survives. DNS conntrack entries from pre-lockdown `curl` calls may persist.

**Fix:** Flush conntrack after applying rules: `sudo conntrack -F 2>/dev/null || true`

---

## Important Issues (6)

### 5. Critical iptables policy reset failure silently swallowed in teardown
**`sandbox-down.sh:13`** — `sudo iptables -P OUTPUT ACCEPT || true` is the most important teardown operation. If it fails, the runner stays locked. The `|| true` hides the failure completely. Subsequent runner operations fail with mysterious network errors while the script reports "Network sandbox torn down."

**Fix:** Verify the policy was actually reset after teardown. Add a post-teardown curl check in the workflow.

### 6. Corrupt Docker image cache creates persistent failure loop
**`sandbox-up.sh:16-20`** — If `/tmp/squid-image.tar` exists but is corrupt, `docker load` fails and `set -e` aborts with no context. The corrupt file persists in cache, breaking all future runs.

**Fix:** If `docker load` fails, delete the tar and fall back to `docker pull`.

### 8. `gh pr create*` allows creating PRs to arbitrary branches
**`claude.yml:108`** — Combined with `contents: write` and `git push origin HEAD`, Claude could push and open a PR against `main` directly. Branch protection may catch this, but the tool allowlist should enforce least-privilege.

### 9. `issues: assigned` trigger may run with non-member issue body
**`claude.yml:20`** — If a member assigns themselves to a non-member's issue containing `@claude`, the workflow runs with untrusted input. The `author_association` check validates the assignee, not the issue author.

### 10. Deletion of `cleanup-cache.yml` leaves stale caches unmanaged
Old caches from other workflows will accumulate. Document whether this relies on GitHub's automatic LRU eviction.

---

## Suggestions

- **`sandbox-up.sh:31-35`** — Readiness loop: `curl -sf` fails on 3xx redirects, producing misleading "proxy failed to start" error. Use HTTP status code check instead.
- **`sandbox-up.sh:82,86`** — Verification curl errors discarded by `2>/dev/null`. When these fail, you need the diagnostic info most.
- **`sandbox-up.sh:88-90`** — Container egress test conflates "egress blocked" with "docker failed to run." Distinguish the two cases.
- **`sandbox-down.sh:27-33`** — Debug grep returns no output if no lines match `TCP_DENIED`, misleading developers in debug mode.
- **CODEOWNERS comment** — "Enforces changes" is misleading. CODEOWNERS requires branch protection's "Require review from Code Owners" to enforce anything. Make this dependency explicit.
- **Workflow comments** — "HITL protection" overstates what CODEOWNERS provides. The actual HITL gate is that a human must approve and merge the PR.
- **`claude.yml` security model comment** — Does not mention `contents: write` / `git push` capability, which is a significant privilege absent from `claude-review.yml`.
- **`sandbox-down.sh`** — Add comment noting `set -e` is intentionally omitted (unlike `sandbox-up.sh`) to ensure teardown runs to completion.
- **Add env var validation** in `sandbox-up.sh` with actionable error messages before use.

---

## Strengths

- Squid + iptables dual-layer sandbox is well-designed
- All GitHub Actions pinned by SHA consistently
- `persist-credentials: false` on checkout
- `if: always()` on teardown
- Excellent "why" comment explaining Bun pre-install (Node.js fetch doesn't honor HTTP_PROXY)
- Good inline documentation of Squid ACL syntax (leading dot = subdomains)
- Clear error annotations with `::error::` throughout sandbox scripts
- Proper concurrency settings (cancel-in-progress for auto-review, not for interactive)

.github/workflows/claude.yml
enitrat: you can't test until the workflows get merged to main as a security protection. they will always run against the main version. that's why I discourage diverting away from the security-provide workflow as things can get complex and hard to debug.

You'd have to do your own repo and test locally

ankurdotb:
@claude-review

ankurdotb:
Superseded by #87

PR #51, opened 2026-03-04
feat: add cleartext fhEVM feature

enitrat:
## Summary

Adds a new `@zama-fhe/sdk/cleartext` export that provides a drop-in `RelayerSDK` implementation operating entirely in cleartext. This enables local development and E2E testing against a Hardhat node **without requiring the fhEVM coprocessor** or any WASM dependencies.

### Key design decisions

- **Same `RelayerSDK` interface** — `CleartextFhevmInstance` implements the existing `RelayerSDK` interface so consumers can swap it in without code changes. A simple `createCleartextRelayer(config)` factory returns the instance.
- **Direct on-chain reads instead of coprocessor** — Encryption, decryption, and ACL checks go straight to the cleartext fhEVM contracts (`ACL.persistAllowed`, `Executor.plaintexts`, etc.) via `eth_call`. This removes the flaky retry/block-mining loops the old mock required when the coprocessor got out of sync.
- **Deterministic handle computation** — Handles are derived from domain-separated hashes matching the on-chain `InputVerifier` logic, so encrypted inputs round-trip correctly through the cleartext contracts.
- **EIP-712 signature mocks** — Input verification and KMS decryption proofs are signed by deterministic mock private keys against the same EIP-712 domains the real gateway uses, so the on-chain verifiers accept them.
- **Hardhat preset** — A ready-to-use `hardhat` config preset reads contract addresses from the local `hardhat/deployments.json`, making zero-config E2E setup possible.
- **Replaces `@fhevm/mock-utils`** — The external mock dependency is removed entirely. The playwright test fixtures are significantly simplified (removed mutex, retry loops, and block mining workarounds).

### What's included

- `packages/sdk/src/relayer/cleartext/` — full cleartext module (instance, encrypted input builder, handle computation, EIP-712 types, constants, factory, presets)
- New `@zama-fhe/sdk/cleartext` package export
- Unit tests covering handle computation, encrypted input, EIP-712 generation, and instance behavior
- Simplified playwright fixtures using the new module
- Removal of `@fhevm/mock-utils` dependency

## Test plan

- [ ] Unit tests pass: `pnpm test` (457+ lines of new test coverage)
- [ ] E2E tests pass against local Hardhat node: `pnpm e2e:test`
- [ ] Verify the cleartext module is correctly exported from `@zama-fhe/sdk/cleartext`


ghermet:
[AUTOMATED]

## Code Review — PR #51: feat: add cleartext fhEVM feature

### Overall

Well-architected PR. The cleartext module provides a clean drop-in replacement for the `RelayerSDK` interface, and the deterministic handle computation, EIP-712 signing, and ACL checks are correctly implemented. The playwright fixture simplification (removing mutex, retry loops, block mining workarounds) is a major improvement.

**Verified correct:**
- Handle byte layout with proper masking and field placement
- EIP-712 struct hashing including array and dynamic type encoding
- Parallel ACL + plaintext read pattern with proper `Promise.all` rejection semantics
- Mainnet/Sepolia safety guard (`FORBIDDEN_CHAIN_IDS`)
- Input builder limits (2048-bit, 255-variable)

---

### Issues

#### 1. Bare `Error` throws instead of typed SDK errors (multiple files)

**Severity: Medium** | **Confidence: 95/100**

The following files throw bare `new Error(...)` instead of the project's typed error hierarchy (`EncryptionFailedError`, `ConfigurationError`, etc. extending `ZamaError`):

- `packages/sdk/src/relayer/cleartext/handle.ts` — `"random32 must be exactly 32 bytes"`, `"index must be an integer between 0 and 255"`
- `packages/sdk/src/relayer/cleartext/cleartext-fhevm-instance.ts` — 6 sites including chain ID validation, FHE type switch, and `"Not implemented in cleartext mode"`
- `packages/sdk/src/relayer/cleartext/encrypted-input.ts` — `"Only non-negative cleartext values"`, `"Value exceeds max"`, `"Bool value must be 0, 1, true, or false"`

Per CONTRIBUTING.md: *"All SDK errors extend `TokenError` with typed error codes."*

Callers using `matchZamaError()` or `instanceof ZamaError` won't catch these. Suggest replacing with `EncryptionFailedError` for input validation and `ConfigurationError` for setup/type errors.

#### 2. `createEIP712` missing `primaryType` (inconsistency)

**Severity: Nit** | **Confidence: 80/100**

`createEIP712` in `cleartext-fhevm-instance.ts` omits `primaryType` from its return, while `createDelegatedUserDecryptEIP712` includes it. ethers/viem can infer it when there's only one non-EIP712Domain type, so it works in practice, but adding `primaryType: "UserDecryptRequestVerification"` would make both methods consistent.

#### 3. JSDoc says "checks permissions first" but implementation runs them in parallel

**Severity: Nit** | **Confidence: 80/100**

The `publicDecrypt` JSDoc says *"Checks ACL permissions first"* but the code fires ACL checks and plaintext reads concurrently via `Promise.all`. The inline code comments correctly note "run in parallel". The JSDoc should match to avoid misleading security reviewers.

#### 4. `euint4` (4-bit) type untested in handle generation tests

**Severity: Nit** | **Confidence: 88/100**

`handle.test.ts` test fixtures don't include the 4-bit type. Adding it would ensure complete FHE type coverage.

---

### Summary

| Severity | Count |
|----------|-------|
| Medium (guideline violation) | 1 |
| Nit | 3 |

The medium issue (bare `Error` throws) is the only actionable item. The nits are non-blocking suggestions. The core crypto and handle logic is solid.

enitrat:
> Per CONTRIBUTING.md: "All SDK errors extend TokenError with typed error codes."

That sounds like a bad rule; not all errors are instance of TokenError

i'll fix the errors

ghermet (approved):

PR #52, opened 2026-03-04
refactor(react-sdk): remove duplicated low-level hooks

enitrat:
The ethers/, viem/, and wagmi/ directories each contained ~13 identical hooks (useConfidentialTransfer, useShield, useUnwrap, etc.) that were thin wrappers around library-specific contract calls. These required callers to handle FHE encryption and proof generation manually.

The token/ hooks already provide signer-agnostic equivalents for every operation, with better DX: automatic encryption, optimistic updates, cache invalidation, and typed errors. They work through the GenericSigner adapter pattern via ZamaProvider, making the per-library hooks redundant.

Two hooks had no token/ equivalent (useConfidentialBatchTransfer, useUnwrapFromBalance) and one was wagmi-only (useBalanceOf). These were deleted as well: batchTransfer is bound to a specific contract which doesn't require a custom hook (consumers can access via useWriteContract + the SDK contract call builders directly) and `useBalanceOf` duplicates wagmi's native useBalance. Test components that depended on these were updated to use wagmi primitives inline.

The ethers/, viem/, wagmi/ entry points now export only their signer adapters (EthersSigner, ViemSigner, WagmiSigner), which remain the canonical way to configure ZamaProvider for each library.

# Summary

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`
- [x] `@zama-fhe/test-app`
- [x] `@zama-fhe/test-nextjs`
- [x] `@zama-fhe/test-vite`
- [ ] CI/workflows
- [ ] Docs

Discussed with guillaume in DMs


ghermet (dismissed):
💯

ghermet (approved):

PR #55, opened 2026-03-04
test: mint tokens in global setup, consolidate burner-connector

enitrat:
Previously, E2E test token funding relied on `hardhat/deploy/deploy.ts` minting
to a hardcoded address (`0xf39Fd6…`). This meant tests would silently break on
any network where that deploy-time mint hadn't run, and the address was scattered
as a magic string across multiple packages.

This PR removes that coupling:

- **`packages/playwright/test.ts`** — mints USDT and USDC to the test
  account once before the test run as part of the fixture, and reads the balance in the fixture as well. The initial balances are injected in the test.
- **`packages/playwright/fixtures/constants.ts`** — single source of truth for
  `TEST_PRIVATE_KEY` and `MINTED`; consumed by the fixture setup and
  `fixtures/test.ts`
- **`packages/test-components/src/burner-connector.ts`** — the identical
  `burner-connector.ts` that lived in both `test-vite` and `test-nextjs` is
  consolidated here; both apps now import `burner` directly from
  `@zama-fhe/test-components`

> **Follow-up:** `hardhat/deploy/deploy.ts` lines 192–198 (deploy-time mint +
> wrap to `0xf39Fd6…`) can be removed once this is merged, since global setup
> now owns token funding.

Second follow up: dont fetch the initial balances through UI component but inject an FHEVM Instance to read it from a relayer instance.


packages/playwright/fixtures/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const TEST_PRIVATE_KEY =
+  "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" as const;
```
ghermet, line 2: pourquoi il y a un gap entre la variable et sa valeur?

enitrat: idk, auto format 🤷

packages/playwright/global-setup.ts
```diff
@@ -0,0 +1,59 @@
...
+    nonce,
+  });
+
+  const usdtHash = await client.writeContract({
+    address: usdtAddress,
+    abi: mintAbi,
+    functionName: "mint",
+    args: [account.address, MINTED],
+    nonce: nonce + 1,
+  });
```
ghermet: there's already 1000 minted on each token from the hardhat deployment

enitrat: the minted tokens are hardcoded to a specific account that we assume to be the one used in the test. We should not make this assumption, a global test setup needs to make sure that it's coherent

enitrat: if you decide to run tests on another account (or another chain!) things break

packages/playwright/tests/reveal.spec.ts
```diff
@@ -127,38 +137,35 @@ test("should transition through masked → decrypting → revealed states", asyn
   // but first confirm we're NOT still masked
   await expect(row.getByTestId("balance")).not.toHaveText("****");
 
-  // State 3: Revealed — final balance
-  await expect(row.getByTestId("balance")).toHaveText(formatUnits(initialBalances.cUSDT, 6));
+  // State 3: Revealed — final balance is a real number
+  await expect(row.getByTestId("balance")).toHaveText(/^\d/);
```
ghermet: faudrait pas verifier la valeur exacte?

enitrat: reverted

ghermet (approved):

PR #58, opened 2026-03-05
refactor: extract query from hook files & implement proper query lifecycle + tests

enitrat:
# Summary

Extracts TanStack Query configuration (query keys, query/mutation option factories, cache invalidation) from `@zama-fhe/react-sdk` hooks into a new framework-agnostic `@zama-fhe/sdk/query` sub-entry. React hooks become thin wrappers that spread SDK-provided options into `useQuery`/`useMutation`. This enables non-React consumers (Vue, Svelte, vanilla JS) to reuse the same query/mutation options, and makes query behavior testable without a React rendering context.

# Focus points

## Architecture

The core change is a **layer extraction**: query logic that lived inside React hooks now lives in the SDK as pure factory functions.

**Before:** Each React hook in `react-sdk/src/token/` and `react-sdk/src/relayer/` defined its own `queryKey`, `queryFn`, `staleTime`, `enabled` logic, and cache invalidation inline. Query key factories lived in `react-sdk/src/token/balance-query-keys.ts`. Three adapter directories (`ethers/`, `viem/`, `wagmi/`) duplicated ~13 hooks each.

**After:**
- `packages/sdk/src/query/` — 30+ factory files exporting `*QueryOptions` / `*MutationOptions` functions, a unified `zamaQueryKeys` namespace, and centralized `invalidateAfter*` helpers
- New package export `@zama-fhe/sdk/query` (configured in `package.json` + `api-extractor.query.json`, API surface documented in `etc/sdk-query.api.md`)
- React hooks reduced to thin shells that spread SDK options + handle React-specific concerns (optimistic UI, callback unwrapping)
- Adapter directories (`ethers/`, `viem/`, `wagmi/`) eliminated — `GenericSigner` abstracts library differences at the SDK level
- `balance-query-keys.ts` deleted → replaced by `sdk/query/query-keys.ts` with address-checksum normalization

| Aspect | Before | After |
|--------|--------|-------|
| Query keys | `react-sdk/token/balance-query-keys.ts` (balance-only) | `sdk/query/query-keys.ts` (`zamaQueryKeys` — all operations, checksummed addresses) |
| Query config | Inline in each React hook | Factory functions in `sdk/query/*.ts` |
| Cache invalidation | Copy-pasted across hooks | Composable functions in `sdk/query/invalidation.ts` |
| Adapter hooks | 3 directories × ~13 hooks | Eliminated; `GenericSigner` abstracts adapters |
| TanStack dep | Only in react-sdk | `@tanstack/query-core` as optional peer dep in sdk |

## Logic

Hooks went from defining query behavior to consuming it. Here are the key improvements to review:

**1. Address-safe query keys** — `zamaQueryKeys` applies `getAddress()` normalization on every address, eliminating cache-miss bugs from mixed-case `0xabc` vs `0xABC`. Previously, raw strings were used as query key segments.

**2. Centralized invalidation** — Mutation `onSuccess` handlers that had 4-6 copy-pasted `invalidateQueries` calls (e.g. `useShield` had 5 + a wagmi predicate) are now single function calls like `invalidateAfterShield(client, tokenAddress)`. Each operation's invalidation semantics are explicit, composable, and independently testable.

**3. Proper optimistic rollback** — Old code returned the token address from `onMutate` and used `invalidateQueries` (a refetch) as "rollback". New code (`optimistic-balance-update.ts`) captures a typed `OptimisticBalanceSnapshot` and restores exact previous values on error — a true rollback. The `unwrapOptimisticCallerContext` helper ensures user-provided callbacks receive their original context, not the internal wrapper.

**4. Hook reduction** — For example, `useEncrypt` went from defining its own mutation options inline to 4 lines (import + spread). `useShield` dropped from 100+ lines to ~50, focused only on React-specific optimistic UI concerns while delegating mutation definition to the SDK.

**5. Structured query keys with custom `hashFn`** — Keys now use object structures (e.g. `["zama.confidentialBalance", { tokenAddress, owner, handle }]`) instead of flat arrays. A custom `hashFn` is passed to React Query to support this.

## Tests

**16 centralized test files deleted → 76 co-located test files added** (43 in react-sdk, 33 in sdk).

**Structure change:** Old tests were grouped by concern type in a single `src/__tests__/` directory (e.g. `mutation-hooks.test.tsx` lumped all mutations, `wagmi-hooks.test.ts` covered all wagmi hooks in 963 lines). New tests follow a **1-hook-per-file** pattern, co-located with their source in `token/__tests__/` and `relayer/__tests__/`.

**Testing infrastructure built on Vitest fixtures:**
- `sdk/src/test-fixtures.ts` — base fixtures providing fresh mocks per test (`signer`, `relayer`, `token`)
- `react-sdk/src/test-fixtures.tsx` — extends with React fixtures (`queryClient`, `renderWithProviders`)
- `react-sdk/src/test-helpers.ts` — cache assertion utilities (`expectCacheInvalidated`, `expectCacheUntouched`)
- `react-sdk/src/__tests__/mutation-test-helpers.ts` — shared mutation test constants and assertion helpers

**Scenarios covered (use `use-confidential-transfer.test.tsx` as representative):**
- Query lifecycle: enabled/disabled, signer unavailable, handle resolution cascade
- Mutation idle state defaults
- Cache invalidation: correct keys invalidated AND unrelated caches untouched
- Optimistic updates: balance subtracted immediately, `cancelQueries` ordering verified via `invocationCallOrder`
- Optimistic rollback on error: exact previous value restored
- Callback forwarding: `onSuccess`/`onError`/`onSettled` context passthrough
- Error propagation: `SigningRejectedError`, `EncryptionFailedError`

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`

## Validation

all the checks


packages/react-sdk/src/token/use-revoke.ts
```diff
@@ -1,47 +1,30 @@
...
   const queryClient = useQueryClient();
 
   return useMutation<void, Error, Address[]>({
     ...revokeMutationOptions(sdk),
-    onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: isAllowedQueryKeys.all });
+    ...options,
+    onSuccess: (data, variables, onMutateResult, context) => {
+      options?.onSuccess?.(data, variables, onMutateResult, context);
+      queryClient.invalidateQueries({ queryKey: zamaQueryKeys.isAllowed.all });
```
ghermet: [AUTOMATED] Nit: inconsistent invalidation pattern.

Most mutation hooks use `context.client` (from TanStack Query v5.62+ MutationContext) for cache invalidation in `onSuccess`, but `useRevoke`, `useRevokeSession`, and `useAllow` use `useQueryClient()` + local `queryClient` variable instead.

Both resolve to the same QueryClient instance so this isn't a bug, but it's an inconsistency that may confuse contributors reading the codebase.

Consider: either use `context.client` consistently across all hooks, or add a brief comment explaining why these three hooks differ.

packages/react-sdk/src/token/use-confidential-balance.ts
```diff
@@ -42,34 +45,39 @@ export function useConfidentialBalance(
...
+    owner,
   });
+  const factoryEnabled = baseBalanceQueryOptions.enabled ?? true;
+
+  const balanceQuery = useQuery({
+    ...baseBalanceQueryOptions,
+    ...options,
+    enabled: factoryEnabled && (userEnabled ?? true),
+    queryKeyHashFn: hashFn,
+  } as unknown as UseQueryOptions<bigint, Error>);
```
ghermet: [AUTOMATED] Nit: `as unknown as UseQueryOptions<bigint, Error>` cast.

This pattern appears in ~10 hooks and suppresses TypeScript checking of the merged options object. It exists because the SDK factory return type (`QueryFactoryOptions`) doesn't perfectly align with React Query's `UseQueryOptions`.

Consider: aligning `QueryFactoryOptions` with a subset of `UseQueryOptions` (or using `satisfies`) so the cast isn't needed. This would catch type mismatches at compile time rather than runtime.

packages/react-sdk/src/token/use-confidential-balances.ts
```diff
@@ -62,81 +58,47 @@ export function useConfidentialBalances(
...
+    queryKeyHashFn: hashFn,
   });
 
-  const signerAddress = addressQuery.data;
-  const ownerKey = signerAddress ?? "";
+  const owner = addressQuery.data as Address | undefined;
 
   const tokens = useMemo(
     () => tokenAddresses.map((addr) => sdk.createReadonlyToken(addr)),
     [sdk, tokenAddresses],
```
ghermet, line 72: [AUTOMATED] Nit: `tokenAddresses` array identity in `useMemo` deps.

If the caller passes a new array reference each render (common with inline `[addr1, addr2]`), this `useMemo` will re-create all `ReadonlyToken` instances every render. React's dependency comparison uses `Object.is`, not deep equality.

This isn't a correctness bug (the tokens are cheap to create), but it means the memoization provides no benefit in the common case. The docs/examples could recommend `useMemo` or `useState` for the token addresses array, or the hook could use a serialized key for comparison.

packages/sdk/src/query/utils.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  return rest;
+}
+
+/**
+ * Stable hash function for query keys.
+ * Sorts object keys recursively and converts bigint values to strings.
+ *
+ * @remarks
+ * bigint values are serialized as decimal strings, so `42n` and `"42"` hash to
```
ghermet, line 113: [AUTOMATED] FYI: `bigint` and string collision in `hashFn`.

The comment on L113 correctly documents that `42n` and `"42"` produce the same hash. Just noting this is fine for the current query-key conventions (keys don't mix bigint and string in the same position), but could be surprising if someone adds a query key where the distinction matters.

The explicit documentation here is appreciated.

ghermet (commented):
[AUTOMATED]

## Code Review Summary

This is a well-structured refactoring that cleanly extracts TanStack Query logic from React hooks into a framework-agnostic SDK layer. The architecture is sound, the patterns are consistent, and the test coverage is thorough.

### Architecture

The layer extraction is clean: SDK provides `*QueryOptions` / `*MutationOptions` factories, React hooks spread them and add React-specific concerns (optimistic UI, `queryKeyHashFn`). The `zamaQueryKeys` namespace with address-checksum normalization solves a real class of cache-miss bugs.

### No blocking issues found

Checked for: query key mismatches, invalidation correctness, optimistic update/rollback logic, callback chain integrity, error handling, and guideline compliance. The code is solid.

### Non-blocking observations (Nit)

See inline comments below.

packages/react-sdk/src/token/use-confidential-balance.ts
enitrat: see new `useQuery` util

packages/react-sdk/src/token/use-revoke.ts
enitrat: also has been fixed as per last commit

packages/react-sdk/src/token/use-confidential-balances.ts
enitrat: unrelated to our changes here, so I will disregard for now

ghermet (approved):

PR #67, opened 2026-03-06
refactor: infer contract reads from abi, cleanup duplications, use proper typing

enitrat:
# Summary

Replaces the loosely-typed `ContractCallConfig` / `readContract<T>()` pattern with ABI-inferred generics, so TypeScript derives return types automatically. Alongside this, eliminates duplicated type definitions by re-exporting them from viem and the relayer SDK, and introduces a proper `Handle` branded type to replace ad-hoc `string` / `Address` usage for ciphertext handles.

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] CI/workflows
- [ ] Docs

### 1. ABI-inferred contract read/write types

`GenericSigner.readContract` and `writeContract` (and all implementations: `EthersSigner`, `ViemSigner`, `WagmiSigner`) previously required callers to supply an explicit return-type generic:

```ts
// before
const balance = await signer.readContract<bigint>(confidentialBalanceOfContract(...));
```

The generics are now inferred from the ABI, so callers no longer need manual annotations and TypeScript will catch mismatches between the ABI and the call site:

```ts
// after — return type is inferred from ABI
const balance = await signer.readContract(confidentialBalanceOfContract(...));
```

This required splitting the single `ContractCallConfig` into `ReadContractConfig<TAbi, TFunctionName, TArgs>` and `WriteContractConfig<TAbi, TFunctionName, TArgs>` with the corresponding helper types `ReadFunctionName`, `WriteFunctionName`, `ReadContractArgs`, `WriteContractArgs`, `ReadContractReturnType`, all built on top of viem's `ContractFunctionName` / `ContractFunctionArgs` / `ContractFunctionReturnType`. `ContractCallConfig` is kept as a `ReadContractConfig | WriteContractConfig` union for backwards compatibility.

### 2. Branded `Handle` type for ciphertext handles

FHE ciphertext handles are 32-byte hex values (`bytes32` on-chain). Previously they were typed as `string`, `Address` (40-byte, semantically wrong), or bare `bigint` interchangeably. A `Handle = SDK.Bytes32Hex` branded type is now threaded through the whole codebase:

- `confidentialBalanceOf()` returns `Handle`
- `decryptHandles()`, `decryptBalance()` accept/return `Handle`
- `UnwrapRequestedEvent.encryptedAmount`, `FinalizeUnwrapParams.burnAmountHandle`, `BatchTransferData.encryptedAmount` all use `Handle`
- All decrypt hooks (`useUserDecrypt`, `usePublicDecrypt`, `useDelegatedUserDecrypt`, `useUserDecryptFlow`, `useUserDecryptedValue`, `useUserDecryptedValues`) use `Handle` for keys and `ClearValueType` for values

A `normalizeHandle(value: string | bigint): Handle` utility was added to canonicalise handles coming from different providers (some return a `bigint`, some a hex string).

### 3. Deduplication of SDK types

Several types were hand-rolled duplicates:
- **EIP-1193 provider types** (`ProviderConnectInfo`, `ProviderMessage`, `ProviderRpcError`, `EIP1193EventMap`, `EIP1193Events`, `EIP1193Provider`) are now re-exported from `viem` instead of being maintained locally.
- **`DecryptedValue`** → replaced by `SDK.ClearValueType` from the relayer SDK.
- **`FHEKeypair`** → replaced by `SDK.KeypairType`.
- **`FheType`** → replaced by `SDK.FheTypeName`.
- **`PublicDecryptResult`** is now derived from `SDK.PublicDecryptResults` using `Omit` + override rather than being redefined.
- **`EncryptInput`** is now a proper discriminated union (value type depends on FHE type), e.g. `eaddress` requires an `Address`, `ebool` requires `boolean | bigint`.
- `Hex` type (`0x${string}`) is now re-exported from the relayer SDK bundle instead of being redefined.

### 4. Minor fixes

- `useShield` and `useConfidentialTransfer` had a bug in `onMutate`: they always returned `config.tokenAddress` as the context even when the caller supplied their own `onMutate`, ignoring the caller's return value. This is now fixed.
- `EthersSigner.signTypedData` converts readonly `types` to mutable before passing to ethers (ethers requires a mutable array).
- `gas` override was previously ignored in `EthersSigner.writeContract`; it is now forwarded as `gasLimit`.
- `react-sdk/tsconfig.json` adds path aliases for `@zama-fhe/sdk` → local source so type-checking resolves to the workspace package.

## Validation

```
pnpm typecheck
pnpm lint
pnpm test:coverage
```


packages/sdk/src/relayer/cleartext/cleartext-fhevm-instance.ts
```diff
@@ -313,54 +313,56 @@ export class CleartextFhevmInstance implements RelayerSDK {
...
 
   async #isHandleDelegatedForUserDecryption(
-    delegator: string,
-    delegate: string,
-    contractAddress: string,
-    handle: Hex,
+    delegator: Address,
+    delegate: Address,
+    contractAddress: Address,
+    handle: `0x${string}`,
```
ghermet: `Handle`?

ghermet (dismissed):

vitest.config.ts
```diff
@@ -5,7 +5,7 @@ export default defineConfig({
   test: {
     environment: "jsdom",
     include: ["packages/**/*.test.{ts,tsx}"],
-    exclude: ["**/integration.test.ts"],
+    exclude: ["**/integration.test.ts", "**/node_modules/**"],
```
ghermet, line 8: why is `"**/integration.test.ts"` excluded? it isn't a playwright test?

enitrat: Generally integration tests are ran outside of the default unit test as they require external setup (like anvil nodes), which makes me realize that it's not an integration test as it's mocking everything.
- as an integration test, we want it to run against a real anvil node
- we also need to have a separate entrypoint for integration, like we do for e2e
- but for now none of these are available.

enitrat: let's for now add a test:integration target in package.json. Then, in a followup PR we'll write real integration tests. And we'll wire a vite config specific for integration.

ghermet: integration tests can be done with playwright, I can't really see what sort of integration you'd do with vite

ghermet: unless you find a way to spin a hardhat node process before the tests?

ghermet: wouldn't it be easier to make an integration folder? it'll save you writing `*.integration.*.ts` for each file

enitrat: > unless you find a way to spin a hardhat node process before the tests?

yes, that's the idea. you just spawn a node as part of vitests's global setup for integration tests!

enitrat: we should only be using playwright to check via UI elements ; not for inner logic tests

ghermet (approved):

vitest.config.ts
ghermet: agreed, massive overkill

ghermet: Think I can make  a test fixture for integration tests which spawns a node, snapshots the state before and reverts after the test just like playwright

PR #72, opened 2026-03-06
feat: add delegated decryption support [SDK-12]

ghermet:
## Summary

- Add ACL contract integration with `aclAddress` config and delegation read/write methods (allow/disallow/query)
- Implement `decryptBalanceAs` for delegated balance decryption in the SDK
- Add `useDelegateDecryption` and `useDecryptBalanceAs` React hooks
- Add DelegationPanel test component and full e2e test coverage with Playwright

## Test plan

- [x] Unit tests for delegation SDK methods pass (`pnpm test`)
- [x] E2e delegation tests pass (`pnpm playwright`)
- [x] Verify `useDelegateDecryption` hook allows/disallows/queries delegation
- [x] Verify `useDecryptBalanceAs` decrypts a delegated balance correctly


packages/playwright/fixtures/fhevm.ts
```diff
@@ -14,10 +14,10 @@ const hardhat = {
   gatewayChainId: GATEWAY_CHAIN_ID,
   rpcUrl: "http://127.0.0.1:8545",
   contracts: {
```
ankurdotb: Do you wanna hardcode `rpcUrl` or get it from a variable?

docs/gitbook/src/guides/sdk/delegation.md
```diff
@@ -0,0 +1,128 @@
...
+## Setup
+
+Pass `aclAddress` when creating the SDK. All delegation methods throw `ConfigurationError` without it.
+
+```ts
+const sdk = new ZamaSDK({
+  relayer,
+  signer,
+  storage,
+  aclAddress: "0xACL", // required for delegation
```
ankurdotb: What would be the `aclAddress` when giving delegated access to a portfolio of 2-3 confidential tokens?

Actual example would help I think. I think there's a chance that maybe this term/variable name could be misleading.

docs/gitbook/src/reference/sdk/delegation.md
```diff
@@ -0,0 +1,128 @@
...
+
+The token owner calls `delegateDecryption` to allow a delegate to decrypt their balance for a specific token.
+
+```ts
+// Permanent delegation (no expiration)
+await token.delegateDecryption("0xDelegate");
+
+// Delegation with an expiration date
+await token.delegateDecryption("0xDelegate", {
+  expirationDate: new Date("2025-12-31T00:00:00Z"),
```
ankurdotb: BTW, what format of date does the protocol accept?

- Is it epoch time?
- Or this type of XMLDateTime?
- What happens if someone gives a value not normalised to UTC? Do we then normalise to UTC?

docs/gitbook/src/guides/sdk/delegation.md
```diff
@@ -0,0 +1,128 @@
...
+```ts
+// Check if a delegation is active
+const delegated = await readonlyToken.isDelegated("0xDelegator", "0xDelegate");
+// true if delegation exists and hasn't expired
+
+// Get the raw expiry timestamp
+const expiry = await readonlyToken.getDelegationExpiry("0xDelegator", "0xDelegate");
+// 0n = no delegation
+// 2^64 - 1 = permanent
+// otherwise = Unix timestamp (seconds)
```
ankurdotb: If we're TAKING in time in XmlDateTime, maybe giving UTC normalised XmlDateTime makes more sense than epoch (keep it consistent?)

packages/sdk/src/token/readonly-token.ts
```diff
@@ -62,6 +68,8 @@ export interface ReadonlyTokenConfig {
   credentials?: CredentialsManager;
   /** Address of the confidential token contract. */
   address: Address;
+  /** ACL contract address. Required for delegation methods. */
+  aclAddress?: Address;
```
enitrat: should we make it mandatory instead? so that we require an ACL address to be passed alongside token instanciation. This would avoid surprises at runtime. Normally, we have those addresses for all FHEVM deployments, so I don't really see an issue with making this mandatory.

We could get rid of all `require` here and just use the chainConfig

packages/sdk/src/token/readonly-token.ts
```diff
@@ -519,6 +578,116 @@ export class ReadonlyToken {
...
+   * @throws {@link DecryptionFailedError} if delegated decryption fails.
+   *
+   * @example
+   * ```ts
+   * const balance = await token.decryptBalanceAs("0xDelegator");
+   * ```
+   */
+  async decryptBalanceAs(delegator: Address, options?: { owner?: Address }): Promise<bigint> {
+    const normalizedDelegator = validateAddress(delegator, "delegator");
+    const owner = options?.owner ? validateAddress(options.owner, "owner") : normalizedDelegator;
```
enitrat: nit: prefer `getAddress`

docs/gitbook/src/guides/sdk/delegation.md
ghermet: I might actually refactor this as the contract address for acl is on the Mainnet,Sepolia config

packages/sdk/src/token/readonly-token.ts
```diff
@@ -519,6 +578,116 @@ export class ReadonlyToken {
...
+    if (this.isZeroHandle(handle)) return 0n;
+
+    // Check persistent cache keyed by balance owner.
+    const cached = await loadCachedBalance({
+      storage: this.storage,
+      tokenAddress: this.address,
+      owner,
+      handle,
+    });
+    if (cached !== null) return cached;
```
enitrat: what's the cache invalidation strategy? is it handled via the global TTL setting ? How do we make sure that the value has not changed (e.g. user interacted in another app) and the cache is still correct ?

packages/sdk/src/token/zama-sdk.ts
```diff
@@ -43,6 +43,8 @@ export interface ZamaSDKConfig {
   sessionTTL?: number;
   /** Optional structured event listener for debugging and telemetry. Never receives sensitive data. */
   onEvent?: ZamaSDKEventListener;
+  /** ACL contract address. Required for delegation methods on tokens created by this SDK. */
+  aclAddress?: Address;
```
enitrat: same remark as for token; can this be read from a chainConfig ?somehow?

enitrat (commented):
ok to me just a few comments on cache mechanism and where we get ACL from

docs/gitbook/src/guides/sdk/delegation.md
ankurdotb: Is this the single protocol-level ACL? Then yes, user's should not need to set this.

packages/sdk/src/token/zama-sdk.ts
ankurdotb: +1 to this

packages/sdk/src/token/readonly-token.ts
```diff
@@ -509,6 +516,54 @@ export class ReadonlyToken {
...
 
+  protected async requireAclAddress(): Promise<Address> {
+    try {
+      return await this.relayer.getAclAddress();
+    } catch {
+      throw new ConfigurationError(
+        "Could not resolve ACL address from the relayer. Ensure the relayer is configured with a transport that includes aclContractAddress.",
+      );
+    }
+  }
```
enitrat: still required now that mandatory?

packages/sdk/src/token/readonly-token.ts
```diff
@@ -509,6 +516,54 @@ export class ReadonlyToken {
...
+  async isDelegated(delegator: Address, delegate: Address): Promise<boolean> {
+    const expiry = await this.getDelegationExpiry(delegator, delegate);
+    if (expiry === 0n) return false;
+    // Permanent delegation (uint64 max) — skip the RPC round-trip for block timestamp.
+    if (expiry === 2n ** 64n - 1n) return true;
+    const now = this.signer.getBlockTimestamp
+      ? await this.signer.getBlockTimestamp()
+      : BigInt(Math.floor(Date.now() / 1000));
+    return expiry >= now;
+  }
```
enitrat: all of our signers support `getBlockTimestamp`:
- remove checks on whether it is present or not
- remove optionality of `getBlockTimestamp` in GenericSigner, so that all signers are **required** to support it
- make sure that we properly adapted all cases / docs s.t getBlockTimestamp is considered as present and we have no silent fallbacks.

packages/sdk/src/token/readonly-token.ts
```diff
@@ -519,6 +574,116 @@ export class ReadonlyToken {
...
+      );
+
+      // Sign the delegated EIP-712 with the delegate's signer.
+      // Convert KMS types (bigint chainId, string timestamps) to EIP712TypedData format.
+      const chainId = Number(delegatedEIP712.domain.chainId);
+      if (!Number.isSafeInteger(chainId)) {
+        throw new ConfigurationError(
+          `chainId ${delegatedEIP712.domain.chainId} exceeds safe integer range`,
+        );
+      }
```
enitrat: sounds superfluous here knowing that the chainId will be one of a zama-supported chains

- remove all chainId boundary checks
- considered useless, as we'll only sign for valid zama chains OR mock chains with proper chainId as per ethereum spec which are < 2**64

packages/playwright/fixtures/fhevm.ts
ghermet: It already reads from `process.env.HARDHAT_RPC_URL` with a default fallback (`"http://127.0.0.1:8545"`), so it's not fully hardcoded. Happy to extract it further if you think that's not enough.

docs/gitbook/src/guides/sdk/delegation.md
ghermet: Done — `aclAddress` is now resolved automatically from the relayer transport config. Users no longer pass it manually. Updated docs and removed it from all config examples.

docs/gitbook/src/reference/sdk/delegation.md
ghermet: The SDK accepts a standard JS `Date` object and converts it to a UTC Unix timestamp (seconds since epoch) before sending on-chain. Since `Date.getTime()` always returns UTC milliseconds regardless of local timezone, no manual normalization is needed. Documented this in the delegation guide with examples showing different timezone inputs all producing the same on-chain value.

docs/gitbook/src/guides/sdk/delegation.md
ghermet: We take a `Date` in and store epoch on-chain, so it's consistent on the input side (JS-native) and the storage side (uint64 seconds). The docs now clarify this flow.

packages/sdk/src/token/readonly-token.ts
ghermet: Done — `aclAddress` is now resolved from the relayer transport config automatically (network presets already include it). No user-facing prop needed anymore, so no runtime surprise.

packages/sdk/src/token/readonly-token.ts
ghermet: Done in 261de86.

packages/sdk/src/token/readonly-token.ts
ghermet: The cache key includes the encrypted handle: `(token, owner, handle)`. When the on-chain balance changes (e.g. user interacted in another app), the contract returns a new handle, so the old cache entry is never hit — it's an automatic miss. No TTL or manual invalidation needed. Added inline comments and documented this in the delegation guide.

packages/sdk/src/token/zama-sdk.ts
ghermet: Done — resolved from the relayer transport config now, same as for Token.

packages/sdk/src/token/readonly-token.ts
ghermet: You're right this is overkill

packages/sdk/src/token/readonly-token.ts
ghermet: `getBlockTimestamp` is now mandatory on the GenericSigner interface

packages/sdk/src/token/readonly-token.ts
ghermet: aclAddress is always available as part of the default configs but who knows what would happen when they provide their own custom ones

packages/sdk/src/contracts/acl.ts
enitrat, line 1: seeing a hard constant on FHE_GAS_LIMIT makes me wonder what it's there for. gas estimations should always be delegated to wallets, no ?

packages/sdk/src/token/token.ts
```diff
@@ -673,6 +675,139 @@ export class Token extends ReadonlyToken {
...
+    expirationDate,
+  }: {
+    delegateAddress: Address;
+    expirationDate?: Date;
+  }): Promise<TransactionResult> {
+    const acl = await this.getAclAddress();
+    // uint64 max → no practical expiry
+    const expDate = expirationDate
+      ? BigInt(Math.floor(expirationDate.getTime() / 1000))
+      : 2n ** 64n - 1n;
```
enitrat: nit: replace  `2n ** 64n - 1n` with constants

enitrat (commented):
looks good overall; launching agentic review in //

packages/sdk/src/query/delegation-status.ts
```diff
@@ -0,0 +1,59 @@
...
+const PERMANENT = 2n ** 64n - 1n;
+
+function deriveDelegationFromExpiry(expiryTimestamp: bigint) {
+  if (expiryTimestamp === 0n) {
+    return false;
+  }
+  if (expiryTimestamp === PERMANENT) {
+    return true;
+  }
+  const nowSeconds = BigInt(Math.floor(Date.now() / 1000));
```
enitrat: [AUTOMATED] **Major: Client clock vs chain clock inconsistency**

`deriveDelegationFromExpiry()` uses `Date.now()` to determine if a delegation is active:

```ts
const nowSeconds = BigInt(Math.floor(Date.now() / 1000));
```

Meanwhile, `ReadonlyToken.isDelegated()` (added in this same PR) correctly uses the chain clock:

```ts
const now = await this.signer.getBlockTimestamp();
```

The query layer (which feeds `useDelegationStatus()`) will report incorrect delegation status whenever the client clock is skewed from the chain. Since `isDelegated()` already exists and uses the correct time source, the query factory should either:
1. Call `readonlyToken.isDelegated()` directly, or
2. Accept a `getBlockTimestamp` callback and use chain time

Confidence: 95/100

packages/react-sdk/src/token/use-delegate-decryption.ts
```diff
@@ -0,0 +1,33 @@
...
+export function useDelegateDecryption(
+  config: UseZamaConfig,
+  options?: UseMutationOptions<TransactionResult, Error, DelegateDecryptionParams>,
+) {
+  const token = useToken(config);
+
+  return useMutation<TransactionResult, Error, DelegateDecryptionParams>({
+    ...delegateDecryptionMutationOptions(token),
+    ...options,
+  });
```
enitrat, line 37: [AUTOMATED] **Major: Missing cache invalidation after delegation**

After a successful `delegateDecryption` transaction, this hook does not invalidate `zamaQueryKeys.delegationStatus`. A mounted `useDelegationStatus()` subscriber will stay stale until an unrelated refetch occurs.

Every other mutation hook in the codebase follows this pattern (e.g. `use-allow.ts`):

```ts
  context.client.invalidateQueries({ queryKey: zamaQueryKeys.delegationStatus.all });
},
```

This hook should add an `onSuccess` handler that:
1. Forwards to the caller's `onSuccess` if provided
2. Invalidates `zamaQueryKeys.delegationStatus.all`

Confidence: 97/100

packages/react-sdk/src/token/use-revoke-delegation.ts
```diff
@@ -0,0 +1,30 @@
...
+export function useRevokeDelegation(
+  config: UseZamaConfig,
+  options?: UseMutationOptions<TransactionResult, Error, RevokeDelegationParams>,
+) {
+  const token = useToken(config);
+
+  return useMutation<TransactionResult, Error, RevokeDelegationParams>({
+    ...revokeDelegationMutationOptions(token),
+    ...options,
+  });
```
enitrat, line 37: [AUTOMATED] **Major: Missing cache invalidation after revocation**

Same issue as `useDelegateDecryption` — after a successful revocation, `useDelegationStatus()` subscribers will keep showing the delegation as active because the query cache is never dirtied.

Add an `onSuccess` handler that invalidates `zamaQueryKeys.delegationStatus.all` (same pattern as the other mutation hooks in the codebase).

Confidence: 97/100

packages/sdk/src/token/readonly-token.ts
```diff
@@ -520,6 +569,117 @@ export class ReadonlyToken {
...
+        signature,
+        delegatorAddress: normalizedDelegator,
+        delegateAddress,
+        startTimestamp,
+        durationDays,
+      });
+
+      this.emit({ type: ZamaSDKEvents.DecryptEnd, durationMs: Date.now() - t0 });
+
+      const value = (result[handle] as bigint | undefined) ?? 0n;
```
enitrat: [AUTOMATED] **Medium: Silent 0n coercion for missing relayer payload entry**

In `decryptBalanceAs()`, after the `delegatedUserDecrypt` call:

```ts
const value = (result[handle] as bigint | undefined) ?? 0n;
```

If the relayer returns an incomplete payload (handle not found in result), the SDK silently returns `0n` and persists it to the balance cache. This masks potential relayer/protocol bugs — downstream consumers see a zero balance with no indication that decryption actually failed.

Consider throwing a `DecryptionFailedError` when the handle is absent from the result, similar to how other error paths in this file are handled via `wrapDecryptError`.

Confidence: 88/100

packages/sdk/src/query/delegation-status.ts
```diff
@@ -0,0 +1,59 @@
...
+  ReturnType<typeof zamaQueryKeys.delegationStatus.scope>
+> {
+  return {
+    ...filterQueryOptions(config.query ?? {}),
+    queryKey: zamaQueryKeys.delegationStatus.scope(
+      readonlyToken.address,
+      config.delegatorAddress,
+      config.delegateAddress,
+    ),
+    queryFn: async () => {
```
enitrat: [AUTOMATED] **Medium: `queryFn` captures params from closure instead of queryKey**

Per TanStack Query best practices, `queryFn` should extract parameters from `context.queryKey` rather than closing over them. This guarantees cache consistency — the function fetches for exactly the parameters that determined which cache entry to use.

Current pattern (closure):
```ts
queryFn: async () => {
  const expiryTimestamp = await readonlyToken.getDelegationExpiry({
    delegatorAddress: config.delegatorAddress,
    delegateAddress: config.delegateAddress,
  });
```

Recommended pattern (extract from key):
```ts
queryFn: async (context) => {
  const [, , delegatorAddress, delegateAddress] = context.queryKey;
  const expiryTimestamp = await readonlyToken.getDelegationExpiry({
    delegatorAddress,
    delegateAddress,
  });
```

This is a lower-severity issue since TanStack Query creates new observers when keys change, but it's a deviation from the pattern that protects against stale closures.

Confidence: 82/100

enitrat (changes requested):
[AUTOMATED]

## Code Review — Delegated Decryption

Reviewed the delegation feature against codebase patterns and TanStack Query best practices. Found 4 actionable issues (2 major, 2 medium) and also validated findings from a prior automated review — one of those findings (Finding 9 about `wit(...)` being a typo for `it(...)`) is a **false positive** (`wit` is a valid vitest fixture extension at line 31 of wagmi-signer-subscribe.test.ts).

### Summary

| Severity | Count | Description |
|----------|-------|-------------|
| Major | 2 | Missing cache invalidation in delegate/revoke hooks |
| Major | 1 | `Date.now()` vs chain clock inconsistency |
| Medium | 1 | Silent 0n coercion in `decryptBalanceAs` |
| Medium | 1 | `queryFn` captures params from closure instead of queryKey |

### Note on controlled override surface

The new mutation hooks accept raw `UseMutationOptions` (allowing callers to override `mutationFn`/`mutationKey`). Per TanStack best practices, the type should be `Omit<MutationOptions, 'mutationFn' | 'mutationKey'>`. However, **every existing mutation hook in the codebase uses the same pattern**, so this is a pre-existing convention, not a regression introduced by this PR. Worth tracking as a follow-up but not blocking.

packages/sdk/src/contracts/acl.ts
ghermet: We added this limit because Corentin had gas issues when the sdk with openfort. Maybe these issues were isolated. Maybe we won't need the hard cap on the gas

packages/sdk/src/query/delegation-status.ts
enitrat: @ghermet can you check this one?

packages/sdk/src/token/readonly-token.ts
enitrat: and this one ?

packages/sdk/src/contracts/acl.ts
ghermet: Good catch — this isn't about estimating gas, it's about *bypassing* MetaMask's internal safety cap (2^24 = ~16.7M). FHE operations on Sepolia routinely estimate ~21M gas, which exceeds that cap and causes MetaMask to reject the tx outright. Providing an explicit `gas` value skips the estimate and lets the wallet submit it. Actual consumption is typically under 2M.

That said, this isn't new to this PR — `FHE_GAS_LIMIT` is an existing constant used across all FHE contract calls. Happy to revisit the approach (e.g. per-call estimation with a multiplier) in a separate issue if you think it's worth it.

packages/sdk/src/token/token.ts
ghermet: Already using `MAX_UINT64` from `contracts/constants.ts` — let me double check there's no raw `2n ** 64n - 1n` remaining in token.ts. If there is I'll swap it out.

packages/sdk/src/query/delegation-status.ts
ghermet: Good catch — this is already fixed. The query factory now uses `readonlyToken.signer.getBlockTimestamp()` (chain time) instead of `Date.now()`. You can see it in the current `delegation-status.ts`: expiry is compared against chain block timestamp for non-trivial cases (0n → false, MAX_UINT64 → true, otherwise → compare against chain time).

packages/react-sdk/src/token/use-delegate-decryption.ts
ghermet: Already addressed — `useDelegateDecryption` now has an `onSuccess` handler that forwards to the caller's callback and invalidates `zamaQueryKeys.delegationStatus.all`. You can see it in the current code.

packages/react-sdk/src/token/use-revoke-delegation.ts
ghermet: Same as above — already fixed. `useRevokeDelegation` now invalidates `zamaQueryKeys.delegationStatus.all` on success.

enitrat (dismissed):

ankurdotb (commented):
- Functionally, the spec doc looks good. Review/consider whether it matches the new structure of docs and any Gitbook guidance (see discussion in #81)
- Merge conflict flagged, hence any approval will get dismissed on resolving as the approval will be stale. I can approve after merge conflicts are resolved.

packages/sdk/src/token/__tests__/credentials-manager.test.ts
```diff
@@ -992,8 +992,10 @@ describe("storeKey caching", () => {
     await manager.allow(TOKEN_A);
     expect(computeSpy).toHaveBeenCalledOnce();
 
+    // Change to the same address but with different casing (normalized form is 0x2222222222222222222222222222222222222222)
+    // Test that getAddress normalization means casing change doesn't invalidate cache
     vi.mocked(signer.getAddress).mockResolvedValue(
-      "0x2222222222222222222222222222222222ABCDEF" as Address,
+      getAddress("0x2222222222222222222222222222222222222222"),
```
enitrat: not sure what you want to test here but given that this address is all numbers, you're not testing the address casing

ghermet: true, this is probably was added when there was this getAddress issue you solved.

ankurdotb:
# PR #72 Review Summary — "feat: add delegated decryption support"

**Branch:** `feat/delegated-decryption` → `prerelease`
**Scope:** 72 files changed, ~3.4K insertions
**Reviewed by:** code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer

---

## Existing PR Comment Assessment

| Comment | Status |
|---------|--------|
| @ankurdotb: hardcode rpcUrl vs env var | **Resolved** |
| @ankurdotb: aclAddress confusion for multi-token portfolios | **Resolved** — auto-resolved from config |
| @ankurdotb: Date format for delegation expiry | **Resolved** — uses JS Date, stores epoch |
| @enitrat: make aclAddress mandatory | **Resolved** — auto-resolved from config |
| @enitrat: prefer getAddress | **Resolved** |
| @enitrat: cache invalidation strategy for decryptBalanceAs | **Resolved** — cache keyed by handle |
| @enitrat: read aclAddress from chainConfig | **Resolved** |
| @enitrat: remove getBlockTimestamp optionality checks | **Resolved** — now mandatory |
| @enitrat: remove chainId boundary checks | **Resolved** |
| @enitrat: [AUTOMATED] Client clock vs chain clock | **Fixed** — `delegationStatusQueryOptions` uses `getBlockTimestamp`, no `Date.now()` |
| @enitrat: [AUTOMATED] Missing cache invalidation after delegate/revoke | **Fixed** — both hooks call `invalidateQueries` in `onSuccess` |
| @enitrat: replace `2n ** 64n - 1n` with constants | **Partially fixed** — source uses `MAX_UINT64`, test file still uses inline `2n ** 64n - 1n` |
| @enitrat: FHE_GAS_LIMIT hard constant | **Open** — still present |
| @enitrat: `aclAddress` still required check now that mandatory? | **Open** — defensive check kept for custom configs |

---

## Critical Issues (4)

### 1. `getAclAddress()` returns `undefined` cast as `Address` for unsupported chains
**`packages/sdk/src/relayer/relayer-node.ts:249-253`**, **`packages/sdk/src/relayer/relayer-web.ts:381-385`**

```ts
async getAclAddress(): Promise<Address> {
    const chainId = await this.#config.getChainId();
    const config = Object.assign({}, DefaultConfigs[chainId], this.#config.transports[chainId]);
    return config.aclContractAddress as Address;
}
```

When `DefaultConfigs[chainId]` is `undefined` (unsupported chain) AND user transport config doesn't include `aclContractAddress`, returns `undefined as Address`. The old `mergeFhevmConfig` had an explicit guard for unknown chains — that safety net was removed.

**Fix:** Add guard: `if (!config.aclContractAddress) throw new Error(`No ACL address configured for chain ${chainId}`);`

### 2. Documentation uses wrong parameter shapes — will cause runtime errors if copy-pasted
**`packages/react-sdk/README.md`**, **`packages/sdk/README.md`**, **`docs/gitbook/src/reference/sdk/delegation.md`**

Multiple examples use incorrect API signatures:
- README shows `DelegateDecryptionParams { delegate, options?: { expirationDate } }` — actual is `{ delegateAddress, expirationDate? }`
- README shows `DecryptBalanceAsParams { delegator, options?: { owner } }` — actual is `{ delegatorAddress, owner? }`
- SDK README shows `Token.delegateDecryptionBatch(tokens, "0xDelegate")` — actual takes object `{ tokens, delegateAddress }`
- Gitbook error example calls `readonlyToken.decryptBalanceAs("0xDelegator")` — actual takes object `{ delegatorAddress }`

### 3. `decryptBalanceAs` doesn't pre-check delegation status — wastes user signature
**`packages/sdk/src/token/readonly-token.ts:555-654`**

`decryptBalanceAs` proceeds directly to generate keypair, sign EIP-712, and call relayer without checking `isDelegated()`. If delegation expired/revoked, user signs (wallet popup) then gets generic `DecryptionFailedError`. The PR defines `DelegationExpiredError`/`DelegationNotFoundError` but never uses them in the actual decryption flow.

**Fix:** Add pre-flight `isDelegated()` check before the signature step, throwing the specific error classes.

---

## Important Issues (7)

| # | Issue | Location |
|---|-------|----------|
| 5 | **User `onSuccess` throw prevents cache invalidation** — if user callback throws, `invalidateQueries` is never reached, leaving stale delegation status | `packages/react-sdk/src/token/use-delegate-decryption.ts:33-36`, `use-revoke-delegation.ts:33-36` |
| 6 | **`revokeDelegation` missing from `createMockToken` fixture** — SDK-layer tests using `mockToken.revokeDelegation()` will fail | `packages/sdk/src/test-fixtures.ts:291` |
| 7 | **`isDelegated`/`getDelegationExpiry` missing from `createMockReadonlyToken`** — same fixture gap for readonly token | `packages/sdk/src/test-fixtures.ts:104-121` |
| 8 | **Delegation catch blocks drop non-Error causes** — `outcome.reason` may not be an `Error` with `Promise.allSettled`, cause is lost | `packages/sdk/src/token/token.ts:705-710` |
| 9 | **`DecryptionFailedError` gets double-wrapped** — `wrapDecryptError` doesn't check for `DecryptionFailedError`, wraps it in a second instance losing the specific message | `packages/sdk/src/token/readonly-token.ts:784` |
| 10 | **Gitbook error table missing delegation-specific errors** — `DelegationSelfNotAllowedError`, `DelegationCooldownError`, `DelegationNotFoundError`, `DelegationExpiredError` are exported but undocumented | `docs/gitbook/src/reference/sdk/delegation.md:155-158` |
| 11 | **SDK README method table uses wrong signatures** — shows `delegateDecryption(delegate, options?)` but actual API is `delegateDecryption({ delegateAddress, expirationDate? })` | `packages/sdk/README.md:396` |

---

## Test Coverage Gaps

| Priority | Gap | Files |
|----------|-----|-------|
| **Critical** | **No unit test for `delegationStatusQueryOptions` queryFn** — core delegation status derivation logic (zero/MAX_UINT64/timestamp branching, chain clock) untested at query layer | `packages/sdk/src/query/delegation-status.ts:35-53` |
| **Critical** | **No test for "relayer returns undefined for handle"** — distinct failure path in `decryptBalanceAs` where relayer returns success but omits the handle | `packages/sdk/src/token/readonly-token.ts:629-633` |
| **Critical** | **No test for `EthersSigner.getBlockTimestamp`** — Viem and Wagmi signers tested, Ethers is not | `packages/sdk/src/ethers/ethers-signer.ts:136` |
| Important | No unit test for `revokeDelegationMutationOptions` | `packages/sdk/src/query/revoke-delegation.ts` |
| Important | Test uses inline `2n ** 64n - 1n` instead of `MAX_UINT64` constant | `packages/sdk/src/token/__tests__/delegation.test.ts:70,134` |

---

## Suggestions (5)

| # | Suggestion | Location |
|---|-----------|----------|
| 12 | `isDelegated` boundary `expiry >= now` — verify ACL contract uses `>=` not `>` to avoid off-by-one | `packages/sdk/src/token/readonly-token.ts:502` |
| 13 | `getBlockTimestamp` JSDoc minimal across signer implementations — add `@inheritDoc` or brief note | viem-signer.ts, ethers-signer.ts, wagmi-signer.ts |
| 14 | `decryptBalanceAs` JSDoc example uses positional syntax `token.decryptBalanceAs("0xDelegator")` — should be object form | `packages/sdk/src/token/readonly-token.ts` |
| 15 | Gitbook doesn't mention `durationDays = 1` hardcoded for delegated EIP-712 request validity | `docs/gitbook/src/reference/sdk/delegation.md` |
| 16 | Gitbook batch example references `ZamaError` which is `ae-forgotten-export` — needs explicit import path | `docs/gitbook/src/reference/sdk/delegation.md:82` |

---

## Strengths

- **`delegation.test.ts` is thorough** — 16 tests covering reads, writes, error wrapping, ACL resolution failures, permanent delegation short-circuit, caching by owner, batch with partial failure
- **Playwright e2e tests are excellent** — 10 tests covering full lifecycle including time-advance-based expiration, revocation, custom expiry, contract rejection of too-short expiry
- **Chain clock used correctly** — `delegationStatusQueryOptions` uses `getBlockTimestamp` not `Date.now()`, consistent with `isDelegated()` on ReadonlyToken
- **Cache invalidation implemented** — both `useDelegateDecryption` and `useRevokeDelegation` invalidate `delegationStatus` queries in `onSuccess`
- **All 4 delegation error classes tested** — `instanceof`, code, and name verified
- **`MAX_UINT64` constant extracted** — avoids inline magic numbers in source code
- **Permanent delegation optimization** — skips RPC round-trip for `getBlockTimestamp` when `expiry === MAX_UINT64`

---

## Recommended Action

1. **Fix Critical #1** — `getAclAddress()` guard for unsupported chains (security regression)
2. **Fix Critical #2** — Documentation parameter shape mismatches (developer-facing)
3. **Fix Critical #3** — Add pre-flight delegation check in `decryptBalanceAs`
4. **Fix Critical #4** — Update past dates in examples
5. **Fix Important #5** — Reorder `invalidateQueries` before user callback
6. **Fix Important #6-#7** — Complete test fixtures
7. **Fix Important #9** — Add `DecryptionFailedError` to `wrapDecryptError` passthrough
8. **Add tests** — `delegationStatusQueryOptions` queryFn, `EthersSigner.getBlockTimestamp`, "no handle" path

---

🤖 *Reviewed with [Claude Code](https://claude.com/claude-code) using code-reviewer, silent-failure-hunter, pr-test-analyzer, and comment-analyzer agents*

ankurdotb (changes requested):
I think the main ones I care about would be the ones under "critical" section, especially 2 and 3 as that feels like docs misalignment. The rest, I leave you to check.

packages/react-sdk/src/token/use-delegation-status.ts
```diff
@@ -0,0 +1,57 @@
...
+ * ```tsx
+ * const { data } = useDelegationStatus({
+ *   tokenAddress: "0xToken",
+ *   delegatorAddress: "0xDelegator",
+ *   delegateAddress: "0xDelegate",
+ * });
+ * // data?.isDelegated, data?.expiryTimestamp
+ * ```
+ */
+export function useDelegationStatus(config: UseDelegationStatusConfig) {
```
enitrat: [AUTOMATED] **Structural inconsistency (Low):** This hook accepts no `options` parameter, unlike peer query hooks:

- `useWrapperDiscovery(config, options?)` — accepts `Omit<UseQueryOptions<...>, 'queryKey' | 'queryFn'>`
- `useConfidentialIsApproved(config, options?)` — same pattern
- `useUnderlyingAllowance(config, options?)` — same pattern

This breaks the convention and prevents consumers from passing `refetchInterval` or `staleTime` — which matters here because `isDelegated` is derived from time and can go stale after natural expiry (see body comment).

**Suggested signature:**
```ts
export function useDelegationStatus(
  config: UseDelegationStatusConfig,
  options?: Omit<UseQueryOptions<DelegationStatusData, Error>, 'queryKey' | 'queryFn'>,
)
```

Confidence: 92/100

docs/gitbook/src/reference/sdk/delegation.md
```diff
@@ -0,0 +1,160 @@
...
+```ts
+// Decrypt the delegator's balance (owner defaults to delegator)
+const balance = await readonlyToken.decryptBalanceAs({
+  delegatorAddress: "0xDelegator",
+});
+
+// Decrypt a specific owner's balance (when owner differs from delegator)
+const balance = await readonlyToken.decryptBalanceAs({
+  delegatorAddress: "0xDelegator",
+  owner: "0xOwner",
```
enitrat, line 137: [AUTOMATED] **Nit (docs):** This example shows `owner` differing from `delegatorAddress`. Worth noting that the relayer validates ACL permissions against `delegatorAddress` (not `owner`) — see `userDecrypt.ts:340` in `relayer-sdk`. So this only works if the delegator is already ACL-authorized on the owner's encrypted handle (e.g. an approved operator).

The common case is `owner === delegator`. A brief note would prevent users from assuming any delegate can read any owner's balance via any delegator.

Confidence: 85/100

enitrat (commented):
[AUTOMATED]

## Code Review — Delegated Decryption

Overall this is a well-structured feature. The SDK layer, query layer, React hooks, and tests all follow established patterns. Address normalization via `getAddress` is consistently applied at public entrypoints. The ACL contract integration, batch delegation, and cache-by-owner strategy are sound.

I found **1 real bug**, **1 cache-correctness concern**, and **2 structural/docs issues**. Details below.

### What looks good

- `getAddress()` normalization at all public entrypoints
- Query keys use the existing `zamaQueryKeys` pattern with proper address normalization
- Mutation hooks follow the `onSuccess → invalidateQueries` pattern consistently
- Test coverage is thorough — unit tests, React hook tests, E2E Playwright tests
- Batch delegation with `Promise.allSettled` for partial-success semantics
- Cache keyed by `(token, owner, handle)` — stale entries never served

---

### Finding 1 — Medium: Browser worker drops HTTP status for delegated decrypt errors

**File:** `packages/sdk/src/worker/relayer-sdk.worker.ts:587` (not in diff — existing file)

`handleDelegatedUserDecrypt` does not call `extractHttpStatus(error)` before `sendError()`, unlike `handleUserDecrypt` at line 403 which extracts and passes `statusCode`.

This means relayer HTTP errors (400 for `not_ready_for_decryption`, 429 for rate limiting, 503 for service unavailable) from `/delegated-user-decrypt` are lost before they reach `wrapDecryptError()` in `readonly-token.ts:784`. The SDK always throws a generic `DecryptionFailedError` instead of the more specific `NoCiphertextError` (400) or `RelayerRequestFailedError` (with status code).

**Fix (2 lines):**
```ts
// relayer-sdk.worker.ts, line 584-587
const statusCode = extractHttpStatus(error);
console.error('[Worker] DelegatedUserDecrypt error:', message);
sendError(id, type, message, statusCode);
```

Note: the node worker (`relayer-sdk.node-worker.ts:323`) has the same omission, but it also doesn't extract status for regular `userDecrypt` — so at least it's consistently missing there.

---

### Finding 2 — Low-Medium: `delegationStatus` query caches time-derived boolean with no refresh

**File:** `packages/sdk/src/query/delegation-status.ts:35` (not in diff — new file added by PR)

The `queryFn` computes `isDelegated` from `expiryTimestamp >= blockTimestamp`. Once cached by TanStack Query, this boolean can go stale: if a component mounts while the delegation is active (expiry in 10 minutes), the cached `isDelegated: true` persists after expiry until something invalidates the query.

Mutation hooks handle grant/revoke invalidation, but **natural expiry** (time passing) has no trigger.

Consider:
1. `refetchInterval` derived from time-to-expiry
2. Exposing `query` options in the React hook (see finding 3)
3. Documenting the limitation

enitrat:
@ghermet can you check this one (idk why, wasnt properly posted):

> Finding 1 — Medium: Browser worker drops HTTP status for delegated decrypt errors
> File: packages/sdk/src/worker/relayer-sdk.worker.ts:587 (not in diff — existing file)
> 
> handleDelegatedUserDecrypt does not call extractHttpStatus(error) before sendError(), unlike handleUserDecrypt at line 403 which extracts and passes statusCode.
> 
> This means relayer HTTP errors (400 for not_ready_for_decryption, 429 for rate limiting, 503 for service unavailable) from /delegated-user-decrypt are lost before they reach wrapDecryptError() in readonly-token.ts:784. The SDK always throws a generic DecryptionFailedError instead of the more specific NoCiphertextError (400) or RelayerRequestFailedError (with status code).
> 
> Fix (2 lines):
> 
> // relayer-sdk.worker.ts, line 584-587
> const statusCode = extractHttpStatus(error);
> console.error('[Worker] DelegatedUserDecrypt error:', message);
> sendError(id, type, message, statusCode);
> Note: the node worker (relayer-sdk.node-worker.ts:323) has the same omission, but it also doesn't extract status for regular userDecrypt — so at least it's consistently missing there.

ankurdotb:
## PR Review v4.1 — Delegation-Specific Findings Only

Filtered to issues directly related to the delegated decryption feature.

---

### 🔴 Critical

**C-1. Spec doc inverts `sessionTTL: 0` semantics**
`docs/superpowers/specs/2026-03-12-batch-decrypt-balances-as-design.md:84` — Claims `sessionTTL: 0` means "never expire". Actual behavior in `session-signatures.ts:74`: `ttl === 0` returns `true` (= expired → always re-sign). Will mislead developers.

---

### 🟡 Important

**I-1. `batchDecryptBalancesAs` skips pre-flight delegation check**
`decryptBalanceAs` validates delegation before signing (lines 717-736), but `batchDecryptBalancesAs` → `#batchDecryptCore` does not. A batch call with an expired delegation wastes a wallet signature before failing.

**I-2. `onError` callback not wrapped in try-catch**
`readonly-token.ts:438` — In `#batchDecryptCore`, if the user-provided `onError` callback throws, it propagates unhandled. Wrap in try-catch.

**I-3. `saveCachedBalance` in batch `.then()` lacks try-catch**
`readonly-token.ts:427-433` — If cache write throws, it's reported as a *decryption* failure. The balance was decrypted successfully — guard the side effect.

**I-4. Non-existent `token.on()` event API in docs**
`docs/gitbook/src/reference/sdk/delegation.md:195` — `Token`/`ReadonlyToken` has no public `.on()` method. Example will fail at runtime.

**I-5. `wrapSigningError` drops non-Error causes**
`errors.ts:209-210` — When `error` is not an `Error` instance, `cause` is `undefined`. Affects delegation catch blocks in `token.ts:711-721, 753-755`. Fix: always pass `cause: error`.

**I-6. Delegation catch blocks in `token.ts` drop non-Error causes**
`token.ts:711-721, 753-755` — Same pattern as above, compounds with `wrapSigningError`.

---

### 🟠 Documentation & Tests

**D-1. React hooks called inside `.map()` — violates Rules of Hooks**
`docs/gitbook/src/reference/react/useBatchDecryptBalancesAs.md:33` — `useReadonlyToken(addr)` inside `.map()` will crash at runtime.

**D-2. Past dates in delegation examples**
`docs/gitbook/src/reference/sdk/delegation.md` — Examples use `2025-12-31`. With past-date validation, copy-pasting throws `ConfigurationError`. Update to future dates.

**D-3. Phantom `@param options` in delegation JSDoc**
`token.ts:683` — JSDoc `@param options` doesn't match actual parameter bag.

**T-1. No unit test for `delegationStatusQueryOptions` queryFn**
3-way branching (0n / MAX_UINT64 / finite) is untested.

**T-2. Batch delegation edge cases not covered**
No test for: empty token array in `batchDecryptBalancesAs`, all-fail scenario, `onError` callback that throws.

---

### ✅ Delegation Strengths

- Pre-flight delegation check in `decryptBalanceAs` — avoids wasting wallet signatures
- `#batchDecryptCore` generic — clean DRY unifying both batch methods
- Chain clock (`getBlockTimestamp`) over `Date.now()` for expiry — correct
- Good error taxonomy: `DelegationSelfNotAllowedError`, `DelegationCooldownError`, `DelegationNotFoundError`, `DelegationExpiredError`
- Past-date validation in `delegateDecryption`
- Playwright e2e tests for delegation flows

---

### Priority Actions

1. Fix spec doc `sessionTTL: 0` semantics (C-1)
2. Add pre-flight check to batch path or document why skipped (I-1)
3. Wrap `onError` in try-catch (I-2)
4. Guard `saveCachedBalance` in batch (I-3)
5. Remove `token.on()` from docs or implement it (I-4)
6. Fix `wrapSigningError` to preserve all causes (I-5)
7. Fix hooks-in-loop example (D-1)
8. Update past dates in examples (D-2)
9. Add delegation-specific unit tests (T-1, T-2)

*v4.1 — delegation-focused: 1 critical, 6 important, 4 doc/test issues.*

ankurdotb:
### Code review

Found 1 issue:

1. `decryptBalanceAs` and `batchDecryptBalancesAs` JSDoc `@throws` lists are missing `DelegationNotFoundError` and `DelegationExpiredError`. Both methods call `#assertDelegationActive` (line 762 and line 322 via `preFlightCheck`), which throws these errors *before* the try block. They propagate directly to callers undocumented. Tests at `delegation.test.ts:357-397` and `batch-decrypt-as.test.ts:135-175` confirm these throws occur.

https://github.com/zama-ai/sdk/blob/698f0e7094c531f500a74fb21460fd35a6f92a22/packages/sdk/src/token/readonly-token.ts#L721-L728

https://github.com/zama-ai/sdk/blob/698f0e7094c531f500a74fb21460fd35a6f92a22/packages/sdk/src/token/readonly-token.ts#L293-L296

ankurdotb (dismissed):

ankurdotb (approved):

zama-cremaud (approved):

PR #73, opened 2026-03-07
refactor: type hygiene, address normalization, query correctness

enitrat:
## Summary

Remove defensive fallbacks and placeholder-string patterns that masked real bugs: tighten address types to viem's `Address`, replace empty-string sentinels in query keys with conditional spreading, and stop lowercasing addresses before hashing.

### Type hygiene & cleanup
The codebase had accumulated several layers of indirection: local re-exports of `Address`/`Hex`/`Handle` from internal modules, a custom `validateAddress`/`assertAddress` pair that duplicated what `viem`'s `getAddress` already does, and backward-compat re-exports from `token.types.ts`. All of these have been collapsed — types now import directly from `viem`, `normalizeHandle`/`validateAddress` helpers are gone, and the public surface is cleaner. The Chrome extension detection in `CredentialsManager` also drops an `any` cast in favour of `Reflect.get` + a type guard.

### Query keys: no more empty-string sentinels
`query-keys.ts` used `normalizeAddressIfPresent` to substitute `""` when an optional address was absent, so a missing `owner` and an owner of `""` produced the same key. Query functions would then silently return `0n` or a default value when required params were missing. Both problems are fixed: optional addresses are now conditionally spread into the key object (absent → key omitted entirely), and query functions throw explicit errors when called without required params — while the `enabled` flag prevents them from firing in the first place.

### Address normalization consistency
Several internal paths (credential store key computation, account-change detection in `eip1193Subscribe`, pending-unshield tracking) compared or stored addresses via `.toLowerCase()`. This was inconsistent with the rest of the codebase that relies on EIP-55 checksumming. All these sites now use `getAddress()` from viem, and `computeStoreKey` no longer lowercases before hashing (test fixtures updated accordingly).


packages/react-sdk/package.json
```diff
@@ -53,11 +53,13 @@
   },
   "devDependencies": {
     "@zama-fhe/sdk": "workspace:*",
+    "viem": "^2.47.0",
     "rolldown": "^1.0.0-rc.6",
     "rolldown-plugin-dts": "^0.22.3"
   },
   "peerDependencies": {
     "@zama-fhe/sdk": "^1.0.0-alpha.16",
+    "viem": "^2.47.0",
```
ghermet:
```suggestion
    "viem": ">=2",
```

packages/sdk/src/contracts/encryption.ts
```diff
@@ -15,8 +14,6 @@ import { FHE_GAS_LIMIT } from "./gas";
  * ```
  */
 export function confidentialBalanceOfContract(tokenAddress: Address, userAddress: Address) {
-  assertAddress(tokenAddress, "tokenAddress");
-  assertAddress(userAddress, "userAddress");
```
ghermet, line 19: Don't we wanna find out at runtime if the values are not valid addresses?

enitrat: The runtime checks should only be done on the public entry points so that they fail fast, for the rest the typing should ensure that they're indeed adresses. In all entrypoint with user-provided inputs you should find getAddress() checks

ghermet (commented):
## PR #73 Review: Type Hygiene, Address Normalization, Query Correctness

**+1381 / -1605** across ~120 files — a solid cleanup PR. Here's the consolidated review.

---

### Overview

This PR systematically:
1. Removes custom `validateAddress`/`assertAddress`/`normalizeHandle` in favor of viem's `getAddress()`
2. Eliminates empty-string sentinels in query keys, replacing them with conditional spreading
3. Switches address storage/comparison from `.toLowerCase()` to EIP-55 checksumming
4. Cleans up type re-exports, removing indirection layers

---

### Critical Issues

**1. Storage key format changes break existing persisted data (Confidence: 95)**

Three storage paths change their key format from `.toLowerCase()` to checksummed `getAddress()`:

- **`computeStoreKey`** in `credentials-manager.ts` — SHA-256 hash input changes from `${address.toLowerCase()}:${chainId}` to `${address}:${chainId}`. All existing stored credentials become inaccessible. Users must re-sign EIP-712 credentials.
- **`balance-cache.ts`** — Storage key changes from lowercased to mixed-case addresses. Orphaned cache entries that never get cleaned up.
- **`pending-unshield.ts`** — Storage key changes similarly. In-progress unshields could be lost after upgrade.

Since the AES-GCM derived key in credentials-manager also uses the address as salt, even if the store key somehow matched, decryption would fail with different casing.

**Recommendation:** Since this is alpha (1.0.0-alpha.16), this may be acceptable as a breaking change, but it should be explicitly documented. Alternatively, keep `.toLowerCase()` in storage key computation for backward compatibility, or add a migration step.

**2. Removed runtime validation at public API boundaries (Confidence: 85)**

All `assertAddress()` calls removed from contract helper functions (`encryption.ts`, `erc20.ts`, `fee-manager.ts`, etc.) and Token methods. TypeScript types enforce `Address` at compile time, but JS consumers or incorrect `as Address` casts will now get obscure downstream errors instead of clear validation messages.

**Recommendation:** Consider keeping lightweight runtime validation at top-level public API entry points (Token/ZamaSDK methods), even if internal helpers no longer validate.

---

### Minor Issues

**3. `prefixHex` in `#decrypt` (Confidence: 82)**

`#decrypt` now wraps output with `prefixHex()`. If older stored data was saved without `0x` prefix, behavior changes subtly. The function also doesn't validate the content is actually hex — it just casts `as Hex`.

**4. Double space in JSDoc (Confidence: 80)**

In `use-create-eip712.ts`: `/** Contract  addresses... */` — extra space between "Contract" and "addresses".

**5. API Extractor warnings (Confidence: 82)**

`sdk-ethers.api.md` and `sdk-viem.api.md` show new `ae-forgotten-export` warnings for `GenericSigner`, `ReadContractConfig`, `EIP712TypedData`, `Handle`, etc. that are no longer re-exported from sub-package entry points. Consumers importing from `@zama-fhe/sdk/ethers` or `@zama-fhe/sdk/viem` may see TypeScript errors.

---

### What Looks Good

- **Query key refactoring** is excellent. The conditional spreading pattern is clean, type-safe, and eliminates false cache hit risk from empty-string sentinels.
- **`enabled` flags** are comprehensive and correctly aligned with `queryFn` invariant checks across all query functions.
- **`Reflect.get` for Chrome extension detection** properly replaces `(globalThis as any).chrome` with a type-safe narrowing function.
- **Invalidation safety** is maintained — prefix-based key matching still works correctly with the new key structure.
- **Test coverage** is thorough, with fixtures updated consistently for checksummed addresses.

---

### Summary

A well-structured refactor that significantly improves type safety and query correctness. The main concern is the **storage key format changes** that silently break persisted data on upgrade. Since this is pre-1.0 alpha, documenting it as a breaking change in the release notes is the minimum; adding migration logic would be the gold standard.

enitrat:
> 1. Storage key format changes break existing persisted data (Confidence: 95)

Not a problem in prerelease and it would just invalidate cache — no big deal

> 2. Removed runtime validation at public API boundaries (Confidence: 85)

Have I missed any validation at public APIs? 

> 3. prefixHex in #decrypt (Confidence: 82)

Same as 1.

> 4. Double space in JSDoc (Confidence: 80)

Fixed

> 5. API Extractor warnings (Confidence: 82)

not sure this is still relevant, I don't see the warnings when running api-report: I am not sure I understand properly what it's about either

enitrat:
From what i've seen 13 public methods still have runtime validation via getAddress(), so I think we're fine.

ghermet:
> From what i've seen 13 public methods still have runtime validation via getAddress(), so I think we're fine.

all the contract config are public apis, like you can import them all from the sdk. To check this run `pnpm api-report` and you'll see in the `sdk/etc` folder 
```md
// @public (undocumented)
export function readConfidentialBalanceOfContract(client: PublicClient, tokenAddress: Address, userAddress: Address): Promise<`0x${string}`>;

// @public (undocumented)
export function readSupportsInterfaceContract(client: PublicClient, tokenAddress: Address, interfaceId: Address): Promise<boolean>;

// @public (undocumented)
export function readUnderlyingTokenContract(client: PublicClient, wrapperAddress: Address): Promise<`0x${string}`>;
```

enitrat:
as per your review I re-added the checks on public entrypoints

ghermet (dismissed):

ghermet (approved):

PR #75, opened 2026-03-09
feat: rework public APIs via `createFhevmConfig`

enitrat:
wip


ghermet (commented):
Hey @enitrat — solid work on this draft! Here's a review of where things stand. Since it's WIP, flagging things early so they're on your radar.

## Strengths

- The config-first API (`createFhevmConfig` + `wagmiAdapter()`) is a huge DX win — the provider setup goes from ~20 lines to 4.
- ACL fix in `#assertDecryptAuthorization` is correct — now checks both actor AND contract with the `actor !== contract` guard, matching relayer-sdk behavior.
- Good test coverage for the new modules (`resolveRelayer`, `resolveWallet`, `config`, provider chain resolution).
- Clean separation: `config.ts` is inert, resolution logic is isolated.

## Things to address before merge

### High

1. **`createFhevmConfig` creates a new `MemoryStorage()` on every call.** If someone writes `<ZamaProvider config={createFhevmConfig({...})}>` inline without memoizing, the storage ref changes every render → SDK recreation loop. Worth either documenting prominently or returning a singleton.

2. **`viem` added as a peer dep of react-sdk.** `resolve-wallet.ts` imports `createPublicClient`/`http` directly. Non-wagmi users who don't use viem must now install it. Is this intentional? Could be a dynamic import instead.

3. **`ConfiguredChainSigner` (`configured-chain-signer.ts`) appears unused.** It's defined but never referenced in `provider.tsx` or `resolve-wallet.ts`. Leftover from a future step?

### Medium

4. **`IMPLEMENTATION_PLAN.md` + `docs/plans/` (685 lines of planning docs)** — probably should be removed before merge.

5. **`resolveRelayer` throws on unknown chain IDs**, but the design doc says "Unknown chains default to cleartext mode." Intentional divergence for now?

6. **Breaking exports removed from react-sdk barrel:** `RelayerWeb`, `RelayerWebConfig`, `RelayerWebSecurityConfig`, and `createCleartextRelayer` from sdk/cleartext. Fine for a breaking prerelease, just make sure the migration guide covers these.

7. **`usePublicKey`/`usePublicParams` now explicitly set `queryKey`** instead of spreading. Looks like a workaround — worth a comment explaining why.

### Low

8. Hardhat preset addresses in `presets.ts` are now hardcoded rather than derived from `HardhatConfig` — won't match custom deployments. Test apps still use `deployments.json` so it works, but the preset is less flexible.

9. The `queryFn` context type annotations across ~15 query files (`async (context: { queryKey: typeof queryKey })`) — is this fixing a TS inference issue from the rolldown migration? Might be worth a one-liner comment.

## Security

ACL changes look correct. `persistAllowed` checked for both actor + contract, delegate intentionally not checked (authorized by EIP-712). No new injection vectors.

---

Overall this is heading in a great direction. The before/after on the test app providers alone makes this worth it. Let me know if you want me to dig into anything specific.

PR #78, opened 2026-03-09
feat(sdk): cleartext flat config refactor

enitrat:
- refactor(cleartext): flatten CleartextInstanceConfig into a single config type; remove encrypted-input.ts and factory.ts; migrate fhe-type.ts from relayer-SDK to use the exact same types, decouple constants from relayer-utils by inlining values

- fix api-extractor that had tsconfig with `"node"`, that ignores package.json exports, causing subpath imports to resolve to raw .ts source files instead of .d.ts declarations. updated to `bundler`


packages/sdk/src/relayer/cleartext/fhe-type.ts
```diff
@@ -0,0 +1,191 @@
...
+  8: 2,
+  16: 3,
+  32: 4,
+  64: 5,
+  128: 6,
+  160: 7,
+  256: 8,
+} as const;
+
+Object.freeze(FheTypeNameToId);
```
ghermet, line 107: tu sais que `as const` dans typescript c'est `Object.freeze`

enitrat: copié de relayer-sdk intentionnellement sans changements

packages/sdk/package.json
```diff
@@ -67,6 +71,9 @@
     "lint": "eslint .",
     "typecheck": "tsc --noEmit"
   },
+  "dependencies": {
+    "viem": ">=2"
+  },
```
enitrat, line 72: as discussed

packages/sdk/src/relayer/cleartext/__tests__/fixtures.ts
```diff
@@ -15,14 +15,11 @@ export const USER_ADDRESS = "0x1000000000000000000000000000000000000001";
...
 export const CLEAR_TEXT_MOCK_CONFIG: CleartextConfig = {
-  chainId: 31_337n,
+  network: "http://127.0.0.1:8545",
+  chainId: 31_337,
   gatewayChainId: GATEWAY_CHAIN_ID,
-  contracts: {
-    acl: TEST_FHEVM_ADDRESSES.acl,
-    executor: TEST_FHEVM_ADDRESSES.executor,
-    inputVerifier: TEST_FHEVM_ADDRESSES.inputVerifier,
-    kmsVerifier: TEST_FHEVM_ADDRESSES.kmsVerifier,
```
enitrat, line 24: intentionnaly removed as they're not useful for the SDK

packages/sdk/package.json
```diff
@@ -21,6 +21,10 @@
       "types": "./dist/cleartext/index.d.ts",
       "import": "./dist/cleartext/index.js"
     },
+    "./chains": {
+      "types": "./dist/chains/index.d.ts",
+      "import": "./dist/chains/index.js"
+    },
```
ghermet: To be moved to #75

ghermet (commented):
## PR #78 Review: Cleartext Flat Config Refactor, Chains Module

**+834 / -835** across 17 files — a well-structured refactor with a new module addition.

---

### Overview

1. **Cleartext config flattening** — Replaces `CleartextChainConfig` + `CleartextContracts` + `createCleartextRelayer()` factory with a single flat `CleartextConfig` type and direct `new CleartextFhevmInstance(config)` construction
2. **Chains module** — New `@zama-fhe/sdk/chains` export with `FhevmChain` type and presets (mainnet, sepolia, hardhat, hoodi)
3. **API-extractor fix** — `moduleResolution: "node"` → `"bundler"` to properly resolve package.json exports

---

### Critical Issues

**1. `viem` moved from optional peerDependency to hard dependency (Confidence: 92)**

The chains module imports `viem/chains` at runtime. This forces `viem` as a required dependency for all consumers, even those using only the ethers adapter or core module.

**Recommendation:** Either hardcode chain ID constants to avoid the runtime import, or accept this as intentional and document it as a breaking change.

---

### Important Issues

**2. Inconsistent `chainId` type passed to EIP-712 domain factories (Confidence: 82)**

In `cleartext-fhevm-instance.ts`:
- `createEIP712` passes `this.#config.chainId` (a `number`)
- `createDelegatedUserDecryptEIP712` passes `this.#chainIdBigInt` (a `bigint`)

Both work since `DomainFactory` accepts `number | bigint`, but the inconsistency is confusing.

**3. `FhevmChain` type is disconnected from `CleartextConfig` (Confidence: 80)**

- `FhevmChain.id` uses `FhevmChainId` (literal union `1 | 11155111 | 31337 | 560048`)
- `CleartextConfig.chainId` uses plain `number`
- No type-level link between `fhevmHoodi` (chains) and `hoodiCleartextConfig` (cleartext)

A consumer could accidentally mix `fhevmSepolia` with `hoodiCleartextConfig` with no compile-time error. Consider having `CleartextConfig` reference `FhevmChainId` or enriching `FhevmChain` with contract addresses so configs can be derived from chain presets.

**4. `chainId: number` is too permissive**

Accepts floats, negative numbers, `NaN`, `Infinity`. The constructor converts via `BigInt()` which throws on non-integers with unhelpful error messages. Consider explicit validation.

**5. Scattered `as Address` casts in implementation**

`this.#config.aclContractAddress as Address` appears in multiple places despite `CleartextConfig` already typing these fields as `Address` from viem. This suggests a type import mismatch between `viem.Address` and `relayer-sdk.types.Address`. Should be resolved so casts are unnecessary.

---

### Minor Notes

- **`ebool` encryption bitwidth changed from 1 to 2** — Safe because `normalizeEncryptValue` separately enforces `0n | 1n` for ebool, but the subtlety warrants a comment.
- **FHE type duplication risk** — `fhe-type.ts` is explicitly copied from `@zama-fhe/relayer-sdk`. No mechanism to detect drift if the relayer SDK adds new types.
- **No defensive copy of config** — Constructor stores `this.#config = config` directly. Caller mutation after construction would corrupt state. Low risk with const presets but breaks encapsulation.

---

### What Looks Good

- **Config flattening is a genuine simplification** — 3 types + factory → 1 type + constructor. Cleaner API.
- **Forbidden chain ID check** prevents cleartext mode on mainnet/Sepolia — important safety invariant.
- **FHE type system** is well-structured with frozen maps, bidirectional lookups, and proper type guards.
- **Test coverage significantly improved** — new tests for encrypt validation, bit-width overflow, bool validation, authorization checks, signer-equals-contract edge cases.
- **API-extractor fix** is correct — matching `"bundler"` moduleResolution with tsconfig.
- **Removal of `kmsVerifier`/`inputVerifier`** from config honestly reflects what cleartext mode actually uses.

---

### Breaking Changes to Document

- `createCleartextRelayer()` removed → `new CleartextFhevmInstance(config)`
- `CleartextChainConfig` / `CleartextContracts` types removed → flat `CleartextConfig`
- `chainId` changed from `bigint` to `number`
- `viem` becomes a hard dependency
- `euint4` (type ID 1) deprecated and removed from type maps

enitrat:
> 1. viem moved from optional peerDependency to hard dependency (Confidence: 92)
seen, it's ok

> 2. Inconsistent chainId type passed to EIP-712 domain factories (Confidence: 82)
normalized everything to take the chainId (number) from config. EIP-712 domain factories wrap it in a Bigint for signing.

> 3. FhevmChain type is disconnected from CleartextConfig (Confidence: 80)
removed entirely as discussed

> 4. chainId: number is too permissive
acknowledged, wontfix

> 5. Scattered as Address casts in implementation
cleaned up 


for the rest, all good

packages/playwright/fixtures/fhevm.ts
```diff
@@ -2,29 +2,12 @@ import { Page } from "@playwright/test";
...
-    inputVerifier: deployments.fhevm.inputVerifier,
-    kmsVerifier: deployments.fhevm.kmsVerifier,
-    verifyingInputVerifier: VERIFYING_CONTRACTS.inputVerification,
-    verifyingDecryption: VERIFYING_CONTRACTS.decryption,
-  },
-} satisfies CleartextChainConfig;
-
 export async function mockRelayerSdk(page: Page, baseURL: string) {
-  const fhevm = createCleartextRelayer(hardhat);
+  const fhevm = new CleartextFhevmInstance(hardhatCleartextConfig);
```
ghermet: Il faut enlever tous les mocks playwright

packages/sdk/src/relayer/cleartext/__tests__/cleartext-fhevm-instance.test.ts
```diff
@@ -10,128 +8,112 @@ import {
...
 
 type MockClientOptions = {
   persistAllowed?: (handle: string, account: string) => boolean;
   isAllowedForDecryption?: (handle: string) => boolean;
-  isHandleDelegatedForUserDecryption?: (
-    delegator: string,
-    delegate: string,
-    contractAddress: string,
-    handle: string,
-  ) => boolean;
```
ghermet, line 41: I've got the feature lined up should we keep this?

packages/playwright/fixtures/fhevm.ts
enitrat: done

ghermet (dismissed):

ghermet (dismissed):

ghermet (approved):

PR #87, opened 2026-03-09
ci: add claude review job

enitrat:
Adds claude review job


.github/workflows/claude.yml
enitrat, line 1: can we keep a single file?

.github/scripts/install-claude.sh
enitrat, line 1: wouldn't it be simpler to just `npm i -g` ?

ankurdotb: I agree. I was sticking to the SHA verification in the original setup from Roger. Can keep the claude version pinned and install instead of installing `@latest` (which fair enough, we pin actions as well). And switch to NPM as NPM already does that validation I think.

enitrat: you can pin the installation of claude code CLI to a specific rev even via NPM I think ? would that make sense?

ankurdotb: This has been switched to install via NPM

.github/workflows/claude.yml
ankurdotb: Merged to single file

ankurdotb (dismissed):

enitrat (commented):
ok for me but generally i'd just do a 1:1 of the FHEVM config and call it a day. I don't think we need anything more than the default setup there that has already been accepted

enitrat:
closing in favor of new shared workflows

PR #88, opened 2026-03-09
perf: speed up vitest by 40%

enitrat:
## Summary

Vitest was spending most of its time on jsdom environment setup (~78s) rather than running tests (~17s). Three changes bring wall time from **15.7s → 9.5s**:

- **Split into projects** — SDK tests now run in `node` env instead of jsdom (−27%)
- **vmThreads pool** for SDK — shares module caches across test files (−5%)
- **happy-dom** replaces jsdom for react-sdk — lighter DOM implementation (−13%)

## Test plan

- [x] `pnpm vitest run` — 974 tests pass
- [x] Benchmarked each change individually (3 runs each, median reported)


vitest.config.ts
```diff
@@ -1,13 +1,76 @@
...
-    include: ["packages/**/*.test.{ts,tsx}"],
-    exclude: ["**/integration.test.ts", "**/node_modules/**"],
-    globals: true,
-    setupFiles: ["./vitest.setup.ts"],
+    projects: [
+      {
+        test: {
+          name: "sdk",
+          environment: "node",
+          pool: "vmThreads",
```
ghermet, line 51: is vmThreads suitable for CI?

vitest.config.ts
```diff
@@ -1,13 +1,76 @@
...
+          exclude: ["**/integration.test.ts", "**/node_modules/**", "**/worker/__tests__/**"],
+          globals: true,
+          setupFiles: ["./vitest.setup.ts"],
+        },
+        resolve: sharedResolve,
+      },
+      {
+        test: {
+          name: "react-sdk",
+          environment: "happy-dom",
```
ghermet, line 62: Muy bien 💯

ghermet (approved):

PR #92, opened 2026-03-10
test: improve branch coverage for storage, worker, and event modules

ghermet:
## Summary
- Add new test suites for `ChromeSessionStorage` (0% → 100%) and `AsyncLocalMapStorage` (0% → 100%)
- Expand branch coverage for `eip1193Subscribe`, `BaseWorkerClient`, and `NodeWorkerPool`
- Overall coverage: 94.43% → 95.53% statements, 83.4% → 84.87% branches

## Test plan
- [x] All 67 new/updated tests pass (`pnpm vitest run`)
- [x] Typecheck passes (`pnpm typecheck`)
- [x] Lint passes (via pre-commit hook)
- [ ] CI passes


enitrat (commented):
[AUTOMATED]

## Code Review - Test Coverage PR

Good coverage improvement overall (94.43% → 95.53% statements, 83.4% → 84.87% branches). Tests are well-structured and cover real code paths. A few items to address:

### Issues

1. **Zero-assertion test** in `eip1193-subscribe.test.ts` — "uses default no-op callbacks" will never fail (see inline comment)
2. **Inconsistent payload verification** in `worker.base-client.test.ts` — some domain method tests verify payload, others don't (see inline comment)
3. **Duplicate test** in `chrome-session-storage.test.ts` — last test is functionally identical to the first

### Missing edge cases worth adding

- `eip1193Subscribe`: `accountsChanged` with multiple accounts in the array (only `accounts[0]` is read)
- `eip1193Subscribe`: `chainChanged` with decimal vs hex string (implementation uses `parseInt(chainId, 16)`)
- `AsyncLocalMapStorage`: error propagation from `run` callback
- `NodeWorkerPool`: `terminate()` before `initPool()`

### What's good

- AsyncLocalMapStorage isolation test with `Promise.all` is well-designed
- NodeWorkerPool init lifecycle tests (concurrent init, failure cleanup, re-init) are valuable
- eip1193Subscribe case-insensitive address comparison and getAddress rejection tests catch real branches
- Proper use of `expect.unreachable`, `vi.waitFor`, extended fixtures

packages/sdk/src/token/__tests__/eip1193-subscribe.test.ts
```diff
@@ -106,4 +110,101 @@ describe("eip1193Subscribe", () => {
...
+  it("uses default no-op callbacks when not provided", async () => {
+    const provider = createFakeProvider();
+    eip1193Subscribe(provider, () => Promise.resolve(ADDR_A), {});
+    await vi.waitFor(() => {});
+
+    provider.emit("accountsChanged", []);
+    provider.emit("accountsChanged", [ADDR_B]);
+    provider.emit("disconnect");
+  });
 });
```
enitrat, line 211: [AUTOMATED] This test has **zero assertions** — it will never fail regardless of code changes. It's pure coverage padding.

Consider adding assertions, e.g.:
```ts
it("uses default no-op callbacks when not provided", async () => {
  const provider = createFakeProvider();
  const unsub = eip1193Subscribe(provider, () => Promise.resolve(ADDR_A), {});
  await vi.waitFor(() => {});

  // These should not throw with default no-op callbacks
  expect(() => provider.emit("accountsChanged", [])).not.toThrow();
  expect(() => provider.emit("accountsChanged", [ADDR_B])).not.toThrow();
  expect(() => provider.emit("disconnect")).not.toThrow();
});
```

Or better yet, verify that the function completes without error by wrapping in `expect(async () => { ... }).not.toThrow()`.

packages/sdk/src/worker/__tests__/worker.base-client.test.ts
```diff
@@ -332,4 +332,184 @@ describe("BaseWorkerClient", () => {
...
+      expect((err as Error & { statusCode?: number }).statusCode).toBe(429);
+    }
+  });
+
+  it("terminate is a no-op when no worker exists", () => {
+    const client = new TestWorkerClient();
+    client.terminate();
+  });
+
+  it("handleWorkerError without existing worker does not throw", () => {
```
enitrat, line 365: [AUTOMATED] **Nit:** Several domain method tests (this one, `createEIP712`, `createDelegatedUserDecryptEIP712`, `delegatedUserDecrypt`) only verify the request `type` but skip payload verification. Meanwhile `publicDecrypt` and `getPublicParams` correctly verify both type AND payload.

For consistency and to catch payload forwarding bugs, consider adding:
```ts
expect(lastCall.payload).toEqual(params);
```
to each domain method test.

packages/sdk/src/token/__tests__/chrome-session-storage.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    await expect(storage.delete("missing")).resolves.not.toThrow();
+  });
+
+  it("handles complex objects", async () => {
+    const obj = { nested: { array: [1, 2, 3] } };
+    await storage.set("complex", obj);
+    expect(await storage.get("complex")).toEqual(obj);
+  });
+
+  it("returns null when chrome.storage.session.get returns empty object", async () => {
```
enitrat: [AUTOMATED] **Nit:** This test is functionally identical to "returns null for missing keys" (line 37). Both exercise the same code path: key absent from store → `get` returns `{}` → `result[key]` is `undefined` → `?? null`.

Consider removing this duplicate, or differentiating it (e.g., test with a key that exists but has `undefined` as its value, to verify the `?? null` coalescing behavior with falsy values).

enitrat (approved):

PR #96, opened 2026-03-10
docs: minor updates on cleartext mode

enitrat:
I intentionally did not mention how to spin up a new deployment of the cleartext contracts on a hardhat node: I am still not exactly sure how this should be done; if we should add an `anvil` config; and how to link it to a durable solution. For now it's in https://github.com/zama-ai/forge-fhevm, but let's set this for a future task. Maybe it should be in the main docs instead.

### DIFF
--- a/docs/gitbook/src/guides/local-development.md
+++ b/docs/gitbook/src/guides/local-development.md
@@ -5,15 +5,15 @@ description: How to use the cleartext relayer for local Hardhat nodes and custom
 
 # Local Development
 
-The SDK ships `RelayerCleartext`, a drop-in replacement for `RelayerWeb` and `RelayerNode` that operates in cleartext mode. Values are stored as plaintext on-chain — no KMS, no gateway, no WASM. Use it for local Hardhat nodes, custom testnets, or any chain where you deploy fhEVM contracts in cleartext mode.
+The SDK ships `RelayerCleartext`, a drop-in relayer that replaces FHE operations with cleartext operations. Values are stored as plaintext on-chain — no KMS, no gateway, no WASM. Use it for local Hardhat nodes, custom testnets, or any chain where you deploy fhEVM contracts in cleartext mode.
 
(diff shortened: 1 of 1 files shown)


ghermet (approved):

ghermet:
> I intentionally did not mention how to spin up a new deployment of the cleartext contracts on a hardhat node: I am still not exactly sure how this should be done; if we should add an `anvil` config; and how to link it to a durable solution. For now it's in https://github.com/zama-ai/forge-fhevm, but let's set this for a future task. Maybe it should be in the main docs instead.

People should be able to use the RelayerCleartext with their hardhat / hoodi setups reading the docs. If we don't explain them how to connect the cleartext config with their deployment they will come back to us with questions we already had the answer to. If you think this is enough let's see what bitgo can do with it

enitrat:
> People should be able to use the RelayerCleartext with their hardhat / hoodi setups reading the docs.

My interrogation here was more as to _how_ do we present people with how to deploy a local cleartext stack on any chain. I'm not sure it belongs in the SDK documentation itself OR inside the zama docs like https://docs.zama.org/protocol/solidity-guides/getting-started/setup. I am thinking of delegating this part of the docs to devx team.

ghermet:
> > People should be able to use the RelayerCleartext with their hardhat / hoodi setups reading the docs.
> 
> My interrogation here was more as to _how_ do we present people with how to deploy a local cleartext stack on any chain. I'm not sure it belongs in the SDK documentation itself OR inside the zama docs like https://docs.zama.org/protocol/solidity-guides/getting-started/setup. I am thinking of delegating this part of the docs to devx team.

The solidity guide is a general guide, it does not convey any information about how to use the deployment with the sdk. I think there should be a simple guide in the sdk doc where we point to `https://docs.zama.org/protocol/solidity-guides/getting-started/setup` as a starting point then instruct them how to plug the deployment variables to the cleartext config.

PR #97, opened 2026-03-10
fix(sdk): extend contract addresses instead of regenerating keypair

ghermet:
## Summary
- When `allow()` is called with contracts not covered by existing credentials (e.g. individual token decrypt followed by "Reveal All"), the keypair is now reused and re-signed with the merged contract set instead of doing a full keypair regeneration
- Added `#extendContracts`, `#mergeContracts`, `#persistExtended`, `#signWithContracts` helper methods to `CredentialsManager`
- Split `#isValid` into `#isTimeValid` and `#coversContracts` for finer-grained checks
- Uses lowercase comparison for address matching to avoid EIP-55 checksum inconsistencies
- Added 9 comprehensive tests covering contract extension scenarios (active session, revoke, page reload, caching, persistence, incremental extension, subset, time-expiry, EIP-712 verification)

## Test plan
- [x] All 967 tests pass across 112 test files
- [x] New contract extension tests cover: active session reuse, revoke path, page reload path, caching after extend, cross-instance persistence, incremental A→A,B→A,B,C extension, subset no-op, time-expiry full regen, EIP-712 merged set verification

### DIFF
--- a/packages/sdk/src/token/__tests__/credentials-manager.test.ts
+++ b/packages/sdk/src/token/__tests__/credentials-manager.test.ts
@@ -59,7 +59,7 @@ describe("CredentialsManager", () => {
     expect(relayer.generateKeypair).toHaveBeenCalledOnce();
   });
 
-  it("re-signs when new contract not in signed list", async ({
+  it("re-signs when new contract not in signed list but reuses keypair", async ({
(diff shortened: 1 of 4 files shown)


ghermet:
See #99

PR #99, opened 2026-03-11
refactor(sdk): simplify internals, add caching, and extend contract addresses

ghermet:
## Summary
- Extract shared `toError` helper to `utils.ts` (was duplicated in 3 files)
- Deduplicate IndexedDB transaction boilerplate via `#withTransaction` helper
- Simplify `balanceOf` to delegate to `decryptBalance`, eliminating ~40 lines of duplicated decrypt logic
- Parallelize cache lookups in `batchDecryptBalances` using `Promise.all` and skip `allow()` for cached balances
- Add SHA-256 store key and PBKDF2 derived key caching to `CredentialsManager` to avoid redundant crypto operations
- Unify `#isValid` / `#isValidWithoutDecrypt` into a single method with widened parameter type, then split into `#isTimeValid` + `#coversContracts` for finer-grained checks
- Extend contract address set instead of regenerating keypair when `allow()` is called with new contracts (integrates #97)
- Simplify `#mergeContracts` to use `Set` with `getAddress` canonical checksums

## Test plan
- [x] All 977 tests pass across 112 test files
- [x] 8 new caching tests (storeKey caching, deriveKey caching, unified #isValid)
- [x] 9 new contract extension tests (active session, revoke, page reload, caching, persistence, incremental, subset, time-expiry, EIP-712 verification)
- [x] 2 new batch decrypt tests (all cached skip allow, partial cached calls allow with uncached only)
- [x] Typecheck clean

### DIFF
--- a/packages/sdk/src/token/__tests__/credentials-manager.test.ts
+++ b/packages/sdk/src/token/__tests__/credentials-manager.test.ts
@@ -813,31 +813,52 @@ describe("contract address extension", () => {
     expect(normalized).toContain(getAddress(TOKEN_A));
     expect(normalized).toContain(getAddress(TOKEN_B));
   });
-});
 
-describe("storeKey caching", () => {
-  it("caches the SHA-256 store key across multiple allow() calls", async ({
+  it("concurrent extensions don't drop contract addresses", async ({
     relayer,
     signer,
     credentialManager,
   }) => {
     setupMocks(relayer, signer);
-    const digestSpy = vi.spyOn(crypto.subtle, "digest");
 
     await credentialManager.allow(TOKEN_A);
-    const digestCallsAfterFirst = digestSpy.mock.calls.filter(
-      ([algo]) => algo === "SHA-256",
-    ).length;
+
+    // Launch two extensions concurrently with different contracts
+    const [resultB, resultC] = await Promise.all([
+      credentialManager.allow(TOKEN_A, TOKEN_B),
+      credentialManager.allow(TOKEN_A, TOKEN_C),
+    ]);
+
+    // The last result should cover all three contracts (no address dropped)
+    const finalContracts = resultC.contractAddresses.map((a) => getAddress(a));
+    expect(finalContracts).toContain(getAddress(TOKEN_A));
+    expect(finalContracts).toContain(getAddress(TOKEN_B));
+    expect(finalContracts).toContain(getAddress(TOKEN_C));
+
+    // First concurrent result covers at least A and B
+    const firstContracts = resultB.contractAddresses.map((a) => getAddress(a));
+    expect(firstContracts).toContain(getAddress(TOKEN_A));
+    expect(firstContracts).toContain(getAddress(TOKEN_B));
+  });
+});
+
+describe("storeKey caching", () => {
+  it("caches the store key across multiple allow() calls", async ({
+    relayer,
+    signer,
+    credentialManager,
+  }) => {
+    setupMocks(relayer, signer);
+    const computeSpy = vi.spyOn(CredentialsManager, "computeStoreKey");
 
     await credentialManager.allow(TOKEN_A);
-    const digestCallsAfterSecond = digestSpy.mock.calls.filter(
-      ([algo]) => algo === "SHA-256",
-    ).length;
+    expect(computeSpy).toHaveBeenCalledOnce();
 
-    // Second allow() should not trigger another SHA-256 digest for the store key
-    expect(digestCallsAfterSecond).toBe(digestCallsAfterFirst);
+    await credentialManager.allow(TOKEN_A);
+    // Second allow() should not recompute the store key
+    expect(computeSpy).toHaveBeenCalledOnce();
 
-    digestSpy.mockRestore();
+    computeSpy.mockRestore();
   });
 
   it("reuses cached store key for isExpired() after allow()", async ({
@@ -846,22 +867,16 @@ describe("storeKey caching", () => {
     credentialManager,
   }) => {
     setupMocks(relayer, signer);
+    const computeSpy = vi.spyOn(CredentialsManager, "computeStoreKey");
 
     await credentialManager.allow(TOKEN_A);
-
-    // Reset call counts to measure only isExpired() calls
-    vi.mocked(signer.getAddress).mockClear();
-    vi.mocked(signer.getChainId).mockClear();
-    const digestSpy = vi.spyOn(crypto.subtle, "digest");
+    expect(computeSpy).toHaveBeenCalledOnce();
 
     await credentialManager.isExpired();
+    // isExpired() should reuse the cached store key
+    expect(computeSpy).toHaveBeenCalledOnce();
 
-    // getAddress/getChainId are still called (to build identity string for comparison),
-    // but SHA-256 digest should NOT be called again because the cache hits
-    const sha256Calls = digestSpy.mock.calls.filter(([algo]) => algo === "SHA-256").length;
-    expect(sha256Calls).toBe(0);
-
-    digestSpy.mockRestore();
+    computeSpy.mockRestore();
   });
 
   it("invalidates store key cache when signer address changes", async ({
@@ -878,23 +893,21 @@ describe("storeKey caching", () => {
       sessionStorage: createMockStorage(),
       keypairTTL: 86400,
     });
+    const computeSpy = vi.spyOn(CredentialsManager, "computeStoreKey");
 
     await manager.allow(TOKEN_A);
+    expect(computeSpy).toHaveBeenCalledOnce();
 
     // Change the signer address
     vi.mocked(signer.getAddress).mockResolvedValue(
       "0xdDdDddDdDdddDDddDDddDDDDdDdDDdDDdDDDDDDd" as Address,
     );
 
-    const digestSpy = vi.spyOn(crypto.subtle, "digest");
-
     await manager.allow(TOKEN_A);
+    // Should have recomputed the store key for the new address
+    expect(computeSpy).toHaveBeenCalledTimes(2);
 
-    // Should have recomputed the store key (new SHA-256 call)
-    const sha256Calls = digestSpy.mock.calls.filter(([algo]) => algo === "SHA-256").length;
-    expect(sha256Calls).toBeGreaterThanOrEqual(1);
-
-    digestSpy.mockRestore();
+    computeSpy.mockRestore();
   });
 
   it("invalidates store key cache when chain ID changes", async ({
@@ -911,20 +924,18 @@ describe("storeKey caching", () => {
       sessionStorage: createMockStorage(),
       keypairTTL: 86400,
     });
+    const computeSpy = vi.spyOn(CredentialsManager, "computeStoreKey");
 
     await manager.allow(TOKEN_A);
+    expect(computeSpy).toHaveBeenCalledOnce();
 
     // Change the chain ID
     vi.mocked(signer.getChainId).mockResolvedValue(1);
 
-    const digestSpy = vi.spyOn(crypto.subtle, "digest");
-
     await manager.allow(TOKEN_A);
+    expect(computeSpy).toHaveBeenCalledTimes(2);
 
-    const sha256Calls = digestSpy.mock.calls.filter(([algo]) => algo === "SHA-256").length;
-    expect(sha256Calls).toBeGreaterThanOrEqual(1);
-
-    digestSpy.mockRestore();
+    computeSpy.mockRestore();
   });
 
   it("reuses the same store key when signer address casing changes", async ({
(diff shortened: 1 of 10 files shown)


packages/sdk/src/token/credentials-manager.ts
```diff
@@ -70,6 +70,10 @@ export class CredentialsManager {
   #onEvent: ZamaSDKEventListener;
   #createPromise: Promise<StoredCredentials> | null = null;
   #createPromiseKey: string | null = null;
+  #cachedStoreKey: string | null = null;
+  #cachedStoreKeyIdentity: string | null = null;
+  #cachedDerivedKey: CryptoKey | null = null;
+  #cachedDerivedKeyIdentity: string | null = null;
```
enitrat, line 77: [AUTOMATED] **Medium**: These new cache fields (`#cachedStoreKey`, `#cachedStoreKeyIdentity`, `#cachedDerivedKey`, `#cachedDerivedKeyIdentity`) are never nulled by `revoke()` (line 253) or `clear()` (line 280). After explicit revocation, the PBKDF2-derived AES key and store key remain in memory for the lifetime of the `CredentialsManager` instance.

The derived key is non-extractable so this isn't a direct key-leak, but it's weaker secret hygiene than before this PR. A long-lived SDK instance (e.g. SPA) retains cryptographic material that the user explicitly asked to discard.

Suggested fix — add to both `revoke()` and `clear()`:
```ts
this.#cachedStoreKey = null;
this.#cachedStoreKeyIdentity = null;
this.#cachedDerivedKey = null;
this.#cachedDerivedKeyIdentity = null;
```

Also consider clearing `#cachedDerivedKey` in `ZamaSDK.#revokeByTrackedIdentity()` on account/chain switches, since that path bypasses `CredentialsManager.revoke()`.

Confidence: 90/100

packages/sdk/src/token/credentials-manager.ts
```diff
@@ -316,33 +345,88 @@ export class CredentialsManager {
...
-  #isValidWithoutDecrypt(encrypted: EncryptedCredentials, requiredContracts: Address[]): boolean {
-    const nowSeconds = Math.floor(Date.now() / 1000);
-    const expiresAt = encrypted.startTimestamp + encrypted.durationDays * 86400;
-    if (nowSeconds >= expiresAt) return false;
-    const signedSet = new Set(encrypted.contractAddresses.map(getAddress));
-    return requiredContracts.every((addr) => signedSet.has(getAddress(addr)));
+  /** Check if the signed address set covers all required addresses. */
+  #coversContracts(signedAddresses: Address[], requiredContracts: Address[]): boolean {
+    const signedSet = new Set(signedAddresses.map((a) => a.toLowerCase()));
+    return requiredContracts.every((addr) => signedSet.has(addr.toLowerCase()));
```
enitrat: [AUTOMATED] **Minor**: `#coversContracts` switched from `getAddress()` normalization (which validates EIP-55 checksums and rejects malformed addresses) to `.toLowerCase()` (which accepts any string). A tampered IndexedDB entry with a non-address string would now pass the coverage check.

Downstream paths (`#mergeContracts` at line 389, EIP-712 signing) still use `getAddress()`, so this isn't directly exploitable — but it's a defense-in-depth regression on a security-sensitive boundary.

Consider: either restore `getAddress()` here, or add address format validation in `#assertEncryptedCredentials()` for each element of `contractAddresses`.

Confidence: 85/100

packages/sdk/src/token/indexeddb-storage.ts
```diff
@@ -60,52 +60,37 @@ export class IndexedDBStorage implements GenericStorage {
...
     const db = await this.#getDB();
     return new Promise((resolve, reject) => {
-      const tx = db.transaction(this.#storeName, "readonly");
+      const tx = db.transaction(this.#storeName, mode);
       tx.onabort = () => reject(tx.error ?? new Error("Transaction aborted"));
-      const store = tx.objectStore(this.#storeName);
-      const request = store.get(key);
-      request.onsuccess = () => resolve(request.result?.value ?? null);
+      const request = fn(tx.objectStore(this.#storeName));
+      request.onsuccess = () => resolve(request.result);
```
enitrat: [AUTOMATED] **Medium**: `#withTransaction` resolves on `request.onsuccess`, but for `readwrite` transactions, `onsuccess` fires before the transaction commits to disk. If the transaction subsequently aborts (quota exceeded, DB closed by `onversionchange` mid-commit), the data is rolled back but the caller already received a resolved promise — silent data loss.

Note: the old code had the same pattern for `set`/`delete`, so this is a **pre-existing** issue inherited by the refactor. However, since you're touching this code anyway, it's a good opportunity to fix it.

For writes, resolve on `tx.oncomplete` instead:
```ts
if (mode === "readonly") {
  request.onsuccess = () => resolve(request.result);
} else {
  tx.oncomplete = () => resolve(request.result);
}
```

Confidence: 82/100

packages/sdk/src/token/__tests__/credentials-manager.test.ts
```diff
@@ -601,6 +607,500 @@ describe("session allow/revoke", () => {
...
+    expect(relayer.generateKeypair).toHaveBeenCalledOnce();
+    // Only 2 signatures: initial + extend. Third call is cached.
+    expect(signer.signTypedData).toHaveBeenCalledTimes(2);
+  });
+
+  it("persists extended credentials across instances", async ({
+    relayer,
+    signer,
+    storage,
+    sessionStorage,
```
enitrat: [AUTOMATED] **Minor**: This test ("persists extended credentials across instances") shares the same `sessionStorage` between `manager1` and `manager2`. The test can pass purely because the in-memory session signature survives — it doesn't exercise the higher-risk reload path where session storage is fresh.

Compare with the test at line 116 which correctly uses `createMockStorage()` for manager2's sessionStorage to simulate a page reload.

Suggested: use a fresh `createMockStorage()` for manager2's sessionStorage to truly test the persist-and-reload path with extended credentials.

Confidence: 90/100

packages/sdk/src/token/__tests__/credentials-manager.test.ts
```diff
@@ -601,6 +607,500 @@ describe("session allow/revoke", () => {
...
+});
+
+describe("storeKey caching", () => {
+  it("caches the SHA-256 store key across multiple allow() calls", async ({
+    relayer,
+    signer,
+    credentialManager,
+  }) => {
+    setupMocks(relayer, signer);
+    const digestSpy = vi.spyOn(crypto.subtle, "digest");
```
enitrat: [AUTOMATED] **Nit**: The storeKey and deriveKey caching tests (lines 829-1027) assert internal `crypto.subtle.digest`/`deriveKey` call counts. These will break on harmless implementation changes (e.g. switching hash algorithms, adding salting steps) without testing user-visible behavior.

Also, the deriveKey test at line 975 depends on `signTypedData` returning the exact same `"0xsig789"` across calls — a real wallet may return different signatures for the same payload, which would cause the derived key cache to miss and invalidate the test's assumption.

Consider testing the observable caching effect instead (e.g. verify that the second call doesn't trigger a wallet prompt, or measure that no additional keypair generation occurs).

Confidence: 85/100

enitrat (changes requested):
[AUTOMATED]

## Code Review — `refactor(sdk): simplify internals, add caching, and extend contract addresses`

Overall the refactoring is well-structured: the contract extension flow avoids keypair regeneration, the caching is identity-keyed, and the `#withTransaction` dedup is a nice cleanup. Found **2 medium** and **3 minor** issues that should be addressed before merge.

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Medium | credentials-manager.ts | Cache fields not cleared on `revoke()`/`clear()` |
| 2 | Medium | indexeddb-storage.ts | `#withTransaction` resolves before transaction commits |
| 3 | Minor | credentials-manager.ts | `#coversContracts` uses `.toLowerCase()` instead of `getAddress()` |
| 4 | Minor | credentials-manager.test.ts | Reload persistence test shares sessionStorage |
| 5 | Nit | credentials-manager.test.ts | Caching tests couple to internal crypto.subtle call counts |

No private-key logging, plaintext persistence, or exfiltration paths found.

enitrat:
[AUTOMATED]

## Follow-up: Additional finding from security analysis

**Medium-High: Race condition in `#extendContracts` can silently drop contract authorizations**

`packages/sdk/src/token/credentials-manager.ts`, lines 149-154 and 169-176

Two concurrent `allow()` calls with different contract addresses (e.g., `Promise.all([creds.allow("0xA"), creds.allow("0xB")])`) can both read the same base credentials, each merge only their own address, then persist. Last write wins, silently dropping the other address's authorization.

The `#createPromise` / `#createPromiseKey` deduplication guard at lines 188-201 only protects the fresh `create()` path. The two `#extendContracts` entry points (line 150 with active session, line 172 without session) have no serialization.

**Mitigation**: The standard usage pattern (`await creds.allow("0xA", "0xB")` — all addresses in one call) is safe, and `ReadonlyToken.allow(...tokens)` also uses this pattern. However, the API surface allows the racy pattern, and it's not documented as unsafe.

**Suggested fix**: Add a promise-coalescing guard around the extend path, similar to the existing `#createPromise` pattern. Alternatively, document that concurrent `allow()` calls with different addresses are not supported.

Confidence: 88/100

packages/sdk/src/token/credentials-manager.ts
ghermet: Addressed in e05debbe — added `#clearCryptoCache()` helper that nulls all four cache fields, called from both `revoke()` and `clear()`.

packages/sdk/src/token/credentials-manager.ts
ghermet: Addressed in fb722cd1 — `#coversContracts` now uses `getAddress()` on both sides. Also added `isAddress()` validation in `#assertEncryptedCredentials` for each element of `contractAddresses`.

packages/sdk/src/token/indexeddb-storage.ts
ghermet: Addressed in e05debbe — `#withTransaction` now resolves on `tx.oncomplete` for write transactions and `request.onsuccess` for reads, as suggested.

packages/sdk/src/token/__tests__/credentials-manager.test.ts
ghermet: Addressed in e05debbe — the "persists extended credentials across instances" test now uses `createMockStorage()` for manager2's sessionStorage to properly simulate a page reload.

packages/sdk/src/token/__tests__/credentials-manager.test.ts
ghermet: Addressed in 9b919223 — replaced `crypto.subtle.digest` spies with `CredentialsManager.computeStoreKey` spies (resilient to hash algorithm changes). Improved deriveKey test names and tightened assertions. The spy-on-`deriveKey` approach is kept for the derivation tests since it's the most direct way to verify caching, but the spy is now scoped tightly (set up after baseline calls, asserting `not.toHaveBeenCalled`).

ghermet:
Addressed in 9b919223 — added `#extendLock` serialization guard on the `#extendContracts` path. Concurrent `allow()` calls now chain: the second waits for the first to complete, then either short-circuits if the previous extension already covers the required contracts, or uses the previous result as its base to avoid dropping addresses.

Added a test ("concurrent extensions don't drop contract addresses") that validates this with `Promise.all([allow(A, B), allow(A, C)])` and asserts all three addresses are present in the final result.

enitrat (changes requested):
## Race condition: session signature visible before matching ciphertext

`#performExtend` updates the session signature **before** persisting the re-encrypted ciphertext. A concurrent `allow()` entering during that window reads the new signature, fails to decrypt the old ciphertext, hits the `catch` at L185, deletes stored creds, and regenerates a fresh keypair covering only its own contracts — silently dropping addresses from the first caller.

**Repro:** delay `storage.set` in the first extension until after the second `allow()` reads the session → `generateKeypair` fires twice, second result loses addresses from the first.

**Test gap:** the regression test at L817 races two extensions but doesn't force the interleaving between session update and ciphertext persist, so it passes while the bug exists.

packages/sdk/src/token/credentials-manager.ts
```diff
@@ -311,38 +355,134 @@ export class CredentialsManager {
...
+    const merged = this.#mergeContracts(credentials.contractAddresses, requiredContracts);
+    const signature = await this.#signWithContracts(credentials, merged);
+    await this.#setSessionEntry(storeKey, signature);
+
+    const extended: StoredCredentials = {
+      ...credentials,
+      contractAddresses: merged,
+      signature,
+    };
+    await this.#persistCredentials(storeKey, extended);
```
enitrat: Persist ciphertext **before** updating the session signature. The current order creates a window where a concurrent reader sees `sig2` but finds ciphertext encrypted with `sig1`, fails decrypt, and regenerates — dropping the first caller's addresses.

Swapping the two writes is safe: a reader that still sees the old signature will decrypt successfully with the old key, find missing contracts, and enter the extend path itself (correct behavior). The dangerous direction is the current one (new signature, old ciphertext).

```suggestion
    await this.#persistCredentials(storeKey, extended);
    await this.#setSessionEntry(storeKey, signature);

    const extended: StoredCredentials = {
      ...credentials,
      contractAddresses: merged,
      signature,
    };
```

Note: this also needs a regression test that forces the interleaving — delay `storage.set` in the first extension until after the second `allow()` reads the session, then assert no extra `generateKeypair` call and that all addresses survive.

ghermet: Addressed in 24f0961d — swapped persist order in both `#performExtend` and `create()`: ciphertext is now written before the session signature. A concurrent reader that still sees the old signature will decrypt successfully; one that sees the new signature will find matching ciphertext. All 979 tests pass.

enitrat (approved):

PR #103, opened 2026-03-11
fix(sdk): inline worker code via blob URL and optimize build

ghermet:
## Summary
- Inline worker code via blob URL to eliminate Vite `optimizeDeps` workaround for consuming apps
- Refactor build to on-the-fly IIFE build, removing the two-pass rolldown config
- Switch sdk vitest pool from `vmThreads` to `vmForks` for ~18% faster tests

## Test plan
- [x] All 747 sdk tests pass
- [x] All 211 react-sdk tests pass
- [ ] Verify worker loads correctly in Vite and Next.js test apps
- [ ] Verify no `optimizeDeps` config needed in consuming apps


packages/sdk/src/worker/worker.client.ts
```diff
@@ -7,6 +7,11 @@ import type {
   WorkerResponse,
 } from "./worker.types";
 import { BaseWorkerClient } from "./worker.base-client";
+import workerCode from "./relayer-sdk.worker.ts?iife";
+
+const workerBlobUrl = /* @__PURE__ */ URL.createObjectURL(
+  new Blob([workerCode], { type: "application/javascript" }),
```
enitrat: [AUTOMATED] **Major — import-time browser-only side effect (95/100)**

`URL.createObjectURL(new Blob(...))` executes at module scope. The import chain is:

```
index.ts  →  export { RelayerWeb }
  → relayer-web.ts  →  import { RelayerWorkerClient } from "../worker/worker.client"
    → worker.client.ts  →  URL.createObjectURL(...)   // runs at import time
```

So `import { RelayerWeb } from "@zama-fhe/sdk"` now eagerly allocates a blob URL before any class is instantiated. In Node/SSR environments `URL.createObjectURL` doesn't exist, turning this into an import-time crash.

The `/* @__PURE__ */` annotation helps consuming bundlers tree-shake this away when `RelayerWeb` is unused, but it doesn't help consumers who actually import `RelayerWeb` (the primary use case).

**Before (base branch):** `new URL("./relayer-sdk.worker.js", import.meta.url)` lived inside `createWorker()` — lazy, only evaluated when a worker was actually needed.

**Suggested fix:** Move the blob URL computation inside `createWorker()` (or use a lazy getter that caches on first access):

```ts
let _workerBlobUrl: string | undefined;
function getWorkerBlobUrl() {
  return (_workerBlobUrl ??= URL.createObjectURL(
    new Blob([workerCode], { type: "application/javascript" }),
  ));
}
```

packages/sdk/src/worker/worker.client.ts
```diff
@@ -31,7 +36,7 @@ export class RelayerWorkerClient extends BaseWorkerClient<Worker, WorkerClientCo
   }
 
   protected createWorker(): Worker {
-    return new Worker(new URL("./relayer-sdk.worker.js", import.meta.url));
+    return new Worker(workerBlobUrl);
   }
```
enitrat, line 47: [AUTOMATED] **High — `blob:` URL breaks browser-extension environments (90/100)**

`createWorker()` now always loads the worker from a `blob:` URL. Browser extensions with `worker-src 'self'` CSP will reject this at runtime.

Notably, the worker file itself (`relayer-sdk.worker.ts:162-166`) already documents this exact problem and includes an `isBrowserExtension()` fallback for CDN script loading inside the worker. But with a `blob:` worker URL, the worker can't even boot — the extension-awareness code inside becomes unreachable.

The old `new Worker(new URL("./relayer-sdk.worker.js", import.meta.url))` loaded from a same-origin file URL that extensions allow.

**Suggestion:** Provide a fallback path (e.g., detect extension context before choosing blob vs file-based worker, or accept an optional worker URL/factory in the config).

vitest.config.ts
```diff
@@ -1,6 +1,21 @@
-import { defineConfig } from "vitest/config";
+import { defineConfig, type Plugin } from "vitest/config";
 import path from "path";
 
+/** Stub `?iife` imports in tests — returns an empty string instead of file contents. */
+function iifeStub(): Plugin {
+  return {
+    name: "iife-stub",
```
enitrat, line 7: [AUTOMATED] **Medium — `?iife` stub means the riskiest new code path is untested (98/100)**

`iifeStub()` returns `export default '';` for all `?iife` imports, so tests never exercise:
- `iife-plugin.ts` actually bundling the worker as IIFE
- The blob URL being created with real worker code
- The worker booting from that blob URL

The new bundling pipeline (`iife-plugin` → inline string → blob URL → `new Worker(...)`) is the core change in this PR, yet it has zero end-to-end test coverage. The existing worker tests validate message-passing control flow against mocks, which is fine, but they can't catch regressions in the bundling/loading path itself.

Consider adding at least a build-level smoke test that verifies `dist/index.js` contains a valid worker payload and that `new Worker(blobUrl)` doesn't throw.

enitrat (changes requested):
[AUTOMATED]

## Code Review Summary

The PR solves a real problem (Vite `optimizeDeps` friction for consuming apps) but introduces regressions for SSR, browser-extension, and CSP-restricted environments. The two unchecked test-plan items ("Verify worker loads correctly in Vite and Next.js test apps" / "Verify no optimizeDeps config needed") are exactly the scenarios where these regressions would surface.

### Findings

| # | Severity | Issue | Confidence |
|---|----------|-------|------------|
| 1 | Major | Import-time `URL.createObjectURL` is a browser-only side effect | 95/100 |
| 2 | High | `blob:` URL breaks browser-extension CSP | 90/100 |
| 3 | Medium | `?iife` vitest stub doesn't exercise production bundling path | 98/100 |

### Recommendation

Request changes. Consider keeping the worker instantiation lazy (inside `createWorker()`) — e.g. compute the blob URL on first call and cache it — so that importing the SDK doesn't eagerly touch browser-only APIs. For the extension CSP issue, a fallback to a file-based worker URL (or a configurable worker factory) would preserve the current portability.

enitrat (dismissed):

ankurdotb (dismissed):

ankurdotb (approved):

PR #105, opened 2026-03-11
test: make test apps compatible with injected providers

enitrat:
Current way of handling injected providers would cause some issues on connection via injected provider (MM window not opening); so I propose instead using the burnWallet.pk from localstorage as discriminator to know whether to use the injected provider or the burner wallet.

### DIFF
--- a/packages/test-components/src/connect-wallet.tsx
+++ b/packages/test-components/src/connect-wallet.tsx
@@ -2,6 +2,22 @@
 
 import { useAccount, useConnect, useDisconnect, useConnectors } from "wagmi";
 
+const BURNER_PK_KEY = "burnerWallet.pk";
+
+/**
+ * Picks the right connector based on environment:
+ * - If a burner private key is in localStorage, use the burner connector (Playwright / E2E)
+ * - Otherwise, use the injected connector (MetaMask / browser wallet)
+ */
+function pickConnector(connectors: readonly { id: string; name: string }[]) {
+  const hasBurnerPk = typeof window !== "undefined" && !!localStorage.getItem(BURNER_PK_KEY);
+
+  if (hasBurnerPk) {
+    return connectors.find((c) => c.id === "burnerWallet") ?? connectors[0];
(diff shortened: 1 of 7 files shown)


packages/test-components/src/connect-wallet.tsx
```diff
@@ -26,7 +42,11 @@ export function ConnectWallet() {
 
   return (
     <button
-      onClick={() => connect({ connector: connectors[0]! })}
+      onClick={() => {
+        const connector = pickConnector(connectors);
+        if (connector)
+          connect({ connector: connector as Parameters<typeof connect>[0]["connector"] });
+      }}
```
ghermet, line 49: Doesn't it auto connect if `burnerWallet.pk` exists?

packages/test-nextjs/package.json
```diff
@@ -7,7 +7,7 @@
   "author": "Zama",
   "scripts": {
     "dev": "next dev --port 3100",
-    "dev:e2e": "NEXT_PUBLIC_NETWORK=hardhat next dev --port 3100",
+    "dev:e2e": "next dev --port 3100",
```
ghermet: if you're not gonna pin the network in this command you can remove it

packages/test-vite/package.json
```diff
@@ -4,7 +4,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite --port 3200",
-    "dev:e2e": "VITE_NETWORK=hardhat vite --port 3200",
+    "dev:e2e": "vite --port 3200",
```
ghermet: if you're not gonna pin the network in this command you can remove it

packages/test-components/src/connect-wallet.tsx
enitrat: yes, that's was my idea, do you think it's not a good one? because it will always be injected by PW but never be there in human-controlled sessions

ghermet (approved):

PR #109, opened 2026-03-13
refactor: remove zaiffer-contracts dep, parallelize e2e tests

enitrat:
## Summary

Replace the private `hardhat` submodule (`zaiffer-smart-contracts`) with a self-contained Foundry setup using the public `forge-fhevm` submodule. The test contracts themselves are unchanged — only the tooling that compiles and deploys them changes. As a side effect, each Playwright project (nextjs, vite) now gets its own isolated Anvil node, allowing both test suites to run in parallel and cutting overall e2e time roughly in half.

## High-level things to review

- **CI workflow** (`.github/workflows/playwright.yml`): Hardhat install/compile/cache steps (~50 lines) are gone, replaced by Foundry install + `forge build` with cache. The `SUBMODULE_TOKEN` secret is no longer required since `forge-fhevm` is a public submodule.
- **Fixture refactor** (`packages/playwright/fixtures/test.ts`): `viemClient` promoted to a **worker-scoped** fixture so each Playwright worker connects to its own Anvil port. `readErc20Balance` is now a factory closure that captures the right client.
- **Playwright config** (`packages/playwright/playwright.config.ts`): `workers` raised from 1 to 2. Each project (`nextjs`, `vite`) declares `workers: 1` and receives a dedicated `anvilPort` fixture value. The `hardhat node` web server entry is replaced by a simple anvil node on which contracts are auto-deployed.
- **Forge contracts** (`contracts/`): New `foundry.toml`, `remappings.txt`, and `Deploy.s.sol` deploy script. Contract source files (ERC7984, wrappers, FeeManager, etc.) are identical to what was in the private Hardhat submodule — no logic changes.

## Key decisions

| Decision | Rationale |
|---|---|
| **Keep contract logic identical** | The goal is a tooling migration, not a contract change. Reviewers don't need to re-audit business logic — only deployment wiring. |
| **Two Anvil nodes instead of one shared node** | Playwright's two projects previously shared a single Hardhat node (serialised). Giving each its own node removes all cross-project state pollution and lets them run in parallel. |
| **Same deployer key + same nonce on both nodes → same addresses** | Both nodes start from genesis with Anvil account #0 as deployer. Because nonces are identical, `forge` produces the same contract addresses on both chains, so a single `contracts/deployments.json` is valid for both test projects — no per-project address config needed. |
| **`forge-fhevm` as a public submodule** | Removes the `SUBMODULE_TOKEN` CI secret and the dependency on the internal `zaiffer-smart-contracts` repo. |
| **Worker-scoped `viemClient` fixture** | Playwright worker scope is the right granularity: one client per parallel worker, each talking to its own Anvil port. Test-scoped would recreate the client per test, which is unnecessary overhead. |
| **`forge build` cache keyed on `src/**/*.sol` + `foundry.toml`** | Mirrors the previous Hardhat cache strategy — only rebuilds when Solidity sources or compiler config change. |


packages/sdk/src/relayer/relayer-utils.ts
```diff
@@ -97,7 +97,7 @@ export const HardhatConfig = {
   network: "http://127.0.0.1:8545",
   aclContractAddress: "0x50157CFfD6bBFA2DECe204a89ec419c23ef5755D",
   inputVerifierContractAddress: "0x36772142b74871f255CbD7A3e89B401d3e45825f",
-  kmsContractAddress: "0xbE0E383937d564D7FF0BC3b46c51f0bF8d5C311A",
+  kmsContractAddress: "0x901F8942346f7AB3a01F6D7613119Bca447Bb030",
```
enitrat, line 102: two wrongs made a right, but the correct value here is the one that aligns with the `ZamaConfig.sol` contract (that I corrected here)

.github/workflows/playwright.yml
```diff
@@ -133,6 +107,8 @@ jobs:
 
       - name: Build Vite test-vite
         run: pnpm build:test-vite
+        env:
+          VITE_ANVIL_PORT: 8546
```
ghermet, line 111: Why is this no 8545?

ghermet: Are you spinning 2 instances for parallel testing?

ghermet (commented):
## PR Review Summary

### Critical Issues (2 — must fix)

**1. Unused `readFileSync` import**
`packages/playwright/global-setup.ts:2`
```ts
import { readFileSync } from "node:fs"; // never used
```
Remove it.

**2. Unnecessary `env` spread in `execSync`**
`packages/playwright/global-setup.ts:70-74`
```ts
env: {
  ...process.env,  // identical to omitting env entirely
},
```
`execSync` inherits `process.env` by default. Remove the `env` option.

### Important Issues (3 — should fix)

**3. `mockRelayerSdk` import removed — verify replacement**
`packages/playwright/fixtures/test.ts` — The old `mockRelayerSdk` import from `./fhevm` was dropped. Confirm the FHE relayer mock is now handled by `forge-fhevm`'s `deploy-local.sh`, otherwise tests may silently skip mocking.

**4. Anvil spawn failure gives 30s timeout instead of immediate error**
`packages/playwright/global-setup.ts:39-49` — If `anvil` isn't on PATH, `proc.on("error")` only writes to stderr. `waitForPort` waits the full 30s before timing out with a confusing "port not ready" error. Consider propagating spawn errors to fail fast.

**5. Duplicate private key**
`packages/playwright/global-setup.ts:10` — `ANVIL_DEPLOYER_PK` is the same key as `TEST_PRIVATE_KEY` in `fixtures/constants.ts`. Import from constants instead of duplicating.

### Suggestions (nice to have)

- `global-setup.ts:12-34` — `waitForPort` could be simplified to an async/await retry loop (eliminates the `settled` flag)
- `global-setup.ts:37` — Anvil `stdio: "ignore"` loses crash diagnostics. Consider piping stderr to a log file for CI debugging.
- `playwright.config.ts:28,39` — Redundant `testDir: "./tests"` on both projects (already set at top level)
- `deployments.json` — Committed but also generated at runtime. Add a comment explaining why it's checked in (deterministic addresses).
- `playwright.yml:99,111` — Hardcoded ports `8545`/`8546` duplicate values from `constants.ts`. Add a sync comment.
- `test-nextjs` vs `test-vite` providers — Near-identical files. Consider extracting shared config into `@zama-fhe/test-components`.

### Strengths

- Clean global-setup/teardown with proper error handling — no orphaned Anvil processes
- Deterministic addresses via same-nonce strategy — elegant single `deployments.json`
- Worker-scoped `viemClient` fixture with closure-captured `readErc20Balance`
- Public submodule eliminates `SUBMODULE_TOKEN` secret
- Proper CI caching keyed on `.sol` + `foundry.toml`
- No silent error swallowing anywhere

Overall this is a well-architected migration. The critical items are trivial cleanups, and the important items are worth addressing before merge.

packages/playwright/global-setup.ts
```diff
@@ -0,0 +1,89 @@
...
+import net from "node:net";
+import { resolve, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+import { NEXTJS_ANVIL_PORT, VITE_ANVIL_PORT } from "./fixtures/constants";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const contractsDir = resolve(__dirname, "../../contracts");
+const ANVIL_DEPLOYER_PK = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"; // Anvil account #0
+
+function waitForPort(port: number, timeoutMs = 30_000): Promise<void> {
```
ghermet: there's a wait for port and wait for stdout baked in playwright you can use.

ghermet:
```sh
#!/usr/bin/env bash
# Start an anvil instance, deploy contracts, then keep anvil in the foreground.
# Usage: ./start-anvil.sh <port>
set -euo pipefail

PORT="${1:?Usage: start-anvil.sh <port>}"
SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
CONTRACTS_DIR="$(cd "$SCRIPT_DIR/../../contracts" && pwd)"
FORGE_FHEVM_DIR="$CONTRACTS_DIR/lib/forge-fhevm"
DEPLOYER_PK="0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"

# Start anvil in the background so we can deploy to it
anvil --port "$PORT" --chain-id 31337 --silent &
ANVIL_PID=$!
trap "kill $ANVIL_PID 2>/dev/null" EXIT

# Wait for anvil to be ready
for i in $(seq 1 150); do
  if nc -z 127.0.0.1 "$PORT" 2>/dev/null; then
    break
  fi
  sleep 0.2
done

if ! nc -z 127.0.0.1 "$PORT" 2>/dev/null; then
  echo "Anvil on port $PORT failed to start" >&2
  exit 1
fi

# Deploy fhevm host stack
"$FORGE_FHEVM_DIR/deploy-local.sh" --anvil-port "$PORT"

# Deploy project contracts
(cd "$CONTRACTS_DIR" && forge script script/Deploy.s.sol \
  --rpc-url "http://127.0.0.1:$PORT" \
  --broadcast --silent \
  --sender 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 \
  --private-key "$DEPLOYER_PK")

echo "Anvil ready on port $PORT"

# Keep the process alive — Playwright will kill it on teardown.
# Remove the trap so we don't double-kill.
trap - EXIT
wait $ANVIL_PID
```

```ts
  webServer: [
    {
      command: `./start-anvil.sh ${NEXTJS_ANVIL_PORT}`,
      port: NEXTJS_ANVIL_PORT,
      wait: {
        stdout: /Anvil ready on port (\d+)/,
      },
      reuseExistingServer: !CI,
    },
    {
      command: `./start-anvil.sh ${VITE_ANVIL_PORT}`,
      port: VITE_ANVIL_PORT,
      wait: {
        stdout: /Anvil ready on port (\d+)/,
      },
      reuseExistingServer: !CI,
    },
]

File: .github/workflows/playwright.yml (hunk 2 above)
Comment (enitrat) [reply .github/workflows/playwright.yml:111]: > Global test setup (packages/playwright/global-setup.ts): New file that spawns two Anvil processes (ports 8545 / 8546), waits for them to be ready, deploys the fhevm host stack via forge-fhevm's deploy-local.sh to both nodes in parallel, then runs forge script Deploy.s.sol against each node individually.

File: packages/playwright/global-setup.ts (hunk 3 above)
Comment (enitrat) [reply packages/playwright/global-setup.ts]: great, I didn't know about`wait`

File: package.json [hunk 4]
```diff
@@ -27,10 +27,10 @@
     "format:check": "oxfmt --check",
     "release:ci": "pnpm exec semantic-release",
     "release:dry-run": "pnpm exec semantic-release --dry-run --no-ci --debug",
+    "postinstall": "git submodule update --init --recursive && (cd contracts/lib/forge-fhevm && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install) && (cd contracts && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install)",
     "prepare": "husky",
-    "submodule:init": "git submodule update --init --recursive",
     "submodule:update": "git submodule update --remote --merge",
-    "contracts:deps": "cd contracts/lib/forge-fhevm && forge soldeer install",
+    "contracts:deps": "(cd contracts/lib/forge-fhevm && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install) && (cd contracts && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install)",
```
ghermet: looks like we need 2 new scripts:
```sh
"forge-fhevm:install": "cd contracts/lib/forge-fhevm && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install",
"contracts:install": "cd contracts && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install"
```

enitrat: I collapsed them as post-install

package.json
```diff
@@ -27,12 +27,13 @@
     "format:check": "oxfmt --check",
     "release:ci": "pnpm exec semantic-release",
     "release:dry-run": "pnpm exec semantic-release --dry-run --no-ci --debug",
+    "postinstall": "git submodule update --init --recursive && pnpm forge-fhevm:install && pnpm contracts:install",
+    "forge-fhevm:install": "cd contracts/lib/forge-fhevm && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install",
+    "contracts:install": "cd contracts && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install",
```
ghermet, line 32: What happens if you don't have foundry installed?

enitrat: that will fail. see CONTRIBUTING.md

ghermet (approved):

PR #110, opened 2026-03-16
feat(sdk): persist FHE public key and params with artifact-level revalidation

ghermet:
## Summary

- Adds **persistent caching** for FHE public key and CRS params via the new `ArtifactCache` class, so they are not re-downloaded on every app load
- Adds **artifact-level revalidation** using conditional HTTP requests (ETag / Last-Modified) to detect stale cached material after key rotation
- **Defaults RelayerWeb to IndexedDB** artifact cache — zero config needed for browser apps
- Renames config fields to `artifactStorage` and `artifactCacheTTL` for clarity
- Adds **GitBook reference page** for `ArtifactCache`

## Problem

Every time the app is instantiated (e.g. page reload), the SDK re-downloads the FHE public key and public params from the relayer. These are large binary payloads that rarely change, resulting in unnecessary network overhead and slower startup times.

### `ArtifactCache` class (`artifact-cache.ts`)
- **Two-layer cache**: in-memory (instant) + persistent storage (survives app restarts)
- Uses the existing `GenericStorage` abstraction (IndexedDB in browser, any async key-value store)
- `Uint8Array` data is base64-encoded for storage, using chunked encoding (8KB chunks) to avoid stack overflow on large arrays
- Cache keys scoped by chain ID (`fhe:pubkey:{chainId}`, `fhe:params:{chainId}:{bits}`)
- Inflight request deduplication prevents concurrent duplicate fetches
- Best-effort storage writes — persistence failures don't break the flow

### Artifact-level revalidation (`revalidateIfDue`)
- Fetches `/keyurl` manifest to discover current artifact URLs
- Uses `ETag`/`If-None-Match` and `Last-Modified`/`If-Modified-Since` conditional HTTP requests
- Checks both public key and all cached CRS param bit-sizes
- Returns `true` (stale) if any artifact has changed, triggering full worker teardown and re-init
- **Fail-open**: network errors use short retry (5 min) instead of full TTL to prevent retry storms
- Configurable via `artifactCacheTTL` (default: 24h, set to 0 for every-startup checks)

### Default IndexedDB storage
- `RelayerWeb` now defaults to `new IndexedDBStorage("FheArtifactCache", 1, "artifacts")` — no config needed
- `IndexedDBStorage` is now configurable (database name, version, store name)

### Config options (non-breaking)
- `RelayerWebConfig.artifactStorage?: GenericStorage`
- `RelayerWebConfig.artifactCacheTTL?: number`
- `RelayerNodeConfig.artifactStorage?: GenericStorage`
- `RelayerNodeConfig.artifactCacheTTL?: number`

Fully backward compatible — defaults to IndexedDB in browser.

### Documentation
- New GitBook reference page: `reference/sdk/ArtifactCache.md`
- Covers constructor, methods, relayer config, revalidation behavior, and storage key patterns

## Test plan

- [x] Unit tests for `ArtifactCache`: cache hit/miss, persistence, chain isolation, null handling, storage failure resilience, concurrent deduplication, revalidation (304, ETag change, Last-Modified change, URL change, network error fail-open, chain isolation)
- [x] Integration test for `ArtifactCache` with real HTTP
- [x] All existing tests pass
- [x] API reports updated

## Related

Closes SDK-7


packages/sdk/etc/sdk-node.api.md
```diff
@@ -540,6 +540,8 @@ export interface RelayerNodeConfig {
     logger?: GenericLogger;
     // (undocumented)
     poolSize?: number;
+    revalidateIntervalMs?: number;
```
ankurdotb: Is this in milliseconds? Our time-related settings are in seconds (e.g., keypairTTL, sessionTTL) so let's stick that as a convention.

Secondly, it's better to be more descriptive. Revalidate what? Better to call it something like `fheKeyUrlCacheTTL` or similar no?

ankurdotb:
# PR Review: Persistent Caching for FHE Public Key and CRS Params

## Critical Finding: Revalidation is Non-Functional in Production

The revalidation logic compares `dataId` fields from the `/keyurl` manifest against cached values. However, **in production the `dataId` fields are static placeholders**:

```json
{
  "fhePublicKey": { "dataId": "fhe-public-key-data-id", "urls": ["https://...s3..."] },
  "crs": { "2048": { "dataId": "crs-data-id", "urls": ["https://...s3..."] } }
}
```

`"fhe-public-key-data-id"` and `"crs-data-id"` are hardcoded strings that **never change**, making the entire revalidation flow a no-op. Key rotation will never be detected.

**What's needed (as SDK-7 originally specified):** Artifact-level validation using HTTP conditional requests against the actual S3/CDN artifact URLs:

1. Fetch manifest from `/keyurl` to get artifact URLs
2. For each cached artifact, issue a conditional GET with `If-None-Match: "{stored_etag}"` to the artifact URL
3. **304** → cache is fresh, update `lastValidatedAt`
4. **200** → artifact changed, replace cached bytes + stored `etag`/`lastModified`

S3 URLs reliably serve `ETag` and `Last-Modified` headers, so this approach is robust regardless of CDN backend.

**Stored metadata per artifact should include** (per SDK-7):
- `artifactUrl` — the S3/CDN URL from the manifest
- `etag` — from the HTTP response
- `lastModified` — from the HTTP response
- `lastValidatedAt` — epoch timestamp of last successful check

The PR description claims "Uses `ETag`/`If-None-Match` and `Last-Modified`/`If-Modified-Since` conditional HTTP requests" but the implementation does not use any HTTP headers.

---

## Critical Issues

### 1. Cold-start CRS rotation is undetectable

**`public-params-cache.ts:337-351`**

`#collectParamEntries()` iterates `this.#publicParamsMem.keys()`, which is empty after a process restart. On cold start, `revalidateIfDue()` runs before `getPublicParams()` is called, so the in-memory map has no entries. Result: CRS rotation is never detected until `getPublicParams()` happens to be called first in the new session.

The test suite masks this because `seedAndPrime()` always calls `getPublicParams(2048, ...)` before any revalidation test, priming the in-memory map. A test that creates a fresh `PublicParamsCache`, seeds only storage (not memory), and then calls `revalidateIfDue()` would expose this.

**Fix:** Persist a params index key (e.g. `fhe:params-index:{chainId}`) recording all stored `bits` values, so `#collectParamEntries` can discover them without relying on the in-memory map.

### 2. All error logging is silent without the optional logger

**Every catch block in `public-params-cache.ts`**

Every log call uses `this.#logger?.warn(...)`. When no logger is provided (it's optional in both `RelayerWebConfig` and `RelayerNodeConfig`), all error paths — storage corruption, network failures, quota exceeded — produce zero output.

**Fix:** Use a fallback logger that writes to `console.warn`/`console.error` when none is provided. There's precedent in `balance-cache.ts` which uses `console.warn` directly.

### 3. Fail-open pushes timestamps by full TTL on transient failure

**`public-params-cache.ts:276-284, 314-334`**

When the manifest fetch returns 500 or a network error occurs, `#updateValidationTimestamps` pushes `lastValidatedAt` forward by the full `revalidateIntervalMs` (default 24h). This means a transient relayer outage suppresses revalidation for 24 hours. For cryptographic material, that's too long.

**Fix:** On failure, use a shorter retry interval (e.g. 5 minutes) instead of the full revalidation interval.

---

## Important Issues

### 4. No integrity validation of cached data

**`public-params-cache.ts:109-118, 177-186`**

`this.#storage.get<CachedPublicKey>(key)` is an unchecked type assertion. Corrupt or schema-mismatched storage entries (e.g. from a previous SDK version) propagate silently or cause opaque `atob` errors.

**Fix:** Add a lightweight validation guard: `typeof v.publicKeyId === 'string' && typeof v.publicKey === 'string'` before trusting stored data. On failure, delete the corrupt entry and fall through to network fetch.

### 5. `#updateValidationTimestamps` mutates storage-returned objects in place

**`public-params-cache.ts:375-377`**

`storedPk.lastValidatedAt = now` mutates the object returned by `storage.get()`. For `MemoryStorage` (which returns the same object reference), this mutates the live stored value before confirming the write succeeded. Use `{ ...storedPk, lastValidatedAt: now }` instead.

### 6. No concurrency guard on `revalidateIfDue`

**`public-params-cache.ts:230`**

Unlike `getPublicKey`/`getPublicParams` which use inflight deduplication, `revalidateIfDue` has none. While the relayer classes serialize via `#ensureLock`, the cache class should be self-protecting. Add a `#revalidationInflight: Promise<boolean> | null` guard.

### 7. Broad catch block masks programming errors

**`public-params-cache.ts:314-334`**

The outer try-catch wraps storage reads, network fetch, JSON parsing, and writes. A `TypeError` from a bug is treated identically to a transient network error. Log the full stack trace and consider distinguishing error categories.

### 8. Worker teardown on stale cache has no user-visible logging

**`relayer-web.ts:172-177`, `relayer-node.ts:117-122`**

When revalidation triggers `#tearDown()`, users experience sudden WASM re-initialization delay with no explanation. Add `logger.info("Cached FHE artifacts are stale — reinitializing")`.

### 9. Extract manifest type to a named interface

**`public-params-cache.ts:285-288`**

The inline `as` cast is hard to discover and maintain. Extract to a private `RelayerManifest` interface alongside the other cache types.

---

## Naming and Convention Issues

### 10. `revalidateIntervalMs` should be `fheKeyUrlCacheTTL` in seconds

The project convention is seconds for time-related config (`keypairTTL: 86400`, `sessionTTL: 2592000`). This field should follow suit:

```typescript
/** Seconds cached /keyurl artifacts (FHE public key + CRS) are trusted
 *  before re-checking. Default: 86400 (1 day). 0 = check every operation.
 *  Ignored when storage is not set. */
fheKeyUrlCacheTTL?: number;
```

### 11. PR description references `/v2/keyurl` but code uses `/keyurl`

Minor but worth aligning.

### 12. PR description claims 19 tests, actual count is 23

Test plan section is out of date.

---

## Suggestions

- **`lastValidatedAt` should be required** (default `0`) in `CachedPublicKey`/`CachedPublicParams` — eliminates `undefined`/`number` ambiguity
- **Document the three-state semantics** of `#publicKeyMem` (`undefined` = not loaded, `null` = fetcher returned null, non-null = loaded)
- **Use `Promise.allSettled`** in `#collectParamEntries` so one corrupt entry doesn't block all revalidation
- **Add `RelayerNode` cross-instance persistence test** to match the existing `RelayerWeb` one
- **Add unit test for `fheKeyUrlCacheTTL: 0`** (always-check mode)

---

## Strengths

- Persistent caching design is solid — two-layer cache (memory + storage) eliminates repeated large downloads
- Inflight deduplication for `getPublicKey`/`getPublicParams` prevents thundering herd
- Fail-open pattern is correct in principle (availability over freshness)
- Fully backward compatible — no storage = no caching
- Chain isolation via scoped storage keys
- Thorough test coverage for the implemented behavior (23 unit + integration tests)
- Good use of `readonly` private fields for immutable config
- Base64 chunking in `toBase64` correctly avoids call-stack overflow on large buffers

---

## Summary

| Priority | Count | Key Items |
|----------|-------|-----------|
| **Critical** | 3 | Revalidation is non-functional (dataId is a static placeholder), cold-start CRS detection bug, silent error logging |
| **Important** | 6 | No cache integrity validation, object mutation, no concurrency guard, broad catch, no teardown logging, inline manifest type |
| **Convention** | 3 | Rename to `fheKeyUrlCacheTTL` in seconds, PR description inaccuracies |
| **Suggestion** | 5 | Required `lastValidatedAt`, three-state docs, `Promise.allSettled`, additional tests |

### Recommended Action

1. **Implement artifact-level HTTP revalidation** per SDK-7 — conditional GET with `If-None-Match` against S3/CDN artifact URLs. The `dataId` comparison does not work with the current relayer.
2. **Rename** `revalidateIntervalMs` → `fheKeyUrlCacheTTL` (seconds, default `86400`)
3. **Fix** cold-start CRS detection, add fallback console logger, use shorter retry on failure
4. **Update PR description** to accurately reflect implementation
5. Address important issues before merge

ankurdotb (changes requested):
This PR completely misses the context of what's specified in Linear and will never work, because `dataId` from the keyurl endpoint literally never changes currently.

packages/sdk/src/relayer/__tests__/fhe-artifact-cache.integration.test.ts
```diff
@@ -0,0 +1,234 @@
...
+  baseUrl: string;
+  rotatePk: () => void;
+  rotateCrs: () => void;
+}
+
+function createTestServer(): { server: http.Server; start: () => Promise<TestServer> } {
+  let pkEtag = '"pk-etag-v1"';
+  let crsEtag = '"crs-etag-v1"';
+  const pkPath = "/artifacts/pk.bin";
+  const crsPath = "/artifacts/crs-2048.bin";
```
gasp, line 24: do these file really exist? I can't see them 🤷🏼‍♂️ 
you don't need to upload the file, you can also generate it with something like

```
import { writeFileSync } from "fs";
import { randomBytes } from "crypto";

export function generateFakeBinFile(filePath: string, size: number): void {
  if (size <= 0) throw new Error("Size must be greater than 0");
  writeFileSync(filePath, randomBytes(size));
}
```

gasp: or even don't store them in the fs and generate them on the fly 🐦‍⬛

packages/sdk/etc/sdk.api.md
```diff
@@ -4108,10 +4108,12 @@ export class RelayerWeb implements RelayerSDK {
 
 // @public
 export interface RelayerWebConfig {
+    fheArtifactCacheTTL?: number;
     getChainId: () => Promise<number>;
     logger?: GenericLogger;
     onStatusChange?: (status: RelayerSDKStatus, error?: Error) => void;
     security?: RelayerWebSecurityConfig;
+    storage?: GenericStorage;
```
gasp: The ZamaProvider.storage that users pass (e.g. `indexedDBStorage`) is for **credentials only**. The artifact cache storage on `RelayerWebConfig.storage` is a **separate optional field** : if omitted, artifacts are re-fetched every session. If provided in a browser context, it must be `indexedDBStorage` because public params can be many MBs.

The problem is that nothing prevents a browser user from passing `localStorage`-backed storage to `RelayerWebConfig.storage`. The docs currently just say `GenericStorage` without guidance.

The fix should be: document (and ideally default) `RelayerWebConfig.storage` to `indexedDBStorage` when used in the browser. But maybr thid should be done on a higher level (like sdk-web) because it doesn't affect nodejs where storage hardly have size limitations, event [Redis can support 512M keys](https://redis.io/docs/latest/develop/using-commands/keyspace/#:~:text=The%20maximum%20allowed%20key%20size%20is%20512%20MB.)

gasp: maybe in the types, having a LargeStorage alias would help avoid confusion

packages/sdk/src/relayer/relayer-web.ts
gasp, line 37: (unrelated), how can you have a package integrity SHA if ${RELAYER_SDK_VERSION} is a variable? what's this wizardy 🧙🏼 ?

packages/sdk/etc/sdk.api.md
gasp: or enforcing it (I'm not sure that it's a good idea…)
Image

ghermet: Agreed, let's make the `indexedDBStorage` as the default storage unless specified otherwise.

packages/sdk/src/relayer/relayer-web.ts
ghermet: the SHA-384 hex digest was precalculated for the given version. However, This is something that needs to be addressed at some point as the CDN might to become legacy.

packages/sdk/etc/sdk.api.md
ghermet: we could also set indexedDBStorage as the storage for FheArtefactCache

packages/sdk/src/relayer/__tests__/fhe-artifact-cache.integration.test.ts
ghermet: Updated code generates artifacts with `randomBytes` on the server

packages/sdk/etc/sdk.api.md
```diff
@@ -4106,6 +4106,8 @@ export class RelayerWeb implements RelayerSDK {
 
 // @public
 export interface RelayerWebConfig {
+    artifactCacheTTL?: number;
```
ankurdotb: I honestly don't think this is obvious enough. Yes, the TTL relates to an artefact but the way most developers who've been using Relayer SDK so far is that this is the FHE public key they need to download/cache.

I'm open to suggestions though. I did consider `fhePublicKeyCacheTTL` and that works too I suppose, I was trying to account for the fact that it applies to both the key and CRS. The other one I suggested earlier was `fheKeyUrlCacheTTL` since a lot of the existing developers are also familiar of fetching from the keyurl path. I didn't think of calling it _just_ `keyUrlCacheTTL` since it doesn't clarify which sort of key URL.

packages/sdk/src/relayer/relayer-web.ts
ankurdotb: Do developers/apps currently guard against a supply chain attack by cross-checking the content integrity of the relayer package published by CDN currently? I mean for example in Relayer SDK.

Trouble is, according to our package file, we should already be on v0.4.2, not v0.4.1!

I'm inclined to not have it? Open to thoughts from @gasp if this is something you guys used to do as a check or still do as a check.

ankurdotb (changes requested):
See comments. Also, what about documentation?

packages/sdk/src/relayer/relayer-web.ts
ghermet: @ankurdotb I've address this issue in the following [PR](https://github.com/zama-ai/sdk/pull/127), it bundles the cdn script from the relayer-sdk dependency

packages/sdk/etc/sdk.api.md
ghermet: So would `fheArtifactCacheTTL` be a better option?

ankurdotb: I would use `fheKeyUrlCacheTTL` as the least ambiguous and most likely to make sense to developers.

ankurdotb: Never mind, I'm fine with this

docs/gitbook/src/reference/sdk/FheArtifactCache.md
```diff
@@ -0,0 +1,205 @@
...
+```ts
+import { RelayerWeb, SepoliaConfig } from "@zama-fhe/sdk";
+
+// RelayerWeb uses IndexedDB artifact cache by default — no config needed
+const relayer = new RelayerWeb({
+  getChainId: () => signer.getChainId(),
+  transports: {
+    [SepoliaConfig.chainId]: {
+      ...SepoliaConfig,
+      relayerUrl: "https://your-app.com/api/relayer/11155111",
```
ankurdotb: Bad example. Either don't set, or show relayer.testnet.zama.org as the example. And if it's proxy scenario, then it should be via the `proxyUrl` parameter right?

docs/gitbook/src/reference/sdk/FheArtifactCache.md
```diff
@@ -0,0 +1,205 @@
+---
+title: FheArtifactCache
+description: Persistent cache for FHE public key and public params, avoiding re-downloads across sessions.
+---
+
+# FheArtifactCache
+
+Persistent cache for the FHE network public key and public params (CRS). Stores large binary artifacts in a `GenericStorage` backend (e.g. IndexedDB) so they are not re-downloaded on every page load. Cache keys are scoped by chain ID.
```
ankurdotb: _nitpick_ Use "parameters" rather than "params".

docs/gitbook/src/reference/sdk/FheArtifactCache.md
```diff
@@ -0,0 +1,205 @@
...
+
+```ts
+import { RelayerWeb, IndexedDBStorage, SepoliaConfig } from "@zama-fhe/sdk";
+
+const relayer = new RelayerWeb({
+  getChainId: () => signer.getChainId(),
+  transports: {
+    [SepoliaConfig.chainId]: {
+      ...SepoliaConfig,
+      relayerUrl: "https://your-app.com/api/relayer/11155111",
```
ankurdotb: Same as above

docs/gitbook/src/reference/sdk/FheArtifactCache.md
ankurdotb, line 1: This is overall a good document, nice work!

However, it is SO important that it deserves to be mentioned in the primary example of getting started.

From what I understand, this is instantiated by default for browsers (frontend scenarios), and for server-side scenarios when storage specified. Make it part of the default examples shown in the doc for relayerwb/node config.

ankurdotb:
## Remaining items

**1. Docs: RelayerNode caching is silent opt-in**

`RelayerWeb` defaults to IndexedDB caching with zero config. `RelayerNode` silently does nothing unless you pass `fheArtifactStorage`. The `FheArtifactCache.md` docs say the default is `new IndexedDBStorage(...)` without clarifying this is RelayerWeb-only. RelayerNode users will re-download multi-MB artifacts on every startup with no indication they're missing out.

Fix: clarify the RelayerWeb vs RelayerNode default difference in the docs page, and add a Node.js tab to the usage examples showing explicit opt-in with a storage backend.

**2. Docs: mention artifact caching in the getting-started guide**

Even though it's on by default for RelayerWeb, artifact caching deserves a mention in the primary getting-started example. Users should know: (a) multi-MB FHE artifacts are being cached in IndexedDB, (b) how to configure TTL, (c) that Node.js requires explicit opt-in. This is a significant performance feature that's invisible otherwise.

**3. HEAD-only revalidation — no GET fallback**

`#checkArtifactFreshness` only issues HEAD requests. A 405 (Method Not Allowed) triggers fail-open with 5-min retry, meaning the cache would *never* successfully revalidate against a CDN that doesn't support HEAD. S3 supports HEAD so this is fine today, but a one-line retry with GET on 405 would make it robust:

```ts
let res = await globalThis.fetch(url, { method: "HEAD", headers });

// Fallback to GET if HEAD is not supported (e.g. some CDN configs)
if (res.status === 405) {
  res = await globalThis.fetch(url, { headers });
}
```

Up to you whether this is worth doing now or tracking as a follow-up.

docs/gitbook/src/reference/sdk/FheArtifactCache.md
```diff
@@ -0,0 +1,205 @@
...
+```ts
+cache.revalidateIfDue(): Promise<boolean>
+```
+
+Checks whether cached artifacts are still fresh by issuing HTTP conditional requests (`ETag` / `If-None-Match`, `Last-Modified` / `If-Modified-Since`) against the artifact CDN URLs discovered from the relayer manifest.
+
+- Returns `true` if the cache was invalidated (caller should re-fetch).
+- Returns `false` if artifacts are still fresh or revalidation is not yet due.
+
+Concurrent calls are coalesced. On transient failures (network errors, 5xx), the cache fails open and retries after 5 minutes.
```
ankurdotb, line 187: Separate comment provided, currently assumes that HEAD is always supported. It might not be. Even if the server supports it, a WAF in the way might not.

ankurdotb (changes requested):

docs/gitbook/src/reference/sdk/FheArtifactCache.md
ghermet: we can swap it for `relayer.testnet.zama.org` here as the proxy is not what we're showcasing anyway.

ankurdotb (dismissed):

ankurdotb (dismissed):

ankurdotb (approved):

PR #111, opened 2026-03-16
feat(examples): add example-hoodi — ERC-7984 demo on Hoodi testnet

zama-cremaud:
Demonstrates all four confidential token operations (shield, unshield, confidential transfer, balance decrypt) on the Hoodi testnet using RelayerCleartext and an ethers-based MetaMask integration.

Key implementation notes:
- Hybrid EIP-1193 provider routes eth_call to a direct JsonRpcProvider, bypassing MetaMask network routing issues for read-only calls
- Chain detection + chainChanged listener with wrong-network banner and Switch to Hoodi button to prevent cross-chain transaction submissions
- Two token pairs (USDT Mock / Test Token) selectable at runtime

# Summary

## Scope

- [ ] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] CI/workflows
- [ ] Docs


examples/example-hoodi/src/app/client-providers.tsx
```diff
@@ -0,0 +1,14 @@
+"use client";
+
+import dynamic from "next/dynamic";
+import type { ReactNode } from "react";
+
+// ssr: false — Providers accesses window.ethereum (MetaMask) which is unavailable server-side.
+// next/dynamic with ssr: false must be declared inside a Client Component.
+const Providers = dynamic(() => import("../providers").then((m) => ({ default: m.Providers })), {
+  ssr: false,
+});
```
ghermet: You don't need dynamic import if you wrap the signer creation in useMemo?

examples/example-hoodi/src/app/page.tsx
```diff
@@ -0,0 +1,333 @@
...
+import {
+  useConfidentialBalance,
+  useShield,
+  useConfidentialTransfer,
+  useUnshield,
+  useMetadata,
+} from "@zama-fhe/react-sdk";
+import type { Address } from "@zama-fhe/react-sdk";
+
+const ERC20_ABI = ["function balanceOf(address) view returns (uint256)"];
```
ghermet: I think this can be imported from the SDK

examples/example-hoodi/src/app/page.tsx
```diff
@@ -0,0 +1,333 @@
...
+    };
+    const handleChainChanged = (...args: unknown[]) => setChainId(args[0] as string);
+
+    window.ethereum.on("accountsChanged", handleAccountsChanged);
+    window.ethereum.on("chainChanged", handleChainChanged);
+    return () => {
+      window.ethereum?.removeListener("accountsChanged", handleAccountsChanged);
+      window.ethereum?.removeListener("chainChanged", handleChainChanged);
+    };
+  }, []);
```
ghermet: This handled by the sdk via its lifecycle methods

ghermet (changes requested):
# PR Review Summary

**Scope:** New standalone Next.js example app (`examples/example-hoodi/`) — 13 files, +2325 lines

Overall the architecture is solid — SSR-safe dynamic import, hybrid EIP-1193 provider, clean hook usage, and excellent partner-facing documentation (README, WALKTHROUGH). A few issues need attention before merge.

---

## Critical Issues (3)

### 1. `window.ethereum` undefined crashes entire app
**`providers.tsx:37`** — `createHybridEthereum(window.ethereum)` is called without a null check inside `useMemo`. If MetaMask is not installed, this throws `TypeError: Cannot read properties of undefined` during render, crashing the entire React tree. The `ssr: false` dynamic import only prevents server-side access — it doesn't guard against a missing extension on the client.

**Fix:** Guard `window.ethereum` before constructing signer/relayer. Render a "Please install MetaMask" message if absent.

### 2. `BigInt()` crashes on invalid input — evaluated on every render
**`page.tsx:144`** — `const parsedAmount = BigInt(amount || "0")` is executed during render. If the user types `1.5`, `-3`, or `abc` into the `<input type="number">`, `BigInt()` throws an uncaught `SyntaxError` that crashes the component tree.

**Fix:** Wrap in try/catch, or validate input. Add `min="0" step="1"` to the input element.

### 3. "Switch to Hoodi" button — fire-and-forget promise, no error handling
**`page.tsx:133-139`** — The wrong-network banner's switch button calls `wallet_switchEthereumChain` with no `.catch()` and no fallback to `wallet_addEthereumChain` (error code 4902). Unlike `connect()` which correctly handles this, this inline handler silently produces unhandled promise rejections.

**Fix:** Extract to an async handler with the same add-chain fallback logic used in `connect()`.

---

## Important Issues (5)

### 4. `connect()` catch swallows all errors silently
**`page.tsx:94-96`** — User rejection (code 4001), network failures, etc. are only `console.error`'d. User gets zero feedback.

### 5. `lastTxHash` shows stale success from previous operations
**`page.tsx:147-151`** — After a successful shield, `shield.isSuccess` stays `true` even if a subsequent transfer fails. The UI shows "Transaction confirmed!" alongside an error, which is confusing.

### 6. `package-lock.json` in a pnpm monorepo
The monorepo uses pnpm. Including an npm lockfile with pinned alpha versions (`1.1.0-alpha.3`) is confusing and will become stale quickly. Consider removing it or documenting it as intentional.

### 7. `ignoreBuildErrors: true` in next.config.ts
Suppresses all TypeScript errors at build time. For a partner-facing demo, this masks real type errors introduced during customization.

### 8. `JsonRpcProvider` instantiated inside `useQuery` callback
**`page.tsx:123-126`** — A new provider is created on every query execution. Should be hoisted to module scope.

---

## Suggestions

- **`page.tsx:51-56`** — Initial wallet detection (`eth_accounts`, `eth_chainId`) has no `.catch()`. Silent failure if MetaMask is locked.
- **`page.tsx:209-213`** — Only one mutation error displayed at a time due to `??` chain.
- **`client-providers.tsx:8-10`** — No `loading` fallback for `dynamic()` import.
- **WALKTHROUGH.md** — `hoodiCleartextConfig` excerpt omits fields (partners may think it's the full shape). "Single EIP-712 signature" claim for `useConfidentialBalances` is misleading in cleartext mode. Unshield "two MetaMask confirmations" may be inaccurate with `RelayerCleartext`.
- **`.env.example`** — Alchemy URL uses `hoodi-mainnet` but Hoodi is a testnet. Verify actual URL pattern.

---

## Strengths

- Hybrid EIP-1193 provider pattern is clever and well-documented (JSDoc explains the *why*)
- Clean component layering: `client-providers.tsx` → `providers.tsx` → `page.tsx`
- WALKTHROUGH.md is excellent partner-facing documentation (ASCII diagrams, troubleshooting table, step-by-step flow)
- README comparison table vs `react-ethers` example is immediately useful

ghermet:
## SDK Usage Review

The example correctly uses all SDK APIs (imports, hook signatures, provider setup). However, it **reimplements several things the SDK already provides**:

### 1. Account & chain lifecycle — SDK already handles this

`EthersSigner.subscribe()` provides `onAccountChange`, `onChainChange`, and `onDisconnect` callbacks via `eip1193Subscribe()`. `ZamaProvider` already wires this up automatically to invalidate queries on wallet events.

The example reimplements this manually with raw `window.ethereum.on("accountsChanged")` / `on("chainChanged")` + `useState` in `page.tsx` (lines 6–26). This is redundant — the SDK is already listening and reacting to these events internally.

### 2. ERC-20 balance — SDK exports `balanceOfContract()`

The SDK exports `balanceOfContract(tokenAddress, account)` from `@zama-fhe/react-sdk`. The existing test-components use it (see `token-table.tsx`).

The example instead creates a standalone ethers `JsonRpcProvider` + manual react-query query to call `balanceOf`. It could use `signer.readContract(balanceOfContract(token.erc20, address))` via `useZamaSDK()` — staying within the SDK abstraction and avoiding the extra provider.

### 3. Underlying ERC-20 address — SDK exports `underlyingContract()`

The example hardcodes ERC-20 addresses in the `TOKENS` map. The SDK exports `underlyingContract(wrapperAddress)` which resolves the underlying ERC-20 from the wrapper contract on-chain. The example could derive the ERC-20 address instead of hardcoding it.

### 4. Redundant `JsonRpcProvider` in `page.tsx`

The hybrid EIP-1193 provider in `providers.tsx` is a legitimate MetaMask workaround (the SDK doesn't handle `eth_call` routing). But `page.tsx` then creates a **second** standalone `new JsonRpcProvider(HOODI_RPC)` for the ERC-20 balance query — this would be unnecessary if using `signer.readContract(balanceOfContract(...))`.

### Summary

| What the example does manually | What the SDK already provides |
|---|---|
| `window.ethereum.on("accountsChanged/chainChanged")` + useState | `signer.subscribe({ onAccountChange, onChainChange })` — auto-wired by `ZamaProvider` |
| Raw ethers `Contract.balanceOf()` + custom react-query | `balanceOfContract()` + `signer.readContract()` via `useZamaSDK()` |
| Hardcoded ERC-20 addresses in TOKENS | `underlyingContract(wrapperAddress)` resolves on-chain |
| Standalone `new JsonRpcProvider()` for balance reads | Unnecessary if using `signer.readContract()` |

The chain detection banner / `wallet_switchEthereumChain` is legitimately app-level code — the SDK doesn't provide network switching helpers.

### Other notes

- `typescript: { ignoreBuildErrors: true }` in `next.config.ts` — not ideal for a reference example
- Uses `1.1.0-alpha.3` — should be updated before merging to main

ankurdotb (changes requested):
Notes from trying to run this:

1. If you copy the example .env file to `.env.local` and keep the Hoodi RPC URL bank, instead of taking default it assumes blank value and throws an exception when the program starts. We should fix this.
2. On loading the app, it tells me "Wrong network - Switch to Hoodi" and shows a button, but clicking the button does nothing.
3. We should allow people to enter human values. Decimals for a token are easy to check.
4. Our default/example apps should have better UI/UX since we want to recommend this as a demo developers can actually execute. Let's discuss on how we make this better. We could start with this example and port it to the other example apps, or start from top and trickle it down to this one. I don't particularly care which way round, but we should be thinking very similarly of portfolio/token balance display, shield/unshield, transfers.

examples/example-hoodi/src/app/page.tsx
```diff
@@ -0,0 +1,346 @@
...
+export default function Home() {
+  const [address, setAddress] = useState<string | null>(null);
+  const [chainId, setChainId] = useState<string | null>(null);
+  const [selectedToken, setSelectedToken] = useState<TokenKey>("usdt");
+  const [amount, setAmount] = useState("");
+  const [recipient, setRecipient] = useState("");
+  const [connectError, setConnectError] = useState<string | null>(null);
+
+  const token = TOKENS[selectedToken];
+  const isHoodi = chainId !== null && parseInt(chainId, 16) === HOODI_CHAIN_ID;
```
gasp: this works but it's a little convoluted:

let's say chain id is `0x99` or `153`
`parseInt('99', 16)` returns 153 because it's on base 16
and `0x99 === 153` returns true

on chain change, metamask returns `"0x88bb0"` so I would recommend:


```suggestion
  const isHoodi = chainId === HOODI_CHAIN_ID_HEX;
```

examples/example-hoodi/src/providers.tsx
```diff
@@ -0,0 +1,68 @@
+"use client";
+
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { ZamaProvider, indexedDBStorage } from "@zama-fhe/react-sdk";
+import { RelayerCleartext, hoodiCleartextConfig } from "@zama-fhe/sdk/cleartext";
+import { EthersSigner } from "@zama-fhe/sdk/ethers";
+import { JsonRpcProvider } from "ethers";
+import { useMemo, type ReactNode } from "react";
+
+const HOODI_RPC_URL = process.env.NEXT_PUBLIC_HOODI_RPC_URL ?? "https://rpc.hoodi.ethpandaops.io";
```
gasp:
```suggestion
const HOODI_RPC_URL = process.env.NEXT_PUBLIC_HOODI_RPC_URL || "https://rpc.hoodi.ethpandaops.io";
```

Next.js replaces NEXT_PUBLIC_* environment variables at build time. When a variable is not defined, it substitutes an empty string "" in the compiled bundle — not undefined.
So `process.env.NEXT_PUBLIC_HOODI_RPC_URL` ?? "https://..." evaluates to "", which then gets passed to viem's http("") transport, which treats it as "no URL provided" and throws UrlRequiredError.

gasp (commented):
some small comments added

gasp (dismissed):

examples/example-hoodi/src/app/client-providers.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import dynamic from "next/dynamic";
+import type { ReactNode } from "react";
+
+// Providers accesses window.ethereum which is unavailable server-side.
+// next/dynamic with ssr:false must be declared inside a "use client" component,
+// which is why this thin wrapper exists — layout.tsx is a Server Component.
+const Providers = dynamic(() => import("../providers").then((m) => ({ default: m.Providers })), {
+  ssr: false,
+});
```
ghermet, line 11: Is this needed if you wrap the signer in a useMemo?

ankurdotb (changes requested):
Sorry forgot to give the review here instead of Slack:

1. On initial connect, it asks for access to ALL possible accounts including Bitcoin, Solana, etc.  This is overkill. Can we scope this to just Ethereum Hoodi?
2. On first load with no USDT mock or cUSDT, it says “Decryption failed” which is a bit misleading.
3. Shielding fails, even after approving the spend cap limit? It might be because the tx was not confirmed yet on Hoodi but app tries to check immediately and throws a “Shield failed”. This happened in both USDT mock and Test token 1.
4. After I got the confirmation from metamask that the initial spending cap txn had succeeded, then retrying shielding worked.
5. After a transfer, tried shield and it failed (I think also because of pending txn)
6. Asks me to sign a decrypt txn after each action. Aren’t we asking for decryption rights and storing signatures to reuse?
7. Unshield worked fine
8. Wondering if we can incorporate such steps into tests? Of course you can’t do on Hoodi for real without constantly topping up funds, but atm I see no Playwright tests. Maybe if we had some, we could simulate things like Mint -> Try to shield -> do step 1 of shield -> Make a pause -> do step 2 etc. What do you think? (I also imagine they will be useful across the example apps?) Having tests will also aid any AI-assisted coding you’re doing.

See my account: https://hoodi.etherscan.io/address/0x248d5447785152527f7dc9489fbfdbff4607b01e

examples/example-hoodi/src/app/client-providers.tsx
zama-cremaud: Seems like useMemo still runs during SSR, and `ssr: false` prevents the component and its SDK imports from executing on the server at all

ankurdotb (changes requested):
1. On shield, the approval tries to get approval for the **exact** amount as spend cap. Meaning, EVERY sign needs a two-step process. This is sometimes what apps do, but a better answer in this case is to set the spent cap to their ERC-20 balance. This will only trigger spend cap approval when necessary, which is if the shield amount exceeds current spend cap. (Which also means you should check if the existing spend cap already lets you spend the amount asked to shield.)

ankurdotb:
### Code review

Found 1 issue:

1. **Stale `activeUnshieldToken` on mutation failure causes phantom IndexedDB records**

   `setActiveUnshieldToken(tokenAddress)` is called before `unshield.mutate()`, but only cleared when `ZamaSDKEvents.UnshieldPhase1Submitted` fires. If the mutation fails early (wallet rejection, gas estimation error), `_wrapperAddress` is never reset to `null`. A subsequent `UnshieldPhase1Submitted` event would call `savePendingUnshield` with the stale/wrong token address, writing a phantom pending-unshield record into IndexedDB. Fix: add `setActiveUnshieldToken(null)` in an `onError` callback.

   https://github.com/zama-ai/sdk/blob/a6d7694e2eb490587c090b174329fc83472cbff0/examples/example-hoodi/src/components/UnshieldCard.tsx#L48-L51

   https://github.com/zama-ai/sdk/blob/a6d7694e2eb490587c090b174329fc83472cbff0/examples/example-hoodi/src/providers.tsx#L249-L260

.github/workflows/example-hoodi-e2e.yml
```diff
@@ -0,0 +1,64 @@
+name: E2E — example-hoodi
+
+on:
+  pull_request:
+    paths:
+      - "examples/example-hoodi/**"
```
ankurdotb, line 7: If it remains in independent workflow, this should also have a branches filter for main/prerelease?

ankurdotb:
### Code review

Found 7 issues:

---

**1. Chain-switching errors fully swallowed -- no user feedback on failure**

`switchToHoodi` silently swallows all non-4902 errors (including internal wallet errors, rate limiting, -32002 "request already pending"). Then `handleSwitchToHoodi` wraps it in an empty `catch {}` that discards `wallet_addEthereumChain` failures too. The user clicks "Switch to Hoodi", something fails, and the button just stops spinning with no indication.

Fix: in `switchToHoodi`, re-throw anything that is not 4902 (unknown chain) or 4001 (user rejection). In `handleSwitchToHoodi`, replace empty catch with `console.error`. User rejection is fine to swallow -- unexpected errors should not be.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/src/app/page.tsx#L58-L83

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/src/app/page.tsx#L103-L116

---

**2. WALKTHROUGH claims `useUnshield` supports `optimistic: true` -- it does not**

The "Going further" section says "for hooks like `useUnshield` and `useConfidentialTransfer`, pass `optimistic: true`". Only `useShield` and `useConfidentialTransfer` support this option. A partner developer following this advice for `useUnshield` would get a TypeScript error or a silently ignored option.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/WALKTHROUGH.md#L484

---

**3. `Promise.all` for initial wallet detection has no `.catch()`**

The mount effect calls `Promise.all([eth_accounts, eth_chainId]).then(...)` with no catch handler. If the wallet is locked, the extension crashed, or RPC times out, this is an unhandled promise rejection. The UI stays on "Connect Wallet" with no indication the wallet was detected but failed.

Fix: add `.catch((err) => console.error("Failed to detect wallet state:", err))`.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/src/app/page.tsx#L126-L139

---

**4. `handleSwitchToHoodi` captured in stale closure in `useEffect([])`**

`handleSwitchToHoodi` is an inline async function (not wrapped in `useCallback`) but is called from a `useEffect` with an empty dependency array. It works today because it only reads stable setters and calls `getEthereumProvider()` fresh, but it violates React's rules of hooks and will become a real bug if the function body ever references non-stable values.

Fix: either wrap in `useCallback` or inline the switch logic directly in the effect.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/src/app/page.tsx#L103-L139

---

**5. Missing `"type": "module"` in package.json**

Per CONTRIBUTING.md: "ESM-only -- all packages use `"type": "module"`". The example app's package.json does not include this field.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/package.json

---

**6. Workflow missing `branches` filter and does not trigger on SDK changes**

The e2e workflow triggers on `pull_request` with only a `paths` filter (`examples/example-hoodi/**`). Two issues:
- No `branches: [main, prerelease]` -- runs on PRs to any branch, including feature-to-feature PRs.
- SDK package changes (e.g. breaking change in `@zama-fhe/react-sdk`) will not trigger this workflow since the path filter does not cover `packages/`.

Consider also merging this into the existing `playwright.yml` as a separate job to reduce workflow duplication.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/.github/workflows/example-hoodi-e2e.yml#L1-L7

---

**7. `eth_accounts` seed failure silently discarded in providers.tsx**

The error branch of the `.then(success, failure)` pattern sets `refSeededRef.current = true` but discards the error entirely -- not even a `console.error`. If the wallet extension errors for a reason other than "not connected", developers have no way to diagnose it.

Fix: add `console.error("[Providers] Failed to seed accounts:", err)` in the failure handler.

https://github.com/zama-ai/sdk/blob/ff01a63e5c969c017a68b51a4c130ce50126632e/examples/example-hoodi/src/providers.tsx#L204-L212

---

examples/example-hoodi/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "test:e2e": "playwright test",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@tanstack/react-query": "^5.90.0",
+    "@zama-fhe/react-sdk": "1.2.0-alpha.5",
+    "@zama-fhe/sdk": "1.2.0-alpha.5",
```
ankurdotb: Worth upgrading to latest alphas?

ankurdotb (changes requested):

zama-cremaud:
On the latest Claude code review :

```
 ┌─────┬────────────────────────────────────────────┬──────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │  #  │                   Issue                    │     Decision     │                                                                 Reason                                                                 │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 1   │ Chain-switching errors silently swallowed  │ ✅ Fixed         │ Added console.error — consistent with how connect() already handles errors                                                             │
  │     │ in handleSwitchToHoodi                     │                  │                                                                                                                                        │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 2   │ WALKTHROUGH claims useUnshield supports    │ ✅ Fixed         │ Real doc bug — removed useUnshield from the mention                                                                                    │
  │     │ optimistic: true                           │                  │                                                                                                                                        │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 3   │ Promise.all wallet detection has no        │ ✅ Fixed         │ Added .catch((err) => console.error(...))                                                                                              │
  │     │ .catch()                                   │                  │                                                                                                                                        │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 4   │ handleSwitchToHoodi captured in stale      │ ⏭️  Skipped       │ The function only references stable values (state setters, module-level functions) — no real bug today or tomorrow unless the function │
  │     │ closure in useEffect([])                   │                  │  body changes. Adding useCallback would increase complexity with no benefit in a quickstart context                                    │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 5   │ Missing "type": "module" in package.json   │ ⏭️  Skipped       │ The CONTRIBUTING.md rule targets SDK packages under packages/ — not example apps. Adding "type": "module" to a Next.js app             │
  │     │                                            │                  │ package.json can break the build; Next.js handles ESM internally                                                                       │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 6   │ Workflow missing branches filter           │ ⏭️  Skipped       │ branches: [main, prerelease] was added in a previous commit. The SDK path filter point is by design — the example consumes published   │
  │     │                                            │ (already done)   │ npm packages, not local SDK sources                                                                                                    │
  ├─────┼────────────────────────────────────────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 7   │ eth_accounts seed failure silently         │ ✅ Fixed         │ Added console.error("[Providers] Failed to seed accounts:", err)                                                                       │
  │     │ discarded in providers.tsx                 │                  │                                                                                                                                        │
  └─────┴────────────────────────────────────────────┴──────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

ankurdotb (approved):

PR #112, opened 2026-03-16
chore: re-enable lints in config

enitrat:
## Summary
- add a shared `tsconfig.base.json` baseline plus split repo configs for package sources, Vitest coverage, app fixtures, e2e, and examples instead of relying on one root catch-all config
- tighten Oxlint around TS, imports, promises, and runtime safety with `denyWarnings`, explicit React/Next/Vitest plugins, and narrower overrides for tests, workers, and legacy hot spots
- update package, app, and example `tsconfig.json` files to extend the shared config and add scoped `typecheck` scripts where we actually want typed gates
- fix existing TS and Oxlint violations across the SDK, React SDK, test components, examples, and fixtures so the stricter default gates are green
- add `oxlint-tsgolint` so the repo is ready for a later type-aware Oxlint rollout without blocking the default fast path today

## Validation
- `pnpm lint`
- `pnpm run typecheck`
- `pnpm run typecheck:apps`

## Scope Notes
- `pnpm run typecheck:tests`, `pnpm run typecheck:apps:fixtures`, `pnpm run typecheck:e2e`, and `pnpm run typecheck:examples` stay outside the default gate for now because they still depend on heavier test globals, generated artifacts, or example-local installs


ghermet (dismissed):

enitrat:
@ghermet i'll let it draft until we merge the big PRs you have to avoid you some conflicts

ghermet (approved):

PR #116, opened 2026-03-17
feat(sdk,react-sdk): add WrappersRegistry class and on-chain registry hooks

ghermet:
## Summary

- Add `WrappersRegistry` class to query native↔confidential token pairs from the on-chain `ConfidentialTokenWrappersRegistry` contract
- Automatic address resolution per chain via `DefaultWrappersRegistryAddresses`
- `ZamaSDK.createWrappersRegistry()` factory method
- Contract builders for both ethers and viem with 7 read helpers
- Query option factories for TanStack Query integration (all 7 registry operations)
- React hooks: `useTokenPairsRegistry`, `useTokenPairsLength`, `useTokenPairsSlice`, `useTokenPair`, `useConfidentialTokenAddress`, `useTokenAddress`, `useIsConfidentialTokenValid`, `useWrappersRegistryAddress`
- 43 unit tests across contracts, class, and query layers
- GitBook documentation (WrappersRegistry reference, ZamaSDK, contract-builders, network-presets updates)
- PR review fixes: credential error handling, cache write best-effort, pLimit validation

## Test plan

- [ ] Unit tests pass (`pnpm test`)
- [ ] 43 new WrappersRegistry tests (contracts, class, query layers)
- [ ] Size-limit check passes
- [ ] API reports updated


zama-cremaud:
## Issues

### [MEDIUM] Breaking rename in `useWrapperDiscovery` — not in CHANGELOG

**Files:** `packages/sdk/src/query/wrapper-discovery.ts`, `packages/react-sdk/src/token/use-wrapper-discovery.ts`

The `coordinatorAddress` parameter has been renamed to `erc20Address`. This is a **public API breaking change** for any consumer who calls `useWrapperDiscovery({ tokenAddress, coordinatorAddress })` or `wrapperDiscoveryQueryOptions({ tokenAddress, coordinatorAddress })`.

TypeScript will catch this at compile time, but the migration path is unclear without a CHANGELOG entry. There is currently no CHANGELOG entry for this branch.

**Action required:** Add a CHANGELOG entry under `⚠ BREAKING CHANGES` for the `coordinatorAddress` → `erc20Address` rename in both `wrapperDiscoveryQueryOptions` and `useWrapperDiscovery`.

---

### [MEDIUM] `listPairsQueryOptions.queryFn` creates a new `WrappersRegistry` instance per execution, bypassing instance-level cache

**File:** `packages/sdk/src/query/wrappers-registry.ts`

The `queryFn` inside `listPairsQueryOptions` (and the other query factories that accept `signer`) constructs a new `WrappersRegistry` instance on each invocation:

```ts
queryFn: async () => {
  const registry = new WrappersRegistry({
    signer,
    wrappersRegistryAddresses: ...,
    registryTTL: config.registryTTL,
  });
  return registry.listPairs({ page, pageSize, metadata });
},
```

`WrappersRegistry` maintains an internal `Map`-based TTL cache on `this.#cache`. A freshly constructed instance starts with an empty cache every time. As a result, the TTL caching that is the core value proposition of `WrappersRegistry` is completely inert when accessed through these query factories — each `queryFn` execution results in a fresh contract call regardless of `registryTTL`.

The TanStack Query layer provides `staleTime`/`gcTime` caching for the serialized result, but that caching is at the query level (per React tree, keyed by page/pageSize). The class-level cache is intended to serve cross-query, cross-component lookups (e.g., `getConfidentialToken` and `listPairs` sharing a warmed cache). Neither benefit is realized.

**Suggested approach:** Pass the `ZamaSDK.registry` singleton (or a shared `WrappersRegistry` instance managed by the SDK) into the query factories rather than constructing one inside `queryFn`. Alternatively, move registry instantiation to a stable reference outside the `queryFn` closure.

---

### [MEDIUM] `useWrapperDiscovery` post-rename API is misleading

**File:** `packages/react-sdk/src/token/use-wrapper-discovery.ts`

After the rename, the hook accepts two address parameters:

```ts
useWrapperDiscovery({ tokenAddress, erc20Address })
```

The JSDoc describes `tokenAddress` as "Address of the confidential token" and `erc20Address` as "Address of the ERC-20 token". Both are required. However:

- `erc20Address` is passed to `WrappersRegistry.getConfidentialToken(erc20Address)` — it drives the actual lookup.
- `tokenAddress` is used **only as a query key discriminator** and plays no role in the registry call.

A consumer reading the JSDoc will reasonably expect that `tokenAddress` (the "confidential token address") is what the hook resolves, and `erc20Address` is the input. In fact, the direction is the reverse: `erc20Address` is the input, and the hook's `data` field returns the confidential token address. `tokenAddress` in the key just allows the query to be invalidated per-token — it is not semantically meaningful to the consumer.

**Recommendation:** Either rename `tokenAddress` to something that signals its role (e.g., `cacheScope` or document it explicitly as "used only for query key scoping"), or drop it entirely if `erc20Address` is already a sufficient discriminator for the query key.

---

### [LOW] No `staleTime` on wrappersRegistry queries — data is stale immediately

**File:** `packages/sdk/src/query/wrappers-registry.ts`

None of the query option factories set a `staleTime`. TanStack Query defaults to `staleTime: 0`, meaning every query result is considered stale on the next render and will trigger a background refetch when the component re-mounts or the window re-focuses.

The `WrappersRegistry` class accepts a `registryTTL` in seconds (defaulting to 86400 — one day). This value is used for the in-memory class cache. It would be natural to propagate the same value as `staleTime` on the query options, avoiding unnecessary background refetches for data that is already valid according to the registry's own TTL contract.

---

### [LOW] `ReadonlyToken.discoverWrapper` creates a fresh `WrappersRegistry` with no TTL, not sharing cache with `sdk.registry`

**File:** `packages/sdk/src/token/readonly-token.ts`

```ts
async discoverWrapper(erc20Address: Address): Promise<Address> {
  const registry = new WrappersRegistry({ signer: this.signer });
  return registry.getConfidentialToken(erc20Address);
}
```

This creates a throwaway `WrappersRegistry` instance with default TTL (86400s), separate from `sdk.registry`. Each call to `discoverWrapper` starts with an empty cache. If `discoverWrapper` is called multiple times — or if the same address was previously resolved through `sdk.registry` — the cached result is not reused.

If the `ReadonlyToken` is created from a `ZamaSDK` instance, passing `sdk.registry` (or the SDK's `registryTTL`) through would eliminate redundant on-chain calls.

---

### [LOW] No tests for React hooks in `packages/react-sdk/src/wrappers-registry/`

**Directory:** `packages/react-sdk/src/wrappers-registry/`

The nine new React hooks (`useListPairs`, `useTokenPairsRegistry`, `useTokenAddress`, `useConfidentialTokenAddress`, `useIsConfidentialTokenValid`, `useTokenPair`, `useTokenPairsLength`, `useTokenPairsSlice`, `useWrappersRegistryAddress`) have no dedicated tests. The query option factories they wrap are tested in `packages/sdk/src/query/__tests__/wrappers-registry.test.ts`, but the React layer (provider integration, address resolution via `useWrappersRegistryAddress`, `enabled` gating when address is absent) is untested.

The `useWrappersRegistryAddress` stale-time of `30_000 ms` and the comment "chain switches may take up to 30s to reflect" in particular warrant a test.

---

## What Is Good

**`WrappersRegistry` class design is clean.** The class correctly separates concerns: read-only contract calls, an internal TTL cache keyed by method + normalized args, and a `refresh()` method to clear the cache on demand. The `#ttlMs` is derived once from config (`(registryTTL ?? 86400) * 1000`) rather than on each cache read. The `staleAt` approach (recording expiry time at write) is straightforward and correct.

**`DefaultWrappersRegistryAddresses` auto-populates from `DefaultConfigs`.** Consumers of `relayer-utils.ts` get registry addresses for supported chains out of the box via `ExtendedFhevmInstanceConfig`. No manual per-chain configuration is needed for standard deployments.

**`ZamaSDK.registry` is a lazy singleton.** The registry is only instantiated on first access and reuses the SDK's `signer` and `registryTTL`. This is the right pattern for a resource-intensive object.

**`listPairs` pagination is explicit and safe.** The `page` parameter is 1-indexed with a clear error thrown for `page < 1`. The `total` field in `PaginatedResult<T>` is correctly derived via `Number(bigint)` from the contract return. The `hasMore` flag is cleanly computed from `fromIndex + items.length < Number(total)`.

**Unit test coverage is comprehensive.** `packages/sdk/src/token/__tests__/wrappers-registry.test.ts` covers address resolution, TTL expiry, `refresh()`, cache hit/miss, `listPairs` multi-page, and `ZamaSDK.registry` singleton behavior. `packages/sdk/src/query/__tests__/wrappers-registry.test.ts` validates all query factories.

**`wrapSigningError` improvement is a good fix.** Including the original error message (`${context}: ${originalMessage}`) in `SigningRejectedError` and `SigningFailedError` makes debugging wallet failures significantly easier — the original message was silently dropped before this change.

**`CredentialsPersistFailed` is the right observability pattern.** Elevating credential persist failures from a `console.warn` to a structured `CredentialsPersistFailedEvent` (with `cause`) allows host applications to surface or log these failures through their own error-reporting pipelines rather than relying on console output.

**`pLimit` guard against `maxConcurrency <= 0` prevents a confusing infinite hang.** Without it, passing `0` would silently queue all tasks indefinitely. The explicit `throw` is the correct behavior.

---

## Summary

Two items to address before merge:

1. **Add a CHANGELOG entry** for `⚠ BREAKING CHANGES`: `coordinatorAddress` → `erc20Address` rename in `wrapperDiscoveryQueryOptions` and `useWrapperDiscovery`.
2. **Fix `listPairsQueryOptions.queryFn`** (and other query factories): pass a shared `WrappersRegistry` instance rather than constructing one inside the function, so that the class-level TTL cache is actually effective.

Two design points worth resolving:

3. **Clarify `useWrapperDiscovery` parameter semantics** — either rename `tokenAddress` to reflect its role as a cache-scoping key, or drop it if `erc20Address` already uniquely identifies the query.
4. **Propagate `registryTTL` as `staleTime`** on query option factories, so TanStack Query and the registry class operate under the same freshness contract.

---

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased] — feat/wrappers-registry
+
+### ⚠ BREAKING CHANGES
+
+- **wrappers-registry:** `listPairsQueryOptions` first argument changed from `signer: GenericSigner` to `registry: WrappersRegistry`. Callers should pass `sdk.registry` (the lazy singleton on `ZamaSDK`) instead of `sdk.signer`. The `registryTTL` field has been removed from `ListPairsQueryConfig` — the registry's own TTL is used automatically.
+
+- **wrappers-registry:** `useListPairs` no longer accepts a `registryTTL` option. The TTL is derived from the `ZamaSDK` configuration.
+
```
ghermet: Is this necessary?

ghermet:
@ankurdotb thoughts on the `wrappersRegistryAddress` parameter name for the object containing containing registry addresses keyed by chain?

CHANGELOG.md
zama-cremaud: Good point, it is deleted now

ankurdotb:
> @ankurdotb thoughts on the `wrappersRegistryAddress` parameter name for the object containing containing registry addresses keyed by chain?

I would call it `registryAddress` as suggested in Linear, rather than `wrapperRegistryAddress`. In the future, the Registry is very likely to expand to covering native confidential tokens as well as wrapped tokens. However, a `WrapperRegistry` is still a valid idea as a class. But I'm not too fussed, since other classes of tokens in such a registry don't exist yet.

zama-cremaud:
> > @ankurdotb thoughts on the `wrappersRegistryAddress` parameter name for the object containing containing registry addresses keyed by chain?
> 
> I would call it `registryAddress` as suggested in Linear, rather than `wrapperRegistryAddress`. In the future, the Registry is very likely to expand to covering native confidential tokens as well as wrapped tokens. However, a `WrapperRegistry` is still a valid idea as a class. But I'm not too fussed, since other classes of tokens in such a registry don't exist yet.

Renaming done

packages/sdk/src/zama-sdk.ts
```diff
@@ -159,6 +168,27 @@ export class ZamaSDK {
...
+   */
+  get registry(): WrappersRegistry {
+    if (!this.#registryInstance) {
+      this.#registryInstance = new WrappersRegistry({
+        signer: this.signer,
+        registryTTL: this.#registryTTL,
+      });
+    }
+    return this.#registryInstance;
+  }
```
ghermet: Feels like this should be defined as class attribute in the constructor. #registryTTL won't be needed as an attribute
```ts
constructor({ signer, registryTTL}) {
this.#wrappersRegistry = new WrappersRegistry({
        signer,
        registryTTL,
      });
}

Comment (zama-cremaud) [reply packages/sdk/src/zama-sdk.ts]: Yes, good point, updated now!

File: packages/sdk/etc/sdk-node.api.md [hunk 3]
```diff
@@ -590,6 +592,7 @@ export const SepoliaConfig: {
     readonly inputVerifierContractAddress: "0xBBC1fFCdc7C316aAAd72E807D9b0272BE8F84DA0";
     readonly verifyingContractAddressDecryption: "0x5D8BD78e2ea6bbE41f26dFe9fdaEAa349e077478";
     readonly verifyingContractAddressInputVerification: "0x483b9dE06E4E4C7D35CCf5837A1668487406D955";
+    readonly registryAddress: "0xDEbdfa2568dEd84044e9807500ee523acDe9308b";
```
ankurdotb: This is the old Zaiffer registry, not the correct registry listed in Linear from the [Protocol Apps docs](https://github.com/zama-ai/protocol-apps/blob/main/docs/addresses/testnet/sepolia.md).

docs/gitbook/src/reference/sdk/WrappersRegistry.md
ankurdotb, line 1: Where are any of the high level functions as described in Linear?

ankurdotb:
### Code review (updated at df5c85db — fifth pass)

**Previously fixed (13/13):** wrong Sepolia address, toIndex clamping, stale coordinatorAddress in gitbook docs, missing Hoodi default, lowercase/duplicated addresses, query key missing registry address, `createWrappersRegistry` not forwarding `registryTTL`, `discoverWrapper` throwaway registry, stale "lazily instantiated" JSDoc, `ZamaProvider` missing registry props, `Object.assign` mutation, `enabled` missing `registryAddress` check, stale `discoverWrapper` in ReadonlyToken.md, stale `coordinatorAddress` in react-sdk README, misleading `tokenAddress` JSDoc. All resolved.

Found 3 remaining issues:

1. **`useRef` for `registryAddresses` swallows legitimate updates.** The `useRef` stabilization prevents spurious re-creation from inline literals, but it also makes real `registryAddresses` changes invisible after mount. `ZamaSDK` only reads this value at construction time, and the `useMemo` no longer depends on the prop, so if a parent swaps to a different registry mapping the provider silently keeps the stale addresses. This breaks dynamic local/custom-chain setups. Consider a shallow-compare memo instead so real changes trigger re-creation but identical literals don't.

https://github.com/zama-ai/sdk/blob/df5c85db13a7b40fb4c1fb676a12b1fa979615ce/packages/react-sdk/src/provider.tsx#L89-L117

2. **`packages/sdk/src/contracts/deployment-coordinator.ts` JSDoc still uses `coordinatorAddress`.** The parameter name was renamed elsewhere in this PR but the JSDoc examples on lines 10 and 29 still reference the old name.

https://github.com/zama-ai/sdk/blob/df5c85db13a7b40fb4c1fb676a12b1fa979615ce/packages/sdk/src/contracts/deployment-coordinator.ts#L9-L30

3. **`packages/sdk/README.md` still shows `discoverWrapper(coordinatorAddress)`.** The method was removed and the parameter renamed, but the SDK README table (line 234) still references both.

https://github.com/zama-ai/sdk/blob/df5c85db13a7b40fb4c1fb676a12b1fa979615ce/packages/sdk/README.md#L233-L235

ankurdotb (changes requested):
In addition to any automated reviews:

- There are still leftovers for `COORDINATOR_ADDRESS`, in the example apps. Worth fixing in this, or create a sub-task/follow-up once this one is released (if a released version is necessary to fix those).
- Sepolia Registry is incorrect (it's the old Zaiffer one)

packages/sdk/etc/sdk-node.api.md
ghermet: This is stale documentation, it's the correct one in the code

docs/gitbook/src/reference/sdk/WrappersRegistry.md
ghermet: The high-level methods are now documented: `listPairs` (with pagination + metadata enrichment), `getConfidentialToken`, `getUnderlyingToken`, and `refresh` — see lines 75–135 of `WrappersRegistry.md`. This was added in 7d3b8c1f.

zama-cremaud:
### [LOW] `sdk.registry` JSDoc says "Lazily instantiated" — it is not

**File:** `packages/sdk/src/zama-sdk.ts`

The field declaration reads:

```ts
/**
 * A {@link WrappersRegistry} instance auto-configured for the current chain.
 * Lazily instantiated on first access. Uses default registry addresses …
 */
readonly registry: WrappersRegistry;
```

The constructor body contains `this.registry = new WrappersRegistry(…)` — an eager instantiation, not a lazy one (commit `40e2873`). The JSDoc is stale from a prior design iteration. Any consumer reading it may unnecessarily guard for `undefined` or rely on first-access semantics that no longer apply.

**Fix:** Update to "Eagerly instantiated in the constructor" or simply remove the instantiation-timing claim.

packages/sdk/src/relayer/cleartext/presets.ts
```diff
@@ -23,4 +23,5 @@ export const hoodiCleartextConfig = {
   verifyingContractAddressDecryption: HardhatConfig.verifyingContractAddressDecryption,
   verifyingContractAddressInputVerification:
     HardhatConfig.verifyingContractAddressInputVerification,
+  registryAddress: "0x1807aE2f693F8530DFB126D0eF98F2F2518F292f",
```
ankurdotb, line 26: I think this should remain as Hardhat, as that's one of the networks that developers will likely want to use.

packages/sdk/src/wrappers-registry.ts
```diff
@@ -0,0 +1,500 @@
...
+
+/**
+ * Default wrappers registry addresses derived from {@link DefaultConfigs}.
+ * Only includes chains where a registry is deployed (excludes Hardhat).
+ */
+export const DefaultRegistryAddresses: Record<number, Address> = {
+  [MainnetConfig.chainId]: MainnetConfig.registryAddress,
+  [SepoliaConfig.chainId]: SepoliaConfig.registryAddress,
+  [hoodiCleartextConfig.chainId]: hoodiCleartextConfig.registryAddress,
+};
```
ankurdotb, line 29: And here, I would treat Hoodi as a similar level like `MainnetConfig` and `SepoliaConfig` as `HoodiConfig`.

Hoodi is currently supported via cleartext, but will some day (in Q2 potentially?) be supported as an actual implementation rather than via cleartext.

ankurdotb (changes requested):
Also see the additional review from Claude Code (original post from it was updated)

packages/sdk/src/relayer/cleartext/presets.ts
ghermet: okay sure

ghermet: registryAddress because `undefined`

packages/sdk/src/wrappers-registry.ts
ghermet: We should only create `HoodiConfig` once it's supported then. I'll replace with:
```ts
export const DefaultRegistryAddresses: Record<number, Address> = {
  [mainnet.id]: MainnetConfig.registryAddress,
  [sepolia.id]: SepoliaConfig.registryAddress,
  [hoodi.id]: "0x1807aE2f693F8530DFB126D0eF98F2F2518F292f",
};
```

ankurdotb (approved):

PR #117, opened 2026-03-17
build(sdk): add dual ESM/CJS build for CommonJS compatibility

ghermet:
## Summary

- Add CJS build output (`dist/cjs/`) alongside ESM (`dist/esm/`) so projects using `"module": "CommonJS"` / `"moduleResolution": "node"` can resolve subpath exports like `@zama-fhe/sdk/node`
- Add `require` conditions to all subpath exports in `package.json`
- Add `typesVersions` field for TypeScript legacy module resolution
- Track ESM and CJS bundle sizes separately in size-limit

## Context

Projects using `"moduleResolution": "node"` (the default for CommonJS) cannot resolve the `exports` field in `package.json`. The old `@zama-fhe/relayer-sdk` had `require` conditions that made this work. The new SDK was ESM-only, breaking these consumers.

## Test plan

- [x] `pnpm build` produces both `dist/esm/` and `dist/cjs/` with correct file extensions
- [x] CJS `require()` works for all entry points (root, node, ethers, viem, cleartext, query)
- [x] ESM `import()` works for all entry points
- [x] TypeScript resolves `@zama-fhe/sdk/node` with `moduleResolution: "node"` + `module: "CommonJS"`
- [x] All 873 tests pass
- [x] Size-limit check passes (ESM: 36.83 KB, CJS: 31.16 KB, both under 40 KB)


ankurdotb:
# PR Review Summary — #117: Dual ESM/CJS Build

Reviewed by 3 specialized agents: **Code Reviewer**, **Silent Failure Hunter**, **Test Analyzer**.

---

## Critical Issues (1 found)

### 1. `/node` subpath CJS runtime failure
**Files:** `rolldown.config.ts`, `package.json`

The `./node` export has a `require` condition pointing to `dist/cjs/node/index.cjs`, but:
- The node worker file (`relayer-sdk.node-worker.js`) only exists in `dist/esm/`, not `dist/cjs/`
- CJS `node/index.cjs` resolves the worker path relative to its own location → file not found → runtime crash

**Recommendation:** Remove `require` from the `"./node"` export and remove `"node/index"` from the CJS `entryPoints` in rolldown config. The node entry depends on `node:worker_threads` which is inherently ESM-oriented.

---

## Important Issues (3 found)

### 2. Missing `"default"` condition in exports map
**File:** `package.json` (lines 15-45)

No `"default"` fallback in any export block. Node.js docs recommend it as a catch-all for environments that don't match `import`/`require` (Webpack 4, Jest, edge runtimes).

**Fix:** Add `"default"` pointing to ESM entry as the last condition in each export block.

### 3. No CJS build smoke tests
The core feature of this PR (CJS distributable) has zero test coverage. The existing `worker.bundler.test.ts` only validates ESM output.

**Recommendation:** Add CJS smoke tests verifying:
- Each CJS entry point `.cjs` file exists at declared paths
- Files are valid CJS format
- `package.json` exports map resolves to existing files on disk
- Node worker is NOT in CJS output

### 4. Sourcemap files not in `files` array
**File:** `package.json` (lines 66-72)

Both builds have `sourcemap: true` but `dist/**/*.map` is not in the `files` array. Published `.cjs` files will reference `.cjs.map` that don't exist in the npm package.

**Fix:** Add `"dist/**/*.map"` to `files`.

---

## Suggestions (2 found)

### 5. CJS `require` conditions lack nested `types` for `node16`/`nodenext` resolution
For maximum TypeScript compatibility, consider nested conditions:
```json
"import": { "types": "...", "default": "..." },
"require": { "types": "...", "default": "..." }
```

### 6. `delegatedCredentials` bug fix lacks test coverage
The fix is correct — `createToken` was missing `delegatedCredentials` propagation. But no test validates the sharing, so a regression could re-introduce the bug.

---

## Strengths

- Build config refactor is well-structured — shared `entryPoints` avoids duplication
- CJS correctly excludes `dts` plugin and node worker entry
- `delegatedCredentials` fix ensures tokens share the SDK-level credentials manager
- API extractor paths consistently updated — no stale `dist/` references
- Size limit tracking for both ESM and CJS bundles is a good CI guardrail
- `typesVersions` correctly handles legacy `moduleResolution: "node"` consumers

---

ankurdotb (changes requested):
Major gap I see is updated to documentation and making sure that it's consistent with the documentation style + Gitbook formatting notes on how to import into ESM projects vs CJS. I don't think this is automatic or covered under existing docs, right?

Have also asked for a fix + tests to review comments, but any independent fixes to that are also fine as long as they resolve those issues.

ankurdotb (approved):
Really good update on docs, very useful!

PR #125, opened 2026-03-19
feat(react-sdk): unify useUserDecrypt with reactive cache and auto-decrypt

ghermet:
## Summary

Introduces a persistent, requester-scoped `DecryptCache`, lifts decrypt orchestration into `sdk.userDecrypt()`, and converts the React `useUserDecrypt` hook from a mutation to an auto-firing query gated on `isAllowed`.

### What changed

**Core SDK**
- **`DecryptCache`** — new storage-backed cache keyed by `(requester, contractAddress, handle)`. Survives page reloads; different signers cannot read each other's cached decryptions (mirrors on-chain ACL). Index writes are serialised to prevent concurrent `set` calls from losing entries.
- **`sdk.userDecrypt()`** — new top-level method on `ZamaSDK`. Orchestrates: cache check → `credentials.allow()` → contract-grouped relayer calls → cache write. Accepts `DecryptOptions` with `onCredentialsReady` / `onDecrypted` callbacks.
- **`ReadonlyToken`** now delegates to the shared `DecryptCache` instead of the old per-token `balance-cache` module (deleted).
- **`isAllowed` contract coverage** — `checkAllowed()` now verifies that cached credentials cover the requested contract addresses, preventing the decrypt query from firing before credentials are ready.
- **Revoke clears cache** — `revoke()` / `revokeSession()` call `cache.clearForRequester()`. Signer lifecycle events (disconnect, account/chain change) call `cache.clearAll()`.

**React SDK**
- **`useUserDecrypt` converted from mutation to query** — accepts `handles` in config, auto-fires when `isAllowed` is true. No more `mutate()` call needed; decrypted values appear reactively once `useAllow` has been called. Chains `signerAddressQuery` → `isAllowedQuery` → `userDecryptQuery`.
- **Removed `useUserDecryptedValue` / `useUserDecryptedValues`** — deleted entirely; the query-based `useUserDecrypt` now serves cached reads directly.
- **Revoke invalidates decryption queries** — `useRevoke` and `useRevokeSession` now invalidate `zamaQueryKeys.decryption.all`.

**Query layer**
- **`userDecryptQueryOptions`** replaces `userDecryptMutationOptions` — thin query factory calling `sdk.userDecrypt()` with `staleTime: Infinity`.
- **`isAllowedQueryOptions`** now accepts optional `contractAddresses` and delegates to `sdk.credentials.isAllowed()`.

### Design decisions

**Why `(requester, contract, handle)` as cache key?**
The plaintext for a handle is always the same, but the on-chain ACL controls who can decrypt. The `requester` dimension ensures revoking user A's session only wipes user A's entries, and user B can't get a cache hit on handles they aren't ACL-allowed to decrypt.

**Why query gated on `isAllowed`, not mutation?**
Once credentials exist, decryption is effectively a pure read (cache hit or relayer fetch). Converting to a query means values appear automatically when the prerequisite `useAllow` completes — no imperative `mutate()` call needed. The `isAllowed` gate with contract coverage check prevents the query from firing before credentials are ready.

**Why `sdk.userDecrypt()` on ZamaSDK?**
Lifts orchestration out of React so Node SDK, vanilla JS, and tests use the same cache-enabled flow. The React hook is a thin wrapper around `userDecryptQueryOptions`.

## Linear

- Closes [SDK-44](https://linear.app/zama/issue/SDK-44/remove-redundant-useuserdecryptedvalue-useuserdecryptedvalues-hooks) — remove redundant `useUserDecryptedValue` / `useUserDecryptedValues` hooks
- Closes [SDK-77](https://linear.app/zama/issue/SDK-77/critical-eip-712-signature-contract-scope-not-validated-in-credential) — CRITICAL: EIP-712 signature contract scope not validated in credential checks (parent)
- Closes [SDK-78](https://linear.app/zama/issue/SDK-78/enforce-eip-712-contract-scope-in-session-credential-checks) — enforce EIP-712 contract scope in session credential checks
- Closes [SDK-80](https://linear.app/zama/issue/SDK-80/redgreen-test-suite-eip-712-contract-scope-enforcement-in-credential) — red/green test suite for EIP-712 contract scope enforcement

## Test plan

- `decrypt-cache.test.ts` — unit tests for isolation, persistence, clearing, error resilience, concurrency
- `zama-sdk.test.ts` — `sdk.userDecrypt()` cache hit/miss/mixed, grouping by contract, revoke clears cache
- `user-decrypt.test.ts` — query factory tests for `userDecryptQueryOptions`
- `use-user-decrypt.test.tsx` — React hook integration (auto-fire, isAllowed gate, no signature prompt when gated false)
- `readonly-token.test.ts` — token balance decryption uses DecryptCache
- `credentials-manager.test.ts` — `isAllowed` contract coverage check, empty-array runtime guard, keypair-vs-session expiry, valid/invalid invariants
- `delegated-credentials-manager.test.ts` — delegation scope mismatch, wrong delegator, empty-array runtime guard
- `allow-revoke.test.ts` — revoke invalidates decryption queries


zama-cremaud:
## Issues

### [HIGH] Breaking changes not documented in CHANGELOG

The following public API exports are **removed** by this branch:

| Removed export | Replacement |
|---|---|
| `useUserDecryptedValue(handle)` | `useUserDecrypt({ handles: [...] }).values[handle]` |
| `useUserDecryptedValues(handles)` | `useUserDecrypt({ handles: [...] }).values` |
| `type DecryptCallbacks` | `type UserDecryptCallbacks` (re-exported unchanged) |
| `type DecryptParams` | `type UserDecryptMutationParams` (re-exported unchanged) |

The `useUserDecrypt` return type also changed: it was `UseMutationResult<Record<Handle, ClearValueType>, Error, UserDecryptMutationParams>` and is now a custom object with overridden `mutate`/`mutateAsync` signatures plus the new `values` field. Any consumer relying on the old `mutate(params: UserDecryptMutationParams)` call signature (required params, no default) will get a TypeScript type mismatch.

None of this is in the CHANGELOG. Given this is `prerelease`, breaking changes are expected — but they need to be documented so that downstream maintainers and partners can adapt.

---

### [MEDIUM] Stale `DecryptParams` reference in README

**File:** `packages/react-sdk/README.md`, line 842

The "All Encryption & Decryption Hooks" table still references `DecryptParams` as the input type for `useUserDecrypt()`:

```
| `useUserDecrypt()` | `DecryptParams` | `Record<string, bigint>` | ...
```

`DecryptParams` is the type alias that was removed in this branch. The correct type is `UserDecryptMutationParams`. This will cause confusion for any developer who looks at the README and tries to import `DecryptParams`.

---

### [MEDIUM] `mutateAsync()` returns `{}` when all handles are cached — not the actual cached values

**File:** `packages/react-sdk/src/relayer/use-user-decrypt.ts`

When `mutateAsync()` is called with all handles already in the cache, it resolves to an empty object `{}` rather than the actual cached values:

```ts
if (params.handles.length === 0) {
  return Promise.resolve({} as DecryptResult);
}
```

The caller receives `{}` and the mutation's `isSuccess` and `data` fields are not updated. The real values are in the `values` map, but a caller doing:

```ts
const result = await decrypt.mutateAsync(); // all cached
doSomethingWith(result); // result === {} — empty, not { "0xH": 100n }
```

…gets a silently empty result with no error thrown. This is documented in the JSDoc, but it's a semantic inconsistency that will likely confuse consumers.

**Suggested fix:** return the cached values for the requested handles when the mutation is skipped:

```ts
if (params.handles.length === 0) {
  // Return cached values so the caller always gets a consistent result
  const cached: DecryptResult = {};
  for (const h of handles) {
    const state = queryClient.getQueryState(zamaQueryKeys.decryption.handle(h.handle));
    if (state?.data !== undefined) {
      cached[h.handle] = state.data as ClearValueType;
    }
  }
  return Promise.resolve(cached);
}
```

---

### [MEDIUM] No exported return type for `useUserDecrypt`

**File:** `packages/react-sdk/src/relayer/use-user-decrypt.ts`

The hook returns `{ ...mutation, mutate, mutateAsync, values }`, which TypeScript infers as a discriminated union of the four TanStack Query mutation states (idle / pending / error / success). This verbose union is visible in the API report (`react-sdk.api.md`) but there is no named type that consumers can reference.

Any consumer trying to store or annotate the hook's result:

```ts
// No type to use here — no exported UseUserDecryptResult
const [decryptState, setDecryptState] = useState<???>();
```

A single line would fix this:

```ts
export type UseUserDecryptResult = ReturnType<typeof useUserDecrypt>;
```

---

### [LOW] `getUncachedHandles` uses default TanStack hash function, not `hashFn`

**File:** `packages/react-sdk/src/relayer/use-user-decrypt.ts`

```ts
const state = queryClient.getQueryState(zamaQueryKeys.decryption.handle(h.handle));
```

`queryClient.getQueryState()` does not accept a `queryKeyHashFn` option — it uses the client's default hash function (`JSON.stringify`). The `useQueries` entries above it use `queryKeyHashFn: hashFn`. For the current query key shape (`["zama.decryption", { handle: "0xH" }]` — single-key object), both produce identical output so there is no practical bug today.

However, if `zamaQueryKeys.decryption.handle()` is ever extended with additional keys in the object component, the default hasher's non-deterministic key ordering could cause a hash mismatch: `getQueryState` would return `undefined` even for cached handles, breaking the no-op optimization. This is worth noting as a fragility for future maintainers.

---

### [LOW] No test for per-call `options` callbacks passed to `mutate(params, options)` / `mutateAsync(params, options)`

**File:** `packages/react-sdk/src/__tests__/use-user-decrypt.test.tsx`

The wrapped `mutate` / `mutateAsync` functions forward a second `options` parameter to the underlying `mutation.mutate()`. This enables per-call `onSuccess` / `onError` / `onSettled` callbacks (distinct from the config-level `onDecrypted` callback). There is no test verifying that these per-call options are forwarded correctly. The path is exercised implicitly but could silently break without detection.

---

## What Is Good

**The design is the right call.** Replacing three hooks with one is a genuine ergonomic improvement: the old pattern required consumers to pair a stateful `useUserDecrypt` with a separate `useUserDecryptedValue` read hook and keep them in sync manually. The new `handles` + `values` API co-locates both concerns in a single hook call, which maps naturally to how these hooks are used in practice.

**`getUncachedHandles` correctly handles falsy cached values (`0n`).** The function uses `state.dataUpdatedAt === 0` as the "not cached" check rather than `!state.data` or `state.data === undefined`. This correctly treats `0n`, `false`, and `""` as valid cached results. There is a dedicated test for this edge case.

**The `enabled: false` / error-throwing `queryFn` defensive pattern is clean.** The `useQueries` entries are subscriptions to the cache, not fetchers. Using `enabled: false` with an error-throwing `queryFn` makes the intent explicit and would surface any accidental refetch as a loud error rather than a silent network call.

**The behavior when `handles` changes between renders is correctly reactive.** `handles` is read from the hook config on each render, so adding or removing handles at runtime correctly updates the `useQueries` subscriptions and the default argument to `mutate()`.

**Test coverage for the new features is comprehensive.** The 7 new tests cover: reactive `values` map population, skipping cached handles, full no-op, `mutateAsync` with cached/uncached mix, explicit params bypassing config handles, falsy cache values, and callbacks. The deleted `useUserDecryptedValue` and `useUserDecryptedValues` test files were cleanly removed.

**Documentation is updated.** The GitBook guide, `useUserDecrypt.md` reference page, and `README.md` all show the new API pattern with the correct import and `handles` + `values` usage. `use-public-decrypt.ts` JSDoc is also updated to reflect the new `values` map instead of the removed hooks.

---

## Summary

Two items to fix before merge:
1. **Add a CHANGELOG entry** with `⚠ BREAKING CHANGES` for the removed hooks and type aliases, and a note on the `mutate` signature change.
2. **Fix the stale `DecryptParams` reference** in `README.md` (line 842).

Two quality improvements worth addressing:
3. **`mutateAsync()` when all cached** should return the cached values rather than `{}` so callers get a consistent result regardless of whether decryption was triggered.
4. **Export `UseUserDecryptResult`** so consumers can annotate the hook's return value in their TypeScript code.

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,171 @@
...
+
+  return {
+    ...mutation,
+    /** Trigger decryption (fire-and-forget). Without args, decrypts uncached handles from config. */
+    mutate,
+    /** Trigger decryption (returns promise). Without args, decrypts uncached handles from config. */
+    mutateAsync,
+    /** Reactive map of handle → decrypted cleartext (undefined if not yet decrypted). */
+    values,
+  };
```
enitrat: Here you are actively bypassing the Tanstack system we put in place, where react hooks are a thin wrapper around properly build queries/mutations that live in the SDK.

This hook here should be a three-liner wrapper around a mutation, like what was there before:

```ts
export function useUserDecrypt(config?: UseUserDecryptConfig) {
  const sdk = useZamaSDK();
  return useMutation<Record<Handle, ClearValueType>, Error, UserDecryptMutationParams>(
    userDecryptMutationOptions(sdk, config),
  );
}
```

Adding more logic here is a strong antipattern.

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,171 @@
...
+      },
+      enabled: false,
+    })),
+  });
+
+  const values: Record<Handle, ClearValueType | undefined> = {};
+  for (let i = 0; i < handles.length; i++) {
+    // oxlint-disable-next-line typescript/no-non-null-assertion
+    values[handles[i]!.handle] = cacheResults[i]?.data as ClearValueType | undefined;
+  }
```
enitrat: This logic of "don't trigger decryption on already cached handles" is spiritually good but belongs in the mutation options factory.

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,171 @@
...
+import { hashFn, userDecryptMutationOptions, zamaQueryKeys } from "@zama-fhe/sdk/query";
 import { useZamaSDK } from "../provider";
 
-export type {
-  UserDecryptCallbacks as DecryptCallbacks,
-  DecryptHandle,
-  UserDecryptMutationParams as DecryptParams,
-};
+/** A map of handles to their decrypted clear-text values. */
+export type DecryptResult = Record<Handle, ClearValueType>;
```
enitrat: should not be defined at the react-sdk layer. if we decide to create an explicit type, should be in the `sdk`, and we would want to use it in the queries for consistency (see `userDecryptMutationOptions`)

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,171 @@
 "use client";
 
-import { useMutation } from "@tanstack/react-query";
+import { useMutation, useQueries, useQueryClient } from "@tanstack/react-query";
```
enitrat: `import { useQueries } from "../utils/query";`

enitrat (changes requested):

packages/react-sdk/src/relayer/use-user-decrypt.ts
ghermet: so what about the handles?

enitrat:
Giving this another thought: I think the real problem here is that what we want is a behavior similar to the SDK's `balance-cache.ts`, not rely on Tanstack Query to supply cached values from previous calls to the `useUserDecrypt` mutation hook.



What you should do instead is have `userDecryptMutationOptions` either read from the cache (hit) or call `await sdk.relayer.userDecrypt({` (miss). that's pretty much it

docs/gitbook/src/reference/sdk/ZamaSDK.md
```diff
@@ -204,6 +204,18 @@ const sdk = new ZamaSDK({
 
 ## Properties
 
+### cache
+
+`Map<Handle, ClearValueType>` (read-only reference)
+
+In-memory cache of decrypted handle values, shared with all tokens created by this SDK instance. Keyed by the ciphertext handle — when the on-chain handle changes (e.g. after a transfer), the old entry is automatically a miss.
```
enitrat: "shared with all tokens" might not be the right formulation here

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,64 @@
...
+ *
+ * // mutate() without args decrypts all config.handles (skips cached)
+ * <button onClick={() => mutate()} disabled={isPending}>Decrypt all</button>
  * ```
  */
 export function useUserDecrypt(config?: UseUserDecryptConfig) {
   const sdk = useZamaSDK();
-
-  return useMutation<Record<Handle, ClearValueType>, Error, UserDecryptMutationParams>(
+  return useMutation<DecryptResult, Error, UserDecryptMutationParams | void>(
```
enitrat: this is not going in the direction we discussed: decryptions should be queries, not mutations; s.t. a cached decryption could be automatically displayed without being triggered by a user. If you make this a mutation that needs explicit triggers, you can't get cached decryptions to be shown on page load.

enitrat (changes requested):
I dont think we can merge it as is, as quite a few underlying issues surfaced:
- balance cache must be made a generic decryption cache for any handle/ClearValue pair at the SDK level (not token level)
- userDecrypt should be a query, not a mutation
- userDecrypt query should not have cache logic; but rather trivially delegate to the sdk cache


I will draft a PR that explains a bit better how I think we should have proceeded.

enitrat:
globally what you need is a `decrypt-cache` indexed by handle & requester address. you should probably share the public&user decryption caches, just build the key differently if it's public or user-based.

once you have that decrypt-cache at the sdk layer, you call it with a method like `sdk.decrypt` (tbd, sounds like a good place to put it for now). Then, SDK consumers can simply call `decrypt` and get cached-optimisations directly, and you wrap this up in a simple tanstack query mirroring the config for confidential balances.

finally you just wrap that query as a react hook.

enitrat:
see: https://github.com/zama-ai/sdk/pull/195

packages/react-sdk/src/relayer/use-user-decrypt.ts
ghermet: I see the appeal of query semantics for auto-display on page load, but I think mutation is the right choice here — here's why:

**Decryption has a real side effect.** The first call for uncached handles triggers `credentials.allow()`, which prompts a wallet signature. TanStack Query expects query functions to be pure reads — side-effectful queries break assumptions around retries, refetches, and window focus re-validation. A failed wallet prompt retried automatically by TanStack is a bad UX.

**The "call `sdk.allow()` upfront" workaround shifts the problem, it doesn't solve it.** If the app must call `allow()` before queries work without side effects, you've just moved the mutation to a different place. And if the user hasn't called `allow()` yet, the query silently triggers a wallet popup on mount — which is worse than an explicit button.

**Auto-display of cached values doesn't require query semantics.** With the persistent `DecryptCache` we just landed, a component can read cached values directly on mount:

```tsx
const cached = await sdk.cache.get(signer, contract, handle);
```

A thin `useDecryptedValue(handle)` read-only hook over the cache would give auto-display without conflating reads with the decrypt-and-sign flow.

**Separation of concerns is cleaner:**
- `useUserDecrypt` = mutation = "go decrypt these handles" (may sign)
- `useDecryptedValue` = query = "give me the cached value for this handle" (pure read, `staleTime: Infinity`)

This keeps the wallet-prompting path explicit and the cache-reading path automatic — best of both worlds without bending query semantics.

enitrat: disagree

enitrat: if you can propose me a solution using a mutation that would auto-trigger on load under certain conditions, I could get behind that. Otherwise, strong no on having separate cached/uncached hooks (this is not something the consumer wants to know about), and strong no on not having auto-enabled decryptions hooks

packages/sdk/src/zama-sdk.ts
```diff
@@ -272,6 +291,90 @@ export class ZamaSDK {
...
+   *
+   * @example
+   * ```ts
+   * const values = await sdk.decrypt([
+   *   { handle: balanceHandle, contractAddress: cUSDT },
+   * ]);
+   * console.log(values[balanceHandle]); // 1000n
+   * ```
+   */
+  async decrypt(
```
enitrat: userDecrypt as you'll have the public equivalent later on

packages/sdk/src/query/is-allowed.ts
```diff
@@ -1,24 +1,26 @@
...
   sdk: ZamaSDK,
   config: IsAllowedQueryConfig,
 ): QueryFactoryOptions<boolean, Error, boolean, ReturnType<typeof zamaQueryKeys.isAllowed.scope>> {
+  const { account, contractAddresses = [], query = {} } = config;
   return {
-    ...filterQueryOptions(config.query ?? {}),
-    queryKey: zamaQueryKeys.isAllowed.scope(config.account),
-    queryFn: () => sdk.credentials.isAllowed(),
+    ...filterQueryOptions(query),
+    queryKey: zamaQueryKeys.isAllowed.scope(account),
```
enitrat: query key is only scoped by account. you need to scope with contractAddresses as well.

packages/sdk/src/query/is-allowed.ts
```diff
@@ -1,24 +1,26 @@
...
-import { filterQueryOptions } from "./utils";
 import { zamaQueryKeys } from "./query-keys";
-import type { Address } from "viem";
+import { filterQueryOptions } from "./utils";
 
 export interface IsAllowedQueryConfig {
   account: Address;
-  query?: Record<string, unknown>;
+  /** When provided, also checks that cached credentials cover these contracts. */
+  contractAddresses?: Address[];
```
enitrat: why make this optional?

packages/sdk/src/token/readonly-token.ts
```diff
@@ -385,36 +390,22 @@ export class ReadonlyToken {
...
       const handle = resolvedHandles[i]!;
-      const cached = cachedValues[i];
-
-      if (cached !== null && cached !== undefined) {
+      if (token.isZeroHandle(handle)) {
+        results.set(token.address, 0n);
+        continue;
+      }
+      const cached = await firstToken.cache.get(ownerAddress, token.address, handle);
+      if (cached !== null) {
```
enitrat: you used to have parallel operations in the previous logic which you lost here.

packages/sdk/src/credentials/credentials-manager-base.ts
```diff
@@ -295,12 +295,27 @@ export abstract class BaseCredentialsManager<
     } as ZamaSDKEventInput);
   }
 
-  protected async checkAllowed(key: string): Promise<boolean> {
+  protected async checkAllowed(key: string, contractAddresses?: Address[]): Promise<boolean> {
```
enitrat: why make contractAddresses optional here?

packages/sdk/src/query/is-allowed.ts
```diff
@@ -1,24 +1,26 @@
...
+  query?: { enabled?: boolean };
 }
 
 export function isAllowedQueryOptions(
   sdk: ZamaSDK,
   config: IsAllowedQueryConfig,
 ): QueryFactoryOptions<boolean, Error, boolean, ReturnType<typeof zamaQueryKeys.isAllowed.scope>> {
+  const { account, contractAddresses = [], query = {} } = config;
```
enitrat: avoid destructuration here for consistency

packages/sdk/src/query/is-allowed.ts
```diff
@@ -1,24 +1,26 @@
...
   return {
-    ...filterQueryOptions(config.query ?? {}),
-    queryKey: zamaQueryKeys.isAllowed.scope(config.account),
-    queryFn: () => sdk.credentials.isAllowed(),
+    ...filterQueryOptions(query),
+    queryKey: zamaQueryKeys.isAllowed.scope(account),
+    queryFn: () => sdk.credentials.isAllowed(...contractAddresses),
     staleTime: 30_000,
-    enabled: config.query?.enabled !== false,
+    enabled: query?.enabled,
```
enitrat: restore the explicit false check, unless it's intentional?

packages/sdk/src/activity.ts
```diff
@@ -20,6 +20,8 @@ import {
 import type { Address, Hex } from "viem";
 import type { Handle } from "./relayer/relayer-sdk.types";
 import { ZERO_HANDLE } from "./token/readonly-token";
+import type { ClearValueType } from ".";
```
enitrat: import from the proper original location

packages/sdk/src/query/user-decrypt.ts
```diff
@@ -10,66 +10,42 @@ export interface DecryptHandle {
...
-  onCredentialsReady?: () => void;
-  /** Fired after decryption completes. */
-  onDecrypted?: (values: Record<Handle, ClearValueType>) => void;
-}
-
-export function userDecryptMutationOptions(
+/**
+ * Query factory for user decryption of FHE handles.
+ *
+ * Calls `sdk.decrypt()` which checks the persistent cache first, then
```
enitrat: you should not mention this in the item docs. it is going to be a nightmare to maintain all of this docs. Stay succint with the bare minimum.

packages/react-sdk/src/authorization/use-revoke.ts
```diff
@@ -28,6 +28,7 @@ export function useRevoke(options?: UseMutationOptions<void, Error, Address[]>)
     onSuccess: (data, variables, onMutateResult, context) => {
       options?.onSuccess?.(data, variables, onMutateResult, context);
       void context.client.invalidateQueries({ queryKey: zamaQueryKeys.isAllowed.all });
+      void context.client.invalidateQueries({ queryKey: zamaQueryKeys.decryption.all });
```
enitrat: after a revoke, what you want is deletion of the data, not simple invalidation

enitrat (changes requested):

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,76 @@
...
+  const allowedOpts = signerAddress
+    ? isAllowedQueryOptions(sdk, { account: signerAddress, contractAddresses })
+    : ({
+        queryKey: zamaQueryKeys.isAllowed.all,
+        queryFn: skipToken,
+        enabled: false,
+      } as const);
+  const allowedQuery = useQuery<boolean>(allowedOpts);
+  const isAllowed = allowedQuery.data === true;
+
```
enitrat: i am still not convinced this is a good thing. the SDK integrator owns the UX decision of when decryption should happen. Whether that hook should trigger or not is the responsibility of an integrator, that, upon its state being ready, will decide to flip the hook as "enabled".

By adding an allowance check in the hook, you make the lifecycle harder to reason about, and you enforce an integrator to follow your "allow-and-only-then-decrypt", instead of being able to, on purpose, do a simple call to `useUserDecrypt`. 

An integrator of the SDK should only be thinking about: "should this run now?", when its component is ready, and not have to think about any possible internal behavior that could contradict its behavior.

enitrat: after second thought I am actively leaning against this. You are forcing an integrator to explicitly allow before calling the hook, decoupling the behavior of a `useUserDecrypt` from the underlying `sdk.userDecrypt` operation, and instead of empowering an integrator with primitives you are restricting it. That should not be a behavior imposed by the SDK, as you cannot bypass it in the current form.

packages/sdk/src/credentials/credentials-manager-base.ts
ghermet: I guess based on what you said before all signatures are tied to the contracts so we should enforce it.

packages/sdk/src/query/is-allowed.ts
ghermet: helps with the defaults thought, but happy to circle back to this topic another time

packages/sdk/src/query/is-allowed.ts
ghermet: `config.query?.enabled !== false` is flawed as any other falsy value would be truthy here. it'd be better to type `enabled` as a boolean in the config and let the query option infer whether its true or false.

packages/sdk/src/query/is-allowed.ts
ghermet: it'd be a breaking change, but maybe that's not a biggie

packages/react-sdk/src/relayer/use-user-decrypt.ts
ghermet: Well let's put it this way, should we be the guardians of bad UX? or that's something we're not concerned about and the burden is on the integrator?

ghermet: the original aim of the locked query by allowed contracts was to prevent surprise signature requests. There's no way of getting around it in its current form. What will the integrator think upon realising that this doesn't work out of the box and that he needs to tie allowed contracts to the user decrypt query?

ghermet: If the query threw an error instead of surprise signature, would that convince you?

enitrat: > the original aim of the locked query by allowed contracts was to prevent surprise signature requests

This is making an opinionated claim about how integrators should be using the SDK without taking into consideration that different people might have different needs or views into how it should be used. You're locking users into an opinionated pattern that might not be relevant to their use case. When the query is enabled, it should run the flow; regardless of whether a credential is cached or a handle pre-approved. Enabling the query is the responsibility of an integrator: on button click, on component mount, on transaction success, whatever it decides. If a signature triggers at that time, well, that's the intended behavior. It's up to the UI to make this clear why it's happening.

packages/sdk/src/query/is-allowed.ts
enitrat: > config.query?.enabled !== false is flawed as any other falsy value would be truthy here

That's because the intended behavior is to have the query enabled, by default, UNLESS specifically disabled.

packages/sdk/src/query/is-allowed.ts
enitrat: yes, the e2e flow related to allowance must consider that contract addresses are supplied. It doesn't make sense to check an allowance without providing an associated contract

packages/react-sdk/src/authorization/use-revoke-session.ts
```diff
@@ -23,6 +23,7 @@ export function useRevokeSession(options?: UseMutationOptions<void>) {
     onSuccess: (data, variables, onMutateResult, context) => {
       options?.onSuccess?.(data, variables, onMutateResult, context);
       void context.client.invalidateQueries({ queryKey: zamaQueryKeys.isAllowed.all });
+      void context.client.invalidateQueries({ queryKey: zamaQueryKeys.decryption.all });
```
enitrat:
```suggestion
      void context.client.removeQueries({ queryKey: zamaQueryKeys.decryption.all });
```

same reason as comment below

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -247,38 +248,40 @@ function ConfidentialAction() {
...
 
 ```tsx
-import { useUserDecrypt, useUserDecryptedValue } from "@zama-fhe/react-sdk";
+import { useAllow, useIsAllowed, useUserDecrypt } from "@zama-fhe/react-sdk";
 
 function DecryptExample() {
-  const decrypt = useUserDecrypt();
-  const { data: decryptedValue } = useUserDecryptedValue("0xabc123...");
+  const { mutate: allow, isPending: isAllowing } = useAllow();
+  const { data: isAllowed } = useIsAllowed();
```
enitrat: missing contract addresses

packages/react-sdk/README.md
```diff
@@ -826,15 +826,15 @@ const result = await decrypt.mutateAsync({
   ],
 });
 // result: { "0xabc...": 500n, "0xdef...": 1000n }
-// Results are automatically cached for useUserDecryptedValue
+// Results are automatically cached — read via useDecryptedValue / useDecryptedValues
```
enitrat: no longer existts

packages/react-sdk/README.md
```diff
@@ -854,30 +854,22 @@ const result = await decrypt.mutateAsync({
...
-// Single handle
-function useUserDecryptedValue(handle: string | undefined): UseQueryResult<bigint>;
-
-// Multiple handles
-function useUserDecryptedValues(handles: string[]): {
-  data: Record<string, bigint | undefined>;
-  results: UseQueryResult<bigint>[];
-};
-```
+`useUserDecrypt` populates the persistent decrypt cache. Use `useDecryptedValue` / `useDecryptedValues` to read cached values on mount (even after page reload):
```
enitrat: no longer exists

packages/react-sdk/README.md
```diff
@@ -854,30 +854,22 @@ const result = await decrypt.mutateAsync({
...
 
 ```tsx
-// First, trigger decryption (populates the cache)
 const decrypt = useUserDecrypt();
-await decrypt.mutateAsync({
+const { data: balance } = useDecryptedValue({
```
enitrat: ^

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,21 @@
...
 
-export type {
-  UserDecryptCallbacks as DecryptCallbacks,
-  DecryptHandle,
-  UserDecryptMutationParams as DecryptParams,
-};
-
-/** Configuration for {@link useUserDecrypt}. */
-export type UseUserDecryptConfig = UserDecryptCallbacks;
+/** @deprecated Use {@link UserDecryptQueryConfig} directly. */
```
enitrat: do we keep, do we delete later? if so what's the strategy?

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,21 @@
...
- * });
- * decrypt.mutate({
- *   handles: [{ handle: "0xHandle", contractAddress: "0xContract" }],
- * });
- * ```
+ * React hook for FHE user decryption. Thin wrapper around
+ * `userDecryptQueryOptions` with `useQuery` semantics.
  */
-export function useUserDecrypt(config?: UseUserDecryptConfig) {
+export function useUserDecrypt(config: UserDecryptQueryConfig, options?: { enabled?: boolean }) {
```
enitrat: the options passed here must be the Tanstack Query options — except queryKey and queryFn which we inject ourselves.

```
Omit<UseQueryOptions<T>, "queryKey" | "queryFn">,
```

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -16,22 +16,24 @@ Here is a complete flow that encrypts a value, sends it to a custom FHE contract
...
 function ConfidentialRoundTrip() {
   const sdk = useZamaSDK();
   const encrypt = useEncrypt();
-  const decrypt = useUserDecrypt();
-  const [handle, setHandle] = useState<string>();
-  const { data: decryptedValue } = useUserDecryptedValue(handle);
+  const { mutate: allow } = useAllow();
+  const [handles, setHandles] = useState<{ handle: string; contractAddress: `0x${string}` }[]>([]);
+
+  // useUserDecrypt fires automatically once allow() is called
```
enitrat: no longer true

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -288,99 +291,32 @@ function DecryptExample() {
...
 
-`useUserDecrypt` automatically groups handles by contract address and issues one decryption request per contract:
+`useUserDecrypt` automatically groups handles by contract address and issues one decryption request per contract. Make sure all contracts are authorized via `useAllow`:
 
 ```tsx
-const result = await decrypt.mutateAsync({
+// Authorize all contracts upfront
+allow(["0xTokenA", "0xTokenB"]);
+
+// useUserDecrypt fires once credentials cover both contracts
```
enitrat: untrue

enitrat (changes requested):

packages/react-sdk/src/relayer/use-user-decrypt.ts
ghermet: must delete this

packages/sdk/src/credentials/credentials-manager-base.ts
```diff
@@ -295,6 +295,8 @@ export abstract class BaseCredentialsManager<
     } as ZamaSDKEventInput);
   }
 
+  // Optional because DelegatedCredentialsManager calls without contract addresses
+  // (delegated credentials aren't scoped per-contract).
```
enitrat: this is weird. are you sure this is true?

ghermet: thats wrong

enitrat: how did this even make it into a commit? I don't know how you are driving claude, but if claims as false as this make it into the code, I cannot trust anything that is being submitted in PRs.

You are rationalizing an API inconsistency that is demonstrating a clear bug in the SDK by adding comments over internal function rather than questioning why `DelegatedCredentialsManager` would not be checking contract addresses and we'd require to pass optional values. This is clearly incorrect and the root cause should be fixed, rather than justifying bad behavior with a comment...

packages/sdk/src/credentials/credentials-manager-base.ts
```diff
@@ -295,8 +295,8 @@ export abstract class BaseCredentialsManager<
     } as ZamaSDKEventInput);
   }
 
-  // Optional because DelegatedCredentialsManager calls without contract addresses
-  // (delegated credentials aren't scoped per-contract).
+  // Optional because DelegatedCredentialsManager.isAllowed only checks session
```
enitrat: why? there's no fundamental difference between delegated and user decryptions. both signed data are over a specific set of contracts

ghermet: yes indeed thats the same issue as above

packages/sdk/src/token/readonly-token.ts
```diff
@@ -752,16 +742,9 @@ export class ReadonlyToken {
...
-    // Check persistent cache keyed by (token, owner, handle).
-    // When the on-chain balance changes the encrypted handle changes too,
-    // so stale entries are never served — no TTL needed.
-    const cached = await loadCachedBalance({
-      storage: this.storage,
-      tokenAddress: this.address,
-      owner: normalizedOwner,
-      handle,
-    });
+    const cached = await this.cache.get(normalizedOwner, this.address, handle);
```
enitrat, line 745: as i'm reading this part of the code I have a few questions to address in a subsequent PR (as it's not directly related to changes here:

1. `normalizedOwner` used here is the explicit `owner` OR the `normalizedDelegator`. Should this not be `owner` or `signer` ? the doc says:

"The connected signer acts as the delegate who has been granted permission by the delegator to decrypt their balance". So the balance you should be checking here is the `normalizedDelegator`, right? not the `owner`? 

2. I am a bit confused by the "owner" here as the terminology used throughout the flow is `delegator/delegatee`. If I understand correctly, the cache key used should for delegated decryptions should always be the delegatee's address, right?

packages/sdk/src/credentials/credentials-manager.ts
```diff
@@ -106,9 +106,9 @@ export class CredentialsManager extends BaseCredentialsManager<
     await this.revokeSession(key);
   }
 
-  /** Whether a session signature is currently cached. */
-  async isAllowed(): Promise<boolean> {
-    return this.checkAllowed(await this.#storeKey());
+  /** Whether a session signature is currently cached and covers the given contracts. */
+  async isAllowed(...contractAddresses: Address[]): Promise<boolean> {
```
enitrat: Would this not be better to not use a variadic parameter here, but enforce a concrete `contractAddresses: Address[]` type? what I'm thinking is:

- it doesnt make sense to check allowance for no contracts (what do you even check?): the value passed is a collection of contract addresses
- it's better when callers already have an array, which is the case here, as they can pass it directly

the same holds for `allow` and `revoke` which also use variadic parameters

packages/sdk/src/credentials/credentials-manager-base.ts
```diff
@@ -295,12 +295,27 @@ export abstract class BaseCredentialsManager<
...
+        if (!stored) {
+          return false;
+        }
+        this.assertEncrypted(stored);
+        return coversContracts(stored.contractAddresses, contractAddresses);
+      } catch {
+        return false;
+      }
+    }
+    return true;
```
enitrat: nit: early `return true` on `contractAddresses.length  == 0` instead to remove one nesting layer and reduce cognitive complexity

packages/sdk/src/credentials/credentials-manager.ts
ghermet: If we consider that people will have more than one contract anyway we could ditch the variadic

enitrat: ah, so it's about `isAllowed(tokenA)` vs `isAllowed([tokenA])`. no strong opinon on that then... perhaps better to keep `isAllow(tokenA)`

packages/sdk/src/token/readonly-token.ts
enitrat: you can create a linear ticket to follow up on that

ghermet: Here's the [Linear ticket](https://linear.app/zama/issue/SDK-62/clarify-owner-vs-delegator-semantics-in-readonlytoken-balance-cache) to iron this out

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,22 @@
...
+export function useUserDecrypt(
+  config: UserDecryptQueryConfig,
+  options?: Omit<UseQueryOptions<DecryptResult>, "queryKey" | "queryFn">,
+) {
   const sdk = useZamaSDK();
-
-  return useMutation<Record<Handle, ClearValueType>, Error, UserDecryptMutationParams>(
-    userDecryptMutationOptions(sdk, config),
-  );
+  return useQuery<DecryptResult>({ ...userDecryptQueryOptions(sdk, config), ...options });
```
enitrat: passing `...options` here means that the `enabled` behavior will only be dependent on the user's passed value, not the factory guard. you should align with our pattern where `enabled: factoryEnabled && (options?.enabled ?? true)`

packages/sdk/src/zama-sdk.ts
```diff
@@ -268,6 +287,90 @@ export class ZamaSDK {
...
+      }
+    }
+
+    if (uncached.length === 0) {
+      onDecrypted(result);
+      return result;
+    }
+
+    // Get unique contract addresses and acquire credentials
+    const contractAddresses = [...new Set(uncached.map((h) => h.contractAddress))];
```
enitrat, line 335: sanitize user input to avoid eventual duplicated values passed
```suggestion
    const contractAddresses = [...new Set(uncached.map((h) => getAddress(h.contractAddress)))];
```

packages/sdk/src/token/readonly-token.ts
```diff
@@ -878,22 +844,12 @@ export class ReadonlyToken {
...
-          owner: signerAddress,
-          handle,
-          value,
-        });
-      } catch (cacheError) {
-        // oxlint-disable-next-line no-console
-        console.warn("[zama-sdk] Cache write failed (non-blocking):", cacheError);
-      }
+      await this.cache.set(signerAddress, this.address, handle, value);
+      assertBigint(value, "decryptBalance: result[handle]");
```
enitrat, line 851: asserting after setting the cache means that all `cache.get` will fail, better to move it before

packages/sdk/src/token/readonly-token.ts
enitrat, line 866: to address in a future PR: should this be a method of `readonly-token` ?

enitrat, line 894: this should go though the main `sdk.userDecrypt` if we decide to keep it. to address in a future PR, please link a linear ticket

packages/sdk/src/query/query-keys.ts
```diff
@@ -174,7 +174,14 @@ export const zamaQueryKeys = {
 
   isAllowed: {
     all: ["zama.isAllowed"] as const,
-    scope: (account: Address) => ["zama.isAllowed", { account: getAddress(account) }] as const,
+    scope: (account: Address, contractAddresses: Address[]) =>
+      [
+        "zama.isAllowed",
+        {
+          account: getAddress(account),
+          contractAddresses: normalizeAddresses(contractAddresses),
```
enitrat: you probably want to sort the handles here, as you do for zama.decryption

enitrat (changes requested):
last set of things to address, then we'll be good

packages/sdk/src/zama-sdk.ts
```diff
@@ -268,6 +287,90 @@ export class ZamaSDK {
...
+    }
+
+    // Get unique contract addresses and acquire credentials
+    const contractAddresses = [...new Set(uncached.map((h) => getAddress(h.contractAddress)))];
+    const creds = await this.credentials.allow(...contractAddresses);
+    onCredentialsReady();
+
+    // Group uncached handles by contract address
+    const byContract = new Map<Address, Handle[]>();
+    for (const h of uncached) {
```
enitrat, line 341:
```suggestion
    for (const h of contractAddresses) {
```

enitrat (dismissed):

enitrat (dismissed):

packages/sdk/src/credentials/credentials-manager-base.ts
```diff
@@ -307,7 +307,7 @@ export abstract class BaseCredentialsManager<
       return false;
     }
     if (contractAddresses.length === 0) {
-      return true;
+      return false;
```
enitrat: why? if you check the allowance of an empty set, it's always true, no?

enitrat (changes requested):

packages/sdk/src/credentials/credentials-manager-base.ts
ghermet: cf Ankur's comment:

> I think there’s a legacy issue that is a fundamental flaw in logic that impacts PR 125 as well well PR 214 (cc: [@Corentin R](https://zama-ai.slack.com/team/U0A79DJUHNW)). I was a bit surprised to see this in checkAllowed at credentials-manager-base.ts:309:
> 
>  if (contractAddresses.length === 0) {
>   return true; // ← just checks "any session exists"             
>  }
> 
> If someone calls credentials.isAllowed() with no arguments, they get true as long as any session exists, regardless of what contracts it covers. I think there’s a similar issue in the batching PR as well on checkAllowed / isAllowed().

enitrat: the query here is "am I allowed to decrypt nothing" and the answer would always be "yes". no ?

ghermet: any session signature has at least one contract so we shouldn't even have to check contract addresses length no?

enitrat: yes, technically if you call  `coversContracts`, which simply checks that the set of contracts is a subset of the set of signed contracts, because an empty set is a subset of all sets, it would always true. with an early return you just know you have nothing to check for.

ghermet: The real question is: _**is there a situation where you wanna check there's a signature regardless of its contracts**_?

ghermet: Even in practice this ` if (contractAddresses.length === 0) {   return true }` is correct I think it allows the developer to dissociate the contracts from their signatures

packages/sdk/src/credentials/credentials-manager.ts
```diff
@@ -107,8 +107,8 @@ export class CredentialsManager extends BaseCredentialsManager<
   }
 
   /** Whether a session signature is currently cached and covers the given contracts. */
-  async isAllowed(...contractAddresses: Address[]): Promise<boolean> {
-    return this.checkAllowed(await this.#storeKey(), contractAddresses);
+  async isAllowed(contractAddress: Address, ...moreAddresses: Address[]): Promise<boolean> {
+    return this.checkAllowed(await this.#storeKey(), [contractAddress, ...moreAddresses]);
```
enitrat: I think `checkAllowed` has a good signature here that properly carries the property that you pass "one non-empty list of contract". the public-facing version with variadic params and "moreAddresses" splits the single "contracts to check" argument among two parameters,  and forces callers into `isAllowed(...contracts)`, which might be problematic if the array is not statistically known to be non-empty. wdyt ?

ghermet: I think this is spot on, maybe `isAllowed(["0X32434"])` is an acceptable concession to make this type safe using `contractAddresses: [Adress,...Address[]]`

ghermet:
**SDK-78 acceptance criteria status** ([Linear](https://linear.app/zama/issue/SDK-78/enforce-eip-712-contract-scope-in-session-credential-checks))

Tightened `checkAllowed` in `packages/sdk/src/credentials/credentials-manager-base.ts` to delegate to `isCredentialValid(stored, contractAddresses)` instead of the direct `coversContracts(...)` call. This unifies the validation logic with `resolveCredentials` and picks up the keypair time-validity check for free.

All five ACs now satisfied:

- **AC 1 — contract scope**: `checkAllowed` runs `coversContracts` via `isCredentialValid`.
- **AC 2 — no context ⇒ not allowed**: `CredentialsManager.isAllowed` and `DelegatedCredentialsManager.isAllowed` take `[Address, ...Address[]]`; the type system rejects zero-arg calls.
- **AC 3 — keypair time validity (SDK-60)**: `isCredentialValid` includes `isTimeValid(stored)`, so a live session over an expired keypair no longer reports allowed.
- **AC 4 — single source of truth**: both `checkAllowed` and the fast-path in `resolveCredentials` go through `isCredentialValid`.
- **AC 5 — internal call sites**: only `ReadonlyToken.isAllowed()` (passes `[this.address]`) and `Token.decryptBalance` (delegates to the token-scoped call) consume these. Both correctly scoped.

ghermet:
**SDK-80 test coverage** ([Linear](https://linear.app/zama/issue/SDK-80/redgreen-test-suite-eip-712-contract-scope-enforcement-in-credential))

Added the missing red/green tests in 76a69ebf so the EIP-712 contract scope rules are now enforced by the suite, not just the type system.

`packages/sdk/src/credentials/__tests__/credentials-manager.test.ts`
- `isAllowed([A]) returns true when signature covers [A, B]` — row 4 (signature scope is a strict superset of the query)
- `isAllowed() returns false when keypair has expired but session has not` — row 7 / SDK-60 (tampers stored `startTimestamp` past the keypair window with the session entry still live; would have caught the bug fixed by routing `checkAllowed` through `isCredentialValid`)
- `invariant: isAllowed true ⇒ allow() reuses cache and does not re-sign` — row 13 (asserts on `signer.signTypedData` call count)
- `invariant: isAllowed false ⇒ allow() produces a fresh EIP-712 signature` — row 14

`packages/sdk/src/credentials/__tests__/delegated-credentials-manager.test.ts`
- `isAllowed() returns false when contract is outside the signed scope` — row 11
- `isAllowed() returns false for a delegator with no stored credentials` — row 12
- `invariant: isAllowed true ⇒ allow() reuses cache and does not re-sign` — row 15

Notes on the rows not added as runtime tests:
- Rows 1 and 9 (no contract context) are unreachable: `isAllowed` requires `[Address, ...Address[]]`, so the type system rejects them at compile time.
- Rows 16 and 17 (React hook integration) live in `use-is-allowed.test.tsx` and are out of scope for this commit.

All 68 tests in the two files pass. Suggest closing SDK-80 once the PR lands.

enitrat (dismissed):

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -1,52 +1,27 @@
...
   const sdk = useZamaSDK();
-
-  return useMutation<Record<Handle, ClearValueType>, Error, UserDecryptMutationParams>(
-    userDecryptMutationOptions(sdk, config),
-  );
+  const queryOpts = userDecryptQueryOptions(sdk, config);
+  return useQuery<DecryptResult>({
+    ...queryOpts,
+    ...options,
+    enabled: queryOpts.enabled && (options?.enabled ?? true),
```
enitrat: this is enabled by default, not disabled by default, is that intended?

ghermet: not intended should be `queryOpts.enabled && options?.enabled`

enitrat: add test "it should not fire by default even when query conditions are met"

ankurdotb:
## Review — remaining items before merge

Reviewed the current state of the branch after the SDK-78/SDK-80 fixes. The EIP-712 contract scope enforcement and test suite look solid. Four items still need attention:

### 1. `decryptHandles` and `decryptBalance` bypass `sdk.userDecrypt()` and the `DecryptCache`

The PR description says it "lifts decrypt orchestration into `sdk.userDecrypt()`" and "unifies all decryption paths through a single cache-enabled entry point." But two methods on `ReadonlyToken` still call `this.relayer.userDecrypt()` directly:

- **`decryptHandles`** (`readonly-token.ts:894`) — doesn't read from or write to the `DecryptCache` at all. Results are lost after the call.
- **`decryptBalance`** (`readonly-token.ts:831`) — manually duplicates the cache read/write logic instead of delegating to `sdk.userDecrypt()`.

Both should delegate to the SDK-level method so there's a single orchestration path. enitrat flagged this in review — this should be resolved in this PR, not deferred, since it contradicts the PR's stated goal.

### 2. README code example is stale

`packages/react-sdk/README.md:746-757` still shows the old mutation pattern:

```tsx
// Wrong — hook is now a query, not a mutation
const decrypt = useUserDecrypt({
  onCredentialsReady: () => setStep("decrypting"),
  onDecrypted: (values) => setStep("done"),
});
const result = await decrypt.mutateAsync({ handles: [...] });
```

Should show query usage with `{ handles, enabled }` config.

### 3. `is-allowed.ts:23` enabled check is flawed

```ts
enabled: config.query?.enabled !== false
```

Any falsy value other than `false` (`undefined`, `0`, `""`) passes through as truthy. Should be `config.query?.enabled ?? true` or typed as a strict boolean.

### 4. SDK-80 minor test gap

Missing the React hook happy-path test (row 17 in the SDK-80 spec): `isAllowed=true` → `useUserDecrypt` fires and decrypts silently without a wallet prompt. The inverse case (`isAllowed=false` → idle) is covered at line 98-120 of `use-user-decrypt.test.tsx`.

enitrat:
> 1. decryptHandles and decryptBalance bypass sdk.userDecrypt() and the DecryptCache

see comment above, this should be given more thought in another task

enitrat (dismissed):

ankurdotb:
> see comment above, this should be given more thought in another task

Agree on `ReadOnlyToken` being truly read-only (if it's used for decrypt, it's not truly read-only). We should definitely think more about the minimal read-only set that `ReadOnlyToken` _should_ do. cc: @ghermet 

Started a follow-up task here but might need to be refined: https://linear.app/zama/issue/SDK-83/readonlytoken-should-be-truly-read-only-move-decrypt-methods-to-token

enitrat (approved):

packages/react-sdk/src/relayer/use-user-decrypt.ts
ghermet: Added a test for this specific feature

ankurdotb (approved):

PR #127, opened 2026-03-19
refactor(sdk): replace CDN with bundled relayer-sdk UMD asset

ghermet:
## Summary

- Replace runtime CDN loading of `@zama-fhe/relayer-sdk` with a build-time copy of the UMD bundle from `node_modules`
- New `relayer-sdk-umd-plugin` copies the UMD into `dist/esm/` at build time; the worker loads it via `importScripts` from a co-located file
- Worker client loads the worker IIFE via `new URL(filename, import.meta.url)` instead of blob URL
- Remove CDN URL validation, SHA-384 integrity checking, and all CDN-related config (`cdnUrl`, `integrity`, `integrityCheck`)
- Fix latent `DataCloneError` bug: `logger` (containing functions) was being spread into `postMessage` payload
- Fix browser extension path: use `runtime.getURL()` for both worker and SDK URLs
- Fix silent 30s hang on unknown worker request types: now sends error response immediately
- Add actionable error messages for `importScripts` failures and missing UMD at build time
- Update all documentation, security model, CSP directives, and D2 diagrams

### Bundle sizes

| File | Before | After |
|---|---|---|
| `index.js` (main) | 40 KB | 22 KB |
| `relayer-sdk.worker.js` | 18 KB (IIFE) | 14 KB (IIFE) |
| `relayer-sdk-js.umd.cjs` | — (CDN) | 783 KB (co-located, loaded on demand) |

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] All 1167 unit tests pass
- [x] `pnpm build` succeeds with correct output structure
- [x] Bundler smoke tests verify worker + UMD sibling assets
- [ ] Playwright E2E tests (CI)


ankurdotb (changes requested):
**Do not merge**. This removes an important defense-in-depth layer.

The current CDN + SHA-384 integrity check ensures the SDK only executes a known-good, reviewed artifact at runtime. This PR replaces that with a build-time copy from `node_modules`, which assumes the local copy hasn't been tampered with: either upstream (compromised registry, dependency poisoning) or locally (malicious postinstall, poisoned lockfile).

I'd prefer we revert this and keep the CDN download with integrity checking. The real problem is the maintenance burden of updating `RELAYER_SDK_VERSION` and `CDN_INTEGRITY` manually. The fix for that might be a commit hook or CI script that:

- Detects when `@zama-fhe/relayer-sdk` is bumped
- Computes the new SHA-384 and updates the constants automatically
- If it can't compute the hash automatically, fails lint/CI and prompts the developer to update it manually

ghermet:
> **Do not merge**. This removes an important defense-in-depth layer.
> 
> The current CDN + SHA-384 integrity check ensures the SDK only executes a known-good, reviewed artifact at runtime. This PR replaces that with a build-time copy from `node_modules`, which assumes the local copy hasn't been tampered with: either upstream (compromised registry, dependency poisoning) or locally (malicious postinstall, poisoned lockfile).
> 
> I'd prefer we revert this and keep the CDN download with integrity checking. The real problem is the maintenance burden of updating `RELAYER_SDK_VERSION` and `CDN_INTEGRITY` manually. The fix for that might be a commit hook or CI script that:
> 
> * Detects when `@zama-fhe/relayer-sdk` is bumped
> * Computes the new SHA-384 and updates the constants automatically
> * If it can't compute the hash automatically, fails lint/CI and prompts the developer to update it manually

if the `@zama-fhe/relayer-sdk` package integrity is comprised the CDN most likely will be too

> **Do not merge**. This removes an important defense-in-depth layer.
> 
> The current CDN + SHA-384 integrity check ensures the SDK only executes a known-good, reviewed artifact at runtime. This PR replaces that with a build-time copy from `node_modules`, which assumes the local copy hasn't been tampered with: either upstream (compromised registry, dependency poisoning) or locally (malicious postinstall, poisoned lockfile).
> 
> I'd prefer we revert this and keep the CDN download with integrity checking. The real problem is the maintenance burden of updating `RELAYER_SDK_VERSION` and `CDN_INTEGRITY` manually. The fix for that might be a commit hook or CI script that:
> 
> * Detects when `@zama-fhe/relayer-sdk` is bumped
> * Computes the new SHA-384 and updates the constants automatically
> * If it can't compute the hash automatically, fails lint/CI and prompts the developer to update it manually

We can do the SHA-384 verification on the bundled file via the rolldown plugin.

ankurdotb:
> if the @zama-fhe/relayer-sdk package integrity is comprised the CDN most likely will be too

If it was compromised at source, i.e., in the Relayer SDK, then sure because the CDN will have that infected copy. I meant a different vector, which is a malware tampering with the copy in `node_modules` _after_ a legitimate version has been downloaded. Is there an integrity check that we can protect from that scenario? 🤔 

I suppose the other point is, with the CDN technique the Relayer SDK is downloaded at runtime and not part of the SDK bundle/package distributed, whereas with this technique, it becomes part of the SDK package's footprint right? (Is that the reason why someone might use it via CDN in the first place?)

ghermet:
> > if the @zama-fhe/relayer-sdk package integrity is comprised the CDN most likely will be too
> 
> If it was compromised at source, i.e., in the Relayer SDK, then sure because the CDN will have that infected copy. I meant a different vector, which is a malware tampering with the copy in `node_modules` _after_ a legitimate version has been downloaded. Is there an integrity check that we can protect from that scenario? 🤔
> 
> I suppose the other point is, with the CDN technique the Relayer SDK is downloaded at runtime and not part of the SDK bundle/package distributed, whereas with this technique, it becomes part of the SDK package's footprint right? (Is that the reason why someone might use it via CDN in the first place?)

It's mostly about whether you want to load the script at runtime or buildtime. Most bundlers don't have a convention to load assets so the CDN would be best. Alexandre B said that the CDN is vowed to be deprecated at some point, I don't know if we should rely on it.

ghermet:
> > > if the @zama-fhe/relayer-sdk package integrity is comprised the CDN most likely will be too
> > 
> > 
> > If it was compromised at source, i.e., in the Relayer SDK, then sure because the CDN will have that infected copy. I meant a different vector, which is a malware tampering with the copy in `node_modules` _after_ a legitimate version has been downloaded. Is there an integrity check that we can protect from that scenario? 🤔
> > I suppose the other point is, with the CDN technique the Relayer SDK is downloaded at runtime and not part of the SDK bundle/package distributed, whereas with this technique, it becomes part of the SDK package's footprint right? (Is that the reason why someone might use it via CDN in the first place?)
> 
> It's mostly about whether you want to load the script at runtime or buildtime. Most bundlers don't have a convention to load assets so the CDN would be best. Alexandre B said that the CDN is vowed to be deprecated at some point, I don't know if we should rely on it.

It's really a nightmare to get the worker bundling right for all bundlers, we'll have a strong case for keeping the CDN

PR #128, opened 2026-03-19
chore: minor SDK fixes (session keys, cache timings)

enitrat:
### SDK Core
- **fix(token): delegate revocation to CredentialsManager** — `revokeSession` and `revokeByTrackedIdentity` in `ZamaSDK` now delegate to `CredentialsManager` instead of manually deleting from storage. This centralizes credential lifecycle logic and only deletes the session signature (not the encrypted FHE keys), so users don't need to regenerate expensive keys on re-auth.

### React SDK
- **fix(react-sdk): swap onSuccess/invalidateQueries ordering** — In `useRevokeDelegation` (and `useDelegateDecryption`), the user's `onSuccess` callback now fires *before* cache invalidation. This matches the standard TanStack Query pattern and lets callbacks read the pre-invalidation cache if needed.
- **feat(react-sdk): add useQueries wrapper with shared queryKeyHashFn** — Ensures all queries in a `useQueries` call use our custom hash function for consistent cache key matching, same as `useQuery` and `useSuspenseQuery` already do.

## Test plan
- [x] Unit tests for `useRevokeDelegation` callback ordering
- [x] Unit tests for `useQueries` wrapper
- [x] Unit tests for `ZamaSDK` revocation delegation
- [ ] CI passes (vitest, api-report, playwright)
- [ ] Solidity contracts need dedicated unit tests before deployment


ghermet (dismissed):

ghermet (commented):
## PR Review Summary

### Important Issues (2)

#### 1. `onSuccess` throwing prevents cache invalidation
**Files:** `use-revoke-delegation.ts:33`, `use-delegate-decryption.ts:33`

The new ordering fires the user's `onSuccess` before `invalidateQueries`. If the user's callback throws, cache invalidation is skipped, leaving stale delegation status in the UI.

**Suggested fix** — wrap in try-finally:
```typescript
  try {
  } finally {
    context.client.invalidateQueries({ queryKey: zamaQueryKeys.delegationStatus.all });
  }
},
```

#### 2. Pre-existing empty catch blocks amplified by this PR
**Files:** `zama-sdk.ts:148-150` (`#initIdentity`), `zama-sdk.ts:118-120, 130-132` (lifecycle handlers)

These empty catches swallow all errors, not just "signer not ready." Combined with the silent return in `#revokeByTrackedIdentity` when identity is null, there's a chain where revocation silently does nothing if identity tracking fails. The delegation refactor centralizes revocation through `CredentialsManager`, making correct identity tracking more important.

**Suggestion:** Add debug-level logging to these catch blocks (can be a follow-up).

### Suggestions (2)

- **`useQueries` wrapper** — `combine` option is accepted but not tested. Low risk since it's passed through via spread.
- **Test cast** `vi.spyOn(sdk.credentials, "clearCaches" as never)` — the `as never` bypasses type checking to spy on a protected method. Fragile if `clearCaches` is renamed.

### Strengths

- Clean delegation pattern — `revokeByKey` is a well-designed public facade over the protected `revokeSession`
- Behavioral improvement — adding `clearCaches()` during revocation fixes stale crypto cache not being cleared
- Good test coverage — ordering tests cleverly verify cache state during callback
- PR claim verified: "only deletes session signature, not FHE keys" — confirmed via `sessionSignatures.delete()`, not `storage.delete()`
- API reports correctly updated

### Verdict

Good to merge once issue #1 (try-finally around `onSuccess`) is considered.

ghermet (approved):

PR #133, opened 2026-03-20
build(sdk): make relayer-sdk a hard dependency

ghermet:
## Summary
- Moved `@zama-fhe/relayer-sdk` from optional peer dependency to a regular dependency in `packages/sdk/package.json`
- Removes the `optional: true` from `peerDependenciesMeta` so consumers get it installed automatically
- Updated README peer dependencies table to reflect the change

Closes SDK-45

## Test plan
- [x] `pnpm build:sdk` passes
- [x] `pnpm typecheck` passes
- [x] `pnpm test:run` — 127 test files, 1210 tests pass

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -26,9 +26,6 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1.7.0
 
-      - name: Install forge-fhevm deps
-        run: cd contracts/lib/forge-fhevm && forge soldeer install
-
       - name: Restore Forge build cache
         id: forge-build-cache
         uses: actions/cache/restore@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
--- a/.github/workflows/release-manual.yml
+++ b/.github/workflows/release-manual.yml
@@ -61,7 +61,7 @@ jobs:
           npm --version
 
       - name: Install dependencies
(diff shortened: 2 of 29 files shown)


ankurdotb (changes requested):
What do you think of adding tests, can that easily be done? @enitrat suggested on Linear:

> Add a regression test that packs the SDK, installs it into a clean temp project with pnpm, and verifies that @zama-fhe/sdk/node either:
> 
> starts successfully when the dependency is present, or
> 
> fails with an explicit, user-facing missing-dependency error when it is absent.

Maybe this can be combined with some other test, e.g., an e2e test?

ghermet:
> What do you think of adding tests, can that easily be done? @enitrat suggested on Linear:
> 
> > Add a regression test that packs the SDK, installs it into a clean temp project with pnpm, and verifies that @zama-fhe/sdk/node either:
> > starts successfully when the dependency is present, or
> > fails with an explicit, user-facing missing-dependency error when it is absent.
> 
> Maybe this can be combined with some other test, e.g., an e2e test?

First we need to fix the bug, then we can write the test-node package with node integration tests

enitrat (dismissed):

enitrat:
I'm not sure how you'd test that either, might be superfluous given that it's not the kind of thing you'd change that often

ghermet:
> I'm not sure how you'd test that either, might be superfluous given that it's not the kind of thing you'd change that often

The same way we test with vite and nextjs, we plug its own anvil node and run tests via the Node SDK

ankurdotb (dismissed):

ankurdotb (approved):

PR #134, opened 2026-03-20
ci: ignore scripts on pnpm install and exclude submodule from linters

ankurdotb:
# Summary
Add --ignore-scripts to CI workflow installs to prevent postinstall hooks from dirtying the forge-fhevm submodule. Exclude contracts/lib from oxlint and oxfmt via ignorePatterns and .prettierignore.

## Scope

- [ ] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [x] CI/workflows
- [ ] Docs

## Validation
Checked and ran locally

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -61,7 +61,7 @@ jobs:
           cache: pnpm
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile
+        run: pnpm install --frozen-lockfile --ignore-scripts
 
       - name: Restore Playwright browsers cache
         id: playwright-cache
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -26,9 +26,6 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@8789b3e21e6c11b2697f5eb56eddae542f746c10 # v1.7.0
 
-      - name: Install forge-fhevm deps
-        run: cd contracts/lib/forge-fhevm && forge soldeer install
-
       - name: Restore Forge build cache
         id: forge-build-cache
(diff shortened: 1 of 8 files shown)


.github/workflows/playwright.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           cache: pnpm
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile
```
enitrat, line 64: I think you want to keep it here, no? You need the contracts to deploy on the node

.github/workflows/playwright.yml
enitrat: What you could remove is this explicit step that is covered by the postinstall (and next one ig)

ankurdotb: Fair point, resolved

ghermet:
@enitrat @ankurdotb We should not install forge or soldeer in postinstall script, this should be done only in the CI and manually in the context of a local e2e test.

ankurdotb:
> @enitrat @ankurdotb We should not install forge or soldeer in postinstall script, this should be done only in the CI and manually in the context of a local e2e test.

That change doesn't come from this PR, it comes from #109. What change are you suggesting?

ghermet:
> > @enitrat @ankurdotb We should not install forge or soldeer in postinstall script, this should be done only in the CI and manually in the context of a local e2e test.
> 
> That change doesn't come from this PR, it comes from #109. What change are you suggesting?

It's a reminder that this is a fix but not a long term solution. There's no change in this PR per say but we need to improve the system

enitrat:
Last time I asked whether to setup with a makefile or a postinstall script I was told to use postinstall, I'm fine either way, there just needs to be a one-liner setup for the required deps when working on the repo

ghermet (approved):

PR #135, opened 2026-03-20
feat(sdk): align delegation API with Delegated User Decryption spec

ghermet:
## Summary
- Add `isHandleDelegatedForUserDecryption` ABI entry and `isHandleDelegatedContract()` builder to check per-handle delegation status
- Add ACL delegation event decoders (`DelegatedForUserDecryption`, `RevokedDelegationForUserDecryption`) with typed interfaces, `AclTopics` constants, decode/find helpers, and `ACL_TOPICS` array
- Enforce minimum 1-hour expiration client-side in `delegateDecryption()` to match on-chain `ExpirationDateBeforeOneHour` revert
- Re-export delegation contract builders (`delegateForUserDecryptionContract`, `revokeDelegationContract`, `getDelegationExpiryContract`, `isHandleDelegatedContract`) from `@zama-fhe/react-sdk`
- Validate all delegation parameters client-side before submitting transactions (SDK-47):
  - Pre-flight: `SenderCannotBeDelegate`, `DelegateCannotBeContractAddress`, `ExpirationDateAlreadySetToSameValue` (RPC read), `NotDelegatedYet` (RPC read before revoke)
  - Catch-and-rethrow: `AlreadyDelegatedOrRevokedInSameBlock` → `DelegationCooldownError`, `SenderCannotBeContractAddress` → `DelegationContractIsSelfError`, `EnforcedPause` → `AclPausedError`
- 4 new error classes + `matchAclRevert()` utility for Solidity revert mapping
- Documentation updates for delegation, event decoders, contract builders, and error types

Closes SDK-47

## Test plan
- [x] `isHandleDelegatedContract` builder test (address, functionName, args)
- [x] `AclTopics` keccak256 hash verification (2 tests)
- [x] `decodeDelegatedForUserDecryption` — valid log, wrong topic, insufficient topics
- [x] `decodeRevokedDelegationForUserDecryption` — valid log, wrong topic, insufficient topics
- [x] `decodeAclEvent` — dispatches to both decoders, returns null for unknown
- [x] `decodeAclEvents` — batch mixed logs
- [x] `findDelegatedForUserDecryption` / `findRevokedDelegationForUserDecryption` — found + not found
- [x] 1-hour minimum expiration rejection (30 min) and acceptance (1h + 1s)
- [x] `DelegationSelfNotAllowedError` when delegate is the signer
- [x] `DelegationDelegateEqualsContractError` when delegate is the token address
- [x] `DelegationExpiryUnchangedError` when new expiry matches current
- [x] `DelegationNotFoundError` when revoking non-existent delegation
- [x] `DelegationNotFoundError` when revoking expired delegation
- [x] `DelegationCooldownError` catch-and-rethrow for `AlreadyDelegatedOrRevokedInSameBlock`
- [x] `AclPausedError` catch-and-rethrow for `EnforcedPause`
- [x] All 951 tests pass, build succeeds for both sdk and react-sdk


ankurdotb (changes requested):
Targets wrong branch

ankurdotb:
### Code review

Found 4 issues (scored 75/100 confidence each -- individually below auto-post threshold, but collectively significant):

---

1. **`revokeDelegation` uses `Date.now()` instead of `getBlockTimestamp()`** -- The codebase explicitly moved away from client clock after PR #72 review feedback. `#assertDelegationActive` in `readonly-token.ts` correctly uses `this.signer.getBlockTimestamp()`. The new pre-flight check reintroduces the client-clock anti-pattern, which fails under clock skew.

https://github.com/zama-ai/sdk/blob/b43c27cebf38a24d4a929a2735a0059e64a9eb45/packages/sdk/src/token/token.ts#L833-L839

2. **Expired delegations throw `DelegationNotFoundError` instead of `DelegationExpiredError`** -- `DelegationExpiredError` already exists in the codebase (errors.ts:198-203) and `#assertDelegationActive` in `readonly-token.ts` correctly distinguishes `0n` (not found) from past-timestamp (expired). The new code collapses both into `DelegationNotFoundError`, breaking callers that catch `DelegationExpiredError`.

https://github.com/zama-ai/sdk/blob/b43c27cebf38a24d4a929a2735a0059e64a9eb45/packages/sdk/src/token/token.ts#L827-L839

3. **`revokeDelegation` pre-flight blocks revoking expired delegations** -- The ACL contract only reverts with `NotDelegatedYet` when expiry is exactly `0` (never delegated). An expired delegation has a non-zero expiry and the contract accepts revocation. The SDK pre-flight is more restrictive than the contract, preventing users from cleaning up expired delegation state.

https://github.com/zama-ai/sdk/blob/b43c27cebf38a24d4a929a2735a0059e64a9eb45/packages/sdk/src/token/token.ts#L827-L839

4. **`matchAclRevert` uses string-includes matching instead of viem's structured error data** -- viem's `ContractFunctionRevertedError` exposes `error.data.errorName` with the decoded ABI error selector. Using `message.includes("AlreadyDelegatedOrRevokedInSameBlock")` is fragile across RPC providers and viem versions. Additionally, the code comment at errors.ts:168-171 says "the SDK does NOT auto-map ACL contract reverts to these" but `matchAclRevert` does exactly that (comment predates the design change and is now stale).

https://github.com/zama-ai/sdk/blob/b43c27cebf38a24d4a929a2735a0059e64a9eb45/packages/sdk/src/token/errors.ts#L287-L338

---

### Linear ticket coverage

**SDK-47** (validate delegation params client-side): 8/10 acceptance criteria met. Strong delivery -- all 8 Solidity error cases covered with tests.

**SDK-48** (docs & API gaps vs spec): 1/6 fully met (ACL event decoders documented). Three criteria not addressed:
- Propagation delay -- no mention of the ~10-block coprocessor delay; integrators will hit `UserDecryptionNotDelegated` errors immediately after `delegateDecryption()`
- `isHandleDelegated` not exposed at Token API level -- only a low-level contract builder, not a method on `Token`/`ReadonlyToken` as specified
- Smart wallet use case guide -- entirely missing (primary use case for custodial integrations)

PR body says "Closes SDK-47" which is accurate. Consider either removing SDK-48 from scope or splitting remaining SDK-48 work into a follow-up.

zama-cremaud:
## Summary

The PR is well-scoped and well-structured. The architecture is sound: client-side pre-flight (saves gas, typed errors) + on-chain revert mapping (two-tier: structured viem cause, then string fallback) + event decoders + documentation is the right approach. No High findings. Three Medium findings, all fixable, mostly in test coverage and documentation accuracy.

---

## Medium

### M1 — Dead first expiration check produces a worse error message for past dates

**File:** `packages/sdk/src/token/token.ts` — `delegateDecryption()`

Two consecutive guards check the expiration date:

```ts
// Check 1
if (expirationDate && expirationDate.getTime() <= Date.now()) {
  throw new ConfigurationError("Expiration date must be in the future");
}
// Check 2
if (expirationDate && expirationDate.getTime() < Date.now() + 3600_000) {
  throw new ConfigurationError("Expiration date must be at least 1 hour in the future");
}
```

Check 1 fires for any date in the past or exactly now; check 2 fires for any date within 1 hour. Since any past date also satisfies `< Date.now() + 3600_000`, check 1 is never reached for the 1-hour-window case, and check 2 can never fire for past dates because check 1 fires first. The practical consequence: a caller passing a past timestamp receives `"Expiration date must be in the future"` instead of the more actionable `"Expiration date must be at least 1 hour in the future"` — inconsistent with the documentation hint, which only mentions the 1-hour rule.

Check 1 is also not tested, and the test for check 2 (`30 minutes in the future`) cannot implicitly cover check 1 because the test date is in the future. The inconsistency is therefore invisible in CI.

**Fix:** Remove check 1. Check 2 is strictly more inclusive and covers all cases with the correct message:

```ts
if (expirationDate && expirationDate.getTime() < Date.now() + 3600_000) {
  throw new ConfigurationError("Expiration date must be at least 1 hour in the future");
}
```

---

### M2 — Documentation incorrectly describes `DelegationNotFoundError` on revoke

**File:** `docs/gitbook/src/reference/sdk/delegation.md` — Client-side pre-flight errors table

The table entry for `DelegationNotFoundError` says: *"Attempting to revoke a non-existent **or expired** delegation."* This is incorrect. The pre-flight check in `token.ts` is:

```ts
const currentExpiry = await this.#signer.readContract(getDelegationExpiryContract(...));
if (currentExpiry === 0n) {
  throw new DelegationNotFoundError(...);
}
```

`currentExpiry === 0n` means the delegation was never established (ACL contract default). An expired-but-set delegation has a non-zero expiry value and passes through — the SDK explicitly permits revoking it (confirmed by test: `"allows revoking an expired delegation (ACL accepts it)"`). The documentation therefore misleads callers into thinking they cannot revoke an expired delegation, which would cause them to add incorrect defensive checks.

**Fix:** Update the table entry to: *"Attempting to revoke a delegation that was never established (has never been delegated)."* Move `DelegationExpiredError` to a separate note clarifying it only applies to the read path (`#assertDelegationActive`, i.e., when querying status), not to revoke.

---

### M3 — Existing success tests do not mock `readContract` for the new pre-flight, relying on implicit `undefined !== 0n`

**File:** `packages/sdk/src/token/__tests__/delegation.test.ts`

Three pre-existing tests now implicitly exercise the new `readContract` pre-flight without mocking it:

- `"delegateDecryption returns TransactionResult"` (line ~143)
- `"revokeDelegation returns TransactionResult"` (line ~233)
- `"revokeDelegation calls ACL correctly"` (line ~147)

In these tests, `signer.readContract` is a bare `vi.fn()` with no configured return value, so it returns `undefined`. The pre-flight checks `currentExpiry === 0n` and `currentExpiry === expirationTimestamp` — both `undefined === 0n` and `undefined === bigint` are `false` in JavaScript due to strict equality semantics (no implicit BigInt coercion). The tests pass, but for the wrong reason: they are not expressing "the delegation exists with a valid expiry", they are accidentally relying on the `undefined !== bigint` edge case.

This creates a maintenance trap: if the pre-flight condition is ever changed to `!currentExpiry` or `currentExpiry == null`, the behaviour changes silently, and these tests would start failing or passing for newly wrong reasons.

**Fix:** Add explicit `readContract` mocks to these three tests:

```ts
// For delegateDecryption: simulate "no existing delegation"
vi.spyOn(signer, "readContract").mockResolvedValue(0n); // getDelegationExpiry returns 0

// For revokeDelegation: simulate "active delegation exists"
vi.spyOn(signer, "readContract").mockResolvedValue(MAX_UINT64);
```

---

## Low

### L1 — ACL revert mappings for `ExpirationDateBeforeOneHour` and `SenderCannotBeContractAddress` have no unit tests

`ACL_ERROR_MAP` defines 7 Solidity error → typed error mappings. The delegation tests cover `AlreadyDelegatedOrRevokedInSameBlock` → `DelegationCooldownError` and `EnforcedPause` → `AclPausedError`. Five mappings — including `ExpirationDateBeforeOneHour` → `ConfigurationError` and `SenderCannotBeContractAddress` → `DelegationContractIsSelfError` — have no test exercising the `matchAclRevert` path. The client-side pre-flight catches most of these before submission, but the on-chain mapping is a separate code path (handles clock skew, concurrent state) and should have at least smoke-test coverage.

### L2 — `ZamaErrorCode` type alias lost its TSDoc, now marked `@public (undocumented)` in the API report

In `packages/sdk/etc/sdk.api.md`, the `ZamaErrorCode` type alias is now listed as `// @public (undocumented)`. The JSDoc comment `/** Union of all... */` was removed from the source in this PR. The `ZamaErrorCode` const object retains its doc. Minor API Extractor visibility regression.

### L3 — No test for `DelegationExpiryUnchangedError` when a permanent (MAX_UINT64) delegation is re-submitted

The `DelegationExpiryUnchangedError` test only covers the finite-timestamp case. The edge case where a caller re-submits `delegateDecryption()` with no `expirationDate` (resulting in `MAX_UINT64`) against an existing `MAX_UINT64` delegation is untested. The guard handles it correctly at runtime, but a test would prevent future regressions.

### L4 — `AclTopics` named object is internal-only; `ACL_TOPICS` array is the only public export

`AclTopics` (the keyed object: `AclTopics.DelegatedForUserDecryption`) is not exported from `packages/sdk/src/index.ts`. Only `ACL_TOPICS` (the flat `[hash1, hash2]` tuple) is public. Consumers building custom `eth_getLogs` filters must either use `ACL_TOPICS[0]` by index or hardcode the hash string. Exporting `AclTopics` would make filters self-documenting. Not a functional issue, but a usability gap worth considering before the API surface is locked.

---

## What Is Done Well

- **Two-tier `matchAclRevert` strategy** is the right design: prefer viem's structured `error.cause.data.errorName` (available when ABI decoding succeeds) and fall back to string matching for wallets that surface revert reasons as raw strings. This handles both on-chain simulation and live RPC paths.
- **`revokeDelegation` pre-flight intentionally allows expired delegations** (`currentExpiry !== 0n` passes regardless of whether it's in the past). This correctly mirrors ACL contract semantics where cleanup revocations on expired grants are accepted.
- **ACL event decoder tests** are thorough: topic hash verification against computed keccak256, valid decode, wrong-topic → null, insufficient topics → null, batch decode with unknowns, and convenience finders. This level of test coverage for ABI-level decoding is exactly right.
- **Documentation update is accurate** for the new behaviour. Splitting the flat error table into client-side / on-chain / other clearly communicates which checks happen where.
- **`matchAclRevert` is not exported from the public API** — it's an internal implementation detail. Correct visibility decision.

ankurdotb:
### Code review

Found 9 issues:

1. **[score: 100]** `delegation.md` error table incorrectly states `DelegationNotFoundError` covers "non-existent **or expired**" delegations. The code only throws it when `currentExpiry === 0n` (never delegated). Expired delegations pass through to the ACL contract and are accepted — confirmed by `token.ts:830` and the test "allows revoking an expired delegation". Suggested fix: "Attempting to revoke a delegation that was never established (expiry is zero)".

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/docs/gitbook/src/reference/sdk/delegation.md#L223-L225

The guard that throws this error (`token.ts:830-833`) only checks `currentExpiry === 0n`:

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/token/token.ts#L829-L834

2. **[score: 85]** Broken import in `event-decoders.md`. The docs tell users to `import { AclTopics }` from `@zama-fhe/sdk`, but `AclTopics` is not exported from the public barrel (`index.ts`). Only `ACL_TOPICS` (the flat array) is exported. This code example will not compile as written.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/docs/gitbook/src/reference/sdk/event-decoders.md#L154-L158

The public barrel only exports `ACL_TOPICS`, not `AclTopics`:

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/index.ts#L169-L173

3. **[score: 75]** Dead first expiration guard in `delegateDecryption`. The check `expirationDate.getTime() <= Date.now()` (line 739) is entirely subsumed by the second check `< Date.now() + 3600_000` (line 742). Any past date triggers the first check, but since the second check is strictly broader, users who pass a past date get the misleading "at least 1 hour in the future" message instead of "must be in the future". The first guard is unreachable dead code.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/token/token.ts#L738-L744

4. **[score: 75]** `matchAclRevert` is marked `@internal` in its JSDoc but is explicitly re-exported from the public API barrel in `index.ts:153`. This will appear in the API report and contradicts the `@internal` intent. Either remove the `@internal` tag or remove the re-export.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/errors/acl-revert.ts#L72-L76

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/index.ts#L152-L154

5. **[score: 75]** Stale comment in `delegation.ts:3-5` says "the SDK does NOT auto-map ACL contract reverts to these." This PR adds `matchAclRevert()` in `acl-revert.ts` which does exactly that. The docs were correctly updated, but the source code comment now directly contradicts the behavior.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/errors/delegation.ts#L3-L5

6. **[score: 75]** `ConfigurationError` is used for the 1-hour expiration violation (both in the pre-flight at `token.ts:740` and in `matchAclRevert` at `acl-revert.ts:59`), but the PR defines specific error classes for every other delegation failure. Consumers catching `ConfigurationError` cannot distinguish expiration-too-soon from unrelated config problems (missing ACL, wrong chain, etc.). Consider a dedicated `DelegationExpirationTooSoonError`.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/errors/acl-revert.ts#L58-L59

7. **[score: 75]** Three success-path tests (`delegateDecryption returns TransactionResult` at line 143, `revokeDelegation calls ACL correctly` at line 148, `revokeDelegation returns TransactionResult` at line 233) don't mock `readContract`. The bare `vi.fn()` returns `undefined`, and guards like `currentExpiry === 0n` evaluate `undefined === 0n` -> `false`, so tests pass for the wrong reason. If guard conditions change to a loose check, tests would silently mask regressions.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/token/__tests__/delegation.test.ts#L143-L144

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/token/__tests__/delegation.test.ts#L232-L234

8. **[score: 50]** `DelegationNotFoundError` is used for two semantically different conditions: the client-side pre-flight (`currentExpiry === 0n`, never delegated) and the `matchAclRevert` mapping of the Solidity `NotDelegatedYet` revert. Currently correct since the contract only fires `NotDelegatedYet` for zero-expiry, but if the contract adds expiration validation, expired delegations would map to `DelegationNotFoundError` instead of `DelegationExpiredError`.

https://github.com/zama-ai/sdk/blob/919923a17814fef35eb9ea0b8a89548b9dfe3987/packages/sdk/src/errors/acl-revert.ts#L64-L65

9. **[score: n/a — feature gap]** `SDK-48` asks for a Token-level `isHandleDelegated(...)` API and a React `useIsHandleDelegated()` hook. The PR adds the low-level builder in `acl.ts` but does not surface it on `Token`/`ReadonlyToken` or add the React hook. SDK-48 also requests documentation for propagation delay, a constraints table, batch delegation gas trade-offs, and a smart wallet guide — none of which are present in this PR.

ankurdotb (changes requested):
This is already in the Claude review, but the main issue that stood out to me is that `DelegationNotFoundError` is used for both never created delegations as well as expired/not active delegations.

Those are two significantly different scenarios.

ghermet:
## Code review fixes — addressed in commits `10b5637c`, `dbadc9a0`, `fa7f5a7a`

| # | Score | Issue | Resolution |
|---|-------|-------|------------|
| 1 | 100 | `DelegationNotFoundError` doc says "non-existent **or expired**" but code only checks `expiry === 0n` | Fixed — doc now reads "delegation that was never established (expiry is zero)" |
| 2 | 85 | `AclTopics` not exported from public barrel, breaking docs import example | Fixed — added `AclTopics` to named exports in `index.ts` |
| 3 | 75 | Dead first expiration guard (`<= Date.now()`) subsumed by second (`< Date.now() + 3600_000`) | Fixed — removed the unreachable first guard |
| 4 | 75 | `matchAclRevert` marked `@internal` but explicitly re-exported | Fixed — changed to `@public` |
| 5 | 75 | Stale comment in `delegation.ts` says SDK does NOT auto-map ACL reverts (it now does via `matchAclRevert`) | Fixed — updated comment |
| 6 | 75 | `ConfigurationError` used for expiration violations — consumers can't distinguish from unrelated config errors | Fixed — created `DelegationExpirationTooSoonError` with dedicated `DELEGATION_EXPIRATION_TOO_SOON` error code, used in both pre-flight and `matchAclRevert` |
| 7 | 75 | 3 tests don't mock `readContract` — pass only because `undefined === 0n` is `false` | Fixed — added explicit `readContract` mocks with realistic values |
| 8 | 50 | `DelegationNotFoundError` reused for client pre-flight and `NotDelegatedYet` revert | **No change** — semantically correct today; would only diverge if the contract adds expiration validation to `NotDelegatedYet` |
| 9 | n/a | SDK-48 feature gaps (Token-level `isHandleDelegated`, React hook, docs) | **Out of scope** — tracked in SDK-48 for follow-up |

ankurdotb:
### Follow-up: additional issues from spec review

**1. Stale doc comment contradicts new behavior**

`errors.md` line 331 says:

> "The SDK does **not** auto-map ACL contract reverts to delegation errors."

But this PR adds `matchAclRevert()` which does exactly that, and `delegation.md` line 278 correctly states the opposite. These contradict each other — the `errors.md` hint is now stale.

https://github.com/zama-ai/sdk/blob/b43c27cebf38a24d4a929a2735a0059e64a9eb45/docs/gitbook/src/reference/sdk/errors.md#L330-L332

**2. No propagation delay documentation**

SDK-48 acceptance criterion #1 requires documenting the ~10-block coprocessor propagation delay (`delegation_block_delay`) before a delegation becomes usable for `decryptBalanceAs`. This is not mentioned anywhere in the delegation docs, error docs, or inline code comments. Integrators will call `delegateDecryption()`, get a confirmed receipt, immediately call `decryptBalanceAs()`, and hit a `UserDecryptionNotDelegated` error with no explanation.

**3. Expired vs revoked are indistinguishable but treated inconsistently**

The SDK has both `DelegationNotFoundError` and `DelegationExpiredError`, but:

- **Revoked** delegations return `0n` on-chain (contract zeros the expiry), so they're indistinguishable from **never-set** — both throw `DelegationNotFoundError`
- **Expired** delegations have a non-zero past timestamp and throw `DelegationExpiredError`, but only in the `#assertDelegationActive` decrypt path (`readonly-token.ts:720`), not in `revokeDelegation`

This is arguably correct (you can't distinguish revoked from never-set via state reads alone — only via `RevokedDelegationForUserDecryption` events). But it's worth documenting explicitly so integrators understand the state model. Consider a short "Delegation states" section in `delegation.md` explaining: never-set (`0n`), active (future timestamp), expired (past timestamp), revoked (reset to `0n`, distinguishable only via events). **Alternatively**, explain that the `DelegationNotFounderError` could happen for EITHER expired or revoked (don't assume without examining events).

ghermet:
All 3 follow-up issues addressed in a2259ccc:

1. **Stale `errors.md` hint** — Updated to correctly state the SDK auto-maps ACL reverts via `matchAclRevert()`, with a link to the delegation error reference.
2. **Propagation delay** — Added a warning callout in `delegation.md` documenting the ~10-block `delegation_block_delay` before delegated decryption works.
3. **Delegation states** — Added a "Delegation states" section with a table covering never-set / active / expired / revoked, explaining that revoked resets expiry to `0n` (indistinguishable from never-set via state reads) and pointing to `RevokedDelegationForUserDecryption` events to differentiate.

ankurdotb (dismissed):

ankurdotb (approved):

PR #139, opened 2026-03-20
feat(playwright): add Node.js SDK e2e tests with mock relayer

ghermet:
## Summary
- Add comprehensive Node.js e2e test suite using Playwright with a mock relayer HTTP server
- Cover SDK lifecycle, sessions, shielding/unshielding, transfers, delegation, approvals, public decryption, and token metadata
- Replace mock worker with mock relayer server for more realistic test isolation
- Colocate node worker with node entrypoint and make `@zama-fhe/relayer-sdk` a hard dependency
- Remove standalone `packages/test-node` in favor of integrated Playwright node tests

## Test plan
- [ ] `pnpm playwright test --config playwright.node.config.ts` passes all node specs
- [ ] Mock relayer server starts and proxies key URL from testnet
- [ ] All FHE operations (encrypt, decrypt, reencrypt) work against local anvil via cleartext relayer


enitrat (changes requested):
I see a few issues here, I think this is not going in the right direction.

The node e2e tests duplicate nearly the entire browser Playwright suite that already exists (in our e2e tests) through the RelayerNode, but this is actually more of a test of the `Token` features (and e2e features of the library), that you can run against wasm/node/cleartext relayer anyway. So here you're duplicating an entire test-suite instead of just testing the minimal surface required.

Re-running all domain-level scenarios through Node.js doesn't validate anything new. The big mock worker file is a strong indicator that we're going in the wrong direction, what we should be doing here is a focused set of transport-layer tests, ensuring that either wasm or node relayer can actually instanciate fine.

If you want to test the RelayerNode properly, you only need:
- unit tests for RelayerNode internals like the pool lifecycle or chain-switch behavior, etc, with a simple mocked pool. Notably you did this well with `server-bootstrap, chain-switching, worker-pool-concurrency` that can stay, be the other are not useful
- one integration test proving the real WASM loads properly in Node.js and can generateKeypair + encrypt.

All the domain logic that is already covered by the browser suite should go, and we can surely drop this from a 2.5k LoC PR to a 400 one.

ghermet:
> I see a few issues here, I think this is not going in the right direction.
> 
> The node e2e tests duplicate nearly the entire browser Playwright suite that already exists (in our e2e tests) through the RelayerNode, but this is actually more of a test of the `Token` features (and e2e features of the library), that you can run against wasm/node/cleartext relayer anyway. So here you're duplicating an entire test-suite instead of just testing the minimal surface required.
> 
> Re-running all domain-level scenarios through Node.js doesn't validate anything new. The big mock worker file is a strong indicator that we're going in the wrong direction, what we should be doing here is a focused set of transport-layer tests, ensuring that either wasm or node relayer can actually instanciate fine.
> 
> If you want to test the RelayerNode properly, you only need:
> 
> * unit tests for RelayerNode internals like the pool lifecycle or chain-switch behavior, etc, with a simple mocked pool. Notably you did this well with `server-bootstrap, chain-switching, worker-pool-concurrency` that can stay, be the other are not useful
> * one integration test proving the real WASM loads properly in Node.js and can generateKeypair + encrypt.
> 
> All the domain logic that is already covered by the browser suite should go, and we can surely drop this from a 2.5k LoC PR to a 400 one.


Fair enough, I conflated domain and environment here, I removed the duplicated domain tests and kept the node transport specific ones.

ghermet:
## Why the mock worker swap (global setup) is needed

The node e2e tests need to swap the built SDK worker before running. Here's why:

**Built worker** (`node_modules/@zama-fhe/sdk/dist/.../relayer-sdk.node-worker.js`):
- Uses `@zama-fhe/relayer-sdk/node` → `createInstance(config)`
- Makes real HTTP calls to `relayerUrl` for FHE operations (encrypt, decrypt, key generation)
- Requires a running relayer service

**Source mock worker** (`packages/sdk/src/node/relayer-sdk.node-worker.js`):
- Uses `@fhevm/mock-utils` → `MockFhevmInstance`
- All FHE operations happen in-process against local anvil
- No external relayer needed

The `node-global-setup.ts` swaps the built worker with the source mock before tests run, and restores it after. This is the same pattern the browser tests use (`fixtures/relayer-sdk.js` mocks the Web Worker). Without this swap, all tests fail with `HTTP error! status: 404` because there's no relayer running.

We considered removing the swap entirely (since the source file already exists in the repo), but the built/published worker is fundamentally different from the source mock — one uses the real SDK transport, the other uses `MockFhevmInstance`.

ghermet:
[AUTOMATED]

## Code Review — PR #139

Reviewed for bugs, guideline compliance, error handling, and silent failures using 4 parallel specialized agents + validation.

### Result: No blocking issues

Two minor observations:

**Nit: Mock worker error handler inconsistency** (`packages/sdk/src/node/relayer-sdk.node-worker.js:286`)
The outer `.catch` uses `err.message` without guarding for non-Error rejections, while the inner `catch` (line 281) correctly does `error instanceof Error ? error.message : String(error)`. If `sendError` itself throws (e.g., port closed), the outer handler could produce `undefined` or throw `TypeError`. Consider aligning:
```js
port.on("message", (request) => {
  handleMessage(request).catch((err) =>
    sendError(request.id, request.type, err instanceof Error ? err.message : String(err))
  );
});
```

**FYI: Batch op cause narrowing** (`packages/sdk/src/token/token.ts`)
The migration from `Promise.allSettled` to sequential `try/catch` introduces `cause: reason instanceof Error ? reason : undefined`, which drops non-Error rejection reasons. The old `allSettled` path passed `outcome.reason` directly. In practice this has negligible impact since the upstream `op` callbacks always reject with `Error` instances, and the pattern is consistent with the rest of the file.

### What was checked
- Bug hunting (2x opus agents): No runtime bugs found
- Guideline compliance: All conventions followed (the `.js` mock worker is intentionally plain JS for the file-swap test strategy)
- Error handling audit: No silent failures in production code
- Double-terminate safety: `using` in fixtures correctly relies on idempotent `terminate()`

enitrat (changes requested):
1. we should remove everything related to playwright here, I don't see a reason to use playwright for those nodejs tests: we're not testing anything browser. any setup we do here should be doable with a native vitest
2. I don't understand why we have a full fake mock of the entire node worker lifecycle in `packages/playwright/fixtures/relayer-sdk.node-worker.mjs`. It seems we already did things right in the `worker.node-pool.test.ts` file, why not reuse a similar system?
3. it seems that actually all scenarios we are adding here are already tested as part of either `relayer.test.ts` or `worker.*.test.ts`.

All the files mentioned above are doing unit tests with mocked workers or pools. If you want to add a "real conditions test", you only need to have a single integration test where you instanciate a `RelayerNode` and verify that you can generate a keypair, which proves you properly loaded the WASM in worker threads and called the FHE operations. So we likely just need to create a `relayer-node.integration.test.ts` with 30lines of test code for that behavior only.

ghermet:
> 2\. worker.node-pool.test.ts

The integration test needs a running anvil node regardless, it's not just 30 lines of code. You can swap playwright for vitests but all we gonna end up doing is duplicating the test infra in vitest.

enitrat:
> The integration test needs a running anvil node regardless

for what?

ghermet:
> > The integration test needs a running anvil node regardless
> 
> for what?

`MockFhevmInstance` reads real contract state (ACL, InputVerifier, KMS properties) from the chain during init, and several tests verify on-chain behavior (token metadata, snapshot/revert). The mock only replaces the WASM/FHE compute, not the blockchain interaction.

ghermet:
We can't really mock the server itself so mocking the worker with mock utils is the closest we can get: This is a "mock the boundary, not the logic" pattern, which is the right tradeoff for e2e tests. The real FHE operations would require a full coprocessor/KMS stack, so mocking at the worker thread boundary keeps tests fast and deterministic while still validating the SDK's message protocol and encryption flow.

zama-cremaud:
# Code Review — `feature/node-sdk-e2e`

**Branch:** `origin/feature/node-sdk-e2e` → `origin/prerelease`
**Scope:** Node.js e2e test infrastructure, SDK changes (`RelayerNode`, `RelayerSDK`, `ZamaSDK`, `token.ts`), Playwright node configuration, CI workflow.

---

## Critical

### 1. CI `test-node` job is missing `pnpm build`

**File:** `.github/workflows/playwright.yml`, `test-node` job (~lines 258–261)

The `test-node` job goes directly from `pnpm install --frozen-lockfile` to `pnpm e2e:test:node` with no build step. Both `test-nextjs` and `test-vite` include an explicit `pnpm build` before running their tests. `node-global-setup.ts` calls `require.resolve("@zama-fhe/sdk/node")` to locate the worker file on disk. On a fresh CI runner where `dist/` does not yet exist, this resolution fails immediately and the entire job crashes at global setup — before any test has a chance to run.

```yaml
# Current (broken on fresh checkout)
- name: Install dependencies
  run: pnpm install --frozen-lockfile
- name: Run Node e2e tests
  run: pnpm e2e:test:node

# Fix
- name: Install dependencies
  run: pnpm install --frozen-lockfile
- name: Build SDK packages
  run: pnpm build
- name: Run Node e2e tests
  run: pnpm e2e:test:node
```

---

## High

### 2. Worker file backup strategy can permanently destroy the original on interrupted runs

**File:** `packages/playwright/fixtures/node-global-setup.ts`, lines 21–24

```typescript
if (existsSync(workerPath)) {
  renameSync(workerPath, backupPath);  // no check whether backup already holds the real original
}
copyFileSync(mockPath, workerPath);
```

If a previous run was interrupted after setup but before teardown, the on-disk state is:

- `workerPath` = mock (written by the interrupted run's `copyFileSync`)
- `backupPath` = the real original SDK worker (saved by the interrupted run's `renameSync`)

A fresh run sees `existsSync(workerPath) === true` and executes `renameSync(workerPath, backupPath)` — **overwriting the backup (which held the real original) with the mock**. Then `copyFileSync(mockPath, workerPath)` copies the mock again. Both paths now hold the mock; the original SDK worker is irreversibly gone from disk.

Recovery requires a full package rebuild or reinstall. This is a real risk in any CI environment where a job can be cancelled mid-run.

**Suggested fix** — guard against overwriting the backup by checking whether the file to be renamed is already the mock:

```typescript
import { readFileSync } from "node:fs";

// A string that exists only in the mock worker, not in the real SDK worker
const MOCK_SENTINEL = "MockFhevmInstance";

function isMockWorker(path: string): boolean {
  try {
    return readFileSync(path, "utf8").includes(MOCK_SENTINEL);
  } catch {
    return false;
  }
}

export default async function globalSetup() {
  if (existsSync(workerPath) && !isMockWorker(workerPath)) {
    renameSync(workerPath, backupPath);
  }
  copyFileSync(mockPath, workerPath);
  // ...
}
```

---

### 3. `RelayerSDK extends Disposable` is an undocumented breaking change

**File:** `packages/sdk/src/relayer/relayer-sdk.ts`

```typescript
- export interface RelayerSDK {
+ export interface RelayerSDK extends Disposable {
```

Any external code that implements the `RelayerSDK` interface — custom transports, test mocks, library wrappers — will now fail to compile because `[Symbol.dispose]()` is a required method. This is a breaking change on a publicly exported interface.

The method is not used by `ZamaSDK` internally: `terminate()` remains the only lifecycle method called from `ZamaSDK.terminate()`. The benefit — enabling `using relayer = new RelayerNode(...)` in user code — is real but narrow, and the cost to downstream implementors is broad.

**Options:**
- Explicitly document this as a breaking change in the PR description and CHANGELOG.
- Or keep `Disposable` as an implementation detail of the concrete classes (`RelayerNode`, `RelayerWeb`) and not extend it on the interface, leaving `terminate()` as the sole lifecycle obligation.

---

### 4. Chain-switching test assertion cannot detect a broken pool re-initialization

**File:** `packages/playwright/tests/node/chain-switching.node.spec.ts`, lines 23–30

```typescript
const kp1 = await relayer.generateKeypair();
// Switch chain
chainId = 99999;
const kp2 = await relayer.generateKeypair();
expect(kp1.publicKey).not.toBe(kp2.publicKey);
```

`MockFhevmInstance.generateKeypair()` returns a freshly-generated random keypair on every call. Two sequential calls on the **same unchanged pool** would also produce different keys. The assertion `kp1 !== kp2` is satisfied regardless of whether the chain switch actually tore down the pool and initialized a new one.

If the chain-detection path in `#ensurePoolInner` were silently broken — for instance, if the `#resolvedChainId` comparison had a bug — this test would still pass.

**Suggested fix:** Assert on a chain-specific observable instead of key uniqueness. The same test file already does this correctly for config reflection by checking `getAclAddress()` against the expected chain's ACL address. Apply the same pattern here: verify that `getAclAddress()` returns the new chain's ACL address after the switch, confirming the pool was reconstructed with the correct configuration.

---

## Medium

### 5. `poolSize: 0` test only checks construction; broken operational state goes untested

**File:** `packages/playwright/tests/node/error-handling.node.spec.ts`, lines 48–57

```typescript
test("zero poolSize defaults gracefully at construction", async ({ transport }) => {
  using relayer = new RelayerNode({ ..., poolSize: 0 });
  expect(relayer).toBeDefined();
});
```

A `NodeWorkerPool` with `poolSize: 0` constructs without error (the init `for` loop runs zero iterations), but any subsequent dispatch call throws `"NodeWorkerPool not initialized. Call initPool() first."` — a message that is also technically incorrect (the pool was initialized; it just has no workers). The test's title claims the relayer "defaults gracefully," but a user who accidentally passes `poolSize: 0` will get a confusing throw on first use with no guidance.

**Suggested fix:** Either add a validation in the `NodeWorkerPool` constructor that rejects `poolSize < 1` with an explicit `"poolSize must be at least 1"` error and test that construction throws, or keep the current behavior and add an assertion that the first operation fails with a descriptive message. Either contract is defensible; the current test covers neither.

---

### 6. Async `handleMessage` not awaited — outer `try/catch` is dead code for async errors

**File:** `packages/playwright/fixtures/relayer-sdk.node-worker.mjs`, lines 287–293

```javascript
port.on("message", (request) => {
  try {
    handleMessage(request);  // async — not awaited
  } catch (error) {
    sendError(request.id, request.type, error instanceof Error ? error.message : String(error));
  }
});
```

`handleMessage` is `async`. The outer `try/catch` in the message handler cannot intercept rejections from the returned Promise; it only catches synchronous throws. `handleMessage` already has its own internal `try/catch` that correctly handles all async errors and calls `sendError()`. The outer `catch` therefore covers only a narrow synchronous window that effectively does not exist.

The one edge case it could theoretically cover — a malformed `request` that is `null` or `undefined` — would cause `sendError(request.id, ...)` inside the `catch` block to throw itself, producing an unhandled exception and a hung pending request on the pool side.

**Suggested fix:**

```javascript
port.on("message", (request) => {
  // handleMessage catches and reports all errors via sendError internally
  void handleMessage(request);
});
```

If defense against malformed messages is genuinely required, add an explicit null-guard at the top of `handleMessage`, before its internal `try` block.

---

### 7. Parallel → sequential change on batch operations has no regression test coverage

**File:** `packages/sdk/src/token/token.ts`

The diff replaces `Promise.allSettled` (parallel execution) with a sequential `for` loop across multi-token batch operations (`batchTransfer`, `batchShield`, etc.). The stated rationale — avoiding nonce contention when multiple transactions share the same signer — is correct.

However, this change has two non-trivial behavioral consequences that are currently untested:

1. **Performance:** a batch of N tokens now takes at minimum `N × single_tx_time` instead of `single_tx_time / concurrency`.
2. **Failure semantics:** an error on token 2 now blocks tokens 3 and 4 from even being attempted, whereas `allSettled` gave each token an independent outcome.

No node e2e test exercises batch operations. Neither the absence of nonce errors (the stated goal of the change) nor the new fail-fast behavior is validated anywhere.

---

## Low

### 8. Redundant and potentially over-broad glob in `tsconfig.node.json`

**File:** `packages/playwright/tsconfig.node.json`

```json
"include": ["fixtures", "tests/node", "**/*.node.spec.ts"]
```

`tests/node` already captures all `*.node.spec.ts` files in that directory. The `**/*.node.spec.ts` glob is redundant in the normal case, and could cause unexpected behavior if a `.node.spec.ts` file ever appears outside of `tests/node/` — it would be type-checked under the Node.js config (no `DOM` lib) rather than the browser config where it belongs.

**Suggested fix:**

```json
"include": ["fixtures", "tests/node"]
```

---

### 9. Double-terminate of `RelayerNode` via fixture composition

**File:** `packages/playwright/fixtures/node-test.ts`, lines 69–92

The `sdk` fixture creates a `ZamaSDK` instance that holds a reference to the `relayer` fixture. When a test ends, Playwright disposes fixtures in reverse dependency order:

1. `sdk` fixture → `sdk[Symbol.dispose]()` → `sdk.terminate()` → **`relayer.terminate()`**
2. `relayer` fixture → `relayer[Symbol.dispose]()` → **`relayer.terminate()` (second call)**

`RelayerNode.terminate()` is currently idempotent (guarded by a `#terminated` flag), so no error surfaces today. This is a latent ownership confusion: `sdk` and `relayer` silently share responsibility for the relayer's lifecycle. If `terminate()` were ever given side effects (metrics, logging, cleanup assertions), they would fire twice.

If the `relayer` fixture is always consumed through `sdk`, consider removing the `using` declaration on the relayer fixture and letting `sdk.terminate()` own the full lifecycle explicitly.

---

### 10. `iife-plugin.ts` fix is correct but the reason is undocumented

**File:** `packages/sdk/iife-plugin.ts`

The diff changes the return value from a raw string to `{ code, moduleType: "js" }`. This is the correct rolldown API for returning transformed module code with an explicit module type from a `load()` hook — without it, rolldown may treat the output as the wrong module kind. The fix is correct. A one-line comment explaining why the previous plain-string return was insufficient would help the next person who touches this file.

---

## Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | **Critical** | `.github/workflows/playwright.yml` | `test-node` job missing `pnpm build` — will fail on every fresh CI checkout |
| 2 | **High** | `fixtures/node-global-setup.ts` | Backup overwrite on interrupted run permanently destroys the original SDK worker |
| 3 | **High** | `relayer/relayer-sdk.ts` | `RelayerSDK extends Disposable` is an undocumented breaking change for interface implementors |
| 4 | **High** | `tests/node/chain-switching.node.spec.ts` | Keypair uniqueness assertion cannot detect a broken pool re-initialization |
| 5 | **Medium** | `tests/node/error-handling.node.spec.ts` | `poolSize: 0` only tests construction; broken operational state untested |
| 6 | **Medium** | `fixtures/relayer-sdk.node-worker.mjs` | `handleMessage` async not awaited; outer `try/catch` is dead code for async errors |
| 7 | **Medium** | `sdk/src/token/token.ts` | Parallel→sequential batch change alters failure semantics with no regression coverage |
| 8 | **Low** | `playwright/tsconfig.node.json` | Redundant `**/*.node.spec.ts` glob |
| 9 | **Low** | `fixtures/node-test.ts` | Double-terminate of `RelayerNode` via fixture composition |
| 10 | **Low** | `sdk/iife-plugin.ts` | Correct fix, reason undocumented |

PR #142, opened 2026-03-22
refactor: reorganize SDK folder structure

ghermet:
## Summary

Reorganize the SDK's internal folder structure from a monolithic `src/token/` catch-all into a domain-driven layout, and separate test infrastructure from published packages.

### Before

```
src/
├── token/               ← catch-all: token logic, credentials, storage, errors, types, SDK entry
│   ├── token.ts
│   ├── readonly-token.ts
│   ├── zama-sdk.ts
│   ├── token.types.ts
│   ├── errors.ts
│   ├── credentials-manager.ts
│   ├── delegated-credentials-manager.ts
│   ├── credential-manager-base.ts
│   ├── credential-crypto.ts
│   ├── credential-validation.ts
│   ├── session-signatures.ts
│   ├── async-local-storage.ts
│   ├── chrome-session-storage.ts
│   ├── indexeddb-storage.ts
│   ├── memory-storage.ts
│   ├── balance-cache.ts
│   ├── eip1193-subscribe.ts
│   └── pending-unshield.ts
├── utils.ts             ← mixed bag: pLimit, getBrowserExtensionRuntime, toError, assertions
├── relayer/
├── query/
├── worker/
├── events/
├── contracts/
├── ethers/
└── viem/
```

### After

```
src/
├── zama-sdk.ts          ← SDK entry point (promoted to root)
├── token/               ← only token logic now
│   ├── token.ts
│   ├── readonly-token.ts
│   ├── balance-cache.ts
│   ├── concurrency.ts   ← pLimit extracted from utils.ts
│   ├── eip1193-subscribe.ts
│   └── pending-unshield.ts
├── credentials/         ← NEW: credential management domain
│   ├── credentials-manager.ts
│   ├── delegated-credentials-manager.ts
│   ├── credentials-manager-base.ts
│   ├── credential-crypto.ts
│   ├── credential-validation.ts
│   └── session-signatures.ts
├── storage/             ← NEW: storage adapters domain
│   ├── async-local-storage.ts
│   ├── chrome-session-storage.ts
│   ├── indexeddb-storage.ts
│   └── memory-storage.ts
├── errors/              ← NEW: split from monolithic errors.ts
│   ├── base.ts
│   ├── credential.ts
│   ├── delegation.ts
│   ├── encryption.ts
│   ├── relayer.ts
│   ├── signing.ts
│   └── transaction.ts
├── types/               ← NEW: split from monolithic token.types.ts
│   ├── callbacks.ts
│   ├── contract.ts
│   ├── credentials.ts
│   ├── signer.ts
│   ├── storage.ts
│   └── transaction.ts
├── utils/               ← NEW: split from monolithic utils.ts
│   ├── assertions.ts
│   ├── error.ts
│   └── hex.ts
├── worker/
│   ├── browser-extension.ts  ← extracted from utils.ts
│   └── ...
├── relayer/
├── query/
├── events/
├── contracts/
├── ethers/
└── viem/
```

### Key changes

| Change | Detail |
|--------|--------|
| **`token/` → `credentials/`** | Credential management files (`credentials-manager`, `delegated-credentials-manager`, `credential-crypto`, `credential-validation`, `session-signatures`, `credentials-manager-base`) moved to their own domain |
| **`token/` → `storage/`** | Storage adapters (`memory-storage`, `indexeddb-storage`, `chrome-session-storage`, `async-local-storage`) moved to their own domain |
| **`token/` → `src/` root** | `zama-sdk.ts` promoted to package root — it's cross-cutting, not token-specific |
| **`token/errors.ts` → `errors/`** | Monolithic error file split into domain-specific modules: `base`, `credential`, `delegation`, `encryption`, `relayer`, `signing`, `transaction` |
| **`token/token.types.ts` → `types/`** | Monolithic types file split into domain-specific modules: `callbacks`, `contract`, `credentials`, `signer`, `storage`, `transaction` |
| **`utils.ts` → `utils/` + relocations** | `pLimit` → `token/concurrency.ts` (only consumer); `getBrowserExtensionRuntime` → `worker/browser-extension.ts` (only consumer); remaining helpers split into `utils/assertions.ts`, `utils/error.ts`, `utils/hex.ts` |
| **Barrel exports** | Added `index.ts` barrel files for `credentials/`, `storage/`, `token/`, `errors/`, `types/`, `utils/`, `events/` |
| **`relayer-sdk` hard dep** | Changed from optional peer dependency to a hard dependency in `package.json` |

## Monorepo restructure

| Change | Detail |
|--------|--------|
| **`packages/` → `test/`** | Moved `playwright`, `test-components`, `test-nextjs`, `test-vite` from `packages/` to `test/` — these are test infrastructure, not published packages |
| **`pnpm-workspace.yaml`** | Added `test/*` to workspace packages |

## Includes merged changes from `prerelease`

This branch includes changes merged from `prerelease` (via PR #143):
- New SDK events: `CredentialsPersistFailed`, `CredentialsCorrupted`, `DelegationSubmitted`, `RevokeDelegationSubmitted`
- `wrapSigningError` rejection heuristics narrowed to match only `"user rejected"` / `"user denied"` (stricter than previous `"rejected"` / `"denied"`)
- `assertBaseEncryptedCredentials` now validates `startTimestamp` and `durationDays` numeric fields
- New event emission tests for `CredentialsCorrupted` and `CredentialsPersistFailed`

## Other

- Fixed all `oxlint` errors (unused imports, empty destructuring, unused parameters)
- Added unit tests for `events/onchain-events.ts` decoders (23 tests)
- ~290 files touched across the SDK, but most changes are import path updates

## Test plan

- [x] `pnpm typecheck` — all packages pass
- [x] `pnpm test` — 146 test files, 1356 tests pass
- [x] `pnpm lint` — 0 errors, 0 warnings
- [x] `pnpm api-report` — API reports up to date


zama-cremaud:
# Code Review — `chores/sdk-folder-structure`

**Branch:** `origin/chores/sdk-folder-structure` → `origin/prerelease`
**Scope:** `packages/sdk/src/` only. The `test/` directory rename is excluded as mechanical.

---

## Summary

The structural refactor is well-executed. Module boundaries are clean, the error hierarchy is sound, and the test coverage for the new modules is thorough. The public API surface (`src/index.ts`) is purely additive — no previously exported identifier was removed or renamed. The findings below are concentrated in two areas: missing exports for new event types, and an inconsistent `cause` wrapping pattern in `token.ts` that silently drops error context.

---

## High

### H1 — `CredentialsPersistFailedEvent` and `CredentialsCorruptedEvent` added to the public union but not individually exported

**Files:**
- `packages/sdk/src/events/sdk-events.ts` — `ZamaSDKEvent` discriminated union
- `packages/sdk/src/index.ts` — event type re-exports

Two new event interfaces, `CredentialsPersistFailedEvent` and `CredentialsCorruptedEvent`, were added to the `ZamaSDKEvent` union and are emitted at runtime via `ZamaSDKEvents.CredentialsPersistFailed` and `ZamaSDKEvents.CredentialsCorrupted`. Both carry a required `error: Error` field. Neither interface name is re-exported from `packages/sdk/src/index.ts`.

This has two compounding consequences:

1. **Exhaustive switch break.** Any consumer with an exhaustive handler over `ZamaSDKEvent` (e.g. `default: assertNever(event)`) gets a TypeScript compile error after upgrading, with no way to name the new members in the switch arms without deep-importing from internal dist paths.

2. **Type narrowing is impossible through the public API.** The `onEvent` callback receives a `ZamaSDKEvent`. To narrow to `CredentialsPersistFailedEvent`, a consumer compares `event.type === ZamaSDKEvents.CredentialsPersistFailed`. That works at runtime, but if they want to reference the interface type explicitly (typed handler map, generic helper function), they cannot do so through `@zama-fhe/sdk`.

`ZamaSDKEvent` is the primary observability surface. Consumers build error loggers, dashboards, and retry handlers against it. Expanding the union without exporting the new member types is a silent, hard-to-diagnose breaking change for any consumer who handles events exhaustively.

**Fix:** Add the two interfaces to the event export block in `src/index.ts`:

```ts
export type {
  // ... existing event types ...
  CredentialsPersistFailedEvent,
  CredentialsCorruptedEvent,
} from "./events/sdk-events";
```

---

### H2 — `token.ts` error wrapping discards non-`Error` causes, inconsistent with the rest of the codebase

**File:** `packages/sdk/src/token/token.ts` — all catch blocks except `#waitAndFinalizeUnshield`

Every catch block in `token.ts` applies:
```ts
throw new TransactionRevertedError("...", {
  cause: error instanceof Error ? error : undefined,
});
```

The ternary silently drops the original cause whenever it is not a native `Error` instance. This affects viem's `ContractFunctionRevertedError` (a structured object), MetaMask's `ProviderRpcError`, and any wallet adapter that throws a plain object — all common in practice.

This is inconsistent with three other places in the same codebase:
- `packages/sdk/src/token/readonly-token.ts`: `{ cause: error }` — raw pass-through
- `packages/sdk/src/errors/signing.ts` (`wrapSigningError`): `{ cause: error }` — raw pass-through
- `token.ts` itself in `#waitAndFinalizeUnshield`: `{ cause: error }` — raw pass-through

`ErrorOptions.cause` accepts `unknown`, so the instanceof narrowing serves no typing purpose. When a transaction reverts, the `cause` on the resulting `TransactionRevertedError` is the only place where the revert reason lives. Dropping it makes `error.cause?.message` logging and programmatic error inspection silent.

**Fix:** Remove the instanceof guard uniformly across all catch blocks in `token.ts`:

```ts
throw new TransactionRevertedError("Transfer transaction failed", { cause: error });
```

---

## Medium

### M1 — `types/transaction.ts` imports from `events/`, reversing the intended layer dependency

**File:** `packages/sdk/src/types/transaction.ts`, line 3

```ts
import type { RawLog } from "../events/onchain-events";
```

`types/` is the leaf layer: it defines framework-agnostic primitives that the rest of the SDK builds on. Importing from `events/` reverses that relationship and means `types/` cannot be understood or reused without pulling in `events/onchain-events`. The `types/__tests__/transaction.test-d.ts` mirrors this by also importing `RawLog` directly from `../../events/onchain-events`.

`RawLog` is a three-field struct (`topics`, `data`, `address`) with nothing event-specific about it — it is a primitive log shape that belongs in `types/`.

**Fix:** Move `RawLog` into `packages/sdk/src/types/transaction.ts` (or `types/log.ts`), then re-export it from `events/onchain-events` to preserve the existing import surface:

```ts
// types/transaction.ts
export interface RawLog { topics: string[]; data: string; address: string; }

// events/onchain-events.ts
export type { RawLog } from "../types/transaction";
```

---

### M2 — `assertBaseEncryptedCredentials` does not validate `startTimestamp` and `durationDays`

**File:** `packages/sdk/src/credentials/credential-validation.ts`

`BaseEncryptedCredentials` declares both `startTimestamp: number` and `durationDays: number`, and `isTimeValid` uses them in arithmetic. However, `assertBaseEncryptedCredentials` validates only `publicKey`, `contractAddresses`, and `encryptedPrivateKey`. The two numeric fields go unchecked on the non-delegated credentials path.

`assertDelegatedFields` — which calls `assertBaseEncryptedCredentials` first — correctly adds its own numeric checks. The gap is isolated to the base assertion.

If storage is corrupted and these fields hold non-number values (`null`, a string timestamp), `isTimeValid` produces `NaN` from the arithmetic, which evaluates as `false` and silently triggers credential regeneration — the user sees an unexpected wallet signing prompt with no logged error explaining why.

Note: this is a pre-existing bug that was moved here without being fixed. This PR is the right place to address it since the validation functions are now isolated and well-tested.

**Fix:** Add to `assertBaseEncryptedCredentials`, after the `encryptedPrivateKey` block:

```ts
assertCondition(
  typeof data.startTimestamp === "number",
  "Expected credentials.startTimestamp to be a number",
);
assertCondition(
  typeof data.durationDays === "number",
  "Expected credentials.durationDays to be a number",
);
```

---

## Low

### L1 — `wrapSigningError` exported from the domain barrel; should be internal

**File:** `packages/sdk/src/errors/index.ts`, line 2

```ts
export { SigningRejectedError, SigningFailedError, wrapSigningError } from "./signing";
```

`wrapSigningError` is an internal factory helper. Exporting it from the domain barrel alongside the public error classes implies it is part of the public API, which it is not. It is not re-exported from `src/index.ts`, so it is not directly reachable through the package entry point today. However, any bundler-level code splitting or consumer importing from a deep dist path can reach it, and the barrel-level export communicates the wrong intent.

**Fix:** Remove `wrapSigningError` from `errors/index.ts`. Consumers inside the package can import directly from `"../errors/signing"`.

---

### L2 — `wrapSigningError` rejection heuristic matches on `"denied"` — too broad

**File:** `packages/sdk/src/errors/signing.ts`, line ~32

```ts
error.message.includes("denied")
```

Messages such as `"Request denied by proxy"`, `"Ledger denied (locked)"`, and `"Permission denied"` all match this pattern and will be classified as `SigningRejectedError` (user-initiated) instead of `SigningFailedError` (infrastructure failure). A user whose Ledger is locked receives an incorrect prompt ("approve in wallet") instead of an actionable error.

**Fix:** Narrow to phrases that unambiguously indicate user action:

```ts
const msg = error.message.toLowerCase();
const isRejected =
  msg.includes("user rejected") ||
  msg.includes("user denied");
```

---

### L3 — `checkExpired` JSDoc contradicts its behavior for missing credentials

**File:** `packages/sdk/src/credentials/credential-manager-base.ts`

The JSDoc reads: *"Returns `true` if credentials are missing, expired, or corrupted."* The implementation returns `false` when `stored` is `null`. The method name `checkExpired` combined with the JSDoc creates a sharp expectation mismatch for anyone reading the interface.

**Fix (pick one):**
1. Correct the JSDoc: *"Returns `true` if stored credentials are expired or corrupted. Returns `false` if no credentials exist yet."*
2. Return `true` when `stored === null` to align behavior with the documented contract, and update callers accordingly.

---

### L4 — New observability events have no emission tests

**File:** `packages/sdk/src/credentials/__tests__/credentials-manager.test.ts`

The test `"regenerates when stored JSON is corrupted"` verifies the outcome (credentials regenerated, storage cleaned) but does not assert that `ZamaSDKEvents.CredentialsCorrupted` is emitted. Similarly, `persistCredentials` failure is not covered by any test that checks for the `CredentialsPersistFailed` event. These are the primary signals for production monitoring.

**Fix:** Add event capture to the existing corruption test:

```ts
const events: string[] = [];
const manager = new CredentialsManager({
  ...,
  onEvent: (e) => events.push(e.type),
});
// ... corrupt storage, trigger regeneration ...
expect(events).toContain(ZamaSDKEvents.CredentialsCorrupted);
```

Add a complementary test that stubs `storage.set` to throw and asserts `CredentialsPersistFailed` is emitted.

---

### L5 — `events/` module has no unit tests

**Directory:** `packages/sdk/src/events/`

Unlike every other new module (`errors/`, `types/`, `credentials/`, `token/`, `utils/`), `events/` received no `__tests__/` directory in this refactor. The `onchain-events.ts` decoders (`decodeConfidentialTransfer`, `decodeUnwrapRequested`, etc.) contain non-trivial ABI decoding logic. High-level coverage exists in `src/__tests__/events.test.ts`, but unit tests for the individual decode functions and the `findUnwrapRequested` helper are missing and would be consistent with the test structure established for every other domain.

---

### L6 — File name inconsistency: `credential-manager-base.ts` vs `credentials-manager.ts`

**Directory:** `packages/sdk/src/credentials/`

The base class file is `credential-manager-base.ts` (singular `credential`), while the other files in the same module use the plural form consistently (`credentials-manager.ts`, `delegated-credentials-manager.ts`). The exported class is `BaseCredentialsManager` (plural). Renaming the file to `credentials-manager-base.ts` would make the pattern uniform. This is a trivial rename and can be deferred to a follow-up.

---

## What Is Done Well

- The error hierarchy in `errors/` is clean and complete. Each class sets `this.name` explicitly, the `Object.setPrototypeOf` pattern preserves `instanceof` across transpilation targets, and `ErrorOptions` threading ensures `cause` chains are preserved end-to-end.
- The `errors/__tests__/errors.test-d.ts` type tests are substantive and non-vacuous. The `matchZamaError` distributive return type tests in particular exercise the conditional type correctly.
- The `BaseCredentialsManager` extraction is a sound abstraction. The deduplication of the create/extend/resolve state machine eliminates real code duplication from `prerelease`, and the abstract hook pattern (`assertEncrypted`, `signForContracts`, `encryptCredentials`, `decryptCredentials`) is a clean template method design.
- `ZamaSDKEventInput` is correctly typed as a distributive mapped type, and the `emit()` helper in the base class is a sensible pattern for stamping `timestamp` automatically.
- Public API surface (`src/index.ts`) is strictly additive relative to `prerelease`. No previously exported identifier was removed or renamed. Existing consumers require no changes.

zama-cremaud (dismissed):

zama-cremaud (dismissed):

ankurdotb:
### Code review

Found 5 issues:

1. **New `ae-forgotten-export` warnings in query entry point** — `CredentialsPersistFailedEvent` and `CredentialsCorruptedEvent` are exported from `src/index.ts` but missing from `src/query/index.ts`, producing two new API Extractor warnings that did not exist on `prerelease`. Consumers of `@zama-fhe/sdk/query` cannot import these event types by name.

https://github.com/zama-ai/sdk/blob/a0f55d42c63b8d25f18851d06e089e8fa16a5ca5/packages/sdk/etc/sdk-query.api.md#L1497-L1498

2. **Behavioral change hidden in refactor: `wrapSigningError` rejection heuristic narrowed** — The old implementation matched any message containing `"rejected"` or `"denied"`. The new implementation requires the `"user "` prefix (`"user rejected"` or `"user denied"`). Errors like `"Transaction rejected by network"` or `"Signature rejected"` that previously produced `SigningRejectedError` will now produce `SigningFailedError`. This changes how callers' `matchZamaError` / `instanceof` handlers classify these errors. Additionally, the `code === 4001` check was broadened from `instanceof Error` objects only to any object — a second behavioral change. Both changes are intentional (new tests document the behavior), but they are functional changes buried in what is framed as a folder-structure refactor.

Old (`prerelease`):
```ts
const isRejected =
  (error instanceof Error && "code" in error && error.code === 4001) ||
  (error instanceof Error &&
    (error.message.includes("rejected") || error.message.includes("denied")));
```

New:
```ts
const hasCode4001 =
  typeof error === "object" && error !== null && "code" in error && error.code === 4001;
const msg = error instanceof Error ? error.message.toLowerCase() : "";
const hasRejectionMessage = msg.includes("user rejected") || msg.includes("user denied");
```

https://github.com/zama-ai/sdk/blob/a0f55d42c63b8d25f18851d06e089e8fa16a5ca5/packages/sdk/src/errors/signing.ts#L23-L31

3. **`@zama-fhe/relayer-sdk` version range tilde accidentally dropped** — On `prerelease`, the dependency is `"~0.4.2"` (allows patch updates). In this PR, it became `"0.4.2"` (exact pin). Based on git history this appears accidental — the tilde was lost during the dependency location shuffling. This prevents automatic pickup of patch releases (e.g. `0.4.3` bug/security fixes).

https://github.com/zama-ai/sdk/blob/a0f55d42c63b8d25f18851d06e089e8fa16a5ca5/packages/sdk/package.json#L131

4. **`@packageDocumentation` says "confidential token operations"** — The SDK serves all FHEVM use cases, not just tokens. The package-level JSDoc says `Core SDK for confidential token operations using Fully Homomorphic Encryption` and lists only token-centric classes. This refactor was a good opportunity to update the description to reflect the broader scope. Pre-existing issue, but worth fixing alongside the folder restructure.

https://github.com/zama-ai/sdk/blob/a0f55d42c63b8d25f18851d06e089e8fa16a5ca5/packages/sdk/src/index.ts#L1-L7

5. **Unresolvable internal reference in public code comment** — `errors/delegation.ts` line 6 says `// See the design doc for the list of ACL revert selectors.` This is a new file created in this PR. External consumers cannot access any internal "design doc". The comment should either link to public documentation, include the relevant selector list inline, or be removed.

https://github.com/zama-ai/sdk/blob/a0f55d42c63b8d25f18851d06e089e8fa16a5ca5/packages/sdk/src/errors/delegation.ts#L3-L6

---

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

packages/sdk/etc/sdk-query.api.md
```diff
@@ -1497,12 +1494,14 @@ export interface ZamaSDKConfig {
     storage: GenericStorage;
 }
 
+// Warning: (ae-forgotten-export) The symbol "CredentialsPersistFailedEvent" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "CredentialsCorruptedEvent" needs to be exported by the entry point index.d.ts
```
ankurdotb: Flagged already in Claude Code review

packages/sdk/package.json
```diff
@@ -128,11 +128,11 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@zama-fhe/relayer-sdk": "~0.4.2",
-    "viem": "^2.47.4"
+    "@zama-fhe/relayer-sdk": "0.4.2",
```
ankurdotb: Deliberately want to pin this dependency or keep it on 0.4.x?

Relayer SDK doesn't follow conventional commits numbering convention fully, the middle version number signifies breaking changes, so anything 0.4.x should be compatible.

ankurdotb (changes requested):
See additional Claude Code review. Since this a major change and API surface diffs mentioned, would like to have a second check from @zama-cremaud on the example apps.

packages/sdk/package.json
zama-cremaud: Indeed, the exact pin looks like an unintentional drop of the ~ during this PR's changes. Probably worth restoring to "~0.4.2" indeed.

zama-cremaud:
### H1 — `CredentialsPersistFailedEvent` / `CredentialsCorruptedEvent` not exported from `@zama-fhe/sdk/query`

*Flagged in v1. Confirmed by @ankurdotb.*

Both types are exported from the main `@zama-fhe/sdk` entry point but missing from `@zama-fhe/sdk/query`, generating two `ae-forgotten-export` warnings in `sdk-query.api.md`. Since `ZamaSDKEvent` (the primary event union) is part of the `/query` surface, consumers importing only `@zama-fhe/sdk/query` cannot type-narrow on the new events by name, and any exhaustive handler over `ZamaSDKEvent` from `/query` will fail to compile post-upgrade.

**Fix:** Re-export both interfaces from `packages/sdk/src/query/index.ts`.

---

### H2 — `token.ts` catch blocks discard non-`Error` causes

*Flagged in v1.*

Every catch block in `packages/sdk/src/token/token.ts` (except `#waitAndFinalizeUnshield`) applies:
```ts
throw new TransactionRevertedError("...", {
  cause: error instanceof Error ? error : undefined,
});
```
This silently drops `cause` for viem's `ContractFunctionRevertedError`, MetaMask's `ProviderRpcError`, and any wallet throwing a plain object — all common in production. `ErrorOptions.cause` accepts `unknown`, so the `instanceof` guard serves no type safety purpose. The inconsistency vs the rest of the file and the rest of the codebase (all of which use `{ cause: error }` directly) confirms this is unintentional.

**Fix:** Remove the `instanceof` guard uniformly across all affected catch blocks.

---

## Open issues — Medium

### M1 — `types/transaction.ts` imports from `events/`, reversing the intended layer dependency

*Flagged in v1.*

`packages/sdk/src/types/transaction.ts` imports `RawLog` from `../events/onchain-events`. `types/` is the leaf layer; it should not depend on `events/`. `RawLog` is a three-field primitive struct with nothing event-specific — it belongs in `types/`. Moving it there and re-exporting from `events/` preserves the existing public surface.

---

### M2 — `assertBaseEncryptedCredentials` does not validate `startTimestamp` and `durationDays`

*Flagged in v1.*

`credential-validation.ts` validates `publicKey`, `contractAddresses`, and `encryptedPrivateKey` but not the two numeric fields that `isTimeValid` uses in arithmetic. If storage is corrupted and these fields are non-numbers, `isTimeValid` silently produces `NaN → false`, triggering unexpected credential regeneration with no logged error. `assertDelegatedFields` is unaffected (it adds its own numeric checks). Pre-existing bug, but this refactor is the right place to fix it.

---

### M3 — `wrapSigningError` behavioral changes are intentional but undocumented in the PR description

*New finding — flagged by @ankurdotb.*

Two functional changes were introduced to `packages/sdk/src/errors/signing.ts` that are not mentioned in the PR description:

1. **Message heuristic narrowed:** `error.message.includes("rejected")` → `msg.includes("user rejected")` and `"denied"` → `"user denied"`. Errors like `"Transaction rejected by network"` or `"Ledger denied (locked)"` that previously mapped to `SigningRejectedError` now map to `SigningFailedError`. The new heuristic is strictly more correct (it was flagged as overly broad in the v1 review), and the new tests document the exact semantics.

2. **`code === 4001` check broadened:** Previously required `instanceof Error`; now accepts any object with `code: 4001`. This correctly handles wallets (MetaMask, WalletConnect) that throw plain objects rather than `Error` instances.

Both changes are improvements. The concern is purely one of PR framing: this is committed as a "folder structure refactor" but includes non-trivial behavioral changes to error classification that affect how callers' `matchZamaError` / `instanceof` error handlers behave. These changes deserve an explicit mention in the PR description or changelog entry so consumers can audit their error handling when upgrading.

**Note:** M3 resolves the L2 concern raised in the v1 review (which flagged `"denied"` as too broad).

---

## Low — Easy fixes

### L-A — `@zama-fhe/relayer-sdk` tilde accidentally dropped

*New finding — flagged by @ankurdotb.*

`packages/sdk/package.json` changed `"~0.4.2"` to `"0.4.2"` (exact pin). Based on git history this is an unintentional loss of the tilde during the dependency shuffle. The `~` should be restored to allow automatic pickup of `0.4.x` patch releases — which per relayer-sdk's own versioning convention are backward-compatible.

### L-B — Unresolvable internal reference in `errors/delegation.ts`

*New finding — flagged by @ankurdotb.*

`packages/sdk/src/errors/delegation.ts` line 6:
```ts
// See the design doc for the list of ACL revert selectors.
```
This is a new file in this PR and this comment references an internal-only document. External consumers reading the source cannot follow this reference. The comment should either link to a public resource, include the selector list inline, or be removed.

ankurdotb (approved):

PR #147, opened 2026-03-23
feat(sdk): validate balances before transfer, shield, and unshield (SDK-52)

ghermet:
## Summary

- Add pre-flight balance validation to `confidentialTransfer`, `shield`, and `unshield` to prevent silent on-chain failures
- `shield`: always validates ERC-20 `balanceOf` (public read, works for all wallet types, even with `skipBalanceCheck: true`)
- `confidentialTransfer` / `unshield`: decrypt and validate cToken balance when credentials are cached; throw clear error if not decryptable
- Add `skipBalanceCheck?: boolean` option (default `false`) for smart wallets that cannot produce EIP-712 signatures
- 3 new error codes: `INSUFFICIENT_CONFIDENTIAL_BALANCE`, `INSUFFICIENT_ERC20_BALANCE`, `BALANCE_CHECK_UNAVAILABLE`

## Design decisions

- **Safe by default** — balance checks are on unless explicitly opted out
- **No surprise popups** — only decrypts when credentials are already cached (`isAllowed()`); throws `BALANCE_CHECK_UNAVAILABLE` otherwise (respects SDK-42 explicit-action pattern)
- **ERC-20 check is unconditional for shield** — public read with no signing, so it always runs
- **Backward compatible** — `TransferOptions`/`UnshieldOptions` extend existing callback interfaces; `shield` options add `skipBalanceCheck` alongside existing fields

## Test plan

- [x] `confidentialTransfer` with amount > cToken balance throws `INSUFFICIENT_CONFIDENTIAL_BALANCE`
- [x] `shield` with amount > ERC-20 balance throws `INSUFFICIENT_ERC20_BALANCE` (all wallet types)
- [x] `unshield` with amount > cToken balance throws `INSUFFICIENT_CONFIDENTIAL_BALANCE`
- [x] `skipBalanceCheck: true` bypasses confidential balance validation
- [x] Shield ERC-20 check runs even with `skipBalanceCheck: true`
- [x] Error messages include requested amount, available balance, and token address
- [x] `BALANCE_CHECK_UNAVAILABLE` thrown when no cached credentials and `skipBalanceCheck` is `false`
- [x] All 1013 existing tests pass
- [x] TypeScript compilation clean

Closes SDK-52


zama-cremaud:
### [HIGH] Behavioral breaking change — `BalanceCheckUnavailableError` is now the default when no credentials are cached

**File:** `packages/sdk/src/token/token.ts` — `#assertConfidentialBalance`

With `skipBalanceCheck` defaulting to `false`, calling `token.confidentialTransfer(to, amount)` or `token.unshield(amount)` without having called `allow()` first now throws immediately:

```
BalanceCheckUnavailableError: Cannot validate confidential balance: no cached credentials.
Call allow() first or use skipBalanceCheck: true for smart wallets
```

Previously, these operations would proceed to the signing step and trigger the EIP-712 prompt at the relayer layer. This is a **semantic breaking change** for any integration that relied on the "transfer triggers the signing popup" flow — which, based on the existing tests (where `skipBalanceCheck: true` had to be sprinkled across dozens of existing test cases), appears to have been the standard pattern.

**Concrete impact:** a dApp that calls `mutate({ to, amount })` without calling `useAllowTokens()` or `token.allow()` beforehand will now surface an error to the user instead of the expected signing prompt.

This may be intentional — enforcing the "explicit allow before decrypt" pattern from SDK-42. But it should be:

1. **Documented prominently** as a breaking change in the CHANGELOG (currently the CHANGELOG has no entry for this branch at all — only a bullet-style reformat commit).
2. **Addressed in the migration guide** for any partner teams already using `useConfidentialTransfer` or `useUnshield`.

---

### [MEDIUM] Public API breaking change on React hook mutation params (`ConfidentialTransferParams`, `UnshieldParams`)

**Files:** `packages/sdk/src/query/transfer.ts`, `packages/sdk/src/query/unshield.ts`

Callbacks were previously nested under a `callbacks` key:

```ts
// Before
mutation.mutate({ to, amount, callbacks: { onTransferSubmitted: () => ... } });
mutation.mutate({ amount, callbacks: { onUnwrapSubmitted: () => ... } });
```

They are now flattened directly into the params object:

```ts
// After
mutation.mutate({ to, amount, onTransferSubmitted: () => ... });
mutation.mutate({ amount, onUnwrapSubmitted: () => ... });
```

Since `ConfidentialTransferParams` and `UnshieldParams` are **exported types** (`index.ts`), any consumer using the old shape will get a runtime no-op (the `callbacks` key is silently ignored instead of being rejected by TypeScript — `callbacks` is just an extra property). This means the break is **silent**: TypeScript will not catch it if the consumer spreads a `callbacks` object, and the callbacks simply won't fire.

This should be in the CHANGELOG as a breaking change and, if needed, a deprecation shim (`callbacks?: T`) could be added with a warning to smooth the migration.

---

### [MEDIUM] `isAllowed()` guard in `#assertConfidentialBalance` — verify it is sufficient to prevent a signing popup in `decryptBalance`

**File:** `packages/sdk/src/token/token.ts` — `#assertConfidentialBalance`

The guard pattern is:

```ts
const hasCredentials = await this.isAllowed();           // (1) session-level check
if (!hasCredentials) throw new BalanceCheckUnavailableError(...);

balance = await this.decryptBalance(handle, userAddress); // (2) internally calls credentials.allow(this.address)
```

`isAllowed()` delegates to `CredentialsManager.checkAllowed(storeKey)` — a session-level (wallet-scoped) check. `decryptBalance` internally calls `credentials.allow(this.address)` which is token-address-scoped. If the cached session was originally created for a different set of contract addresses and the credential manager's `resolveCredentials` treats them as distinct entries, step (2) could still trigger a signing popup even though step (1) returned `true`.

**Recommendation:** confirm (or add a unit test) that `CredentialsManager.checkAllowed(key)` returning `true` guarantees that a subsequent `allow(contractAddress)` call for any address will resolve from cache without calling `create()`. If the invariant holds, add a comment in `#assertConfidentialBalance` stating the guarantee explicitly.

---

### [LOW] `#assertConfidentialBalance` throws for `amount === 0n` when handle is zero

**File:** `packages/sdk/src/token/token.ts`

When `amount === 0n` and `isZeroHandle(handle)` is `true`, the method throws `InsufficientConfidentialBalanceError(requested: 0n, available: 0n)`. Technically `0 >= 0` satisfies the constraint. Zero-amount transfers are probably invalid on-chain anyway, but the error message ("Insufficient confidential balance: requested 0, available 0") could be confusing.

Minor edge case — worth noting but low priority.

---

### [LOW] No test for non-`ZamaError` thrown from `decryptBalance` in `#assertConfidentialBalance`

**File:** `packages/sdk/src/token/__tests__/token.test.ts`

The catch block that wraps an unexpected error from `decryptBalance` in `BalanceCheckUnavailableError`:

```ts
} catch (error) {
  if (error instanceof ZamaError) throw error;
  throw new BalanceCheckUnavailableError(
    `Balance validation failed: could not decrypt confidential balance (token: ${this.address})`,
    { cause: error instanceof Error ? error : undefined },
  );
}
```

This path (non-`ZamaError` thrown by `decryptBalance`) has no dedicated test. The happy-path and `ZamaError` re-throw paths are covered; this third branch is not.

---

## What Is Good

**Typed error classes with structured fields.** `InsufficientConfidentialBalanceError` and `InsufficientERC20BalanceError` expose `requested`, `available`, and `token` as typed `bigint`/`string` fields — much better than having consumers parse error messages. The pattern is consistent with the existing error hierarchy and fully integrates with `matchZamaError` (no changes needed there, as `matchZamaError` uses `Partial<Record<ZamaErrorCode, ...>>`).

**`BalanceCheckUnavailableError` is the right design choice.** Throwing a typed error instead of silently triggering a signing popup mid-flow respects the "explicit action" pattern established in SDK-42. It gives the UI a deterministic signal to show a "Please sign to authorize decryption" prompt rather than having the popup appear at an unexpected moment in the transfer flow.

**ERC-20 balance check for `shield` is unconditional and correct.** The comment ("public read, no signing needed, works for all wallet types") justifies why there is no `skipBalanceCheck` for shield. The ETH shield bypass (`underlying === ZERO_ADDRESS` early return) is also correctly placed before the balance check — the test "skips ERC-20 check for ETH shield" confirms this path.

**`#waitAndFinalizeUnshield` parameter reorder is a clean fix.** Swapping `operationId` and `callbacks` (`(hash, operationId, callbacks)` instead of `(hash, callbacks, operationId)`) and updating all callers consistently is a good cleanup that reduces the chance of future argument-order mistakes.

**Test coverage for the new validation paths is comprehensive.** The new `describe` blocks in `token.test.ts` cover: zero handle, no credentials, insufficient balance, sufficient balance (pass-through), `skipBalanceCheck: true` bypass, and callbacks alongside `skipBalanceCheck` — for all three operations. The documentation in `Token.md` is updated with parameter tables, throw annotations, and examples.

**`lint-staged` fix is sensible.** Removing `*.md` from the `oxfmt` lint-staged target (commit `c3fb817`) prevents pre-commit failures on Markdown files that `oxfmt` doesn't support.

---

## Summary

The core implementation is solid: the new error types are well-designed, the guard logic in `#assertConfidentialBalance` is correct, and test coverage is thorough. The two blocking items before merge are:

1. **Add a CHANGELOG entry** documenting the behavioral breaking change (`BalanceCheckUnavailableError` default) and the flattened callbacks API change.
2. **Confirm or test** that `isAllowed() === true` is a sufficient guard against an unexpected signing popup inside `decryptBalance`.

The low-priority items (zero-amount edge case, missing catch-path test) can be addressed in a follow-up or at the reviewer's discretion.

ghermet (commented):
## PR Review Summary

### Critical Issues (0 found)

No blocking issues.

### Important Issues (3 found)

**1. Redundant `getAddress()` call — second is unguarded**
`#assertConfidentialBalance` calls `this.signer.getAddress()` twice. The second call (before `decryptBalance`) is outside any try-catch. If the wallet disconnects between calls, users get a raw error instead of `BalanceCheckUnavailableError`.
**Fix:** Hoist `userAddress` from the first call and reuse it.

**2. Bare `catch` in `finalizeUnwrap` discards parse error**
`token.ts` ~line 684 — `try { hexToBigInt(...) } catch { throw new DecryptionFailedError(...) }` drops the original parse error. Pre-existing but now on the unshield validation path.
**Fix:** Capture error and add `{ cause: toError(parseError) }`.

**3. Missing boundary tests**
No tests for `amount === balance` (exact boundary). Implementation uses `<` so this should pass, but an off-by-one regression has no guard.

### Suggestions

- **Add `reason` discriminant to `BalanceCheckUnavailableError`** — thrown for 3 distinct failure modes (no handle, no credentials, decryption failed) but distinguished only by message string. A typed `reason` field would enable programmatic handling.
- **Extract shared `BalanceCheckOptions` interface** — both `TransferOptions` and `UnshieldOptions` independently declare `skipBalanceCheck` with identical JSDoc.
- **Inconsistent callback nesting** — `TransferOptions`/`UnshieldOptions` flatten callbacks, `ShieldOptions` nests under `callbacks?`. Intentional for backward compat but worth a code comment.
- **Add tests for `readConfidentialBalanceOf` throwing non-ZamaError** — network failure during balance read is untested.
- **Comment why ETH shield skips balance validation** — intentional but undocumented.

### Strengths

- Clean `ZamaError` re-throw pattern — every catch checks `instanceof ZamaError` before wrapping
- Structured error details (`requested`, `available`, `token`) enable programmatic UI handling
- Cause chains preserved throughout with `toError(error)`
- `skipBalanceCheck` escape hatch well-designed for smart wallets
- Thorough test coverage: zero-handle, no-credentials, insufficient balance, skip, callback forwarding
- API report and type exports properly updated

### Recommended Action

1. Fix the duplicate `getAddress()` call (quick win, real bug potential)
2. Fix the bare `catch` in `finalizeUnwrap` (quick win)
3. Add boundary tests for `amount === balance`
4. Consider the suggestions for a follow-up PR

ankurdotb:
### Code review (pass 2 — 9d6ef769)

Checked against [SDK-52](https://linear.app/zama/issue/SDK-52) acceptance criteria — all 7 acceptance criteria pass.

**Previously raised issues (pass 1, 240f304c):** 6 issues were raised. 4 are resolved, 2 remain (renumbered below), and 1 new issue was found.

---

Found 3 issues:

1. **`react-sdk/README.md` still shows nested `callbacks` pattern** — The gitbook React docs (`useUnshield.md`, `useConfidentialTransfer.md`, `useShield.md`) were updated to the flat callback pattern, but `packages/react-sdk/README.md` (lines 428-431, 439-442, 459-461) still documents `callbacks?: UnshieldCallbacks` and `callbacks: { onUnwrapSubmitted: ... }`. Consumers following the README will have callbacks silently ignored.

https://github.com/zama-ai/sdk/blob/9d6ef76901d6bc441be0a8f3c8b6145042d06980/packages/react-sdk/README.md#L428-L443

2. **`BalanceCheckUnavailableError` reused for ERC-20 read failure** — At `token.ts:378`, a public `balanceOf` network failure throws `BalanceCheckUnavailableError`, which semantically means "no cached credentials / decryption not possible" (see its other 4 usages in `#assertConfidentialBalance`). Callers catching by error type would conflate network errors with missing credentials. Consider a distinct error type for public read failures.

https://github.com/zama-ai/sdk/blob/9d6ef76901d6bc441be0a8f3c8b6145042d06980/packages/sdk/src/token/token.ts#L374-L382

3. **`ShieldOptions` not exported from `sdk/query` entrypoint** — `query/index.ts` was updated to export `TransferOptions` and `UnshieldOptions` but not `ShieldOptions`. Since `Token.shield()` accepts `ShieldOptions` and `Token` is exposed through the query subpath, the generated API report shows a new `ae-forgotten-export` warning. Typed consumers of `@zama-fhe/sdk/query` can see the type in the public surface but cannot import it. (`ShieldOptions` is correctly exported from the main `@zama-fhe/sdk` entrypoint.)

https://github.com/zama-ai/sdk/blob/9d6ef76901d6bc441be0a8f3c8b6145042d06980/packages/sdk/src/query/index.ts#L188-L193

https://github.com/zama-ai/sdk/blob/9d6ef76901d6bc441be0a8f3c8b6145042d06980/packages/sdk/etc/sdk-query.api.md#L5-L6

ankurdotb (approved):

PR #148, opened 2026-03-23
feat(react-ethers): upgrade to SDK 2.1.0-alpha.2 + Sepolia + delegation flows (SDK-35)

zama-cremaud:
- Bump @zama-fhe/react-sdk + @zama-fhe/sdk to 2.1.0-alpha.2
- Replace monolithic page.tsx with component decomposition (BalancesCard, ShieldCard, TransferCard, UnshieldCard, PendingUnshieldCard, DelegateDecryptionCard, RevokeDelegationCard, DecryptAsCard)
- Add Sepolia token addresses (USDC Mock, USDT Mock) with real contract addresses
- Add globals.css shared styles; extract src/lib/{config,ethereum,parseAmount,activeUnshield}.ts
- Wire RelayerWeb through Next.js proxy (/api/relayer) to keep RELAYER_API_KEY server-side
- Add walletKey + refSeededRef pattern for ZamaProvider remount on wallet switch
- Add separate IndexedDBStorage for session signatures
- Add delegation lifecycle: grant / revoke / decrypt-as with useDelegateDecryption, useRevokeDelegation, useDelegationStatus, useDecryptBalanceAs
- Add pending unshield persistence via IndexedDB + ZamaSDKEvents.UnshieldPhase1Submitted
- Add USDT-style approval reset path (front-running guard detection via estimateGas revert)
- Add Playwright e2e tests (5 spec files, auto relayer mock fixture)
- Add WALKTHROUGH.md developer guide and update README.md
- Fix tsconfig.json to standalone (remove extends, fix paths)
- Remove typescript.ignoreBuildErrors from next.config.ts

# Summary

## Scope

- [ ] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] CI/workflows
- [ ] Docs


ankurdotb:
### Code review

Found 1 issue:

1. `NEXT_PUBLIC_SEPOLIA_RPC_URL ?? default` will silently use an empty string as the RPC URL instead of the default. Next.js replaces `NEXT_PUBLIC_*` env vars with `""` (empty string) at build time when not set — not `undefined`. Since `""` is not nullish, `??` does not trigger and the empty string is used as the RPC URL, causing a `UrlRequiredError` at runtime. Use `||` instead of `??`. This same bug was found and fixed in PR #111 for the `example-hoodi` example, which correctly uses `||`.

https://github.com/zama-ai/sdk/blob/3547b455e6768c8d2aba6b131bc4c3fd98a810bc/examples/react-ethers/src/lib/config.ts#L4-L5

Fixed version for reference: https://github.com/zama-ai/sdk/blob/3547b455e6768c8d2aba6b131bc4c3fd98a810bc/examples/example-hoodi/src/lib/config.ts#L7-L8

ghermet:
## Code Review

### Medium Issues (2)

**1. `PERMANENT_DELEGATION` sentinel duplicated from SDK internals**
`examples/react-ethers/src/components/DecryptAsCard.tsx`

`const PERMANENT_DELEGATION = 2n ** 64n - 1n` — the comment says "Not exported by the SDK." This couples to an internal implementation detail. If the SDK changes this sentinel, `formatExpiry` silently breaks. Consider exporting the constant from the SDK.

**2. Implicit coupling between module-level `indexedDBStorage` and context `storage`**
`examples/react-ethers/src/providers.tsx`

`savePendingUnshield` uses the module-level `indexedDBStorage` import, while `PendingUnshieldCard`/`UnshieldCard` use `useZamaSDK().storage`. Both resolve to the same singleton today, but if someone changed the `storage` prop to a different instance, writes and reads would go to different databases.

### Low Issues (5)

3. Relayer proxy forwards client path segments without validation (SSRF risk in non-example contexts)
4. Proxy forwards all client headers (blocklist instead of allowlist)
5. `useEffect` suppresses exhaustive-deps for `queryClient` (safe but masks dependency)
6. No-op signer returns `null` silently — could mask bugs if guard checks are removed
7. `tsconfig.json` removes `.next/dev/types/**/*.ts` from `include` — verify route type generation still works

### Strengths

- Relayer API key properly kept server-side (no `NEXT_PUBLIC_` prefix)
- Thorough error handling (USDT approval fallback, network errors, wallet failures)
- Wallet reactivity pattern (walletKey remount + refSeededRef guard) prevents cross-account state leaks
- Clean component decomposition from monolithic page.tsx
- Comprehensive WALKTHROUGH.md developer guide
- 5 Playwright e2e spec files with auto relayer mock fixture

### Recommended Action

1. Consider exporting `PERMANENT_DELEGATION` from the SDK (or add a comment about the fragility)
2. Use `useZamaSDK().storage` consistently in the event handler instead of the module-level import
3. Remaining low issues are acceptable for an example app — no action required

ghermet:
# PR Review — Comprehensive Automated Analysis

4 specialized agents reviewed this PR in parallel: **code quality**, **silent failure hunting**, **test coverage**, and **comment accuracy**.

---

## Critical Issues (4 — must fix before merge)

### 1. Path traversal via `..` segments escapes `/v2` prefix (SSRF)
**`src/app/api/relayer/[...path]/route.ts`** — `SAFE_SEGMENT = /^[a-zA-Z0-9._-]+$/` allows `..`. A request to `/api/relayer/../v1/admin/foo` resolves to `https://relayer.testnet.zama.org/v1/admin/foo`, escaping the intended `/v2` prefix.

**Fix:** reject `segment === '..' || segment === '.'` after the regex check, or tighten the regex to disallow leading dots: `/^[a-zA-Z0-9_-][a-zA-Z0-9._-]*$/`.

### 2. Hop-by-hop headers not stripped from upstream responses
**`src/app/api/relayer/[...path]/route.ts`** — Request-side stripping uses `HOP_BY_HOP` set, but the response path only strips `RESPONSE_DROP` (content-encoding/length/transfer-encoding). `connection`, `keep-alive`, `proxy-authorization` etc. are forwarded verbatim to the browser.

**Fix:** merge `HOP_BY_HOP` into `RESPONSE_DROP`:
```ts
const RESPONSE_DROP = new Set([...HOP_BY_HOP, "content-encoding", "content-length"]);
```

### 3. No-op signer stub returns `null` instead of throwing
**`src/providers.tsx`** — When no wallet is detected, the fake provider's `request` returns `null` for all RPC calls. If any code path reaches `sdk.signer.*` before connection, it silently propagates `null` through the system producing cryptic downstream failures.

**Fix:** make the stub throw a descriptive error:
```ts
request: async () => { throw new Error("No Ethereum wallet detected"); },
```

### 4. `parseAmount` silently returns `0n` — disabled button with zero user feedback
**`src/lib/parseAmount.ts`** — Bare `catch` swallows all `parseUnits` errors and returns `0n`. User types invalid input (too many decimals, letters, etc.), button stays disabled with no explanation why. Also has **zero unit tests** despite being called by every financial operation.

**Fix:** return a discriminated result (valid/invalid/empty) so callers can show validation messages. Add unit tests.

---

## Important Issues (6 — should fix)

### 5. `savePendingUnshield` failure is fire-and-forget — crash recovery silently broken
**`src/providers.tsx`** — If IndexedDB write fails, only `console.error` is called. User has no indication their in-flight unshield was not persisted. If the tab closes, funds can become stuck with no UI recovery path.

### 6. `clearPendingUnshield` not awaited — stale "Finalize" cards
**`UnshieldCard.tsx` / `PendingUnshieldCard.tsx`** — Fire-and-forget Promise. If clear fails, the card reappears on next load for an already-completed unshield, and clicking "Finalize" reverts on-chain.

### 7. WALKTHROUGH falsely claims "app never calls `wallet_switchEthereumChain`"
**`WALKTHROUGH.md` §2** — `page.tsx` explicitly calls `wallet_switchEthereumChain` in `switchToSepolia()` and via the "Switch to Sepolia" button. The screen table also says "Instructions to switch" when there's actually a programmatic button.

### 8. `switchToSepolia` swallows wallet rejection — no user feedback
**`src/app/page.tsx`** — Non-4902 errors (including user rejection code 4001) are silently ignored. User clicks "Switch to Sepolia", rejects in wallet, button resets with no message.

### 9. WALKTHROUGH "EthersSigner is stateless" is misleading
**`WALKTHROUGH.md` §1** — If truly stateless, remounting ZamaProvider would be unnecessary. The real reason is ZamaProvider caches account-specific session state (EIP-712 signatures) that must be reset on wallet switch.

### 10. README claims "Webpack" but config doesn't enforce it
**`README.md`** — Says "Webpack — Turbopack not yet supported with WASM" but `next.config.ts` is empty and `next dev` uses Turbopack by default. The SDK loads its WASM worker from CDN (not bundled), so the bundler choice is irrelevant here. Either add `--no-turbopack` or remove the claim.

---

## Test Coverage Gaps

| Priority | Gap | Effort |
|----------|-----|--------|
| 1 | `parseAmount` unit tests (edge cases, decimals, invalid input) | Low |
| 2 | Relayer proxy path validation + security tests | Medium |
| 3 | `delegationErrorMessage` mapping unit tests | Low |
| 4 | `activeUnshield` get/set/clear cycle | Low |
| 5 | Transaction error display e2e (mock rejection) | Medium |
| 6 | ShieldCard USDT approval fallback path | Medium |
| 7 | Account switching via `__emitAccountsChanged` | Medium |

All e2e tests verify **UI structure only** — no test exercises a transaction flow or error recovery path.

---

## Cross-PR Note (shared with #150)

Issues **#1 (path traversal)**, **#2 (response headers)**, **#4 (parseAmount)**, **#5 (savePendingUnshield)**, and **#6 (clearPendingUnshield)** are identical between this PR and the viem sister PR #150. Fixes should be mirrored. Issue **#3 (no-op signer)** is ethers-specific and does not apply to the viem PR.

---

## Strengths

- Excellent e2e fixture architecture with EIP-1193 event simulation
- Thorough WALKTHROUGH.md (despite the two inaccuracies noted above)
- Well-documented USDT approval reset path in ShieldCard
- `||` vs `??` Next.js gotcha clearly explained in both code and docs
- Clean component decomposition — 8 focused, props-driven components
- `activeUnshield.ts` bridge pattern well-justified in JSDoc
- Relayer proxy `content-encoding` drop rationale (Node.js auto-decompression) prevents real debugging sessions

---

zama-cremaud:
On latest review, all addressed, except a couple :
                                   
  ▎ 4 — parseAmount discriminated result:
  ▎ The disabled={parsedAmount === 0n} guard already prevents invalid submissions from reaching the chain. A discriminated {valid, invalid, empty} result would require changing the type signature across all six    
  card components that call parseAmount. For a quickstart example the current approach is sufficient. Unit tests for parseAmount are a reasonable separate TODO but not part of this fix scope.                       
                                                                                                                                                                                                                      
  ▎ 5 — savePendingUnshield UI feedback:                                                                                                                                                                             
  ▎ The .catch() handler is already present (logs to console). Adding visible UI feedback requires surfacing an error from inside ZamaProvider's onEvent callback into the React component tree — that callback runs 
  outside the component render cycle and cannot directly set state. The necessary workaround (a ref + a second useEffect to flush it into state) adds non-trivial complexity for an error path that fires only when   
  IndexedDB itself is broken, which in practice doesn't happen.
                                                                                                                                                                                                                      
  ▎ 9 — "EthersSigner is stateless" misleading:                                                                                                                                                                      
  ▎ The sentence is accurate: EthersSigner is stateless in the sense that it reads the current account from the injected provider on each call rather than caching it at construction time. The clause immediately 
  following — "However, ZamaProvider stores session state… To avoid cross-account state leaks, the provider is remounted" — correctly explains why the remount is needed. No change required.                         
                                                            
  ▎ 10 — README claims Webpack but config doesn't enforce it:                                                                                                                                                        
  ▎ The dev script in package.json already has --webpack ("dev": "next dev --webpack"), which forces webpack bundler. The README is accurate. next.config.ts being empty is unrelated — the flag lives in the script, 
  not the config file.

ankurdotb:
### Code review (re-review)

**NOTE**: I haven't been able to test this myself yet due to issue flagged on Slack, to check if the pending unshield thing show up, for example. Don't just fix the issues, fix/add any tests to catch the issue as well to prevent future regressions.

**Previous review status:** Path traversal, header forwarding, no-op signer, `clearPendingUnshield` `.catch()`, `??` vs `||` in config.ts, and `getChainId` error handling have all been addressed. 5 issues remain:

Found 5 issues:

1. **PendingUnshieldCard: success message is never visible** -- `setPendingTxHash(null)` in `onSuccess` causes React 18 to batch it with the mutation state change. On the next render, `pendingTxHash === null` hits the early-return guard at line 59, unmounting the card before the "Unshielded!" alert is ever painted. Users who finalize from the pending recovery card never see confirmation.

https://github.com/zama-ai/sdk/blob/a3bc224cc296a4b1a71e82fd59628241ca52759c/examples/react-ethers/src/components/PendingUnshieldCard.tsx#L37-L59

2. **`handleAccountsChanged` race: sets `address` but not `chainId`** -- MetaMask fires `accountsChanged` on page load for already-connected sites. This sets `address` before the `Promise.all` resolves `eth_chainId`, rendering with `address` set + `chainId` null = persistent "Sepolia Network Required" screen. This was flagged in the previous review and confirmed by a real user report. Still not fixed.

https://github.com/zama-ai/sdk/blob/a3bc224cc296a4b1a71e82fd59628241ca52759c/examples/react-ethers/src/app/page.tsx#L142-L149

3. **ShieldCard: USDT detection try block is too broad** -- `waitForTransactionReceipt` (line 89) is inside the same try block as `writeContract` (line 86). The catch is intended to detect `estimateGas` reverts (USDT front-running guard), but it also catches on-chain tx reverts, network errors, and receipt timeouts. In those cases `needsReset = true` is incorrectly set, masking the real error and entering the 0-then-fullBalance reset path.

https://github.com/zama-ai/sdk/blob/a3bc224cc296a4b1a71e82fd59628241ca52759c/examples/react-ethers/src/components/ShieldCard.tsx#L82-L93

4. **`PERMANENT_DELEGATION` hardcoded, not exported by SDK** -- `DecryptAsCard.tsx` hardcodes `2n ** 64n - 1n`. The equivalent `MAX_UINT64` exists in `packages/sdk/src/contracts/constants.ts` but is not re-exported from the SDK's public API. If the sentinel changes, `formatExpiry` silently displays wrong dates.

https://github.com/zama-ai/sdk/blob/a3bc224cc296a4b1a71e82fd59628241ca52759c/examples/react-ethers/src/components/DecryptAsCard.tsx#L10-L14

5. **Incorrect comment: `mint.reset` is NOT stable in TanStack Query v5** -- The comment claims "`mint.reset` is stable in TanStack Query v5 (callbacks are stabilized internally)". Per TanStack Query v5 docs, `useMutation` returns a new object each render; `reset` is not referentially stable. The effect fires every render (harmless since `reset()` is idempotent), but the comment misleads developers.

https://github.com/zama-ai/sdk/blob/a3bc224cc296a4b1a71e82fd59628241ca52759c/examples/react-ethers/src/app/page.tsx#L247-L251

ankurdotb:
[AUTOMATED]

### Code review

Found 1 issue:

`examples/react-ethers/src/app/page.tsx` currently enables `useConfidentialBalance(...)` as soon as `address && isSepolia` is true. That means the example can trigger the initial EIP-712/session-signature flow immediately on render after connect, before the user has explicitly asked to decrypt anything. For example apps, that is a blind-signing UX problem: the wallet prompt appears as an unsolicited signature request rather than as the result of an explicit user action.

This also conflicts with the guidance captured in SDK-42: example apps should follow the `useIsAllowed()` -> explicit `useAllow()` -> enabled `useConfidentialBalance()` pattern.

Recommended change:

- Use `useIsAllowed()` to check whether decrypt credentials are already cached for the connected wallet.
- Keep `useConfidentialBalance()` disabled until `isAllowed === true`.
- When `isAllowed !== true`, show a locked/encrypted balance state plus an explicit button such as `Decrypt Balance`.
- Wire that button to `useAllow()` so the first wallet signature is clearly user-initiated.
- After `useAllow()` succeeds, enable `useConfidentialBalance()` and let it decrypt normally.

For this specific app, I would also recommend deciding the approval scope up front. Because this is a multi-token demo (`USDC Mock` + `USDT Mock`), the cleanest UX is probably to authorize both wrapper contracts in one explicit user action:

```ts
allow.mutate(Object.values(TOKENS).map((t) => t.confidential));
```

That avoids an extra unexpected signature prompt when the user switches from one demo token to the other, and it makes the example demonstrate the SDK pattern we actually want integrators to copy for multi-token apps. If you prefer stricter least-privilege semantics, authorizing only the currently selected token also works, but then token switching becomes another signature-triggering moment.

That keeps the example aligned with the recommended SDK integration pattern and avoids training users to accept unexpected signature prompts on page load or immediately after connect.

ankurdotb:
### Code review (re-review #3)

**Previous issues status:** All resolved.

- **`isAllowed()` TTL mismatch (SDK-42)** — Fixed in `9ebfc4a1`. `keypairTTL={30 * 24 * 60 * 60}` aligns both TTLs at 30 days, eliminating the window where `isAllowed()` returns true but the keypair is expired. Comment documents the SDK root cause. SDK-level fix tracked in SDK-60.
- **Unshield persistence timing comments** — Fixed in `d0e6906a`. All three comments (`providers.tsx:175`, `UnshieldCard.tsx:55`, `activeUnshield.ts:7`) now correctly say "after Phase 1 is mined but before Phase 2 completes."
- **WALKTHROUGH.md staleness** — Fixed in `9ebfc4a1` + `d0e6906a`. Now documents: initializing screen (Screen 0), token selection via WrappersRegistry (§3), explicit decrypt pattern (§9), corrected UnshieldPhase1Submitted timing (§6), normalizePair ethers compat, renumbered sections.

No new issues found. LGTM.

ankurdotb (dismissed):

ankurdotb (approved):

PR #158, opened 2026-03-24
chore(deps): combine 10 Dependabot dependency bumps

ghermet:
## Summary
- Combines all 10 open Dependabot PRs into a single PR for cleaner review and merge
- **Root**: oxfmt 0.37.0 → 0.41.0, typedoc-plugin-markdown 4.10.0 → 4.11.0
- **test-nextjs**: next 16.1.7 → 16.2.1, @tanstack/react-query 5.90.21 → 5.95.2, typescript 5.9.3 → 6.0.2
- **test-vite**: @tanstack/react-query 5.90.21 → 5.95.2, typescript 5.9.3 → 6.0.2
- **react-wagmi**: socket.io-parser → 4.2.6, h3 1.15.6 → 1.15.9
- **example-hoodi**: next 16.1.6 → 16.1.7

## Supersedes
Closes #132, closes #137, closes #141, closes #151, closes #152, closes #153, closes #154, closes #155, closes #156, closes #157

## Test plan
- [ ] CI passes (lockfile verified locally with `pnpm install`)
- [ ] All dependency bumps are non-breaking (patch/minor)

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,3 +8,6 @@ packages:
 minimumReleaseAge: 4320
 minimumReleaseAgeExclude:
   - "@zama-fhe/relayer-sdk"
+  - "@tanstack/react-query"
+  - "@tanstack/query-core"
+  - "typescript"
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -35,6 +35,10 @@ jobs:
             contracts/cache
           key: forge-build-${{ runner.os }}-${{ hashFiles('contracts/src/**/*.sol', 'contracts/foundry.toml') }}
 
+      - name: Install soldeer dependencies
+        if: steps.forge-build-cache.outputs.cache-hit != 'true'
+        run: cd contracts/lib/forge-fhevm && forge soldeer install
+
       - name: Build contracts
         if: steps.forge-build-cache.outputs.cache-hit != 'true'
         run: cd contracts && forge build
--- a/examples/example-hoodi/package-lock.json
+++ b/examples/example-hoodi/package-lock.json
@@ -10,7 +10,7 @@
         "@zama-fhe/react-sdk": "2.0.0-alpha.2",
         "@zama-fhe/sdk": "2.0.0-alpha.2",
         "ethers": "^6.13.0",
-        "next": "^16.0.0",
+        "next": "^16.1.7",
         "react": "^19.2.0",
         "react-dom": "^19.2.0"
       },
@@ -504,15 +504,15 @@
       }
     },
     "node_modules/@next/env": {
-      "version": "16.1.6",
-      "resolved": "https://registry.npmjs.org/@next/env/-/env-16.1.6.tgz",
-      "integrity": "sha512-N1ySLuZjnAtN3kFnwhAwPvZah8RJxKasD7x1f8shFqhncnWZn4JMfg37diLNuoHsLAlrDfM3g4mawVdtAG8XLQ==",
+      "version": "16.1.7",
+      "resolved": "https://registry.npmjs.org/@next/env/-/env-16.1.7.tgz",
+      "integrity": "sha512-rJJbIdJB/RQr2F1nylZr/PJzamvNNhfr3brdKP6s/GW850jbtR70QlSfFselvIBbcPUOlQwBakexjFzqLzF6pg==",
       "license": "MIT"
     },
     "node_modules/@next/swc-darwin-arm64": {
-      "version": "16.1.6",
-      "resolved": "https://registry.npmjs.org/@next/swc-darwin-arm64/-/swc-darwin-arm64-16.1.6.tgz",
-      "integrity": "sha512-wTzYulosJr/6nFnqGW7FrG3jfUUlEf8UjGA0/pyypJl42ExdVgC6xJgcXQ+V8QFn6niSG2Pb8+MIG1mZr2vczw==",
+      "version": "16.1.7",
+      "resolved": "https://registry.npmjs.org/@next/swc-darwin-arm64/-/swc-darwin-arm64-16.1.7.tgz",
+      "integrity": "sha512-b2wWIE8sABdyafc4IM8r5Y/dS6kD80JRtOGrUiKTsACFQfWWgUQ2NwoUX1yjFMXVsAwcQeNpnucF2ZrujsBBPg==",
       "cpu": [
         "arm64"
       ],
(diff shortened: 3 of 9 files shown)


zama-cremaud:
# Code Review — `ghermet/merge-dependabot-prs`

**Branch:** `origin/ghermet/merge-dependabot-prs` → `origin/prerelease`
**Scope:** Dependency consolidation — merges 10 Dependabot PRs into a single integration branch, plus one CI fix (`fix(ci): regenerate lockfile and add soldeer install to Playwright`).

---

## Summary

The vast majority of updates are low-risk minor/patch bumps to tooling and test-harness dependencies. Two items warrant attention before merge: a permanent policy change to `pnpm-workspace.yaml` that silently broadens supply-chain exposure going forward, and a TypeScript **major** version bump (`5→6`) that is scoped to test packages in the manifest but bleeds into peer resolution across the wider monorepo lock. The CI soldeer fix is correct but has a one-line inconsistency with the existing `postinstall` script that could silently affect CI runs on repos with LFS configured.

---

## Medium

### M1 — `minimumReleaseAgeExclude` is a permanent supply-chain policy change, not a one-time override

**File:** `pnpm-workspace.yaml`

Three packages were added to the `minimumReleaseAgeExclude` list:

```yaml
minimumReleaseAgeExclude:
  - "@tanstack/react-query"
  - "@tanstack/query-core"
  - typescript
```

`minimumReleaseAge` (typically 3 days) is a supply-chain safeguard: it prevents pnpm from automatically resolving to a newly-published version that has not yet had time to be vetted by the community. A compromised or accidentally-broken release discovered within the first few hours of publication is a real scenario ([`event-stream`](https://blog.npmjs.org/post/180565256905/details-about-the-event-stream-incident), [`node-ipc`](https://www.bleepingcomputer.com/news/security/big-sabotage-famous-npm-package-deletes-files-to-protest-ukraine-war/)).

Adding packages to `minimumReleaseAgeExclude` is permanent and inherited by all future `pnpm install` runs. The three packages are being bumped in this PR, so the exclusion is working around the age gate for this specific bump — but it will also suppress the safeguard for all future bumps of these packages indefinitely.

For `typescript` in particular, this is significant: TypeScript releases are frequent (multiple RCs, betas, and patches), and the TypeScript compiler is executed in every build and CI run. Bypassing the release age check for it permanently is a non-trivial policy decision.

**Recommendation:** Consider using a time-bounded override approach (bump the resolved version with a specific pin, then let the exclusion be temporary), or explicitly document in a comment that this is an intentional, permanent policy decision for these packages. At minimum, the PR description should call this out.

---

### M2 — TypeScript 6 is a major bump; its scope in the manifest is narrower than its scope in the lock

**Files:** `packages/test-nextjs/package.json`, `packages/test-vite/package.json`

The manifest changes specify `typescript: ^6.0.2` only in `packages/test-nextjs` and `packages/test-vite`. However, TypeScript 6 is a **major** version release — major TypeScript releases introduce stricter type checking, remove deprecated APIs, and can cause type errors in code that compiled cleanly under TypeScript 5.

Two concrete consequences worth verifying:

1. **Peer resolution cascade.** pnpm resolves the highest compatible version globally. After this bump, `typescript@6.0.2` is used to satisfy the `typescript` peer dep for `viem`, `wagmi`, `rolldown-plugin-dts`, and other packages across the monorepo — even though `packages/sdk` and `packages/react-sdk` do not declare `typescript@6` in their own devDeps. Two TypeScript versions now coexist in the lock (`6.0.2` for build tooling peers, `5.9.3` for typedoc and semantic-release). This is a supported pnpm pattern but deserves an explicit acknowledgement that the SDK packages have been validated under TypeScript 6 type checking.

2. **TypeScript 6 strictness.** Notable TypeScript 6 changes include `noUncheckedSideEffectImports` (potential errors on side-effect-only imports), tightened `override` and `exactOptionalPropertyTypes` enforcement, and removal of `--target ES3` / `--target ES5`. If the SDK or its test harnesses rely on any of the changed behaviors, this bump could silently break type checking in CI or emit new type errors downstream for consumers who use `typescript@6`.

**Recommendation:** Confirm that `tsc --noEmit` (or the equivalent build step) passes cleanly across all packages after this bump, and that no TypeScript 6 type errors are introduced. If that validation was part of CI on this branch, note it in the PR description. If not, add it.

---

## Low

### L1 — `forge soldeer install` CI step omits `GIT_LFS_SKIP_SMUDGE=1` present in `postinstall`

**File:** `.github/workflows/playwright.yml`

The new CI step:

```yaml
- name: Install soldeer dependencies
  if: steps.forge-build-cache.outputs.cache-hit != 'true'
  run: cd contracts/lib/forge-fhevm && forge soldeer install
```

The existing root `postinstall` script runs:

```sh
cd contracts/lib/forge-fhevm && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install
```

The `GIT_LFS_SKIP_SMUDGE=1` flag suppresses Git LFS pointer downloads during `git clone`/`checkout` operations triggered by soldeer. Without it, soldeer may attempt LFS downloads that either fail (if LFS is not configured on the CI runner) or succeed but pull down large binary artifacts unnecessarily (if LFS is configured).

If the CI runner for Playwright jobs has LFS enabled (as it might, given other jobs in the same workflow), this inconsistency could cause the soldeer install to pull LFS objects, increasing job time, or fail entirely if LFS credentials are absent.

**Fix:** Align the CI step with the postinstall script:

```yaml
run: cd contracts/lib/forge-fhevm && GIT_LFS_SKIP_SMUDGE=1 forge soldeer install
```

---

### L2 — `socket.io-parser` commit is a no-op on the resolved version

**Commit:** `8ed1e983 chore(deps): bump socket.io-parser in /examples/react-wagmi`

The resolved version of `socket.io-parser` in `examples/react-wagmi/package-lock.json` is `4.2.6` both before and after this commit. The `"socket.io-parser": ">=4.2.6"` override was already in place and already satisfying the advisory requirement. The Dependabot PR likely detected the range `>=4.2.6` as satisfying but regenerated the lockfile anyway without actually changing any version.

This is harmless, but the commit message implies a version change that did not occur. Anyone auditing the dependency history will need to investigate only to find no actual change.

---

### L3 — `@tanstack/react-query` version divergence between library and test harnesses

After this bump, two versions of `@tanstack/react-query` coexist in the monorepo:

- `5.95.2` — `packages/test-nextjs`, `packages/test-vite` (bumped in this PR)
- `5.90.21` — `packages/react-sdk` (unchanged; its specifier `>=5` was satisfied by the previously installed version)

The test harnesses now test against a newer minor version than the one declared by the library itself. The gap between `5.90.21` and `5.95.2` spans five minor releases; it is unlikely to surface behavioral differences in practice, but the divergence means the e2e tests and the SDK's own `react-query` peer resolution are no longer aligned. If `packages/react-sdk` is eventually upgraded to declare `^5.95.x`, this divergence will close naturally.

---

## What Is Correct

- All other bumps (`next 16.1.6→16.1.7`, `next 16.1.7→16.2.1`, `h3 1.15.6→1.15.9`, `typedoc-plugin-markdown 4.10→4.11`, `oxfmt 0.37→0.41`) are minor or patch updates with no breaking change risk.
- The `pnpm-lock.yaml` regeneration is consistent with the manifest changes; all affected importers reflect the correct resolved versions.
- The `examples/example-hoodi/package.json` constraint change (`^16.0.0` → `^16.1.7`) correctly tightens the minimum and aligns with the actual resolved version.
- Grouping Dependabot PRs into a single integration branch is a sound maintenance pattern that keeps the changelog readable and avoids churn from individual lock-file-only merges.
- The soldeer fix (excluding it from the cache-hit path) mirrors the same pattern used by the forge build step immediately following it, which is correct.

pnpm-workspace.yaml
```diff
@@ -8,3 +8,6 @@ packages:
 minimumReleaseAge: 4320
 minimumReleaseAgeExclude:
   - "@zama-fhe/relayer-sdk"
+  - "@tanstack/react-query"
```
ankurdotb, line 11: Unless there's a VERY solid justification for why the minimum release age check should be skipped for a specific package, we should not add overrides.

For relayer-sdk, this makes sense since we might need to do a coordinated release sometimes in step with them. I'm not convinced of the need to bypass these checks for these three packages.

ankurdotb (changes requested):
This PR might get superseded by #142 ?

PR #161, opened 2026-03-24
feat(react-sdk): allow undefined tokenAddress in useConfidentialIsApproved

ghermet:
## Summary
- `useConfidentialIsApproved` now accepts `tokenAddress: Address | undefined` — the query is automatically disabled (via `skipToken`) when tokenAddress is undefined, instead of throwing `"Address undefined is invalid"`.
- This supports use cases like swap UIs where the token is selected dynamically and may initially be `undefined`.
- Added tests for `tokenAddress: undefined` and the `undefined → defined` transition.

**Note on `until` data:** The issue also requested exposing the operator expiry timestamp. The on-chain `isOperator()` function only returns `bool` — there is no public getter for the raw `uint48 until` value stored in `_operators`. Exposing this would require a contract-level change (adding a view function like `operatorExpiration(holder, spender)`). This can be tracked separately.

Closes SDK-54

## Test plan
- [x] Existing tests pass (8/8 in `use-confidential-is-approved.test.tsx`)
- [x] New test: query disabled when `tokenAddress` is `undefined`
- [x] New test: query activates when `tokenAddress` transitions from `undefined` to a valid address
- [ ] CI passes

### DIFF
--- a/packages/react-sdk/etc/react-sdk.api.md
+++ b/packages/react-sdk/etc/react-sdk.api.md
@@ -1122,9 +1122,10 @@ export type UseConfidentialBalancesOptions = Omit<UseQueryOptions<ConfidentialBa
 export function useConfidentialIsApproved(config: UseConfidentialIsApprovedConfig, options?: Omit<UseQueryOptions<boolean>, "queryKey" | "queryFn">): _tanstack_react_query0.UseQueryResult<unknown, Error>;
 
 // @public
-export interface UseConfidentialIsApprovedConfig extends UseZamaConfig {
+export interface UseConfidentialIsApprovedConfig {
     holder?: Address;
     spender: Address | undefined;
+    tokenAddress: Address | undefined;
 }
 
 // @public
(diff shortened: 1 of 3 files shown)


enitrat (changes requested):
the proper fix is to move the gating to the query options factory and not in the hook directly, because the hook layer should be a minimal wrapper that delegates to the query factory in which the logic is defined.
If we get the gating via the `enabled` flag in the query factories, any consumer gets
the correct behavior without a duplicated the guard.

The issue reported should also be applied to other queries that suffer from the same problem, which also highlights a deeper architectural issue: `ReadonlyToken` requires a concrete address at construction and is a stateful class, but queries should accept `undefined` values that disable their execution. We can't really make the query factory dependent on `ReadonlyToken` which requires explicit tokenAddress

The design we should converge to is:

- shared core functions (like delegation reads, approval reads, etc) 
- query factory that calls these shared core functions; and can accept "undefined" inputs
- ReadonlyToken instance that requires concrete, defined inputs, and delegates to the same core functions
- the react hook are just thin wrappers around the factories

This will require a deeper refactor for stateful functions, like `confidentialBalanceOf`, etc. So for now; let's focus on the stateless functions like and go through direct contract reads in the queries. We can put in a backlog decoupling the ReadonlyToken instance from its inner functions, so that we can work on both stateful classes and stateless functions.

see concrete proposal: superceded by https://github.com/zama-ai/sdk/pull/163

ghermet:
Solved by #163

PR #163, opened 2026-03-24
fix: gate queries on undefined tokenAddress

enitrat:
## Summary
- Query options factories (`confidentialIsApproved`, `delegationStatus`, `wrapperDiscovery`) now accept `tokenAddress: Address | undefined` and auto-disable via `enabled: false` when it's missing — instead of crashing with `Address "undefined" is invalid`.
- React hooks simplified: removed `skipToken` branching in favor of letting the factory handle gating, following the TanStack Query auto-gating pattern.
- `delegationStatus` factory decoupled from `ReadonlyToken` — now takes `(signer, relayer, tokenAddress)` primitives, since it only does plain contract reads (no FHE decryption).

Closes SDK-54

## Motivation
In dynamic UIs (e.g. swap token selector), `tokenAddress` starts as `undefined`. Previously this crashed because `getAddress(undefined)` throws. The fix pushes the `undefined` handling into the Layer 2 query options factories, so any consumer gets correct auto-gating behavior without manual guards.


ghermet (dismissed):

ghermet:
Solves #161

ghermet (approved):

PR #166, opened 2026-03-25
build: upgrade typescript from 5.9 to 6.0

ghermet:
## Summary
- Upgrade TypeScript from `^5.9.3` to `^6.0.2` across root, `test-nextjs`, and `test-vite`
- Set `noUncheckedSideEffectImports: false` in `test-nextjs/tsconfig.json` (TS6 defaults this to `true`)

## Test plan
- [ ] CI passes (build, lint, type-check, tests)
- [ ] Verify `test-nextjs` and `test-vite` apps still compile and run
- [ ] Check for any new type errors introduced by TS6 stricter defaults


pnpm-workspace.yaml
```diff
@@ -9,3 +9,4 @@ packages:
 minimumReleaseAge: 4320
 minimumReleaseAgeExclude:
   - "@zama-fhe/relayer-sdk"
+  - "typescript"
```
enitrat: can we wait 3 days?

ghermet: this isnt' even needed can be on 6.0.1

enitrat: yes let's use 6.0.1 then

ghermet: tomorrow will be 3 days

test/test-nextjs/tsconfig.json
```diff
@@ -6,6 +6,7 @@
         "name": "next"
       }
     ],
+    "noUncheckedSideEffectImports": false,
```
ankurdotb: Is this needed in the `test-vite` tsconfig as well? Or in any other tsconfigs and/or in `tsconfig.base.json`?

ghermet: Actually we don't even need it for nextjs, we'll have a strict ts config for all projects

ankurdotb (approved):

PR #168, opened 2026-03-25
feat(react-sdk): improve query wrappers types

ghermet:
## Summary
- Replace `any`-typed `options` parameter in `useQuery`/`useSuspenseQuery` wrappers with `AnyKeyQueryOptions<TData, TError>` — erases the QueryKey param via `any` in the type alias so callers don't need to spell out specific tuple keys, while keeping full type safety on `TData` and `TError`
- Type `useQueries` wrapper with `QueriesOptions<T>` / `QueriesResults<T>` for deterministic API reports (the untyped return previously caused non-deterministic union ordering in api-extractor)
- Fix `Enabled<bigint>` leaking into `Enabled<Address>` queries in `useConfidentialBalance`/`useConfidentialBalances` by extracting `userEnabled` as `boolean | undefined`
- Split `useIsAllowed` into early-return branches to avoid union type between `skipToken` and factory options
- Add explicit `<boolean>` type param to `useConfidentialIsApproved` and `useIsAllowed` for correct `UseQueryResult` inference

## Test plan
- [x] `pnpm typecheck` passes (0 errors in SDK packages)
- [x] `pnpm test:run` passes (1370 tests, 145 files)
- [x] `pnpm build && pnpm api-report:check` passes (7/7 checks, deterministic)
- [ ] CI passes on this PR

### DIFF
--- a/packages/react-sdk/src/token/use-confidential-balance.ts
+++ b/packages/react-sdk/src/token/use-confidential-balance.ts
@@ -42,7 +42,7 @@ export function useConfidentialBalance(
   options?: UseConfidentialBalanceOptions,
 ) {
   const { tokenAddress, handleRefetchInterval } = config;
-  const userEnabled = options?.enabled;
+  const userEnabled = typeof options?.enabled === "function" ? undefined : options?.enabled;
   const token = useReadonlyToken(tokenAddress);
 
   const addressQuery = useQuery<Address>({
@@ -58,7 +58,7 @@ export function useConfidentialBalance(
   });
   const handleQuery = useQuery<Handle>({
     ...baseHandleQueryOptions,
-    enabled: (baseHandleQueryOptions.enabled ?? true) && (userEnabled ?? true),
+    enabled: Boolean(baseHandleQueryOptions.enabled ?? true) && (userEnabled ?? true),
   });
 
   // Phase 2: Decrypt only when handle changes (expensive relayer roundtrip)
@@ -70,7 +70,7 @@ export function useConfidentialBalance(
   const balanceQuery = useQuery<bigint>({
     ...baseBalanceQueryOptions,
     ...options,
-    enabled: (baseBalanceQueryOptions.enabled ?? true) && (userEnabled ?? true),
+    enabled: Boolean(baseBalanceQueryOptions.enabled ?? true) && (userEnabled ?? true),
   });
 
   return { ...balanceQuery, handleQuery };
--- a/packages/react-sdk/src/token/use-confidential-balance.ts
+++ b/packages/react-sdk/src/token/use-confidential-balance.ts
@@ -19,7 +19,13 @@ export interface UseConfidentialBalanceConfig {
 }
 
 /** Query options for the decrypt phase of {@link useConfidentialBalance}. */
-export type UseConfidentialBalanceOptions = Omit<UseQueryOptions<bigint>, "queryKey" | "queryFn">;
+export interface UseConfidentialBalanceOptions extends Omit<
+  UseQueryOptions<bigint>,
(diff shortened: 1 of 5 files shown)


packages/react-sdk/src/token/use-confidential-balance.ts
```diff
@@ -42,7 +42,7 @@ export function useConfidentialBalance(
   options?: UseConfidentialBalanceOptions,
 ) {
   const { tokenAddress, handleRefetchInterval } = config;
-  const userEnabled = options?.enabled;
+  const userEnabled = typeof options?.enabled === "function" ? undefined : options?.enabled;
```
enitrat: this doesn't look like a correct change: what is the purpose here? userEnabled should be either `true or false`, or a function that returns a boolean, not undefined

here if the user passes a function you bypass it by overwriting with undefined. we need to preserve compatibility with `enabled` passed being functions

packages/react-sdk/src/token/use-confidential-balances.ts
```diff
@@ -58,7 +58,7 @@ export function useConfidentialBalances(
   options?: UseConfidentialBalancesOptions,
 ) {
   const { tokenAddresses, handleRefetchInterval, maxConcurrency } = config;
-  const userEnabled = options?.enabled;
+  const userEnabled = typeof options?.enabled === "function" ? undefined : options?.enabled;
```
enitrat: same comment

enitrat (changes requested):
The type narrowing is good, but it seems that it surface a deeper issue about our approach to composed hooks.

enitrat (commented):
The proper fix here, I think is to narrow the type permissions we have:
- on the query factories, none of our queries use a callback functions for `enabled`, so we could force them to always resolve to booleans
- on the user-supplied flag, it's a bit more complex because:
  - the callback function is typed to the original query; which means you can't pass it to a sub-query that expects another type (e.g. `UseConfidentialBalanceOptions` is typed with a `bigint` but `confidentialHandleQueryOptions` with a `Handle`, which means the callback functions are structurally incompatible
  - we could however consider that this callback form is meaningless when applies to composite queries, and restrict the Options to `enabled?: boolean`.

I think, considering how those hooks are meant to be used, that we should do:
- force factory `enabled` to be `boolean`, and 
- force composite queries to take a `boolean` enabled and not a callback function

packages/react-sdk/src/token/use-is-allowed.ts
```diff
@@ -25,16 +25,17 @@ export function useIsAllowed() {
...
-        queryFn: skipToken,
-      };
-  const factoryEnabled = "enabled" in baseOpts ? (baseOpts.enabled ?? true) : true;
+  if (!account) {
+    return useQuery<boolean>({
+      queryKey: zamaQueryKeys.isAllowed.all,
+      queryFn: skipToken,
+      enabled: false,
+    });
+  }
```
enitrat: react hooks should not be called conditionally; but be called in the same order at each render. what you should do instead is something along the lines of

```
  const baseOpts = account
    ? isAllowedQueryOptions(sdk, { account })
    : ({
        queryKey: zamaQueryKeys.isAllowed.all,
        queryFn: skipToken,
        enabled: false,
      } as const);
```
this way you avoid the conditional call (even though the early return ensures there's no other hook call later on, it's good to keep this in mind)

enitrat (changes requested):

packages/react-sdk/src/token/use-is-allowed.ts
ghermet: Yeah good point

enitrat (approved):

PR #169, opened 2026-03-25
feat(example-node-viem): upgrade to SDK 2.x with full ERC-7984 flows

zama-cremaud:
Mirrors the node-ethers example structure, using viem primitives:
- Two accounts (A + B) via privateKeyToAccount + separate walletClients
- ViemSigner from @zama-fhe/sdk/viem (shared publicClient, separate walletClients)
- Section 1: Setup — RelayerNode + two ZamaSDK instances
- Section 2: Mint — writeContract / waitForTransactionReceipt / readContract
- Section 3: Confidential lifecycle — shield, confidentialTransfer, unshield with callbacks
- Section 4: Delegation — delegateDecryption, decryptBalanceAs, revokeDelegation, isDelegated
- Standalone NodeNext tsconfig (no workspace extends, avoids browser-only path aliases)
- Simplified .env.example: PRIVATE_KEY, DELEGATE_PRIVATE_KEY, SEPOLIA_RPC_URL
- README rewritten with Prerequisites, Setup, Run, and per-section description

# Summary

## Scope

- [ ] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] CI/workflows
- [ ] Docs


ghermet (commented):
# PR Review Summary

## Critical Issues (2)

### 1. `createToken` arguments appear swapped in both examples
**Files**: `examples/node-viem/src/index.ts`, `examples/node-ethers/src/index.ts`

Both examples call:
```ts
sdkA.createToken(TOKEN_ADDRESS, confidentialTokenAddress)
```
where `TOKEN_ADDRESS` is the plain ERC-20. But `ZamaSDK.createToken(address, wrapper?)` documents `address` as "The confidential token contract address", and internally `this.address` is used for `confidentialBalanceOf`, `credentials.allow`, and `supportsInterface` — all of which require a confidential token address.

**Either** the examples have the arguments swapped, **or** the JSDoc + internal usage are wrong. This needs clarification before merge.

### 2. `Promise.allSettled` metadata fallback produces fake data with `decimals: 0`
**File**: `packages/sdk/src/wrappers-registry.ts` (`listPairs` method)

When metadata fetching fails for individual pairs, the code fills in:
```ts
{ name: "Unknown", symbol: "???", decimals: 0, totalSupply: 0n }
```

- `decimals: 0` is **actively dangerous** — downstream formatting code would display `1000000` instead of `1.0 USDT`.
- The `metadataFailed` flag is not part of the `TokenWrapperPairWithMetadata` TypeScript interface, so consumers can't reliably type-check it.
- No logging of which pairs failed or why.

**Suggestions**: Add `metadataFailed` to the TS interface, use `undefined` instead of `0` for decimals, log individual failures.

---

## Important Issues (4)

### 3. Empty catch blocks in ZamaSDK identity/lifecycle
**File**: `packages/sdk/src/zama-sdk.ts`

`#initIdentity`, `onAccountChange`, and `onChainChange` all have empty catch blocks that swallow **all** errors (not just "signer not ready"). If `getChainId()` fails during `onAccountChange`, the SDK retains a stale chain ID — subsequent operations use the wrong chain, and credential revocation may silently no-op.

**Recommendation**: At minimum `console.warn` + emit an event, consistent with patterns elsewhere in the codebase.

### 4. `persistCredentials` silently swallows storage write failures
**File**: `packages/sdk/src/credentials/credentials-manager-base.ts`

Storage quota exceeded, IndexedDB corruption, etc. are caught and reduced to a warning. The caller proceeds as if persistence succeeded. On page reload, credentials are gone — user gets re-signing prompts with no explanation.

### 5. WALKTHROUGH.md contains factual inaccuracies
**File**: `examples/node-viem/WALKTHROUGH.md`

- Claims `createToken(address)` takes a single argument — it takes two.
- Claims "Both addresses are defined as constants in `src/index.ts`" — they come from env vars and registry lookup.
- Claims ViemSigner routes `getBalance` to publicClient — no such method exists on `GenericSigner` or `ViemSigner`.

### 6. `registryTTL` is not validated
**File**: `packages/sdk/src/wrappers-registry.ts`

`keypairTTL` throws on `<= 0`, but `registryTTL` is not validated. A negative or zero TTL silently produces a broken cache.

---

## Suggestions (8)

7. **Derive `this.name` from `new.target.name` in `ZamaError`** (`errors/base.ts`) — eliminates manual `this.name = "FooError"` in all 15 subclasses. Zero risk, prevents copy-paste bugs.
8. **Add `matchZamaError` overload** — when a `_` wildcard is provided, return type should be `R` not `R | undefined`.
9. **`toError` discards diagnostic context** (`utils/error.ts`) — non-Error objects lose all properties. Fix: preserve original as `cause`.
10. **`TransactionResult` fields should be `readonly`** — inconsistent with `TransactionReceipt` which is fully readonly.
11. **`GenericSigner` mixes method and property syntax** (`types/signer.ts`) — inconsistent, affects `this`-binding if methods are destructured.
12. **Add `destroy()` method to `ZamaSDK`** — no way to clean up the signer subscription, causing leaks in long-lived apps.
13. **`createToken` JSDoc is ambiguous** (`zama-sdk.ts`) — says `address` is "the confidential token contract address" but examples pass the ERC-20 address. Related to Critical #1.
14. **Overly broad catch in `resolveCredentials`** (`credentials-manager-base.ts`) — non-corruption errors (e.g. `ConfigurationError`) are treated as "corruption" and trigger unnecessary credential deletion.

---

## Strengths

- Error system design is solid: `ZamaErrorCode` const-object + type-union, `wrapSigningError` with EIP-1193 detection, comprehensive `*.test-d.ts` type-level tests.
- WrappersRegistry caching is well-implemented: negative cache TTL for misses, proper pagination clamping, `refresh()` escape hatch.
- Node-viem example is well-structured: proper env validation, `finally` cleanup, `process.exit(1)` on fatal, good callback usage.
- Contract type system (`ReadContractConfig`/`WriteContractConfig`) with ABI inference is excellent.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

test/playwright/playwright.config.ts
```diff
@@ -50,15 +50,15 @@ export default defineConfig<{}, WorkerFixtures>({
       wait: {
         stdout: /Anvil ready on port (\d+)/,
       },
-      timeout: 60_000,
+      timeout: 120_000,
```
ghermet: why do we need 120_000 for anvil ?

ghermet (approved):

PR #171, opened 2026-03-25
chore: add react/rules-of-hooks lint

enitrat:
Adds `react/rules-of-hooks` to lint hook rules, like no conditional rendering of hooks, etc.

Renamed the `use` from vitest fixtures to avoid false positive triggers.

### DIFF
--- a/packages/react-sdk/src/__tests__/wagmi-signer-subscribe.test.ts
+++ b/packages/react-sdk/src/__tests__/wagmi-signer-subscribe.test.ts
@@ -34,10 +34,10 @@ interface WagmiFixtures {
 
 const wit = base.extend<WagmiFixtures>({
   // eslint-disable-next-line no-empty-pattern
-  wagmiSigner: async ({}, provideFixture) => {
+  wagmiSigner: async ({}, use) => {
     capturedOnChange = undefined;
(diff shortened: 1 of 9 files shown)


packages/react-sdk/src/__tests__/wagmi-signer-subscribe.test.ts
```diff
@@ -34,10 +34,10 @@ interface WagmiFixtures {
 
 const wit = base.extend<WagmiFixtures>({
   // eslint-disable-next-line no-empty-pattern
-  wagmiSigner: async ({}, use) => {
+  wagmiSigner: async ({}, provideFixture) => {
```
ghermet: Can't we just disable the rule for this file?

ghermet: `use` is more concise and the default name for this parameter

ghermet:
Good addition but we can use safely disable the rule for the fixture file

enitrat:
hah, it's not explicit in the [docs ](https://vitest.dev/guide/test-context.html#extend) so I thought it was some name we chose. will revert

ghermet (approved):

PR #172, opened 2026-03-25
chore: enable type awareness in oxlint

enitrat:
enables typeAware linting in oxlint


test/playwright/fixtures/relayer-sdk.js
enitrat, line 1: was unused

packages/sdk/src/types/__tests__/storage.test-d.ts
```diff
@@ -3,7 +3,7 @@ import type { GenericStorage } from "../storage";
 
 describe("GenericStorage", () => {
   test("get returns Promise<T | null>", () => {
-    expectTypeOf<GenericStorage["get"]>().returns.toEqualTypeOf<Promise<unknown | null>>();
+    expectTypeOf<GenericStorage["get"]>().returns.toEqualTypeOf<Promise<unknown>>();
```
enitrat, line 6: unknown swallows null

.github/workflows/vitest.yml
```diff
@@ -34,14 +34,14 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile --ignore-scripts
 
-      - run: pnpm typecheck
+      - name: Build packages
+        run: pnpm build
```
enitrat, line 38: requirement because the types used in the react-sdk actually come from the built `sdk`: oxlint resolves the workspace package types through the `dist` declarations for test files that are outside of the tsconfig surface area.

the other solution would be to include them in the base tsconfig; but this has other effects on running coverage test.

third option was a tsconfig for lints only but it started being a lot; so that's what I think we could settle on 🤔

ghermet: build is really fast I'm not worried about it here

ghermet (approved):

PR #173, opened 2026-03-25
chore: introduce ast-grep based rules to block bad patterns

enitrat:
Adds ast-grep + a rule to block using the tanstack base `useQueries` rather than our wrapper.


.github/workflows/vitest.yml
```diff
@@ -31,6 +31,9 @@ jobs:
           node-version: 22
           cache: pnpm
 
+      - name: Setup ast-grep
+        uses: ast-grep/action@cf62e780f0c88301228978d593a7784427a097a6
+
```
ghermet: should we make a separate workflow with all this linting stuff?
typecheck, linting, formatting, ast ? would make vitest workflow strictly test related just like playwright workflow

ankurdotb:
### Code review

No high-confidence issues (all scored below threshold). Below are lower-confidence observations for consideration:

---

**1. Double ast-grep scan in CI** (confidence: 75/100)

`ast-grep/action` both installs the CLI and runs a full scan with `--format github`. Then `pnpm lint` runs a second scan. If the intent is just installation, replace with a plain install step; if the intent is PR annotations, consider removing the duplicate from `pnpm lint` in CI.

https://github.com/zama-ai/sdk/blob/f4bbd2e2c6664a3bfbc0475d3c78623ce0bf0b82/.github/workflows/vitest.yml#L33-L36

---

**2. `ast-grep` not in `devDependencies`** (confidence: 75/100)

`oxlint` and `oxfmt` are in `devDependencies`, but `ast-grep` is a global-only prerequisite. `pnpm install` alone won't give a working `pnpm lint`. Adding `@ast-grep/cli` to `devDependencies` would match the existing pattern.

https://github.com/zama-ai/sdk/blob/f4bbd2e2c6664a3bfbc0475d3c78623ce0bf0b82/package.json#L24-L25

---

**3. `lint-staged` does not run `ast-grep`** (confidence: 75/100)

`oxlint` runs pre-commit via `lint-staged`, but `ast-grep` does not. Violations are only caught in CI. `ast-grep scan` accepts file path arguments, so it could be added to `lint-staged` for parity.

https://github.com/zama-ai/sdk/blob/f4bbd2e2c6664a3bfbc0475d3c78623ce0bf0b82/package.json#L98-L101

---

**4. Missing version comment on pinned action SHA** (confidence: 50/100)

Every other pinned action in the repo follows `uses: owner/action@<sha> # vX.Y.Z`. The new `ast-grep/action` line is missing the version comment.

https://github.com/zama-ai/sdk/blob/f4bbd2e2c6664a3bfbc0475d3c78623ce0bf0b82/.github/workflows/vitest.yml#L34-L35

---

**5. Relative path in rule message** (confidence: 25/100)

The `message` suggests `"../utils/query"` which is only correct for files one level under `src/`. The `note` field already has the correct absolute path, so low impact.

https://github.com/zama-ai/sdk/blob/f4bbd2e2c6664a3bfbc0475d3c78623ce0bf0b82/tools/ast-grep/rules/no-direct-tanstack-useQueries.yml#L5-L7

---

ankurdotb (changes requested):
Was running through my questions which got posted as a separate review but in short:

- I do agree with adding `ast-grep` as a devDep is the easiest way to include it's bundled, and then can be invoked as a script. Mentioning the `brew` method to install in docs as a separate thing is fine.
- Also agree on lint-staged
- Dumb question is there an ast-grep equivalent of `--fix`? If so, then adding something like that make sense, similar to the `lint:fix`? It then makes some of the fixing more obvious. On lint failure, the `--fix` is my first port of call.
- I'm fine with keeping in same workflow or splitting into a `lint.yml` workflow. 👍🏽  (I don't particularly care.) If you drop the ast-grep action and execute via the `pnpm lint` step with ast-grep added as a devDep, this comment sort of resolves itself as it's then the same as what we currently have.

ankurdotb (approved):

PR #176, opened 2026-03-27
refactor(sdk): remove allow, revoke, isAllowed from ZamaSDK

ghermet:
## Summary
- Removed `allow()`, `revoke()`, and `isAllowed()` convenience methods from `ZamaSDK`
- These should only be accessed via `sdk.credentials.allow()`, `sdk.credentials.revoke()`, `sdk.credentials.isAllowed()`
- Updated query factories and tests to use `sdk.credentials` directly

## Test plan
- [x] TypeScript compiles with no errors
- [x] All 29 affected tests pass


zama-cremaud:
## Issues

### [HIGH] `allowMutationOptions` does not await `credentials.allow()` — fire-and-forget

**File:** `packages/sdk/src/query/allow.ts`

```ts
// Current (branch)
mutationFn: async (contractAddresses) => void sdk.credentials.allow(...contractAddresses),
```

`credentials.allow()` returns `Promise<StoredCredentials>` (not `Promise<void>`). The `void` operator discards that Promise synchronously without awaiting it. Because the arrow function is `async`, it returns a `Promise<undefined>` that resolves immediately — before the actual credential operation has completed or settled. The result:

1. **Errors are silently swallowed.** `credentials.allow()` can throw `SigningRejectedError` (user rejects the wallet prompt) or `SigningFailedError`. With fire-and-forget, these never reach TanStack Query's error state — the mutation always shows `"success"` regardless of outcome.

2. **Query invalidation fires too early.** `useAllow`'s `onSuccess` calls `invalidateQueries({ queryKey: zamaQueryKeys.isAllowed.all })` when the mutation resolves. With this change, the invalidation fires before `credentials.allow()` has completed, so `useIsAllowed`'s refetch may return `false` even after the mutation shows success — a race condition visible in the UI.

The `as never` cast in the updated test (`mockResolvedValue(undefined as never)`) is a symptom of the same type mismatch.

The commit message (`fix(sdk,react-sdk): keep allowMutationOptions return type as void`) indicates the intent was to adapt the return type, but the chosen approach is incorrect. The correct fix is to await the call and let the return value be implicitly discarded:

```ts
// Correct
mutationFn: async (contractAddresses) => {
  await sdk.credentials.allow(...contractAddresses);
},
```

This preserves `Promise<void>` as the resolved type of the mutation function while correctly propagating errors and resolving only after the operation completes. The `as never` cast in the test can then be replaced with a plain `mockResolvedValue({} as StoredCredentials)` or similar.

For context: the old `ZamaSDK.allow()` wrapper did exactly this — `await this.credentials.allow(...)` — so removing the wrapper and going to `sdk.credentials.allow()` directly is correct; the implementation of the factory just needs to mirror that same `await`.

ankurdotb:
### Code review

~~The fire-and-forget issue flagged by @zama-cremaud is fixed (commit fd837f5f). Found 4 remaining issues:~~

~~1. E2E test calls removed `sdk.allow()` — will fail at compile and runtime after merge.~~
~~2. `[Symbol.dispose]` JSDoc example still references removed `sdk.allow(cUSDT)`.~~
~~3. Gitbook reference doc still has full `### allow`, `### revoke`, and `### isAllowed` sections documenting the removed top-level methods.~~
~~4. Gitbook concepts guide references removed methods (`sdk.revoke(...)` and `sdk.allow(...)`).~~

**Re-review (dbe7114d):** All 4 issues resolved in 81ccf4c3 and dbe7114d. No remaining issues found.

ankurdotb (approved):

PR #177, opened 2026-03-27
fix(sdk): add DelegationNotPropagatedError for gateway sync delay

ghermet:
## Summary

- Adds `DelegationNotPropagatedError` — thrown when a delegated decrypt hits a relayer 500, which is the expected behavior during the 1–2 min L1→Arbitrum gateway ACL sync window after granting delegation
- `wrapDecryptError` in `decryptBalanceAs` now detects HTTP 500 in delegated context and produces an actionable error message instead of a generic `RelayerRequestFailedError`
- Documents the propagation delay in JSDoc (`delegateDecryption`, `decryptBalanceAs`) and in `docs/gitbook/src/reference/sdk/delegation.md`

## Context

After calling `delegateForUserDecryption()`, the delegation is recorded on L1 immediately. However, the gateway's `isHandleDelegatedForUserDecryption()` check runs against its own MultichainACL (deployed on Arbitrum), which must sync via cross-chain event propagation. This takes ~1–2 minutes, during which the relayer returns a 500. Previously this surfaced as a generic error with no guidance.

## Test plan

- [x] New unit tests for `DelegationNotPropagatedError` (instanceof, code, name)
- [x] Type test verifying `DelegationNotPropagatedError extends ZamaError`
- [x] Existing delegation and batch-decrypt tests still pass
- [ ] Manual E2E: grant delegation → immediate `decryptBalanceAs` → verify `DelegationNotPropagatedError` is thrown → wait 2 min → verify decrypt succeeds


ankurdotb (commented):
## PR Review

### Strengths

- **Directly addresses a real user pain point** — the error message is actionable and tells the developer exactly what to do
- **Consistent with existing error patterns** — follows the same `ZamaError` subclass + `ZamaErrorCode` enum convention as all other delegation errors
- **Documentation is thorough** — JSDoc on `delegateDecryption`, `decryptBalanceAs`, the GitBook delegation page, and the error table are all updated
- **Minimal, focused change** — only touches what's needed, no scope creep
- **API report files updated** — `sdk.api.md` and `sdk-query.api.md` reflect the new export

---

### Important (should fix before merge)

**1. HTTP 500 assumption is fragile** — `packages/sdk/src/token/readonly-token.ts:1024`

`wrapDecryptError` assumes that *any* HTTP 500 in a delegated context means "delegation not propagated." A 500 could also mean the relayer/gateway genuinely crashed, is overloaded, or hit an unrelated internal bug. This heuristic will produce a misleading error message in those cases, telling the user to "wait 1–2 minutes" when the actual problem is something else entirely.

**Suggestion:** Consider whether the relayer error body contains any signal that distinguishes a delegation-ACL-not-found from a generic server error. If not, the error message should hedge — e.g., *"This is most commonly caused by the delegation not having propagated yet"* rather than stating it as the only possibility. The JSDoc comment on `wrapDecryptError` already says "most likely cause" which is good, but the user-facing error message reads more definitively.

Alternatively, document in the error class JSDoc that this is a best-effort heuristic and may fire for non-propagation 500s.

**2. No integration-level test for the `wrapDecryptError` → `DelegationNotPropagatedError` path**

The existing `error-throwing.test.ts` tests `wrapDecryptError` behavior for `balanceOf()` (non-delegated), including the `statusCode: 500 → RelayerRequestFailedError` case. But there's no corresponding test that exercises the `decryptBalanceAs()` path where `isDelegated = true` and `statusCode === 500` should produce `DelegationNotPropagatedError`. The unit tests in `errors.test.ts` only verify the class constructor, not the actual wrapping logic.

**Suggestion:** Add a test case in `error-throwing.test.ts` (or `delegation.test.ts`) that mocks `relayer.userDecrypt` to reject with a `statusCode: 500` error, then calls `decryptBalanceAs()` and asserts `DelegationNotPropagatedError` is thrown.

---

### Suggestion (nice to have)

**3. Error message could suggest a retry strategy**

The error message says "allow 1–2 minutes" but doesn't mention that the SDK user could implement a polling retry. Since this is a *transient* error by nature, it might be worth noting in the docs error-handling section that a `setTimeout` + retry is the expected pattern.

---

### Summary

| Category | Count |
|----------|-------|
| Critical | 0 |
| Important | 2 |
| Suggestions | 1 |

Overall this is a clean, well-scoped PR that solves a real UX problem. The two important items are about robustness and test coverage rather than correctness — the happy path works as designed.

ankurdotb (approved):

PR #178, opened 2026-03-27
feat: Node+Web SDK E2E test infrastructure

ghermet:
## Summary

- Add comprehensive Node.js E2E test suite running against a local anvil chain
- Replace `RelayerCleartext` with `RelayerWeb` + mock relayer HTTP server in browser E2E tests, exercising the real Web Worker + CDN bundle code path
- Split Playwright CI into parallel Next.js, Vite, and Node jobs
- Add `Disposable` support on `RelayerSDK` and `ZamaSDK`
- Move relayer server launch from Playwright `webServer` config to the test fixture
- Add JSDoc to mock relayer server constants, helpers, and route handlers

## Test plan

- [x] Node E2E tests pass (`pnpm playwright test --config playwright.node.config.ts`)
- [x] Next.js browser E2E tests pass (`pnpm playwright test --config playwright.nextjs.config.ts`)
- [x] Vite browser E2E tests pass (`pnpm playwright test --config playwright.vite.config.ts`)
- [x] CI jobs run in parallel without interference
- [x] CodeQL CORS credential-leak alert is resolved


zama-cremaud:
Small feedback but nothing blocker from my side :

## Issues

### [MEDIUM] `isConfidentialQueryOptions` / `isWrapperQueryOptions` catch ALL errors, not just contract reverts

**File:** `packages/sdk/src/query/is-confidential.ts`

Both query functions now wrap `signer.readContract(...)` in a broad `try/catch` and return `false` on any thrown error. The intent is to handle contracts that don't implement ERC-165 and revert on `supportsInterface`. That is correct. The problem is that the catch also silently converts network errors (connection refused, timeout, malformed response) into `data: false`, which is then cached forever because both queries have `staleTime: Infinity`.

In TanStack Query v5, a `queryFn` that **throws** enters the retry/error pipeline. A `queryFn` that **returns `false`** is treated as a successful result and cached at `staleTime: Infinity`. A transient network hiccup during the first call would therefore permanently mark any token as "not confidential" for the lifetime of the query client, with no recovery path short of a full page reload.

The fix should narrow the catch to actual on-chain reverts. In viem, these come as `ContractFunctionExecutionError`; in ethers, as errors with `.code === 'CALL_EXCEPTION'`. A minimal guard:

```ts
} catch (err) {
  // Only suppress contract execution reverts (non-ERC-165 contracts).
  // Re-throw network/transport errors so TanStack Query's retry logic applies.
  if (isContractRevertError(err)) return false;
  throw err;
}
```

The existing unit test only covers `"execution reverted"` — it does not catch the regression described above. A test for a network-level error (e.g. `new Error("fetch failed")`) would confirm the narrower catch is necessary.

---

### [LOW] Node global setup swaps a build artifact on disk

**File:** `test/playwright/fixtures/node-global-setup.ts`

The global setup renames the built SDK worker (`relayer-sdk.node-worker.js` in the resolved `@zama-fhe/sdk/node` dist directory) and replaces it with the mock. The teardown restores the backup. This is the cleanest approach available today given that `RelayerNode` hardcodes the worker path — there is no injection point to supply an alternate worker at construction time.

The risk: if the test process is terminated with `SIGKILL` (OOM kill, CI timeout, `kill -9`) rather than a normal exit, Playwright's teardown never runs. The real worker is then gone until the next `pnpm build`. The `isMockWorker` guard correctly prevents double-backups on re-run, but it cannot distinguish "previous run was killed" from "previous run succeeded and failed to restore" — in both cases the backup is intact and will be restored correctly on the next run. The real exposure is a killed run followed by a manual `pnpm e2e:test:node` invocation that *also* gets killed: at that point the `.bak` file is the mock, and the backup chain breaks.

The durable fix is to add a `workerPath?: string` option to `RelayerNodeConfig` so tests can supply the mock worker path directly, eliminating the need to mutate build artifacts entirely. Worth tracking as a follow-up SDK ticket.

zama-cremaud (dismissed):
LGTM!

zama-cremaud (dismissed):

zama-cremaud (approved):

PR #193, opened 2026-04-02
fix: improve error handling and code quality

ghermet:
## Summary

Fixes error handling issues, lint violations, and code quality problems across SDK modules.

### Error handling improvements
- **`createCredentials` / `resolveCredentials`**: re-throw `ZamaError` instances instead of wrapping them as `SigningFailedError` — network/relayer errors were being misclassified as wallet signing failures
- **`extractRevertErrorName`**: simplify 12-condition nested `if` chain to early-return style
- **`matchAclRevert`**: remove redundant `?.` after `in` guard
- **`wrapSigningError`**: cache `.toLowerCase()` instead of calling it twice
- **`registryTTL` JSDoc**: fix documentation claiming default is 30 days when actual default is 24 hours (`86400`)

### Lint fixes (no-redundant-type-constituents)
- **`ClearValueType`**: defined locally in `relayer-sdk.types.ts` as `bigint | boolean | \`0x${string}\`` instead of re-exporting from `@zama-fhe/relayer-sdk/bundle` — the external package types resolve to `any` through pnpm's non-hoisted `node_modules`, causing `ClearValueType | undefined` to trigger `no-redundant-type-constituents`
- Updated all internal imports (`query/*.ts`, `worker.types.ts`) to use the local definition
- Regenerated API reports (`sdk.api.md`, `sdk-query.api.md`, `sdk-node.api.md`)

### Worker assertion pattern
- **`relayer-sdk.worker.ts` / `relayer-sdk.node-worker.ts`**: replaced inline `if (!sdkInstance)` guards with `assertSdkInstance()` assertion function using TypeScript's `asserts` narrowing
- Web worker uses an inlined null check (avoids importing `assertNonNullable` into the worker chunk which would increase bundle size)
- Node worker uses `assertNonNullable` from utils (no bundle size concern)

### Browser extension refactor
- **`getBrowserExtensionRuntime`**: replaced exception-based control flow (`try/catch` as `continue`) with `isValidRuntime` type guard — removes 2 local assertion helpers and 2 eslint-disable comments
- Imports shared `assertFunctionProp` from `utils/assertions` instead of duplicating it

### Bundle size
- ESM bundle: 46.92 kB (limit: 50 kB)
- Bumped ESM size limit from 47 KB to 50 KB to accommodate future growth

## Test plan

- [x] All 1481 tests pass
- [x] Lint clean
- [x] Format clean
- [x] Bundle size within limits
- [ ] Verify no regressions in credential creation flow
- [ ] Verify browser extension detection in Chrome/Firefox extensions

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -2,7 +2,7 @@
   {
     "name": "@zama-fhe/sdk (ESM)",
     "path": "packages/sdk/dist/esm/**/*.js",
-    "limit": "50 KB"
+    "limit": "47 KB"
   },
   {
     "name": "@zama-fhe/sdk (CJS)",
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -2,7 +2,7 @@
   {
     "name": "@zama-fhe/sdk (ESM)",
     "path": "packages/sdk/dist/esm/**/*.js",
-    "limit": "47 KB"
+    "limit": "50 KB"
   },
   {
     "name": "@zama-fhe/sdk (CJS)",
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -2,7 +2,7 @@
   {
     "name": "@zama-fhe/sdk (ESM)",
     "path": "packages/sdk/dist/esm/**/*.js",
-    "limit": "47 KB"
+    "limit": "47.1 KB"
   },
   {
     "name": "@zama-fhe/sdk (CJS)",
--- a/packages/react-sdk/src/relayer/use-user-decrypted-values.ts
+++ b/packages/react-sdk/src/relayer/use-user-decrypted-values.ts
@@ -17,9 +17,9 @@ export function useUserDecryptedValues(handles: Handle[]) {
     })),
   });
 
-  const data: Record<Handle, ClearValueType> = {};
+  const data: Record<Handle, ClearValueType | undefined> = {};
   for (let i = 0; i < handles.length; i++) {
-    data[handles[i]!] = results[i]!.data as ClearValueType;
+    data[handles[i]!] = results[i]!.data as ClearValueType | undefined;
   }
 
   return {
--- a/packages/react-sdk/src/relayer/use-user-decrypted-values.ts
+++ b/packages/react-sdk/src/relayer/use-user-decrypted-values.ts
@@ -17,8 +17,10 @@ export function useUserDecryptedValues(handles: Handle[]) {
     })),
   });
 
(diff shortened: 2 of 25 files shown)


zama-cremaud:
## Issues

### [LOW] New `ae-forgotten-export` warning in `sdk-query.api.md` for `ClearValueType`

**File:** `packages/sdk/etc/sdk-query.api.md`

Before this branch, `ClearValueType` in the `@zama-fhe/sdk/query` subpath was sourced from `@zama-fhe/relayer-sdk/bundle`. API Extractor treated it as a third-party type (not owned by the SDK) and did not require it to be re-exported from the subpath entry point. After moving to a locally-defined `ClearValueType`, API Extractor correctly flags it as an internal type that is used in the public query surface but not exported through `packages/sdk/src/query/index.ts`:

```
// Warning: (ae-forgotten-export) The symbol "ClearValueType$1" needs to be exported by the entry point index.d.ts
```

This warning was not present on `prerelease`. The type IS exported from `@zama-fhe/sdk`'s main entry point, so consumers can import it from there. But the `sdk/query` subpath's declared API surface now references a type consumers of that subpath can't import from the same path — which may be unexpected. The fix is to re-export `ClearValueType` from `packages/sdk/src/query/index.ts`.

---

### [LOW] `RelayerSDK` interface still imports `ClearValueType` from `@zama-fhe/relayer-sdk/bundle`

**File:** `packages/sdk/src/relayer/relayer-sdk.ts`

All other files that previously imported `ClearValueType` from `@zama-fhe/relayer-sdk/bundle` were updated to use the locally-defined version in `relayer-sdk.types.ts`. The `RelayerSDK` interface (`relayer-sdk.ts`) was not updated because it doesn't have a `ClearValueType | undefined` usage and therefore did not trigger the lint error being fixed. The result is that two different `ClearValueType` definitions coexist in the public API:
- The locally-defined `ClearValueType = bigint | boolean | \`0x${string}\`` — exported from `sdk/src/index.ts`
- The bundle's version — used in `RelayerSDK.userDecrypt()` and `RelayerSDK.delegatedUserDecrypt()` return types, appearing as `ClearValueType_2` in `sdk.api.md`

Both are structurally identical today, so TypeScript structural typing ensures they are interchangeable. But the API report shows two different names for the same concept, and the `RelayerSDK` interface's return types are inconsistent with the rest of the codebase. Updating `relayer-sdk.ts` to import from `./relayer-sdk.types` would align everything.

zama-cremaud (dismissed):
2 minor feedbacks but nothing blocker from my pov!

zama-cremaud (dismissed):

zama-cremaud (approved):

PR #195, opened 2026-04-03
feat: global decryption cache; userDecrypt refactors

enitrat:
## Summary

Replaces the token-specific `balance-cache` with a generic, handle-level decrypt cache and lifts `userDecrypt` to a first-class SDK method. This unifies all decryption paths (token balances, arbitrary handles, delegated decryption) through a single cache-enabled entry point.

- **Generic decrypt cache** (`decrypt-cache.ts`): replaces `balance-cache.ts`. Any FHE handle from any contract can now be cached, not just token balances. Keyed by `(requester, contractAddress, handle)`.
- **`ZamaSDK.userDecrypt()`**: new top-level method that checks cache → groups uncached handles by contract → batch-decrypts via relayer → writes results back to cache.
- **React simplification**: removes `useUserDecryptedValue` and `useUserDecryptedValues` hooks. `useUserDecrypt` is now the single entry point, backed by `staleTime: Infinity` and the persistent cache.
- **Cache invalidation on revoke**: `revoke()` and `revokeSession()` now clear cached plaintext alongside credentials, preventing stale access after session/account changes.

### Why `(requester, contract, handle)` as cache key?

The decrypted plaintext for a given handle is always the same regardless of who requests it. However, there is an on-chain ACL system — only the handle owner (or their delegate) is authorized to decrypt. The `requester` dimension ensures that:
- Revoking user A's session only wipes user A's cached entries
- User B cannot get a cache hit on handles they haven't been ACL-allowed to decrypt

### Why `userDecrypt` is a query, not a mutation

Decryption *can* trigger a wallet signature (via `credentials.allow`), but the intended UX is that app developers call `sdk.allow()` upfront as a "global unlock". After that, all decryption happens in the background without re-signing. The operation is idempotent (same handles → same plaintext), so TanStack Query semantics (`staleTime: Infinity`, `retry: false`) are the right fit. The wallet prompt is a one-time credential acquisition, not a recurring side effect.

### Why remove `useUserDecryptedValue` / `useUserDecryptedValues`

These hooks were read-only lookups into the query cache by individual handle. With the persistent decrypt cache now built into `sdk.userDecrypt()` and `staleTime: Infinity` on the query, `useUserDecrypt` itself serves this role — call it with the handles you need, and cached values are returned without RPCs.

### `DecryptHandle` type

Pairs a handle with its originating `contractAddress`, enabling decryption of arbitrary handles from any contract (not just token balances where the contract was implicit).

## What changed

| Area | Before | After |
|------|--------|-------|
| Cache | `balance-cache.ts` — token-specific, keyed by `(token, owner, handle)` | `decrypt-cache.ts` — generic, keyed by `(requester, contract, handle)` |
| SDK decrypt | Only via `ReadonlyToken.balanceOf()` | `ZamaSDK.userDecrypt()` + token methods delegate to shared cache |
| React hooks | `useUserDecrypt` + `useUserDecryptedValue` + `useUserDecryptedValues` | `useUserDecrypt` only |
| Revoke | Cleared credentials only | Clears credentials **and** cached plaintext |
| Query key | Per-handle (`decryption.handle`) | Batch (`decryption.batch`) with sorted handles for deterministic hashing |

## Test plan

- [ ] `decrypt-cache.test.ts` — unit tests for load/save/clear/invalidation
- [ ] `zama-sdk.test.ts` — `userDecrypt` cache hit/miss/mixed scenarios
- [ ] `use-user-decrypt.test.tsx` — React hook integration
- [ ] `readonly-token.test.ts` — token balance decryption uses new cache
- [ ] Verify `useUserDecryptedValue` / `useUserDecryptedValues` removed from exports
- [ ] Size limit updated (48 KB → 48.5 KB)


enitrat:
you'll likely want to follow-up doing the same thing for public decrypts, btw.

aquint-zama:
@cla-bot check

PR #205, opened 2026-04-07
refactor(react-sdk): simplify query/mutation hook calls

ghermet:
## Summary
- Pass options objects directly to `useQuery`/`useSuspenseQuery` instead of spreading into a new object literal
- Remove redundant `enabled` merging where the options factory already handles it
- Affects 11 hook files across balance, delegation, relayer, shield, token, transfer, and wrapper-discovery

## Test plan
- [ ] Verify all existing tests pass
- [ ] Confirm hook behavior unchanged (query keys, enabled logic, data flow)


enitrat (changes requested):
removing spread operators in queries/mutations if we have no extra overrides; but the logic changes you made around `enabled` is flawed:  today those hooks do `factoryEnabled && options.enabled`, so user code can disable a query but cannot override missing prerequisites, and you removed that logic, so a query that is missing a prop might un-expectedly fire with missing data.

ghermet:
> removing spread operators in queries/mutations if we have no extra overrides; but the logic changes you made around `enabled` is flawed: today those hooks do `factoryEnabled && options.enabled`, so user code can disable a query but cannot override missing prerequisites, and you removed that logic, so a query that is missing a prop might un-expectedly fire with missing data.

isn't `enabled: (baseOpts.enabled ?? true) && (options?.enabled ?? true)` redundant if with `{ ...baseOpts, ...options }`

enitrat:
> isn't enabled: (baseOpts.enabled ?? true) && (options?.enabled ?? true) redundant if with { ...baseOpts, ...options }

first one means "enabled is true if BOTH baseOpts.enabled AND options.enabled are NOT FALSE"
second one means "enabled is true if options.enabled is true" (that's how object merges work). the second object overrides properties of the first one.

enitrat (approved):

PR #214, opened 2026-04-08
feat(sdk): internally batch credential requests to respect the 10-address fhevm limit [SDK-43]

zama-cremaud:
## Summary

\`allow()\` on both \`CredentialsManager\` and \`DelegatedCredentialsManager\`
previously capped the caller at 10 contract addresses (the fhevm ACL
on-chain limit). Any larger address set required callers to manage
batching themselves.

This PR makes batching fully transparent:

- **Any number of addresses** can be passed to \`allow()\`. The SDK splits
  them internally into batches of ≤ 10 and requests one EIP-712 wallet
  signature per batch, shown sequentially.
- **Shared keypair**: a single FHE keypair is generated per \`allow()\` call
  and reused across all its batches.
- **Bin-packing**: when \`allow()\` is called again with additional addresses,
  new addresses are packed into existing batches that still have spare
  capacity before creating new ones.
- **Fail-fast**: if the user rejects a batch's signature prompt, \`allow()\`
  throws immediately. Already-signed batches are retained in storage so a
  retry only needs to sign the rejected batch.
- **\`CredentialSet<T>\`** (new public type) is returned by \`allow()\` and
  routes each address to its batch via \`credentialFor(address)\`. A
  \`tryCredentialFor(address)\` variant returns \`null\` instead of throwing.

### API changes

| Before | After |
|--------|-------|
| \`allow(...addrs): Promise<StoredCredentials>\` | \`allow(...addrs): Promise<CredentialSet>\` |
| \`allow(delegator, ...addrs): Promise<DelegatedStoredCredentials>\` | \`allow(delegator, ...addrs): Promise<CredentialSet<DelegatedStoredCredentials>>\` |

\`CredentialSet\` and \`MAX_CONTRACTS_PER_CREDENTIAL\` are now exported from
the package entry point.

### Storage key scheme (backward-compatible)

Batch 0 uses the existing key (\`hash(address:chainId)\`); subsequent
batches append an index (\`key:1\`, \`key:2\`, …). Pre-existing single-batch
credentials are loaded transparently without migration.

## Test plan

- [x] \`credential-batching.test.ts\` — 21 new tests covering: 1/2-batch
  splits, 20-address full coverage, shared keypair, cache hits, bin-packing
  (10→11 and 9→11), fail-fast rejection, \`DelegatedCredentialsManager\` batching,
  \`CredentialSet\` semantics (unknown address, case-insensitive lookup,
  duplicate deduplication)
- [x] All 15 existing \`credentials-manager.test.ts\` tests updated to the
  new \`CredentialSet\` return type
- [x] All existing \`delegated-credentials-manager.test.ts\` tests updated
- [x] \`batch-decrypt-as.test.ts\` and \`events.test.ts\` updated
- [x] React SDK \`use-user-decrypt.test.tsx\` updated
- [x] 1542 tests passing, zero type errors, lint and format clean
- [x] API report updated (\`sdk.api.md\`, \`sdk-query.api.md\`)

Ref: https://linear.app/zama/issue/SDK-43/sdk-should-internally-batch-credential-requests-to-respect-relayer


ankurdotb:
### Code review

Found 4 issues (1 resolved since last review):

~1. **`credentialFor` throws plain `Error` instead of `ZamaError`** — Resolved. Now throws `ZamaError(ZamaErrorCode.CredentialNotFound)` and renamed to `batchFor`/`tryBatchFor`.~

2. **`#lastExtendResult` / `#extendPromise` not keyed by batch** — These are single instance fields on the base class. When `resolveMulti` processes batches sequentially, batch N's extension sets `#lastExtendResult`, and when batch N+1 enters `#extendContracts`, the fallback path picks up batch N's result as a "base" — which has the wrong contract addresses. This can cause cross-batch contamination.

https://github.com/zama-ai/sdk/blob/8460fbac5648eaf681a0e27c80965bebd900a15a/packages/sdk/src/credentials/credentials-manager-base.ts#L96-L101

https://github.com/zama-ai/sdk/blob/8460fbac5648eaf681a0e27c80965bebd900a15a/packages/sdk/src/credentials/credentials-manager-base.ts#L666-L676

3. **Session-model docs: stale "new keypair" claim** — Line 170 says "the SDK must generate a new keypair" when adding a contract, but the PR introduced a cached keypair that is reused within its TTL window. Should say a new *signature* (not keypair) is needed.

https://github.com/zama-ai/sdk/blob/8460fbac5648eaf681a0e27c80965bebd900a15a/docs/gitbook/src/concepts/session-model.md#L170-L170

4. **`PartialCredentialError` not documented** — New error type that consumers need to handle when multi-batch `allow()` partially fails, but it is not mentioned in the README or session-model docs. Callers won't know to catch it or access `partialResult`.

5. **Batcher redundant processing** — The accumulation pattern re-invokes `fn` with all previously processed addresses on each iteration. For the credential flow this means `resolveMulti` is called with a growing list each time, causing redundant storage lookups and batch-assignment computation on already-resolved addresses.

https://github.com/zama-ai/sdk/blob/8460fbac5648eaf681a0e27c80965bebd900a15a/packages/sdk/src/utils/batcher.ts#L1-L50

6. **Typo in docs** — `emmiter` should be `emitter` in the Node EventEmitter example in ZamaSDK.md (lines 354, 360, 363, 371).

https://github.com/zama-ai/sdk/blob/8460fbac5648eaf681a0e27c80965bebd900a15a/docs/gitbook/src/reference/sdk/ZamaSDK.md#L354-L371

packages/sdk/src/credentials/credentials-manager.ts
```diff
@@ -74,18 +85,32 @@ export class CredentialsManager extends BaseCredentialsManager<
...
    *
-   * @throws {@link SigningRejectedError} if the user rejects the wallet signature prompt.
-   * @throws {@link SigningFailedError} if the signing operation fails for any other reason.
+   * Addresses are split into batches of ≤ 10 internally to satisfy the fhevm
+   * contract limit. Each batch triggers its own EIP-712 wallet prompt (shown
+   * sequentially). The returned {@link CredentialSet} routes each address to
+   * its batch transparently — callers only need `credSet.credentialFor(address)`.
+   *
+   * Signing failures (e.g. user rejection) are captured in
+   * {@link CredentialSet.failures} rather than thrown — `allow()` itself
```
zama-cremaud: [AUTOMATED]

**Broken TSDoc link + incorrect behavior description.**

`CredentialSet.failures` does not exist on the `CredentialSet` interface — this is a dangling reference that will produce a broken link in generated docs.

More critically, the described behavior is wrong. The implementation is fail-fast: `resolveMulti` propagates `resolveCredentials` errors without catching them, and `Batcher.execute` stops on the first failure. The test at `credential-batching.test.ts` confirms:
```ts
await expect(manager.allow(...makeAddresses(11))).rejects.toThrow("User rejected");
```

Suggested replacement for these lines:
```ts
 * @throws {@link SigningRejectedError} if the user rejects any batch's wallet signature prompt.
 * @throws {@link SigningFailedError} if a batch signing operation fails for any other reason.
 * Previously-signed batches are retained in storage so a retry only needs to sign the failed batch.
```

packages/sdk/src/credentials/delegated-credentials-manager.ts
```diff
@@ -62,22 +73,38 @@ export class DelegatedCredentialsManager extends BaseCredentialsManager<
...
    *
-   * @throws {@link SigningRejectedError} if the user rejects the wallet signature prompt.
-   * @throws {@link SigningFailedError} if the signing operation fails for any other reason.
+   * Addresses are split into batches of ≤ 10 internally (same fhevm limit as
+   * regular credentials). Each batch triggers its own EIP-712 wallet prompt,
+   * shown sequentially. Use {@link CredentialSet.credentialFor} to route each
+   * address to its batch.
+   *
+   * Signing failures (e.g. user rejection) are captured in
+   * {@link CredentialSet.failures} rather than thrown — `allow()` itself
```
zama-cremaud: [AUTOMATED]

Same issue as `CredentialsManager.allow()`: `{@link CredentialSet.failures}` is a dangling reference and `allow()` does throw on signing rejection (fail-fast). The JSDoc should be updated to match the actual behavior:
```ts
 * @throws {@link SigningRejectedError} if the user rejects any batch's wallet signature prompt.
 * @throws {@link SigningFailedError} if a batch signing operation fails for any other reason.
```

packages/sdk/src/index.ts
```diff
@@ -96,6 +98,7 @@ export type {
 } from "./types";
 export type { Address, Hex } from "viem";
 export { ZamaSDKEvents } from "./events";
+export { Batcher } from "./utils";
```
zama-cremaud: [AUTOMATED]

**Unintended public export.** `Batcher` is an internal orchestration utility. The PR summary lists `CredentialSet` and `MAX_CONTRACTS_PER_CREDENTIAL` as the new public exports — `Batcher` is not mentioned. Exporting it:
- Locks the project into a semver guarantee on `Batcher`'s interface.
- Implies to callers that they should manage batching themselves, contradicting the PR's design goal of transparent internal batching.

Suggested fix: remove this line and add `/** @internal */` to `batcher.ts` if deep-import access is still needed internally.

packages/sdk/src/utils/batcher.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @typeParam TItem - The item type being batched.
+ * @typeParam TResult - The return type of each batch call.
+ *
+ * @example
+ * ```ts
+ * const batcher = new Batcher(10);
+ *
+ * // With CredentialsManager
+ * const creds = await batcher.execute(addresses, (addrs) => credentials.allow(...addrs));
+ *
```
zama-cremaud: [AUTOMATED]

**Stale example.** This shows external `Batcher` usage wrapping `credentials.allow()`, but `allow()` now batches internally. If someone follows this example with >10 addresses, the external `Batcher` would call `allow()` once per accumulated slice — each triggering its own internal `Batcher` and `resolveMulti` storage scan — resulting in redundant I/O.

Since `Batcher` is currently exported as `@public` (see parallel comment on `index.ts`), if it is kept public, the example should show a non-credential use case. If it becomes `@internal`, the example can be removed.

zama-cremaud (commented):
[AUTOMATED]

## Requirement Coverage

| Requirement | Status | Evidence |
|---|---|---|
| Transparent batching (≤10 per EIP-712) | ✅ SATISFIED | `resolveMulti` + `Batcher` in `credentials-manager-base.ts` |
| Single shared FHE keypair per `allow()` | ✅ SATISFIED | `#cachedKeypair` / `#cachedKeypairs` in both managers |
| Bin-packing (fill spare capacity first) | ✅ SATISFIED | `#computeBatchAssignments` in `credentials-manager-base.ts` |
| `CredentialSet<T>` returned by `allow()` | ✅ SATISFIED | `credential-set.ts`, both manager signatures updated |
| Backward-compatible storage key scheme | ✅ SATISFIED | Batch 0 uses existing key; batch N uses `key:N` |
| Partial-failure handling | ⚠️ PARTIAL | Implementation is **fail-fast** (throws on rejection), but JSDoc on both managers says `allow()` never rejects (see Finding #1) |

## Linked Sources

- Linear SDK-43: `ACCESS_DENIED` (internal Linear, not publicly accessible)
- PR body: `FETCHED` — used as primary requirement source

## New High-Signal Findings

### Finding #1 — JSDoc on `allow()` describes non-existent partial-failure API (Confidence: 95)

Both `CredentialsManager.allow()` and `DelegatedCredentialsManager.allow()` have this JSDoc:
> _"Signing failures (e.g. user rejection) are captured in `{@link CredentialSet.failures}` rather than thrown — `allow()` itself never rejects."_

This is wrong on two counts:
1. **`CredentialSet.failures` does not exist.** The `CredentialSet` interface (`credential-set.ts`) exposes only `credentialFor`, `tryCredentialFor`, and `batches`. The TSDoc link is dangling and will produce a broken reference in generated docs.
2. **`allow()` does throw.** `resolveMulti` propagates signing errors immediately (fail-fast), and the tests confirm it: `await expect(manager.allow(...makeAddresses(11))).rejects.toThrow("User rejected")`. The hook JSDoc in `use-allow.ts` correctly says _"The mutation rejects if any batch signing fails."_

Callers who read the manager JSDoc and assume `allow()` never throws will omit `try/catch`, producing unhandled rejections in production.

### Finding #2 — `Batcher` is exported as a public API surface (Confidence: 88)

The PR exports `Batcher` from the package entry point (`src/index.ts`, line 101), and it appears as `@public` in `sdk.api.md`. The PR summary only names `CredentialSet` and `MAX_CONTRACTS_PER_CREDENTIAL` as intentional new public exports. `Batcher` is an internal orchestration utility with no documented external use case. Shipping it as `@public` commits the project to a semver guarantee on its interface and signals to callers that they should manage batching themselves — the opposite of this PR's design goal.

### Finding #3 — `Batcher` JSDoc example shows obsolete external-usage pattern (Confidence: 85)

The `@example` block in `batcher.ts` (lines 14–22) shows:
```ts
const creds = await batcher.execute(addresses, (addrs) => credentials.allow(...addrs));
```
Since `allow()` now batches internally, constructing an external `Batcher` around `allow()` is redundant and misleading. If callers followed this example with >10 addresses, the external batcher would call `allow()` once per accumulated slice (O(n) `resolveMulti` scans), while each `allow()` call also spawns its own internal `Batcher`. The example should be removed or replaced with a non-credential use case.

## Blocked Context

- **Linear SDK-43**: not accessible — requirement details inferred from PR body and test coverage only.

## Decision: REQUEST_CHANGES

Finding #1 is a correctness-class documentation bug: the documented behavior (`allow()` never throws, failures captured in `CredentialSet.failures`) contradicts the actual implementation and the tests. Callers relying on the docs will write incorrect error-handling code. Findings #2 and #3 are API-surface issues that would be breaking to reverse once merged.

packages/sdk/src/credentials/delegated-credentials-manager.ts
```diff
@@ -62,22 +73,39 @@ export class DelegatedCredentialsManager extends BaseCredentialsManager<
...
     const normalizedDelegator = getAddress(delegatorAddress);
     const normalized = normalizeAddresses(contractAddresses);
     const key = await this.#storeKey(normalizedDelegator);
-    return this.resolveCredentials({
-      key,
-      contracts: normalized,
-      createKey: `${normalizedDelegator}:${normalized.join(",")}`,
-      createFn: () => this.#create(normalizedDelegator, normalized),
-    });
+    const batcher = new Batcher(MAX_CONTRACTS_PER_CREDENTIAL);
```
ghermet: probably can be a class attribute as there's no variable used

ghermet (commented):
## Code Review: Batch Credential Requests

### Critical Issues (3)

**1. `#scanBatchMeta` stops at first gap — orphaned batches** (`credentials-manager-base.ts`)
If batch 0 is deleted/corrupted but batches 1, 2 exist, the scanner returns empty — higher-index batches become permanently orphaned. New batches could also collide with undiscovered ones. Consider scanning past gaps or storing a batch manifest.

**2. Empty `catch {}` in `#scanBatchMeta`** (`credentials-manager-base.ts`)
Bare catch block swallows all errors from `assertEncrypted` — no logging, no event emission. Bugs in assertion logic or unexpected storage shapes would be silently misinterpreted as "corruption."

**3. Empty `catch {}` in `checkExpired` multi-batch path** (`credentials-manager-base.ts`)
Same pattern — any error from `assertEncrypted` or `isCredentialValid` silently returns `false` (treated as expired), causing unnecessary wallet re-prompts with no diagnostic output.

### Important Issues (5)

**4. `credentialFor` throws plain `Error` instead of `ZamaError`** (`credential-set.ts`)
Breaks the SDK error hierarchy. Consumers using `matchZamaError` won't catch it.

**5. Loss of error context wrapping for `generateKeypair` failures** (confirmed by `use-user-decrypt.test.tsx` change)
Error message changed from `"Failed to create decrypt credentials: keygen failed"` to bare `"keygen failed"` — the contextual prefix from `wrapSigningError` is lost in the batching flow.

**6. README references `credSet.failures` which doesn't exist** (`README.md`)
`CredentialSet` has no `failures` property. Behavior is fail-fast, not partial-success.

**7. `checkExpired`, `checkAllowed`, `clearAll`, `revokeSession` multi-batch paths are untested**
The most complex branching changes in the PR have zero direct test coverage.

**8. `chunkAddresses` has no guard for `size < 1`** (`credential-validation.ts`)
Passing `size = 0` would cause an infinite loop. Every other entry point validates this.

### Suggestions (4)

**9.** Rename `Batcher` to `AccumulatingBatcher` — accumulation semantics are non-obvious from the name.

**10.** Add max-scan-limit to `#scanBatchMeta`/`#scanBatchKeys` — the `for (let i = 0; ; i++)` loops have no upper bound.

**11.** `resolveMulti` error lacks batch context — when signing is rejected in batch 3 of 5, no info about which batch failed.

**12.** O(K²) storage scanning — `Batcher.execute` calls `resolveMulti` K times with progressively larger sets, each re-scanning from scratch. Consider having `resolveMulti` own the loop.

### Strengths

- Clean `CredentialSet<T>` interface with `credentialFor`/`tryCredentialFor` duality
- Fail-fast + retry-from-storage is well-designed for wallet UX
- Shared keypair across batches — `generateKeypair` called once
- Bin-packing fills spare capacity before creating new batches
- Thorough test suite (21 new tests)
- Generic type usage is textbook

ghermet:
## Keypair lifetime concern: `#cachedKeypair` can outlive `keypairTTL`

The `keypairTTL` is correctly enforced when **loading** credentials from storage — `isTimeValid()` gates every reuse path and forces fresh credential creation when the stored keypair has expired.

However, the **in-memory** `#cachedKeypair` (used for new batch creation) is never invalidated by TTL. It lives as long as the manager instance and is only cleared by explicit `clear()` / `revoke()` / `clearCaches()` calls.

This means the following sequence extends the effective keypair lifetime beyond `keypairTTL`:

```
t=0:    allow([A])       → generates keypair K, batch with startTimestamp=now, durationDays=30
t=29d:  allow([B])       → reuses cached K, NEW batch with startTimestamp=now, durationDays=30
                           → keypair K is now usable for decryption for 59 days total
```

Since the purpose of `keypairTTL` is to bound the window during which a compromised keypair can decrypt transport traffic, this is a gap — a single compromise of K covers all future batches created during the manager's lifetime, not just the configured TTL window.

### Suggested fix

Track when `#cachedKeypair` was generated and invalidate it when `keypairTTL` has elapsed:

```typescript
// credentials-manager.ts
#cachedKeypair: { publicKey: Hex; privateKey: Hex; createdAt: number } | null = null;

// in #createBatch:
const now = Math.floor(Date.now() / 1000);
if (!this.#cachedKeypair || (now - this.#cachedKeypair.createdAt) >= this.keypairTTL) {
  const kp = await this.#relayer.generateKeypair();
  this.#cachedKeypair = { ...kp, createdAt: now };
}
```

Same pattern applies to `DelegatedCredentialsManager.#cachedKeypairs`.

This ensures keypair material is never reused beyond its intended rotation window while still allowing efficient reuse within a single batching burst.

ghermet:
## Partial success visibility on batch failure

The sequential batch processing is correct (can't fire multiple wallet popups at once), and the implicit resume behavior is solid — persisted batches are reused on retry via `#scanBatchMeta`.

However, when a mid-batch network failure occurs (e.g. `generateKeypair` or `createEIP712` hitting the relayer while internet is down), the caller gets an opaque error with no visibility into what succeeded.

### Suggestion

When a batch fails, wrap the error to include the partial result:

```typescript
// New error type
class PartialCredentialError<T extends StoredCredentials> extends ZamaError {
  /** Batches that were successfully signed and persisted before the failure. */
  readonly partialResult: CredentialSet<T>;
  /** The original error that caused the batch to fail. */
  readonly cause: Error;
}
```

In `resolveMulti`, catch per-batch errors and attach what succeeded:

```typescript
const results: Array<{ addresses: Address[]; result: TCreds }> = [];
for (const { batchKey, addresses } of assignments) {
  try {
    const creds = await this.resolveCredentials({ ... });
    results.push({ addresses: normalizedAddresses, result: creds });
  } catch (error) {
    throw new PartialCredentialError(
      new CredentialSetImpl(results),  // what succeeded
      toError(error),                   // what failed
    );
  }
}
```

This lets developers make informed retry decisions:

```typescript
try {
  const creds = await sdk.allow(...addresses);
} catch (e) {
  if (e instanceof PartialCredentialError) {
    console.log(`${e.partialResult.batches.length} batches succeeded`);
    // Safe to retry — already-persisted batches are reused automatically
    const creds = await sdk.allow(...addresses);
  }
}
```

The retry path already works correctly today (persisted batches are found by `#scanBatchMeta`), this just makes it explicit to the developer.

zama-cremaud:
Rebased this branch on top of `prerelease` and aligned the credential batching work with the post-SDK-77 / SDK-78 scoped `isAllowed` model.

  Main changes in this pass:
  - kept `isAllowed(...)` contract-scoped across batching
  - fixed multi-batch `checkAllowed(...)` semantics so requested contracts are checked against valid covering batches, instead of treating all stored batches as globally required
  - updated `ZamaSDK.userDecrypt(...)` to use the returned `CredentialSet` correctly and route each contract to the matching batch credential
  - fixed targeted revoke behavior in the batched world so `credentials.revoke(...addresses)` only revokes sessions for batches covering those contracts
  - updated SDK / React SDK README docs to reflect:
    - `CredentialSet` return semantics
    - batching in groups of up to 10
    - `credentials.isAllowed([address, ...])`
    - `useIsAllowed({ contractAddresses })`
  - added / updated batching tests covering:
    - multi-batch `isAllowed(...)`
    - unrelated batch revoke not breaking allowed state for still-covered contracts
    - delegated multi-batch coverage
  - regenerated `packages/sdk/etc/sdk-query.api.md`

  Known note:
  - `pnpm api-report:check:react-sdk` still hits the same API Extractor internal error (`Unable to follow symbol for "ActivityItem"`) that also reproduces on a detached `prerelease` worktree, so this does not appear to be introduced by this branch.

zama-cremaud (commented):
[AUTOMATED]

Review of head `ef060c6ec7` (latest commit: `fix: reconcile credential batching with scoped isAllowed`). Prior reviews: zama-cremaud `2026-04-09`, ghermet `2026-04-10` × 2.

## History-Aware Classification

| Prior finding | Status |
|---|---|
| ghermet #1: `#scanBatchMeta` stops at first gap | Acknowledged, not addressed (by design) — not repeated |
| ghermet #2: empty `catch` in `#scanBatchMeta` | **FIXED** — `console.warn` added |
| ghermet #3: empty `catch` in `checkExpired` | **FIXED** — `console.warn` added |
| ghermet #4: `credentialFor` throws plain `Error` | **NOT FIXED** — see Finding A below (impact escalated) |
| ghermet #5: error context wrapping on `generateKeypair` | Unverifiable from diff — not repeated |
| ghermet #6: README references `credSet.failures` | **FIXED** — `credSet.failures` removed from README |
| ghermet #7: `checkAllowed/clearAll/revokeSession` untested | **FIXED** — 3 new tests in latest commit |
| ghermet #8: `chunkAddresses` size guard | **FIXED** — guard added |
| zama-cremaud: JSDoc `allow()` never rejects | **FIXED** — `@throws` docs added |
| zama-cremaud: `Batcher` unintended public export | **FIXED** — `@internal`, not in `index.ts` |
| zama-cremaud: stale example in `batcher.ts` | **FIXED** — example removed |

## Requirement Coverage

| Requirement (SDK-43 + PR body) | Status | Evidence |
|---|---|---|
| Transparent batching ≤10 per EIP-712 | SATISFIED | `Batcher` + `resolveMulti` in `credentials-manager-base.ts` |
| Single shared FHE keypair per `allow()` burst | SATISFIED | `#cachedKeypair` / `#cachedKeypairs` with `keypairTTL` TTL enforcement |
| Bin-packing (fill spare capacity first) | SATISFIED | `#computeBatchAssignments` in `credentials-manager-base.ts` |
| `CredentialSet<T>` returned by `allow()` | SATISFIED | `credential-set.ts`; both managers updated |
| Backward-compatible storage key scheme | SATISFIED | Batch 0 uses existing key; batch N uses `key:N` |
| Fail-fast: rejection throws immediately | SATISFIED | `Batcher.execute` + tests in `credential-batching.test.ts` |
| `PartialCredentialError` on partial success | SATISFIED | `partial-credential.ts`, `ZamaErrorCode.PartialBatchFailure` |
| Retry safety (persisted batches reused) | SATISFIED | `resolveCredentials` cache + test "retry reuses first batch" |

## New High-Signal Findings

### [A] MEDIUM (confidence: 82) — `credentialFor` throws plain `Error`, not `ZamaError`; impact escalated since ghermet's review

`CredentialSetImpl.credentialFor` (`credential-set.ts` L49):

```ts
throw new Error(`[zama-sdk] No credential found for address ${address}`);
```

This was raised as ghermet finding #4 (not addressed). Since that review, two new call sites were added in the latest commit that make the impact more concrete:

1. `zama-sdk.ts` — fail-fast validation loop before decryption:
```ts
for (const contractAddress of byContract.keys()) {
    credSet.credentialFor(contractAddress); // throws plain Error if missing
}
```

2. `readonly-token.ts` L760, L829, L892 — inline after `allow()`:
```ts
const creds = (await this.credentials.allow(this.address)).credentialFor(this.address);
```

If `credentialFor` throws (e.g. due to a bug or storage inconsistency), the caller receives a plain `Error` that `matchZamaError()` cannot catch, breaking the SDK's documented error-handling contract. Given that these paths are in the hot decrypt flow, the fix cost is low:

```ts
// credential-set.ts — suggested fix
throw new ZamaError(ZamaErrorCode.CredentialNotFound, `No credential found for address ${address}`);
```

Requires adding `CredentialNotFound` to `ZamaErrorCode` (one-liner in `errors/base.ts`).

### [B] LOW (confidence: 62) — `#scanBatchKeys` inherits the gap-stop limitation but is used for destructive cleanup

`#scanBatchKeys` (new function, `credentials-manager-base.ts`) uses the same stop-at-first-absent-slot loop as `#scanBatchMeta`. It is called by `clearAll` (which deletes storage entries) and by `revokeSession` (no-address path). Unlike `#scanBatchMeta`, which is used for read-only discovery, a gap-stop in `clearAll` can leave orphaned session entries uncleaned:

- If slot 0 was overwritten by a `corruptedIndex=0` recovery and batches at `key:1`, `key:2` remain, `clearAll` deletes only `key`'s session, leaving `key:1` and `key:2` sessions dangling in `sessionStorage`.

Given the existing acknowledgment of the gap pattern for `#scanBatchMeta` (ghermet #1), this is consistent. But it's worth tracking separately because `clearAll` / `revokeSession` are user-visible operations that are expected to be "complete".

### [C] LOW (confidence: 60) — `checkAllowed` inner `catch {}` is silent (inconsistent with `checkExpired` fix)

The multi-batch loop in `checkAllowed` (`credentials-manager-base.ts`) still uses a bare `catch { continue; }` after `assertEncrypted`. The analogous block in `checkExpired` was fixed to emit `console.warn`. A silent catch here means corrupted-batch scenarios in the `isAllowed` path produce no diagnostic output, making debugging harder.

## Blocked Context

| Source | Status |
|---|---|
| Linear SDK-43 | FETCHED (no comments) |

## Decision: COMMENT

Finding A is an SDK contract violation that should be fixed before merge — one-liner change with low risk. Findings B and C are low-severity advisory items. The core batching, bin-packing, PartialCredentialError, and keypair-TTL logic all look correct and well-tested.

packages/sdk/src/credentials/credential-set.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @typeParam T - The concrete credential shape (`StoredCredentials` for regular,
+ *   `DelegatedStoredCredentials` for delegated flows).
+ */
+export interface CredentialSet<T extends StoredCredentials = StoredCredentials> {
+  /**
+   * Return the credential batch that covers `address`.
+   * @throws If `address` was not passed to `allow()`.
+   */
+  credentialFor(address: Address): T;
```
ghermet: feels like this should be `batchFor`?

zama-cremaud: Yes! Updated

PR #218, opened 2026-04-09
build: add exports ./package.json subpath and default condition

enitrat:
Adding `"./package.json": "./package.json"` to the `exports` field of both packages. This is a standard practice for npm packages with strict exports — many tools in the ecosystem (bundlers, test frameworks, meta-frameworks) need to read a package's `package.json` at runtime via `require.resolve(pkg + '/package.json')`, which fails when the subpath isn't explicitly exported.

Also adds a `default` condition to `react-sdk` export entries. Without it, CJS resolution (`require.resolve`) can't find the main entry point since only `types` and `import` conditions were defined. The `sdk` package already had `import`/`require`/`default` on all entries.

**Concrete example:** Storybook 10's `sb.mock()` automocking system calls `require.resolve('@zama-fhe/react-sdk/package.json')` during test setup. Without these changes, the entire Storybook Vitest test runner crashes at startup with `Package subpath './package.json' is not defined by "exports"`.

### DIFF
--- a/packages/react-sdk/package.json
+++ b/packages/react-sdk/package.json
@@ -15,12 +15,15 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.js"
+      "import": "./dist/index.js",
+      "default": "./dist/index.js"
     },
     "./wagmi": {
       "types": "./dist/wagmi/index.d.ts",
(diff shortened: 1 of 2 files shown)


ghermet (dismissed):

enitrat:
^ updated to "default" form to satisfy tests

ghermet (approved):

PR #227, opened 2026-04-13
fix(sdk): pass full token set to credentials.allow in batchDecryptBalances

ghermet:
## Summary

- `ReadonlyToken.batchDecryptBalances` (used internally by `useConfidentialBalances`) now passes the **full** token set to `credentials.allow(...)` instead of just the uncached subset. This makes the credentials cache key stable across calls — a parallel `ReadonlyToken.allow(...)` or `useAllow` mutation now dedupes against the hook's internal credential request, so only one wallet signature is ever shown.
- Mirrors the behavior of `userDecryptMutationOptions`.
- Documents the recommended `useAllow` + `useIsAllowed` + `{ enabled: !!allowed }` pattern for `useConfidentialBalances`, matching how `useUserDecrypt` is documented.

### Why

Previously, when a caller did both `ReadonlyToken.allow(...allTokens)` (e.g. for warmup) and rendered `useConfidentialBalances({ tokenAddresses: allTokens })`, two wallet signatures were prompted. The credentials manager dedupes in-flight `allow()` calls by a key derived from the sorted address list, but the batch decrypt path was passing only the uncached subset — so the keys diverged and dedup failed.

## Test plan

- [x] `pnpm vitest run packages/sdk/src/token/ packages/sdk/src/query/ packages/react-sdk/src/balance/` → 507/507 pass
- [x] Updated `calls allow() only with uncached token addresses` → `calls allow() with all token addresses even when some are cached` (asserts both addresses are in the EIP-712 payload)
- [x] Added regression test `dedupes a parallel batchDecryptBalances call to a single signature`
- [ ] Manual verification: pre-call `ReadonlyToken.allow(...tokens)` and render `useConfidentialBalances({ tokenAddresses: tokens })` → only one wallet prompt


docs/gitbook/src/reference/react/useConfidentialBalances.md
```diff
@@ -128,6 +139,27 @@ The `data` property is `Map<Address, bigint> | undefined` -- a map from token ad
 
 {% include ".gitbook/includes/query-result.md" %}
 
+## How It Works
```
enitrat: is documenting internal behaviors that are subject to changes  relevant to have in public documentations?

packages/react-sdk/src/balance/use-confidential-balances.ts
```diff
@@ -41,19 +41,24 @@ export interface UseConfidentialBalancesOptions extends Omit<
  * Returns partial results when some tokens fail — successful balances are
  * always returned alongside per-token error information.
  *
+ * The decrypt phase passes the full token set to `credentials.allow(...)`,
+ * mirroring `useUserDecrypt`. This means a parallel `ReadonlyToken.allow(...)`
+ * or `useAllow` call dedupes against the internal credential request and
+ * triggers only one wallet signature.
```
enitrat: documentation should not describe internals

packages/react-sdk/src/balance/use-confidential-balances.ts
```diff
@@ -41,19 +41,24 @@ export interface UseConfidentialBalancesOptions extends Omit<
...
  * @example
  * ```tsx
- * const { data } = useConfidentialBalances({
- *   tokenAddresses: ["0xTokenA", "0xTokenB"],
- * });
- * const balance = data?.balances.get("0xTokenA");
- * if (data?.isPartialError) {
- *   // some tokens failed — check data.errors
- * }
+ * // Recommended: gate the decrypt phase on explicit authorization.
```
enitrat, line 50: do you want to update the code to align on user decrypt as well?

enitrat (changes requested):
The right fix is to remove `batchDecryptBalances`, or have it just be a very thing wrapper, and just call sdk.userDecrypt which has support for multiple handles. 

see: https://linear.app/zama/issue/SDK-82/route-readonly-token-decrypthandles-through-unified-sdkuserdecrypt

ghermet:
> The right fix is to remove `batchDecryptBalances`, or have it just be a very thing wrapper, and just call sdk.userDecrypt which has support for multiple handles.
> 
> see: https://linear.app/zama/issue/SDK-82/route-readonly-token-decrypthandles-through-unified-sdkuserdecrypt

I was thinking about a shared function user decrypt and batch decrypt could use

enitrat (dismissed):

PR #229, opened 2026-04-13
feat(sdk-61): replace WrappersRegistry with protocol ConfidentialTokenWrappersRegistry contracts

zama-cremaud:
Replace the temporary WrappersRegistry contract with the production-ready protocol contracts (ConfidentialTokenWrappersRegistry, ConfidentialWrapper).

The new registry is UUPS-upgradeable, enforces ERC-165 validation on registration, and supports token revocation.

On the SDK side, update ABIs and event decoders to match the new contract interfaces: UnwrapRequested and UnwrappedFinalized events now include an indexed unwrapRequestId field (replacing burnAmountHandle as identifier), and finalizeUnwrap is updated accordingly. All tests are adjusted to reflect these changes.

https://linear.app/zama/issue/SDK-61/replace-local-zaiffer-contracts-with-currently-deployed-protocol-apps


zama-cremaud (commented):
[AUTOMATED]

## Requirement Coverage

| Requirement (SDK-61) | Status | Evidence |
|---|---|---|
| Local Solidity contracts from protocol-apps | PARTIAL | Registry: `f51845b08cc2` ✓ — Wrapper: `da4afe387420` ✗ (different commit, see Finding A) |
| `finalizeUnwrap` ABI / TS types use `unwrapRequestId` | PARTIAL | ABIs updated ✓ — no source `.ts` defining `FinalizeUnwrapParams`/`Token.finalizeUnwrap` in the diff (see Finding B) |
| No remaining `burnAmountHandle` / `burnAmountHandle` references | PARTIAL | react-sdk tests ✓ — SDK source files not confirmed |
| `deployments.json` regenerated | SATISFIED | `contracts/deployments.json` updated (new registry address) |
| Unit tests updated | SATISFIED | events, activity, wrappers-registry, contracts tests all updated |
| Documentation updated | SATISFIED | `WrappersRegistry.md`, `contract-builders.md`, `Token.md`, React hooks docs updated |

## New High-Signal Findings

### [A] HIGH (confidence: 90) — Wrapper at `da4afe387420` includes Phase 2 breaking changes; SDK-61 requires Phase 1 only (`f51845b`)

`contracts/VERSIONS.md` explicitly records the two source commits:

- **confidential-token-wrappers-registry**: `f51845b08cc2` (December 2025) — matches the "currently deployed" version required by SDK-61 ✓
- **confidential-wrapper**: `da4afe387420` (March 26, 2026) — **different commit**, 3+ months newer ✗

The VERSIONS.md entry for the wrapper documents changes that are explicitly listed in SDK-61 as "upcoming breaking changes" to be excluded:

> - Events `UnwrapRequested` and `UnwrapFinalized` include `bytes32 indexed unwrapRequestId`
> - `unwrap()` returns `bytes32` unwrapRequestId (was `void`)
> - New functions: `unwrapAmount(bytes32)` and `unwrapRequester(bytes32)`
> - ERC-165 interfaceId: `0xf1f4c25a` (changed)

These are reflected throughout the diff: `encrypted.abi.ts`, `wrapper.abi.ts`, `erc165.ts`, `events.test.ts` topic string updates, and `sdk.api.md`.

SDK-61 is explicit:
> "Do not include any Phase 2 changes. The breaking changes are not yet deployed; including them early would break integrations against the current testnet."
> "Upcoming breaking changes (testnet April 21, mainnet April 28) are tracked separately."

With the testnet deployment scheduled for April 21 (7 days away), the current testnet is running the old interface. Including these changes now means the SDK cannot be used against the current testnet deployment.

**Question for the author**: Is this intentional — i.e., is this PR deliberately targeting the April 21 state rather than the current state? If so, the ticket scope note needs updating and the PR description should clarify this is a combined Phase 1+2 migration. If not, the wrapper contract and all ABI/event-signature changes need to be rolled back to `f51845b`.

### [B] MEDIUM (confidence: 75) — Source files defining `FinalizeUnwrapParams` and `Token.finalizeUnwrap` are absent from the diff

`packages/sdk/etc/sdk-query.api.md` reports:

```diff
-    burnAmountHandle: Address;
+    unwrapRequestId: Hex;
```

and

```diff
-    finalizeUnwrap(burnAmountHandle: Handle): Promise<TransactionResult>;
+    finalizeUnwrap(unwrapRequestId: Handle): Promise<TransactionResult>;
```

But none of the 30 changed files is the TypeScript source that defines `FinalizeUnwrapParams` or the `Token.finalizeUnwrap` method body (expected locations: `packages/sdk/src/query/finalize-unwrap.ts`, `packages/sdk/src/token/token.ts`, or similar). The same applies to `resume-unshield.ts`, which SDK-61 explicitly calls out for auditing.

If these source files were not updated, callers passing `{ burnAmountHandle }` will fail at runtime (the contract now expects `unwrapRequestId` as the first `finalizeUnwrap` argument) and TypeScript consumers will see a type mismatch. The `api.md` files suggest the rename is present, but they could reflect a state inconsistent with the committed source if they were regenerated from an earlier build artifact.

**Request**: Please confirm `packages/sdk/src/query/finalize-unwrap.ts` (or the file defining `FinalizeUnwrapParams`) and `resume-unshield.ts` were updated. If they are absent from the diff because they are auto-generated, call that out explicitly.

### [C] LOW (confidence: 65) — No Solidity tests for `ConfidentialTokenWrappersRegistry`

The new `ConfidentialTokenWrappersRegistry.sol` (313 lines, `contracts/src/factory/`) introduces: UUPS upgradeability, `Ownable2StepUpgradeable`, ERC-165 validation on registration, revocation state, and an EIP-7201 storage slot constant (`0xc361bd0b...`). No `.t.sol` test file appears in the diff.

Understandable if the contract is tested in `protocol-apps`, but for the SDK repo the EIP-7201 storage slot value and the upgrade authorization path (`_authorizeUpgrade`) are not covered locally. A minimal Foundry test (`test/ConfidentialTokenWrappersRegistry.t.sol`) verifying registration, revocation, ERC-165 rejection, and double-register revert would close this gap.

## Blocked Context

| Source | Status | Note |
|---|---|---|
| `zama-ai/protocol-apps` @ `da4afe387420` | NOT_REACHABLE | Private repo — cannot verify wrapper source matches PR |
| `zama-ai/protocol-apps` @ `f51845b08cc2` | NOT_REACHABLE | Private repo — cannot verify registry source matches PR |
| Linear SDK-61 | FETCHED | Used as primary requirements source |

## Decision: COMMENT

Finding A is a material scope question (Phase 1 vs Phase 1+2) that requires author confirmation before the PR can be approved. Finding B may block correct TypeScript usage and should be verified. Finding C is advisory. No changes are requested until Finding A is clarified.

zama-cremaud:
In response to the review :

## Wrapper commit is the current deployed state, not Phase 2

The wrapper is pinned to `da4afe387420` (March 26, 2026), which is the version **currently deployed** on testnet and mainnet. This is not the April 21/28 upgrade.

The upcoming breaking changes (April 21 testnet / April 28 mainnet) are at a separate commit (`93c6e7a`, April 7, 2026) and introduce a new interfaceId (`0x1f1c62b2`), `inferredTotalSupply`, and OZ confidential-contracts v0.4.0. Those are tracked in SDK-68 sub-tickets (SDK-69 through SDK-72), entirely separate from this PR.

The SDK-61 ticket was written against the old `f51845b` reference and listed the `unwrapRequestId` additions as "upcoming." They have since been deployed (March 26). This PR aligns the SDK to the live state.

## Finding B: source files are in the diff

`packages/sdk/src/query/finalize-unwrap.ts` and `packages/sdk/src/token/token.ts` are both changed in this PR. `FinalizeUnwrapParams` uses `unwrapRequestId: Hex`, `Token.finalizeUnwrap` is renamed accordingly, and `resume-unshield.ts` forwards to `token.resumeUnshield` unchanged. The api.md files are regenerated from the committed sources via `pnpm api-report`.

ghermet (commented):
# PR Review Summary

**Reviewed by:** 4 specialized agents (code, tests, error handling, comments)
**Scope:** 47 files, +1972 / -472

---

## Critical Issues (2)

### 1. Stale `found` variable in doc example
📍 `docs/gitbook/src/reference/sdk/WrappersRegistry.md:40`

The standalone usage example still uses `const [found, cToken]` instead of `const [isValid, cToken]`. The rest of the document was correctly updated.

### 2. `Wrapped` event ABI/decoder retained but event no longer emitted
📍 `packages/sdk/src/abi/wrapper.abi.ts` ~L688-700, `packages/sdk/src/abi/encrypted.abi.ts` ~L684-688

Both ABIs still define `Wrapped(address indexed to, uint256 amountIn)`, and `Topics.Wrapped` / `decodeWrapped()` / `findWrapped()` remain in `onchain-events.ts`. But `contracts/VERSIONS.md` states `emit Wrapped` was removed from `wrap()` and `onTransferReceived()`. These decoders will never match logs from the new contracts. Either remove them or add a deprecation comment explaining they're retained for backward compatibility with older deployments.

---

## Important Issues (5)

### 3. Revoked tokens now return non-null — silent behavioral change
📍 `packages/sdk/src/wrappers-registry.ts:357-409`

Old code returned `null` for revoked tokens. New code returns `{ isValid: false, address }`. Callers using `if (result)` to check existence will now process revoked tokens. Consider a `@breaking` annotation in the API report or a migration note.

### 4. Shield path produces encrypted amount where old path had cleartext
📍 `packages/sdk/src/activity.ts:135-146`

`ConfidentialTransfer` from `zeroAddress` replaces the old `Wrapped` event for shields. But the old event carried a cleartext `amountIn`, while the new path produces an encrypted handle requiring decryption. Add a code comment documenting this intentional change.

### 5. `FinalizeUnwrapParams.unwrapRequestId` typed as `Hex` but `Token.finalizeUnwrap` expects `Handle`
📍 `packages/sdk/src/query/finalize-unwrap.ts:8`

Type mismatch at the mutation boundary. Either use `Handle` consistently or switch `Token.finalizeUnwrap` to accept `Hex`, since `unwrapRequestId` is semantically a `bytes32` identifier, not an FHE handle.

### 6. No test asserts `publicDecrypt` is called with `unwrapRequestId`
📍 `packages/sdk/src/token/__tests__/token.test.ts`

The most critical behavioral change — passing `unwrapRequestId` instead of `burnAmountHandle` to `relayer.publicDecrypt` — has no assertion guarding it. A regression swapping these values would go undetected.

### 7. `findUnwrapRequested` tests don't assert the extracted `unwrapRequestId` value
📍 `packages/sdk/src/__tests__/events.test.ts:275`

Tests only check `expect(event).not.toBeNull()` without verifying the decoded `unwrapRequestId` equals the expected value.

---

## Suggestions

| # | Location | Issue |
|---|----------|-------|
| 8 | `events/onchain-events.ts:121-126` | `wordAt` silently pads short data with zeros — corrupted logs produce wrong values, not errors |
| 9 | `events/onchain-events.ts:194-233` | Event decoders return `null` on malformed logs with no logging/warning |
| 10 | `wrappers-registry.ts:274-295` | `listPairs` metadata fallback discards error reasons entirely (pre-existing) |
| 11 | `ConfidentialTokenWrappersRegistry.sol:15` | Struct NatSpec says "is revoked" but field is `isValid` (inverse) |
| 12 | `ConfidentialTokenWrappersRegistry.sol:10` | NatSpec phrasing "wrapper extensions of ERC-7984" could be clearer |
| 13 | `react-sdk/.../mutation-test-helpers.ts:26` | Hardcoded topic hash — consider importing `Topics.UnwrapRequested` from SDK |
| 14 | `token/__tests__/token.test.ts` | `UnwrapRequested` log structure copy-pasted ~8 times — extract a helper |

---

## Strengths

- ABI version smoke test pinning `finalizeUnwrap(bytes32)` signature
- Topic keccak256 verification guards against hex copy-paste errors
- Three-state registry tests (`valid` / `revoked` / `not registered`) precisely match new contract semantics
- Cache call count correctly updated (2→1, reflecting eliminated `isConfidentialTokenValid` RPC)
- `contracts/VERSIONS.md` tracking upstream commit hashes and API diffs
- Thorough JSDoc with `@example` blocks on `WrappersRegistry` public methods
- Solid error handling in `finalizeUnwrap` — lifecycle events, typed error wrapping, cause chains

---

## Recommended Action Plan

1. Fix doc example — `found` → `isValid` in `WrappersRegistry.md:40`
2. Annotate or remove `Wrapped` event artifacts (ABI, decoder, README table)
3. Add test assertion that `publicDecrypt` receives `unwrapRequestId` (highest test priority)
4. Add test assertion on `findUnwrapRequested` extracted value
5. Align `unwrapRequestId` type — `Hex` vs `Handle` consistency
6. Add code comment on the shield cleartext→encrypted change in `activity.ts`
7. Consider `@breaking` note for the revoked-token null→non-null semantic change

enitrat (commented):
rather than inlining the entire code of the `ConfidentialWrapper` and `ConfidentialTokenWrappersRegistry` can we import it from https://github.com/zama-ai/protocol-apps/tree/main ? you can add protocol-apps as a foundry soldeer git  dependency and directly pull the relevant contracts from there. 

Our repository should not be defining contracts, apart from mocks required from testing. All contracts used should be imported from their definition source, which is:

-  protocol-apps (we can use main `49b417a4` as the rev)
- openzeppelin-confidential-contracts v0.4.0

this means we can delete _all_ local contracts that were copied from there, and simply install those deps with soldeer, and simply add the required remappings to the foundry project:

this way, we wont risk being out of sync in the future and we can update easily.

enitrat (commented):
inside `onchain-events.ts` we still have hardcoded topics in the old format (see line 26). 

I'm not sure why tests still in `integration.test.ts` passes, perhaps it's not actually testing anything because everything's mocked, either way I don't think our test are properly testing anything atm which we should address in subsequent work. surely we should be able to generate proper bindings from the contract ABIs directly instead of hardcoding stuff

ankurdotb:
> rather than inlining the entire code of the ConfidentialWrapper and ConfidentialTokenWrappersRegistry can we import it from https://github.com/zama-ai/protocol-apps/tree/main ? you can add protocol-apps as a foundry soldeer git dependency and directly pull the relevant contracts from there.

> Our repository should not be defining contracts, apart from mocks required from testing. All contracts used should be imported from their definition source, which is:

Agree with the architectural direction. The SDK repo should not be maintaining copies of protocol-apps contracts. Importing via soldeer is the right long-term approach.

One concern on the rev choice: pinning to `main` at `49b417a4` would pull in the breaking changes from [SDK-68](https://linear.app/zama/issue/SDK-68/adapt-sdk-to-protocol-apps-breaking-changes-april-2026) (April 21 testnet / April 28 mainnet upgrade at `93c6e7a`), which is ahead of what's currently deployed. That would break the scope of this PR, which is to align the SDK with the **live** contracts.

The transition for SDK-68 is the part I want to think through more carefully. For example: after the April 21 testnet upgrade, the wrapper interfaceId changes from `0xf1f4c25a` to `0x1f1c62b2`. Any SDK user on the current version calling `isConfidentialWrapperContract` will get `false` for upgraded wrappers. Similarly, event decoders for `UnwrapRequested`/`UnwrapFinalized` won't match the new topic hashes. How do we want SDK users to handle this?

Some options I can see, open to others:
- **Dual-support SDK release** before the upgrade that accepts both old and new interfaceIds/events, so users upgrade once and are covered
- **Pin-and-bump**: users stay on the current SDK version until their target chain upgrades, then bump SDK version
- **Something else?**

For this PR specifically, I'd suggest switching to the soldeer dep but pinning to the currently-deployed revs (`da4afe387420` for wrapper, `f51845b` for registry) rather than `main`, so we get the dependency hygiene without pulling in unreleased changes. How we handle the SDK-68 bump is a separate decision.

enitrat:
@ankurdotb spot on thx, can we also ask protocol-apps to make proper tags / releases so that we don't have to rely on obscure revs ?

packages/sdk/src/events/onchain-events.ts
```diff
@@ -19,15 +23,20 @@ export type { RawLog } from "../types/transaction";
...
+  // compatibility with older deployments.
   /** `Wrapped(address indexed to, uint256 amountIn)` */
-  Wrapped: "0x4700c1726b4198077cd40320a32c45265a1910521eb0ef713dd1d8412413d7fc",
-  /** `UnwrapRequested(address indexed receiver, bytes32 amount)` */
-  UnwrapRequested: "0x77d02d353c5629272875d11f1b34ec4c65d7430b075575b78cd2502034c469ee",
-  /** `UnwrapFinalized(address indexed receiver, bytes32 encryptedAmount, uint64 cleartextAmount)` */
-  UnwrappedFinalized: "0x2d4edf3c2943002120f53dab3f8940043f34799f4a92ab90f2f81f7dd004a49e",
+  Wrapped: eventTopic("Wrapped(address,uint256)"),
+  /** `UnwrapRequested(address indexed receiver, bytes32 indexed unwrapRequestId, bytes32 amount)` */
+  UnwrapRequested: eventTopic("UnwrapRequested(address,bytes32,bytes32)"),
```
enitrat, line 33: nice ones!

enitrat (dismissed):

packages/sdk/src/events/onchain-events.ts
```diff
@@ -316,6 +336,8 @@ export function findUnwrapRequested(logs: readonly RawLog[]): UnwrapRequestedEve
   return null;
 }
 
+// NOTE: New wrapper contracts no longer emit this event. Retained for backward
+// compatibility with older deployments.
```
enitrat, line 340: might be worth a `@deprecated` jsdoc tag instead of a comment (other cases as well)

enitrat (approved):

PR #234, opened 2026-04-14
refactor(sdk): unify decrypt flows via shared pipelines [SDK-82] [SDK-85 P1]

ghermet:
## Summary

- Extracts all decrypt orchestration into composable pipelines under `packages/sdk/src/pipelines/`
- `ReadonlyToken.decryptBalance`, `decryptHandles`, and `ZamaSDK.userDecrypt` all route through `runUserDecryptPipeline`
- `ReadonlyToken.decryptBalanceAs` routes through `runDelegatedDecryptPipeline`
- `batchDecryptBalances` / `batchDecryptBalancesAs` route through `runBatchDecryptPipeline` which composes over the inner pipelines
- Shared steps extracted: `runCachePartitionPipeline` (cache hit/miss partition), `runGroupByContractPipeline` (group handles by contract)
- Fixes the credential cache key stability issue from PR #227 (full address set passed to `credentials.allow`, not just uncached)
- Breaks the `zama-sdk.ts <> query/user-decrypt.ts` circular dependency
- Moves `ZERO_HANDLE` to `constants.ts` to break `batch-decrypt-pipeline <> readonly-token` cycle
- Adds `sdk.allow(contractAddresses)` public method for pre-authorizing contracts before decrypt calls
- Routes `allowMutationOptions` through `sdk.allow()` instead of `sdk.credentials.allow()` directly
- Guards `credentials.allow()` against empty arrays
- `userDecrypt` pipeline emits `DecryptStart/End/Error` events and wraps relayer errors uniformly

Closes [SDK-82](https://linear.app/zama/issue/SDK-82). Delivers **P1** of [SDK-85](https://linear.app/zama/issue/SDK-85/refactor-sdk-to-use-a-unique-decryption-flow) (new `sdk.userDecrypt` pipeline + `sdk.allow`). Subsumes [PR #227](https://github.com/zama-ai/sdk/pull/227).

## SDK-85 P1 coverage

| Requirement | Status |
|---|---|
| Rewrite `userDecrypt` with full pipeline (normalize → zero → cache → creds → events → relayer → wrapError → cache write → return) | ✅ |
| `DecryptStart/End/Error` events emitted uniformly | ✅ |
| `wrapDecryptError` moved to `errors/decrypt.ts` | ✅ |
| `DecryptOptions` removed from `userDecrypt` signature | ✅ |
| `sdk.allow(contractAddresses)` added | ✅ |
| Contract address set derived from ALL input handles (before cache filtering) | ✅ |

## Pipeline architecture

```
ZamaSDK.userDecrypt ──────────► runUserDecryptPipeline
Token.decryptBalance ─────────►       │
Token.decryptHandles ─────────►       ├─ runCachePartitionPipeline
                                      ├─ runGroupByContractPipeline
                                      └─ relayer.userDecrypt (per contract)

Token.decryptBalanceAs ───────► runDelegatedDecryptPipeline
                                      ├─ runCachePartitionPipeline
                                      ├─ runGroupByContractPipeline
                                      └─ relayer.delegatedUserDecrypt (per contract)

Token.batchDecryptBalances ───► runBatchDecryptPipeline
Token.batchDecryptBalancesAs ─►       ├─ resolve handles
                                      ├─ filter zero/cached
                                      ├─ preflight check
                                      └─ inner pipeline (user or delegated)
```

## Test plan

- [ ] CI green
- [ ] Existing decrypt tests pass (no behavioral change)
- [ ] `sdk.userDecrypt` cache behavior unchanged
- [ ] `batchDecryptBalances` credential dedup unchanged (PR #227 fix preserved)
- [ ] `sdk.allow()` delegates to `credentials.allow` with normalization
- [ ] `allowMutationOptions` routes through `sdk.allow()`


packages/sdk/src/credentials/credentials-manager.ts
```diff
@@ -78,6 +78,9 @@ export class CredentialsManager extends BaseCredentialsManager<
    * @throws {@link SigningFailedError} if the signing operation fails for any other reason.
    */
   async allow(...contractAddresses: Address[]): Promise<StoredCredentials> {
+    if (contractAddresses.length === 0) {
+      throw new Error("credentials.allow() requires at least one contract address");
+    }
```
enitrat, line 83: I thought we fixed this at the type level

packages/sdk/src/pipelines/user-decrypt-pipeline.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  if (handles.length === 0) {
+    return {};
+  }
+
+  let result: Record<Handle, ClearValueType> = {};
+  const nonZero: DecryptHandleEntry[] = [];
+
+  for (const h of handles) {
+    if (h.handle === ZERO_HANDLE || h.handle === "0x") {
```
enitrat, line 58: is `0x` supposed to be a valid input?

enitrat (changes requested):
can you start first with a PR that only focuses on the `userDecrypt` part specifically, so it will be easier to review. Then we'll do migrations of internals to that new userDecypt, so as to minimize the amount of code to review at once.

On the `pipeline` approach: the `pipeline` is not a domain that owns anything nor is responsible for anything, so it should not exist.

- `runCachePartitionPipeline` is a loop on `cache.get()`. The responsibility of this operation is on the cache, and as such this should be a either i) cache method or ii) inlined inside userDecrypt if not reused elsewhere. It's even clearer when we see you need to pass a cache service as parameter. 
- `runGroupByContractPipeline`: I don't see why we'd introduce a new file + function for a three liner bloc of code that's never reused anywhere. should be inlined
- `runUserDecryptPipeline` is the only one that makes _some_ sense **if** we want to have a shared function for delegated and user decryption, but the issue is that it should not be part of a `pipeline`, but rather just an internal function of the SDK.

You don't need an orchestration pipeline with 5 files, what you need is a single function `sdk.userDecypt` that calls the cache + credential + relayer.

By adding `pipelines`, we create a fake module that duplicates responsibilities once again and instead of solving our initial problem, we're just moving it around

About the internals of the code, avoid multiple explicit loop iteration on the values where we could use simple functional operations

gasp (commented):
This impacts how we use the hooks and gitbook doc should be updated too 📙

enitrat:
the general thing to keep in mind is that one 50 lines function that you can read from start to finish in a single pass is way better than split into 5 10-liner internals that you split in 5 different files. I think the only reasonable extraction here is the `DecryptCache.partition`, which you'll reuse in multiple decryption flows, but the rest is not.

ghermet:
see #236

PR #235, opened 2026-04-14
feat(SDK-69): support both old and new wrapper interfaceIds during transition

zama-cremaud:
## Summary

Implements [SDK-69](https://linear.app/zama/issue/SDK-69/support-both-old-and-new-wrapper-interfaceids-during-transition) — dual wrapper interfaceId detection for the April 2026 protocol-apps security upgrade transition.

The upgrade changes the ERC-165 interface ID for `IERC7984ERC20Wrapper` from `0xd04584ba` (legacy) to `0x1f1c62b2` (new). Between testnet upgrade (April 21) and mainnet upgrade (April 28), the SDK must detect wrappers using either ID.

## Changes

**New constant**
- `ERC7984_WRAPPER_INTERFACE_ID_NEW = "0x1f1c62b2"` exported from `@zama-fhe/sdk` and `@zama-fhe/react-sdk`

**Dual-ID detection (core change)**
- `ReadonlyToken.isWrapper()` — checks both IDs in parallel via `Promise.all`, returns `true` if either matches
- `isWrapperQueryOptions()` (used by React `useIsWrapper` / `useIsWrapperSuspense`) — same dual-ID logic

**`isConfidentialWrapperContract()` unchanged** (intentional): it remains a valid low-level builder for the legacy ID. Its JSDoc now carries an explicit transition-period warning pointing callers to the higher-level APIs. This makes Phase C cleanup mechanical: remove the legacy-ID call from the two call sites.

**Documentation**
- `isConfidentialWrapperContract` JSDoc warns that the function only checks the legacy ID
- `supportsInterfaceContract` JSDoc references both interface ID constants
- `ReadonlyToken.isWrapper()` JSDoc describes the dual-ID parallel check

## Tests

- `ReadonlyToken.isWrapper()`: 3 cases — legacy-only match, new-only match, neither match
- `isWrapperQueryOptions()`: 5 cases — legacy-only match, new-only match, both false, contract revert (no ERC-165), network error; extracted into its own `describe` block

## Context

- Parent ticket: [SDK-68](https://linear.app/zama/issue/SDK-68/adapt-sdk-to-protocol-apps-breaking-changes-april-2026)
- Depends on: [SDK-61](https://linear.app/zama/issue/SDK-61/replace-local-zaiffer-contracts-with-currently-deployed-protocol-apps) (must be merged first)
- Phase C cleanup (post-mainnet): [SDK-72](https://linear.app/zama/issue/SDK-72/post-mainnet-cleanup-remove-old-interfaceid-event-signatures-and-deprecated-exports)

## Test plan

- [ ] All unit tests pass (`pnpm test:run`)
- [ ] Verify `isWrapper()` returns `true` for a pre-upgrade wrapper on testnet (responds to `0xd04584ba`)
- [ ] Verify `isWrapper()` returns `true` for a post-upgrade wrapper on testnet (responds to `0x1f1c62b2`)
- [ ] Verify `ERC7984_WRAPPER_INTERFACE_ID_NEW` is importable from both packages


packages/sdk/src/contracts/erc165.ts
```diff
@@ -4,14 +4,26 @@ import { erc165Abi } from "../abi/erc165.abi";
...
+/**
+ * ERC-165 interface ID for IERC7984ERC20Wrapper (confidential wrapper) — post-upgrade.
+ *
+ * Introduced in protocol-apps commit 93c6e7a (April 2026 security upgrade).
+ * During the transition period, both {@link ERC7984_WRAPPER_INTERFACE_ID} and this
+ * constant must be checked to support old and new wrappers simultaneously.
+ *
+ * @see {@link ERC7984_WRAPPER_INTERFACE_ID} for the legacy interface ID.
+ */
+export const ERC7984_WRAPPER_INTERFACE_ID_NEW = "0x1f1c62b2" as const;
```
ghermet: what happens when you deprecate the old interface id? are we gonna rename `ERC7984_WRAPPER_INTERFACE_ID_NEW` to `ERC7984_WRAPPER_INTERFACE_ID`?

should we do the opposite like `ERC7984_WRAPPER_INTERFACE_ID_LEGACY` so that when deprecated we're left with the new one `ERC7984_WRAPPER_INTERFACE_ID`

zama-cremaud: Yes, better indeed, updated

enitrat (approved):

PR #236, opened 2026-04-14
refactor(sdk): refactor userDecrypt with zero-handle filter, events, and error wrapping [SDK-85]

ghermet:
## Summary

- Refactors `ZamaSDK.userDecrypt` with zero-handle filtering, `DecryptStart/End/Error` events, and typed error wrapping via `wrapDecryptError`
- Removes `DecryptOptions` interface (callbacks replaced by the events system)
- Derives contract addresses from ALL input handles (not just uncached) for stable credential cache keys
- Adds `sdk.allow(contractAddresses)` public method with type-level non-empty enforcement
- Extracts `wrapDecryptError` to shared `errors/decrypt.ts` (used by both `zama-sdk.ts` and `readonly-token.ts`)

Addresses review feedback on PR #234. Focused on `userDecrypt` only — ReadonlyToken decrypt methods, batch decrypt, and delegated decrypt are follow-up work.

Closes [SDK-85 P1](https://linear.app/zama/issue/SDK-85/refactor-sdk-to-use-a-unique-decryption-flow)

## Test plan

- [x] 8 new tests: zero handles, mixed handles, DecryptStart/End events, DecryptError event, empty handles no events, stable credential key, incomplete results, allow() delegation
- [x] All 1529 existing tests pass
- [x] Typecheck clean
- [x] Build succeeds


packages/sdk/src/zama-sdk.ts
```diff
@@ -17,13 +18,8 @@ import { WrappersRegistry } from "./wrappers-registry";
...
 
-/** Options for {@link ZamaSDK.userDecrypt}. */
-export interface DecryptOptions {
-  /** Fired after credentials are ready (cached or freshly signed), before relayer calls. Not called when all handles are already cached. */
-  onCredentialsReady?: () => void;
-  /** Fired after all handles have been decrypted, including when all results come from cache. Receives the full result map. */
-  onDecrypted?: (values: DecryptResult) => void;
-}
+/** 32-byte zero handle — uninitialized encrypted balances map to `0n` without a relayer call. */
+const ZERO_HANDLE = "0x0000000000000000000000000000000000000000000000000000000000000000" as const;
```
enitrat: we already have two definitions for ZERO_HANDLE

packages/sdk/src/zama-sdk.ts
```diff
@@ -303,72 +322,120 @@ export class ZamaSDK {
...
+          type: ZamaSDKEvents.DecryptEnd,
+          durationMs: Date.now() - t0,
+          timestamp: Date.now(),
+        });
+        return result;
+      }
+
+      // Derive contract addresses from ALL handles for stable credential cache key
+      const allContractAddresses = new Set(handles.map((h) => getAddress(h.contractAddress)));
+      const creds = await this.credentials.allow(...allContractAddresses);
```
enitrat:
```suggestion
      const creds = await this.allow(...allContractAddresses);
```

packages/sdk/src/zama-sdk.ts
```diff
@@ -303,72 +322,120 @@ export class ZamaSDK {
...
+  async userDecrypt(handles: DecryptHandle[]): Promise<Record<Handle, ClearValueType>> {
     if (handles.length === 0) {
       return {};
     }
 
-    const signerAddress = await this.signer.getAddress();
-    const result: Record<Handle, ClearValueType> = {};
-    const uncached: DecryptHandle[] = [];
+    const t0 = Date.now();
+    this.#onEvent({ type: ZamaSDKEvents.DecryptStart, timestamp: t0 });
```
enitrat: should this be emitted as first part of the function? what is the intended use of those events? do we want emission in all cases, or just non-zero, or just non-cached?

packages/sdk/src/zama-sdk.ts
```diff
@@ -303,72 +322,120 @@ export class ZamaSDK {
...
+      });
+      return result;
+    } catch (error) {
+      this.#onEvent({
+        type: ZamaSDKEvents.DecryptError,
+        error: toError(error),
+        durationMs: Date.now() - t0,
+        timestamp: Date.now(),
+      });
+      throw wrapDecryptError(error, "Failed to decrypt handles");
```
enitrat, line 434: this global try-catch will swallow credential errors. is this intended?

packages/sdk/src/zama-sdk.ts
```diff
@@ -303,72 +322,120 @@ export class ZamaSDK {
...
+        }
+      }
+
+      // Verify completeness
+      for (const h of uncached) {
+        if (!(h.handle in result)) {
+          throw new DecryptionFailedError(
+            `Decryption returned no value for handle ${h.handle} on contract ${h.contractAddress}`,
+          );
+        }
```
enitrat: what is this defending against? a relayer bug where the result is not an error but the handle empty? can this happen?

packages/sdk/src/zama-sdk.ts
```diff
@@ -303,72 +322,120 @@ export class ZamaSDK {
...
+        if (existing) {
+          existing.push(h.handle);
+        } else {
+          byContract.set(h.contractAddress, [h.handle]);
+        }
+      }
+
+      // Decrypt per contract group
+      for (const [contractAddress, contractHandles] of byContract) {
+        const decrypted = await this.relayer.userDecrypt({
```
enitrat: this is running sequential calls for each (contract, handles) pair: should we use the existing pLimit util? that we can move to utils/concurrency.ts then

packages/sdk/src/zama-sdk.ts
```diff
@@ -287,12 +283,35 @@ export class ZamaSDK {
...
+   *
+   * @example
+   * ```ts
+   * // Sign once for three tokens, then decrypt individually
+   * await sdk.allow([cUSDT, cDAI, cWETH]);
+   * const a = await sdk.userDecrypt([{ handle: h1, contractAddress: cUSDT }]);
+   * const b = await sdk.userDecrypt([{ handle: h2, contractAddress: cDAI }]);
+   * ```
+   */
+  async allow(contractAddresses: [Address, ...Address[]]): Promise<void> {
```
enitrat: sorry to come back on this but is this not going to be problematic with dynamic arrays with an unknown amount of elements? 

```
const tokenAddresses: Address[] = [
  "addrA",
  "addrB",
];
allow(tokenAddresses);
```

this would fail because the array is not defined `as const`. so the most elegant signature is just 
```ts
async allow(contractAddresses: Address[]): Promise<void> {
  if (contractAddresses.length === 0) return;
  await this.credentials.allow(...normalized);
}
```

packages/sdk/src/zama-sdk.ts
```diff
@@ -303,72 +322,120 @@ export class ZamaSDK {
...
+        this.#onEvent({
+          type: ZamaSDKEvents.DecryptEnd,
+          durationMs: Date.now() - t0,
+          timestamp: Date.now(),
+        });
+        return result;
+      }
+
+      // Derive contract addresses from ALL handles for stable credential cache key
+      const allContractAddresses = new Set(handles.map((h) => getAddress(h.contractAddress)));
```
enitrat: second normalization, we should have one single top-level normalization of addresses as part of the first instructions

packages/sdk/src/__tests__/zama-sdk.test.ts
```diff
@@ -721,6 +723,142 @@ describe("ZamaSDK", () => {
       expect(result).toEqual({});
       expect(relayer.userDecrypt).not.toHaveBeenCalled();
     });
+
+    it("maps zero handles to 0n without hitting the relayer", async ({ sdk, relayer }) => {
+      const ZERO_HANDLE =
```
enitrat: import from shared constants

enitrat (changes requested):

packages/sdk/src/zama-sdk.ts
ghermet: Fixed — imported `ZERO_HANDLE` from `query/utils` instead of redeclaring.

packages/sdk/src/zama-sdk.ts
ghermet: Can't use `this.allow()` here — it returns `void`, and we need the `StoredCredentials` back from `credentials.allow()`. Kept `this.credentials.allow()` directly.

packages/sdk/src/zama-sdk.ts
ghermet: Fixed — events now only wrap the relayer calls. Zero-handle-only and all-cached paths return without emitting.

packages/sdk/src/zama-sdk.ts
ghermet: Narrowed the try-catch to only wrap the relayer calls. Credential errors (`SigningRejectedError`, etc.) now throw before the try block.

packages/sdk/src/zama-sdk.ts
ghermet: Removed. If the relayer returns partial results, the missing keys just won't be in the result object — callers can check for that themselves.

packages/sdk/src/zama-sdk.ts
ghermet: Added `pLimit` with concurrency of 5, imported from `token/concurrency.ts`.

packages/sdk/src/zama-sdk.ts
ghermet: Reverted to `Address[]` with an early `if (length === 0) return` guard.

packages/sdk/src/__tests__/zama-sdk.test.ts
ghermet: Fixed — tests now import `ZERO_HANDLE` from `query/utils`.

packages/sdk/src/zama-sdk.ts
ghermet: Fixed — addresses are now normalized once at the top of the method via a single `handles.map(h => ({ ...h, contractAddress: getAddress(h.contractAddress) }))`.

packages/sdk/src/zama-sdk.ts
enitrat: can you move token/concurrency.ts to utils/concurrency.ts ?

enitrat (dismissed):
💯

packages/sdk/src/zama-sdk.ts
ghermet: Moved `token/concurrency.ts` → `utils/concurrency.ts` (and its test).

enitrat (dismissed):

enitrat (approved):

PR #242, opened 2026-04-15
refactor(sdk): migrate Token/ReadonlyToken to sdk.userDecrypt [SDK-85]

ghermet:
## Summary

Implements [SDK-85](https://linear.app/zama/issue/SDK-85/refactor-sdk-to-use-a-unique-decryption-flow) **PR 2** — migrate `Token`/`ReadonlyToken` to delegate to the unified `sdk.userDecrypt` pipeline introduced in #236.

- New `ReadonlyToken(sdk, address)` / `Token(sdk, address, wrapper?)` constructors — replaces the 10-field `ReadonlyTokenConfig`/`TokenConfig`. Tokens now hold a single `sdk` reference instead of duplicating credentials/cache/relayer wiring.
- `balanceOf` fetches the on-chain handle and delegates decryption to `sdk.userDecrypt` + `assertBigint`.
- `batchBalancesOf` (static) replaces `batchDecryptBalances` with `{ results, errors }` return type and no `onError` callback.
- `allow()` delegates to `sdk.allow`.
- Removes `decryptBalance`, `decryptHandles`, and `#batchDecryptCore` — the cache/credentials/relayer skeleton lives only in `sdk.userDecrypt` now (invariants I2–I5).
- Hooks (`useConfidentialBalance`, `useConfidentialBalances`, `useActivityFeed`) collapse to single-query polling via the new entrypoints.

## Test plan

- [ ] `pnpm --filter @zama-fhe/sdk test`
- [ ] `pnpm --filter @zama-fhe/react-sdk test`
- [ ] `pnpm --filter @zama-fhe/sdk build` (API report check)


packages/sdk/src/query/confidential-balances.ts
```diff
@@ -53,17 +54,43 @@ export function confidentialBalancesQueryOptions(
...
+            perTokenBalances.set(token.address, value);
+          } catch (error) {
+            // Session-level failures (user rejected signature, SDK misconfigured)
+            // are not per-token failures — rethrow so the whole query errors out.
+            if (isSessionError(error)) {
+              throw error;
+            }
+            perTokenErrors.set(token.address, toError(error));
+          }
+        }),
```
ghermet: We can't delegate all of this to `userDecrypt` because of the per token errors and balances. Should make userDecrypt return a map of results and a map of errors since it's batching internally?

packages/react-sdk/src/balance/use-confidential-balance.ts
enitrat, line 1: internally we should wrap a simple query that just calls `token.balanceOf()`, removing the two internal queries, which is expressed explicitly in the plan

packages/sdk/src/query/confidential-balance.ts
```diff
@@ -36,7 +37,15 @@ export function confidentialBalanceQueryOptions(
...
-      return token.decryptBalance(keyHandle, keyOwner);
+      const decrypted = await token.sdk.userDecrypt([
+        { handle: keyHandle, contractAddress: token.address },
+      ]);
+      const value = decrypted[keyHandle];
+      if (value === undefined) {
+        throw new DecryptionFailedError(`Decryption returned no value for handle ${keyHandle}`);
+      }
+      assertBigint(value, "confidentialBalanceQueryOptions: result[handle]");
+      return value;
```
enitrat: token.balanceOf

packages/react-sdk/src/balance/use-confidential-balances.ts
enitrat, line 1: similar to above:  should be calling a unique useQuery calling ReadonlyToken.batchBalancesOf

packages/sdk/src/query/confidential-balance.ts
enitrat, line 1: I don't see the point in taking a handle as input here: 
- first we need to know whether we want this to be called `balanceOf`, `confidentialBalanceOf`, or something else. 
- then, considering that this is a token-specific query, this should only take owner as input and call internal .balanceOf()

packages/react-sdk/src/balance/use-confidential-balance.ts
enitrat: sorry, internally call the single query that calls token.balanceOf()

packages/sdk/src/query/confidential-balances.ts
```diff
@@ -19,83 +20,81 @@ export interface ConfidentialBalancesData {
...
       assertNonNullable(keyOwner, "confidentialBalancesQueryOptions: owner");
       assertNonNullable(keyHandles, "confidentialBalancesQueryOptions: handles");
-      const perTokenErrors = new Map<Address, Error>();
 
-      const raw = await ReadonlyToken.batchDecryptBalances(tokens, {
-        owner: keyOwner,
-        handles: keyHandles,
-        maxConcurrency: config?.maxConcurrency,
-        onError: (error, address) => {
-          perTokenErrors.set(address, error);
```
enitrat, line 63: why can this query not be a simple `batchBalancesOf` call?

enitrat (changes requested):
massive simplification possibility in the hook/query system by properly applying the simple 3-layer architecture we aim for (see tanstack-best-practices).

on top of that: `assertConfidentialBalance` should be a simple wrapper around `balanceOf`, or `confidentialBalanceOf`, whatever name we decide to use.

packages/sdk/src/query/confidential-balance.ts
ghermet: `handle` is strictly used for caching via the queryKey

packages/react-sdk/src/balance/use-confidential-balance.ts
enitrat: to address

packages/react-sdk/src/balance/use-confidential-balances.ts
enitrat: to address

packages/sdk/src/query/confidential-balance.ts
enitrat: why?

enitrat (commented):
- assertConfidentialBalance feedback not addressed
- useConfidentialBalance / useConfidentialBalances feedback not addressed nor justified properly
- confidentialBalanceQueryOptions still calling userDecrypt instead of going through the token instance

packages/sdk/src/query/confidential-balance.ts
ghermet: It's used as part of the 2 phase polling system in the confidential balance query which periodically polls the handles and only decrypt new ones, the handle passed to the `confidentialBalanceQueryOptions` is then used for caching. Do we still want this system or not?

ghermet: If you skip the on chain read you potentially end up decrypting the same handle twice

ghermet:
`refetchInterval` is now up to the developer for `useConfidentialBalance` and `useConfidentialBalances`

packages/sdk/src/utils/concurrency.ts
```diff
@@ -19,9 +21,11 @@ export async function pLimit<T>(
...
     while (index < fns.length) {
       const i = index++;
-      if (fns[i]) {
-        results[i] = await fns[i]();
-      }
+      // The bounds check above guarantees fns[i] exists; the non-null assertion
+      // makes a missing thunk fail loudly instead of silently leaving a hole
+      // in `results`.
+      assertNonNullable(fns[i], `pLimit: fns[${i}]`);
+      results[i] = await fns[i]();
```
enitrat: what is that for? in which scenario did we need it?

enitrat (dismissed):
👍

packages/sdk/src/utils/concurrency.ts
ghermet: Theoretically should never be an issue as it is wrapped in `while(index < fns.length){}`. when `i` is  bumped the index is still within array bounds. non null assertation asserts the function exists for a given `i` instead of skipping its execution

enitrat: so we can remove ?

ghermet: yes and no, it's either a silent `if(fns[i]){}` or a loud ` assertNonNullable(fns[i], `pLimit: fns[${i}]`);` before `results[i] = await fns[i]();`. assertNonNullable  would prevent an issue early should the while loop be faulty any point in the future

ghermet: There's typescript obviously for the `fns` signature but this would be a good way to enforce it

enitrat: it's an internal to the library that's only used as part of  user decrypt or batches, only called by us with controlled inputs, so I don't think this condition can ever be met. It seems that it's overly defensive programming

second question: have you looked into the `p-map` library?

enitrat (approved):

PR #250, opened 2026-04-16
feat(sdk): add sdk.publicDecrypt() as the primitive for public decryptions

enitrat:
## Summary

Expose `sdk.publicDecrypt()` as a first-class method on `ZamaSDK`, so callers no longer need to reach through `sdk.relayer` directly for public decryptions.

- **New `ZamaSDK.publicDecrypt(handles)`** — delegates to `relayer.publicDecrypt`, with empty-handle short-circuit and `DecryptionFailedError` wrapping
- **Refactor `Token.finalizeUnwrap()`** — replaces ~40 lines of inline decrypt logic + manual event emission with a single `sdk.publicDecrypt()` call
- **Update `publicDecryptMutationOptions`** — now calls `sdk.publicDecrypt()` instead of `sdk.relayer.publicDecrypt()`
- **Tests** — unit tests for the new method (delegation, empty handles, error wrapping) + updated fixtures to return per-handle clear values

## Test plan

- [x] Unit tests for `sdk.publicDecrypt()` (4 cases: delegation, empty handles, error wrap, re-throw)
- [x] Updated `Token.finalizeUnwrap()` tests reflect new error paths
- [x] Mock fixtures return realistic per-handle `clearValues`

Closes SDK-90

### DIFF
--- a/packages/sdk/src/token/token.ts
+++ b/packages/sdk/src/token/token.ts
@@ -1,4 +1,4 @@
-import { type Address, getAddress, type Hex, hexToBigInt } from "viem";
+import { type Address, getAddress, type Hex } from "viem";
 import {
   allowanceContract,
   approveContract,
@@ -48,6 +48,7 @@ import type {
   UnshieldOptions,
 } from "../types";
 import type { ZamaSDK } from "../zama-sdk";
+import { assertBigint } from "../utils/assertions";
 
 /**
  * ERC-20-like interface for a single confidential token.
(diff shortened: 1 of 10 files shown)


packages/sdk/src/token/token.ts
```diff
@@ -627,49 +628,12 @@ export class Token extends ReadonlyToken {
...
-        throw error;
-      }
-      throw new DecryptionFailedError("Failed to finalize unshield", {
-        cause: error,
-      });
-    }
-
+    const result = await this.sdk.publicDecrypt([burnAmountHandle]);
+    const clearValue = result.clearValues[burnAmountHandle];
+    assertBigint(clearValue, "clearValue");
```
ghermet:
```suggestion
    assertBigint(clearValue, "finalizeUnwrap: clearValue");
```

ghermet (approved):

PR #251, opened 2026-04-16
fix: centralize ZERO_HANDLE, use isZeroHandle checks

enitrat:
Closes SDK-91

### DIFF
--- a/packages/sdk/src/query/utils.ts
+++ b/packages/sdk/src/query/utils.ts
@@ -64,6 +64,13 @@ export const ZERO_HANDLE =
   "0x0000000000000000000000000000000000000000000000000000000000000000" as const;
 
 /**
+ * Check whether a handle represents the zero value.
+ */
+export function isZeroHandle(handle: string): boolean {
+  return handle === ZERO_HANDLE || handle === "0x";
+}
+
+/**
  * Remove TanStack behavioral options from a query config object so only domain
  * parameters remain for the lower-level factory.
  *
--- a/packages/sdk/src/query/utils.ts
+++ b/packages/sdk/src/query/utils.ts
@@ -60,16 +60,6 @@ export type StrippedQueryOptionKeys =
   | "query"
   | "pollingInterval";
 
-export const ZERO_HANDLE =
-  "0x0000000000000000000000000000000000000000000000000000000000000000" as const;
-
-/**
- * Check whether a handle represents the zero value.
- */
-export function isZeroHandle(handle: string): boolean {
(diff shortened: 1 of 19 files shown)


packages/sdk/src/query/utils.ts
```diff
@@ -63,6 +63,13 @@ export type StrippedQueryOptionKeys =
 export const ZERO_HANDLE =
   "0x0000000000000000000000000000000000000000000000000000000000000000" as const;
 
+/**
+ * Check whether a handle represents the zero value.
+ */
+export function isZeroHandle(handle: string): boolean {
+  return handle === ZERO_HANDLE || handle === "0x";
+}
```
ghermet: is this a query util? shouldn't this sit under `/utils/handle.ts`?

packages/sdk/src/query/utils.ts
```diff
@@ -63,6 +63,13 @@ export type StrippedQueryOptionKeys =
 export const ZERO_HANDLE =
   "0x0000000000000000000000000000000000000000000000000000000000000000" as const;
 
+/**
+ * Check whether a handle represents the zero value.
+ */
+export function isZeroHandle(handle: string): boolean {
```
ghermet: should this be `boolean` -> `handle is ZERO_HANDLE | handle is "0x"`

packages/sdk/src/query/utils.ts
enitrat: oh yeah I put in the wrong utils thanks

packages/sdk/src/query/utils.ts
enitrat: wdym? it's what it's doing right now

ghermet:
```ts
/**
 * Check whether a handle represents the zero value.
 */
export function isZeroHandle(
  handle: string,
): handle is typeof ZERO_HANDLE | "0x" {
  return handle === ZERO_HANDLE || handle === "0x";
}
```

ghermet: Image

ghermet: it's a conventional way to use `isSomething` in [typecript](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

enitrat: ah, yes ok, well I don't really see a use for having typesafe zero handles right now so I'd tend to say it's better kept as a simple boolean

ghermet (approved):

PR #256, opened 2026-04-16
docs: decrypt UX lifecycle guide (SDK-42)

ghermet:
## Summary

- Add a standalone guide (`decrypt-ux-lifecycle.md`) documenting the `isAllowed` → disabled hook → explicit action pattern as the recommended approach for displaying confidential balances without blind-sign wallet popups
- Cross-reference from `check-balances.md`, `useIsAllowed.md` reference, and guides overview
- Register in `SUMMARY.md`

Closes SDK-42 (documentation and Claude skill portions). Example apps already implement the correct pattern.

**Note:** The `fhevm-developer` Claude skill was also updated with a "Frontend Integration: Decrypt UX Lifecycle" section, but that file lives outside this repo (in the plugins marketplace) and is applied locally.

## Test plan

- [ ] Verify guide renders correctly in GitBook
- [ ] Cross-reference links resolve from check-balances, useIsAllowed, and guides README
- [ ] Code example in guide compiles (uses correct SDK imports and API)


enitrat (commented):
we already have a section:

> #### Pre-authorize once, decrypt anywhere
> 
> A common pattern is to call `useAllow` once — for example right after the user connects their wallet — and then decrypt independently in any component, on any page, without triggering another wallet prompt. Credentials persist in IndexedDB, so they survive page reloads.


inside the encrypt-decrypt guide: rather than another guide I think we should consolidate that one, if its current state is not sufficient.

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -245,42 +245,108 @@ function ConfidentialAction() {
 
 {% endcode %}
 
-### 3. Decrypt with useUserDecrypt
+### 3. Avoid blind-sign wallet popups
```
enitrat:
```suggestion
### 3. Decrypt with useUserDecrypt
```

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -245,42 +245,108 @@ function ConfidentialAction() {
...
 
-`useUserDecrypt` is a query hook that decrypts FHE handles. If no cached credentials exist, it triggers a wallet signature prompt. Use `useAllow` + `useIsAllowed` to pre-authorize and gate the query if you want to control when the prompt appears.
+Any decrypt operation (`useUserDecrypt`, `useConfidentialBalance`) triggers an EIP-712 wallet signature the first time it runs — the SDK needs FHE decrypt credentials. If your app calls these hooks on render without gating, users see an unsolicited MetaMask popup before they have taken any action. In crypto UX this is a **blind-signing anti-pattern**: users are trained to reject unexpected signature requests, and security tools like Blockaid may flag them.
```
enitrat:
```suggestion
Operations performing decryptions like`useUserDecrypt` or `useConfidentialBalance` might trigger EIP-712 wallet signature when ran for the first time. If your app calls these hooks on render without gating, users might an unsolicited MetaMask popup before they have taken any action. In crypto UX this is a **blind-signing anti-pattern**: users are trained to reject unexpected signature requests, and it's confusing to know what is being signed exactly.
```

docs/gitbook/src/guides/encrypt-decrypt.md
ghermet: This isn't about `useUserDecrypt`, the topic here is how to avoid blind signing when interacting with the following queries.

enitrat: The title and content of that section should not be specifically about "avoiding blind signing", but decryption of data altogether, including mentions of proper UX for that. 


other suggestion if not specifically for useUserDecrypt:
```suggestion
### 3. Decryption of the encrypted data
```

enitrat (approved):

PR #257, opened 2026-04-16
refactor(sdk): read cache-identity params from context.queryKey

enitrat:
Closes SDK-92

## Context

Our TanStack Query factories follow a convention where `queryFn` extracts data-fetching parameters from `context.queryKey` rather than from the surrounding closure. This is a defensive best practice for externalized, reusable query factories — a stale closure could otherwise fetch data for parameter X while the cache entry is keyed by parameter Y.

## Changes

Aligns four query factories with the `confidential-balance.ts` reference pattern:

- `packages/sdk/src/query/is-allowed.ts` — destructure `contractAddresses` from `context.queryKey` (cast to non-empty tuple for the SDK signature).
- `packages/sdk/src/query/user-decrypt.ts` — destructure `handles` from `context.queryKey` (cast to `DecryptHandle[]`).
- `packages/sdk/src/query/wrapper-discovery.ts` — destructure `erc20Address` from `context.queryKey`, keep the existing guard.
- `packages/sdk/src/query/wrappers-registry.ts` (`listPairsQueryOptions`) — destructure `page`, `pageSize`, `metadata` from `context.queryKey`; inline the defaults into the key construction to avoid outer-scope shadowing.

## Test Plan

- [x] `pnpm typecheck` passes
- [x] `pnpm test:run packages/sdk/src/query` — 162/162 tests pass (existing tests already invoke `queryFn` with the real `queryKey` via `mockQueryContext`)

### DIFF
--- a/packages/sdk/src/query/wrapper-discovery.ts
+++ b/packages/sdk/src/query/wrapper-discovery.ts
@@ -44,11 +44,12 @@ export function wrapperDiscoveryQueryOptions(
   return {
     ...filterQueryOptions(config.query ?? {}),
     queryKey,
-    queryFn: async () => {
-      if (!config.erc20Address) {
+    queryFn: async (context) => {
+      const [, { erc20Address }] = context.queryKey;
+      if (!erc20Address) {
         throw new Error("erc20Address is required for wrapper discovery query");
       }
-      const result = await registry.getConfidentialToken(config.erc20Address);
+      const result = await registry.getConfidentialToken(erc20Address);
       return result ? result.confidentialTokenAddress : null;
     },
     staleTime: Infinity,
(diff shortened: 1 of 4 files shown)


packages/sdk/src/query/wrapper-discovery.ts
```diff
@@ -44,11 +44,12 @@ export function wrapperDiscoveryQueryOptions(
   return {
     ...filterQueryOptions(config.query ?? {}),
     queryKey,
-    queryFn: async () => {
-      if (!config.erc20Address) {
+    queryFn: async (context) => {
+      const [, { erc20Address }] = context.queryKey;
+      if (!erc20Address) {
         throw new Error("erc20Address is required for wrapper discovery query");
       }
```
ghermet:
```suggestion
     assertNonNullable(erc20Address, "wrapperDiscoveryQueryOptions: erc20Address")
}
```

ghermet (dismissed):

ghermet (approved):

PR #28, opened 2026-02-27
feat: expose WASM thread pool option for RelayerWeb [SDK-10]

ghermet:
## Summary

- Adds a `threads` option to `RelayerWebConfig` that controls the number of WASM threads for parallel FHE operations via `wasm-bindgen-rayon`
- Threads the option through the worker client and into `initSDK({ thread })` in the Web Worker
- Documents the COOP/COEP header requirements for `SharedArrayBuffer` support

## Test plan

- [ ] Verify RelayerWeb initializes correctly without `threads` option (single-threaded default)
- [ ] Verify RelayerWeb initializes with `threads: N` and passes thread count to `initSDK`
- [ ] Confirm existing tests pass with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #261, opened 2026-04-17
fix(security): harden example relayer proxies

zama-cremaud:
## Summary

Harden the three Next.js example relayer proxy routes while keeping the examples simple.

- Validate that the configured relayer URL uses `http` or `https` before proxying requests.
- Keep the existing path-segment allowlist and encode path segments when building the upstream URL.
- Avoid private-network detection in example apps; the remaining SSRF-style alerts were dismissed because `RELAYER_URL` is server-controlled configuration and user-controlled path segments are allowlisted and encoded.

## Security Alerts

Wiz SAST SSRF findings dismissed as false positives:

- #27: examples/react-ethers/src/app/api/relayer/[...path]/route.ts
- #28: examples/react-viem/src/app/api/relayer/[...path]/route.ts
- #29: examples/react-wagmi/src/app/api/relayer/[...path]/route.ts

## Validation

- `pnpm exec oxfmt --check` on modified files
- `git diff --check`
- `pnpm typecheck` in each affected example currently fails on pre-existing `src/app/page.tsx` errors unrelated to these proxy route changes.


enitrat:
I don't think this is desirable as part of the examples which should be kept simple and stragithforward: can we just hide those alerts?

examples/react-ethers/src/app/api/relayer/[...path]/route.ts
```diff
@@ -47,6 +49,54 @@ const RESPONSE_DROP = new Set([...HOP_BY_HOP, "content-encoding", "content-lengt
...
+    host.startsWith("fe80:") ||
+    /^f[cd][0-9a-f]{2}:/u.test(host)
+  ) {
+    return true;
+  }
+
+  const parts = host.split(".").map((part) => Number(part));
+  if (
+    parts.length !== 4 ||
+    parts.some((part) => !Number.isInteger(part) || part < 0 || part > 255)
```
Comment (ghermet) [inline examples/react-ethers/src/app/api/relayer/[...path]/route.ts]: This looks a bit OTT to check localhost in an example app. checking hostname is `localhost` or `127.0.0.1` should be good enough

ghermet (approved):

ghermet (approved):

enitrat (commented):
I still don't get why we're doing env var validation of the supplied proxy URL inside the logic of the code itself. For what purpose? We don't do this in our prod apps.

zama-cremaud:
> I still don't get why we're doing env var validation of the supplied proxy URL inside the logic of the code itself. For what purpose? We don't do this in our prod apps.

Yes, that's fair, closing this PR.

PR #263, opened 2026-04-17
chore: remove lint / typecheck edit hook in claude-setup

enitrat:
why: because not all edit can pass typecheck/lints in the context of the entire codebase. Having a run of typecheck/lint on every edit is both 1) inefficient and 2) injects irrelevant context before we actually _want_ to run those commands, flooding the context window with possible, expected errors. They should be kept as precommit only.

I kept format because it's purely cosmetic; but it's debatable given that it will change the file and will cause subsequent edit to fail on the same blocks.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
 pnpm exec lint-staged
+pnpm typecheck
--- a/claude-setup/settings.json
+++ b/claude-setup/settings.json
@@ -43,24 +43,6 @@
         "hooks": [
           {
             "type": "command",
-            "command": "pnpm typecheck"
-          }
-        ]
-      },
-      {
-        "matcher": "Write|Edit",
-        "hooks": [
-          {
-            "type": "command",
-            "command": "pnpm lint"
-          }
-        ]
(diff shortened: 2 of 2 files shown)


ghermet:
If you leave it up to Claude, it will disregard typescript errors and not follow the oxlint coding guidelines resulting in a failed CI pipeline. You will have to manually instruct Claude to fix the problems

enitrat:
> If you leave it up to Claude, it will disregard typescript errors and not follow the oxlint coding guidelines resulting in a failed CI pipeline

Claude cannot commit something that doesn't pass the precommit hooks, so this will not happen

ghermet:
Lint and typecheck not running in our pre commit hook is it

enitrat:
> Lint and typecheck not running in our pre commit hook is it

you added lints in precommits a few months ago, typechecks could be added as well. It's debatable to add it as claude config so as to have fast feedback loop, but in practice it confuses the model

enitrat:
added typecheck in husky config

ghermet (approved):

PR #268, opened 2026-04-17
feat: introduce concrete signer / provider split

enitrat:
## Summary

Splits the monolithic `GenericSigner` into a read-only `GenericProvider` and a write-only `GenericSigner`, then mirrors that split across every adapter family (viem, ethers, wagmi) and through the React SDK layer. All public chain reads now flow through `sdk.provider.*`; wallet-authority operations stay on `sdk.signer.*`.

Parent epic: SDK-34. `ZamaSDKConfig.signer` remains required — read-only SDK mode is tracked separately.

### New interfaces and errors

- **`GenericProvider`** (`types/provider.ts`): `getChainId`, `readContract`, `waitForTransactionReceipt`, `getBlockTimestamp`.
- **`GenericSigner` narrowed** (`types/signer.ts`): `getChainId`, `getAddress`, `signTypedData`, `writeContract`, optional `subscribe?()`. Read methods removed.
- **`ChainMismatchError`** (`errors/chain.ts`): carries `operation`, `signerChainId`, `providerChainId`. Thrown by `ZamaSDK.requireChainAlignment(operation)` at the top of every write and decrypt method when the signer and provider target different chains.

### Adapter pairs

Each framework adapter is now split into a provider + signer class:

| Framework | Provider | Signer |
|-----------|----------|--------|
| viem | `ViemProvider` (publicClient) | `ViemSigner` (walletClient only, publicClient removed) |
| ethers v6 | `EthersProvider` (Provider or EIP-1193) | `EthersSigner` (Signer or EIP-1193) |
| wagmi | `WagmiProvider` (Config) | `WagmiSigner` (Config) |

### ZamaSDK

- `sdk.provider` field added (required).
- `requireChainAlignment(operation)` gate added at the top of every public write/decrypt entry point: `confidentialTransfer`, `confidentialTransferFrom`, `approve`, `shield`, `unwrap`, `unwrapAll`, `finalizeUnwrap`, `approveUnderlying`, `delegateDecryption`, `revokeDelegation`, `userDecrypt`, `publicDecrypt`, `revokeSession`, `batchBalancesOf`, `batchDecryptBalancesAs`, `decryptBalanceAs`.
- `revokeSession` reuses the aligned chain ID from `requireChainAlignment` instead of calling `signer.getChainId()` again.

### Token / ReadonlyToken

- `Token` holds a non-nullable `signer` field resolved once at construction, collapsing all former `requireSigner()` checks.
- All `readContract` calls migrated to `sdk.provider.readContract`.
- All `waitForTransactionReceipt` calls migrated to `sdk.provider.waitForTransactionReceipt`.
- All `getBlockTimestamp` calls migrated to `sdk.provider.getBlockTimestamp`.

### Query factories

10 query factories migrated from `sdk.signer.readContract` to `sdk.provider.readContract`. Mutation factories simplified to single-expression arrow returns.

### WrappersRegistry

`WrappersRegistry` constructor now takes `{ provider }` instead of the full SDK. All internal reads routed through the provider.

### React SDK

- `ZamaProvider` now requires a `provider` prop alongside `signer`.
- New `useSignerAddress` / `useSignerAddressSuspense` hooks exported for explicit signer-address resolution.
- `useConfidentialIsApprovedSuspense` uses `useSignerAddressSuspense` internally (aligned with `useUnderlyingAllowanceSuspense` pattern).
- All read hooks migrated to `sdk.provider.*`.

### Test infrastructure

- `createMockSigner` returns only write methods; `createMockProvider` returns only read methods — `sdk.signer.readContract` does not typecheck, enforcing the split at compile time.
- `chain-alignment.test.ts` (new): 9 tests covering `requireChainAlignment` mechanism + behavioral tests for `shield`, `confidentialTransfer`, `unwrap`, `delegateDecryption`, `userDecrypt`, `decryptBalanceAs`, `batchBalancesOf`.
- Type-level tests (`signer.test-d.ts`) updated to assert `waitForTransactionReceipt` and `getBlockTimestamp` live on `GenericProvider`, not `GenericSigner`.

### Docs / cleanup

- JSDoc examples in `contracts/*.ts` updated from `signer.readContract` to `provider.readContract`.
- Stale imports, unused fixture params, and dead `ZamaErrorCode.SignerRequired` enum entry removed.

## Out of scope

- Optional signer / read-only SDK mode (signer stays required).
- Explicit `owner`/`holder` on suspense hooks (follow-up aligned with read-only SDK work).
- Provider-scoped query keys (cache key shape unchanged).
- Wrapper-discovery `tokenAddress` cleanup.
- Gitbook / migration guide (SDK-104).


packages/react-sdk/src/__tests__/wagmi-signer-subscribe.test.ts
```diff
@@ -106,9 +107,10 @@ describe("WagmiSigner.subscribe", () => {
   });
 });
 
-describe("WagmiSigner.getBlockTimestamp", () => {
-  wit("returns block timestamp from getBlock", async ({ wagmiSigner }) => {
-    const timestamp = await wagmiSigner.getBlockTimestamp();
+describe("WagmiProvider.getBlockTimestamp", () => {
+  wit("returns block timestamp from getBlock", async () => {
+    const wagmiProvider = new WagmiProvider({ config: {} as unknown as Config });
```
ghermet: I think we need to add `wagmiProvider` to the fixtures so it can reuse base config

packages/react-sdk/src/provider.tsx
```diff
@@ -135,17 +151,19 @@ export function ZamaProvider({
   // and is idempotent.
   useEffect(() => () => sdk.dispose(), [sdk]);
 
-  return <ZamaSDKContext.Provider value={sdk}>{children}</ZamaSDKContext.Provider>;
+  const contextValue = useMemo(() => ({ sdk }), [sdk]);
+
+  return <ZamaSDKContext.Provider value={contextValue}>{children}</ZamaSDKContext.Provider>;
```
ghermet: what is the point of creating an object when `sdk` is the only attribute?

zama-cremaud:
3 topics / small questions but overall LGTM!

  1. Silent wallet errors
  When signer.getAddress() fails, useSignerAddress() returns undefined forever without exposing any error. Balance hooks stay stuck in a loading state with no way for the UI to tell "still resolving" from "wallet disconnected". Should the hook surface error or isError instead of just hiding it?                                                                                                                
                                                                                                                                                                                                                  
  2. The owner !== undefined guard is copy-pasted everywhere
  Every hook that needs the owner address has to manually add && owner !== undefined to its enabled condition. The factory explicitly documents "consumers must do this themselves." One forgotten guard in a future hook = silent broken query. Why not wrap this in a single shared helper?                                                                                                                                 
                                                                                                                                                                                                                  
  3. React Query for address resolution is the wrong tool                                                                                                                                                         
  signer.getAddress() is a one-shot call, not a data-fetching concern. Using React Query brings automatic retries (makes no sense for a disconnected wallet), stale-time management, and manual cache invalidation on account change. A simple useState + useEffect in ZamaProvider would be more predictable and easier to reason about, wdyt ?

packages/sdk/src/zama-sdk.ts
```diff
@@ -478,15 +529,17 @@ export class ZamaSDK {
...
    *
+   * @throws {@link ChainMismatchError} if signer and provider are on different chains.
+   *
    * @example
    * ```ts
    * wallet.on("disconnect", () => sdk.revokeSession());
    * ```
    */
   async revokeSession(): Promise<void> {
+    const chainId = await this.requireChainAlignment("revokeSession");
```
ghermet: [AUTOMATED] **Regression: `revokeSession` now throws on the exact disconnect/chain-switch scenario it documents as its primary use case.**

The method's own JSDoc example (line 536) advertises it as a disconnect handler:

```ts
wallet.on("disconnect", () => sdk.revokeSession());
```

Before this PR, the body was resilient to that scenario:

```ts
const chainId = this.#lastChainId ?? (await this.signer.getChainId());
```

— preferring the cached chain id captured at the last successful identity read, and falling back to a live `signer.getChainId()` only as a last resort.

After this PR, the first line is `await this.requireChainAlignment("revokeSession")`, which unconditionally calls both `signer.getChainId()` and `provider.getChainId()` in parallel and throws `ChainMismatchError` if they disagree. That breaks the documented use case in two ways:

1. **On `wallet.disconnect`**, a wagmi/ethers signer typically rejects `getChainId()` outright (connector torn down). The promise rejects before any cleanup runs. The `disconnect` handler is usually fire-and-forget (`() => sdk.revokeSession()`), so the rejection is silently dropped, leaving the session signature and plaintext cache on disk past logout.
2. **On a chain switch**, the provider and signer will routinely disagree for a short window while the wallet is reconnecting. An app that calls `revokeSession` in `onChainChange` (a natural pattern) will see `ChainMismatchError` and skip the revocation — exactly the state where revocation matters.

`revokeSession` is a local-only cleanup operation — it computes a store key, calls `credentials.revokeByKey`, and `cache.clearForRequester`. None of that touches the chain. Chain alignment is pure overhead here, and it adds a failure mode to a cleanup path that previously had none.

There is no test in `chain-alignment.test.ts` or `zama-sdk.test.ts` covering `revokeSession` with a disconnected signer or a chain mismatch — so the regression isn't caught by the new suite.

**Suggested fix**: drop `requireChainAlignment` from `revokeSession` and restore the cached fallback:

```ts
async revokeSession(): Promise<void> {
  await this.#identityReady;
  const address = this.#lastAddress ?? (await this.signer.getAddress());
  const chainId = this.#lastChainId ?? (await this.signer.getChainId());
  const storeKey = await CredentialsManager.computeStoreKey(address, chainId);
  await this.credentials.revokeByKey(storeKey);
  await this.cache.clearForRequester(address);
}
```

If you want alignment checked on the non-disconnect path, at least `catch` and fall through to the cached identity so disconnect-time cleanup isn't blocked.

A regression test covering `revokeSession` on a mismatched signer/provider would also be worth adding to `chain-alignment.test.ts`.

Confidence: 92/100

ghermet (commented):
[AUTOMATED]

## Code Review Summary

Reviewed the signer/provider split across adapters, the `requireChainAlignment` wiring at every write/decrypt entry point, and the React SDK migration. The split itself is clean: reads consistently flow through `sdk.provider.*`, writes through `sdk.signer.*`, mock fixtures enforce the separation at compile time, and the new `chain-alignment.test.ts` covers the core gate plus key behavioral cases.

One concern worth addressing before merge — flagged inline on `revokeSession`. Everything else looks good.

Checked: bug patterns, silent failures, guideline compliance, `.api.md` consistency, test coverage of the new gate.

packages/react-sdk/src/provider.tsx
enitrat: at some point this was taking both `sdk, signerAddress`, which required memoization, that I reverted and didnt clean here properly. fixed

enitrat:
> When signer.getAddress() fails, useSignerAddress() returns undefined forever without exposing any error...

updated to return all query data

> The owner !== undefined...

it's actually a desired feature; if no holder is provided and the signer can't be resolved, the query must not run (because of a lack of input data). The long term desired shape for this is to remove internal resolving of the owner via the "signer" (because the signer will eventually be optional) and remove internal queries of useSignerAddress altogether.


> signer.getAddress() is a one-shot call, not a data-fetching concern

for this, invalidation is already pluggin in the wallet lifecycle, and overall I don't see a strong reason not to use tanstack query here as it helps with our hooks composition. Note that this hook currently only exists for internal resolution of signer addresses which I am not sure to keep in a future release

packages/react-sdk/src/__tests__/wagmi-signer-subscribe.test.ts
enitrat: done

packages/react-sdk/src/provider.tsx
enitrat: at some point this was taking both sdk, signerAddress, which required memoization, that I reverted and didnt clean here properly. fixed

packages/sdk/src/zama-sdk.ts
```diff
@@ -537,12 +537,9 @@ export class ZamaSDK {
    * ```
    */
   async revokeSession(): Promise<void> {
-    await this.requireChainAlignment("revokeSession");
+    const chainId = await this.requireChainAlignment("revokeSession");
```
ghermet, line 540: This should probably be reverted to ` this.#lastChainId ?? this.signer.getChainId()`

ghermet (approved):

zama-cremaud (approved):

PR #270, opened 2026-04-20
fix(sdk): choose total supply method via ERC-165

zama-cremaud:
# Summary
Uses wrapper ERC-165 interface detection to choose the correct total supply read during the protocol-apps transition.

- Calls `inferredTotalSupply()` for upgraded wrappers.
- Calls legacy `totalSupply()` for pre-upgrade wrappers.
- Restores `totalSupplyContract()` as a legacy `totalSupply()` builder.
- Keeps wrapper version detection internal to avoid expanding the public API surface.
- Documents the upgraded wrapper interface ID source, verified against the recent protocol-apps upgraded wrapper line (`35dbefb...`).


packages/sdk/src/query/wrapper-interface-version.ts
```diff
@@ -0,0 +1,47 @@
...
+    safeSupportsInterface(signer, wrapperAddress, ERC7984_WRAPPER_INTERFACE_ID),
+    safeSupportsInterface(signer, wrapperAddress, ERC7984_WRAPPER_INTERFACE_ID_LEGACY),
+  ]);
+
+  if (supportsUpgraded) {
+    return "upgraded";
+  }
+  if (supportsLegacy) {
+    return "legacy";
+  }
```
ghermet:
```suggestion
  const supportsUpgraded = await safeSupportsInterface(signer, wrapperAddress, ERC7984_WRAPPER_INTERFACE_ID)
  if (supportsUpgraded) {
    return "upgraded";
  }
  
 const supportsLegacy = await safeSupportsInterface(signer, wrapperAddress, ERC7984_WRAPPER_INTERFACE_ID_LEGACY)
  if (supportsLegacy) {
    return "legacy";
  }
```

zama-cremaud: Yes, better indeed, applied.

ankurdotb:
### Code review

Found 3 issues (1 blocker, 2 minor):

1. **Blocker — legacy wrapper detection is silently broken.** `detectWrapperInterfaceVersion` uses `ERC7984_WRAPPER_INTERFACE_ID_LEGACY = 0xf1f4c25a`, but per the [breaking-changes announcement](https://www.notion.so/zamaai/3365a7358d5e809e9a67e3b3a5b01630) and [SDK-68](https://linear.app/zama/issue/SDK-68) the on-chain legacy ID is `0xd04584ba`. For every pre-upgrade wrapper, both `safeSupportsInterface` calls return `false` and this throws `ConfigurationError` instead of falling back to `totalSupply()` — the opposite of Phase B intent. Tracked as [SDK-113](https://linear.app/zama/issue/SDK-113/wrapper-legacy-erc-165-interface-id-is-wrong-0xf1f4c25a-0xd04584ba); the wrong value was introduced in SDK-61 (commit [`dd4364d8`](https://github.com/zama-ai/sdk/commit/dd4364d82d72d9a7774aeb1bd870f8b6f7e38ff8)) and propagated through SDK-69. Unit tests miss it because they mock `supportsInterface` using the same constant they assert against.

https://github.com/zama-ai/sdk/blob/8338d810684655d292e2dd4d5c8c8912ec8485a4/packages/sdk/src/query/wrapper-interface-version.ts#L31-L49

2. **Inline comment is slightly inaccurate.** `total-supply.ts:27` says ERC-165 detection *"adds two RPC calls per refetch"*, but after the short-circuit in [`8338d810`](https://github.com/zama-ai/sdk/commit/8338d810684655d292e2dd4d5c8c8912ec8485a4) upgraded wrappers only incur one call. Suggest: *"adds one or two RPC calls per refetch (one for upgraded wrappers, two for legacy)"*.

https://github.com/zama-ai/sdk/blob/8338d810684655d292e2dd4d5c8c8912ec8485a4/packages/sdk/src/query/total-supply.ts#L24-L32

3. **Worth reconsidering: sequential vs parallel `supportsInterface` calls.** The final `perf` commit replaced `Promise.all` with a sequential short-circuit. Every other dual-ID check in the codebase uses `Promise.all` ([`is-confidential.ts:64`](https://github.com/zama-ai/sdk/blob/8338d810684655d292e2dd4d5c8c8912ec8485a4/packages/sdk/src/query/is-confidential.ts#L62-L70), [`readonly-token.ts:150`](https://github.com/zama-ai/sdk/blob/8338d810684655d292e2dd4d5c8c8912ec8485a4/packages/sdk/src/token/readonly-token.ts#L148-L160)). During the April 21 → April 28 transition window, mainnet wrappers are still legacy, so every `totalSupply` read pays 2× RPC latency. Post-mainnet the short-circuit wins. Worth a second look given that Phase C ([SDK-72](https://linear.app/zama/issue/SDK-72)) removes the legacy branch entirely shortly after.

https://github.com/zama-ai/sdk/blob/8338d810684655d292e2dd4d5c8c8912ec8485a4/packages/sdk/src/query/wrapper-interface-version.ts#L29-L49

Recommend blocking merge on [SDK-113](https://linear.app/zama/issue/SDK-113/wrapper-legacy-erc-165-interface-id-is-wrong-0xf1f4c25a-0xd04584ba) landing first.

packages/sdk/src/query/__tests__/total-supply.test.ts
```diff
@@ -1,14 +1,126 @@
...
+  return vi.fn(async (config: ReadContractConfig) => {
+    if (config.functionName === "supportsInterface") {
+      if (config.args?.[0] === ERC7984_WRAPPER_INTERFACE_ID) {
+        return supportsUpgraded;
+      }
+      if (config.args?.[0] === ERC7984_WRAPPER_INTERFACE_ID_LEGACY) {
+        return supportsLegacy;
+      }
+      return false;
+    }
```
ghermet, line 32:
```suggestion
    if (config.functionName !== "supportsInterface") {
      return false;
    }
    if (config.args?.[0] === ERC7984_WRAPPER_INTERFACE_ID) {
      return supportsUpgraded;
     }
    if (config.args?.[0] === ERC7984_WRAPPER_INTERFACE_ID_LEGACY) {
       return supportsLegacy;
     }
```

zama-cremaud: Implemented, thanks

ghermet (approved):

ankurdotb (approved):

PR #271, opened 2026-04-20
fix(sdk): correct legacy wrapper interface ID

zama-cremaud:
## Summary

- Correct the public legacy wrapper ERC-165 interface ID from `0xf1f4c25a` to `0xd04584ba` for SDK-113.
- Keep the upgraded wrapper ID as `0x1f1c62b2` and update README, VERSIONS, tests, and API report.
- Keep wrapper detection strict: high-level APIs now check only the documented legacy `0xd04584ba` and upgraded `0x1f1c62b2` IDs.

## Validation

- `pnpm exec vitest run packages/sdk/src/contracts/__tests__/contracts.test.ts packages/sdk/src/query/__tests__/is-confidential.test.ts packages/sdk/src/token/__tests__/token.test.ts`
- `pnpm api-report:check`
- `pnpm lint`
- `pnpm typecheck`

Note: `pnpm e2e:test:node` could not complete locally because this temporary worktree's Solidity dependency install is missing nested OpenZeppelin imports; it did not reach an SDK assertion failure. CI should validate this from a clean install.


ankurdotb:
### Code review (updated for 3682412a)

Follow-up to my earlier review on d19420d0: both prior issues are resolved. [`72a0a52d`](https://github.com/zama-ai/sdk/commit/72a0a52d) removed `ERC7984_WRAPPER_INTERFACE_ID_INTERMEDIATE`; [`7aeb98cf`](https://github.com/zama-ai/sdk/commit/7aeb98cf) reverted the `confidentialWrapperInterfaceContracts` helper and restored the original `Promise.all` dual-ID check in `isWrapper()` / `isWrapperQueryOptions` with the corrected `_LEGACY` value. The PR 270 coordination issue falls out automatically.

One new issue found:

1. **Deleting `isWrapper()` e2e assertions regresses the exact coverage SDK-113 was about.** Commit [`3682412a`](https://github.com/zama-ai/sdk/commit/3682412add6ec3975ea7bc20795ce879099ce5b9) drops three pre-existing e2e assertions (introduced in [PR #178](https://github.com/zama-ai/sdk/pull/178), present on `prerelease`) that exercise `isWrapper()` against the deployed local fixture:

   - `readonlyUSDT.isWrapper()` in the Node bootstrap spec
   - `metadata-is-wrapper` on the cUSDT token-metadata page
   - `metadata-is-wrapper` on the cERC20 token-metadata page

   Rationale I can infer from the diffs: the vendored `ConfidentialWrapper` in `protocol-apps-wrapper-da4afe387420` reports an interface ID computed from source XOR that differs from the documented `0xd04584ba`, so the assertion fails against the local fixture. The earlier commit [`32852774`](https://github.com/zama-ai/sdk/commit/3285277483082397aba7bc44a728c570083ae2cb) had the correct fix — a minimal `SdkLocalConfidentialWrapper` that overrides `supportsInterface` to emit `0xd04584ba`, aligning the fixture with the documented deployed baseline (see `contracts/src/SdkLocalConfidentialWrapper.sol` in that commit). `3682412a` reverted that and deleted the assertions instead.

   [SDK-113](https://linear.app/zama/issue/SDK-113/wrapper-legacy-erc-165-interface-id-is-wrong-0xf1f4c25a-0xd04584ba) explicitly diagnoses the bug class as "unit tests mock `supportsInterface` using the same constant they are meant to validate". Deleting the e2e assertions keeps us in that regime — after this PR, `isWrapper()` has no verification against any actual contract, even a fixture. Recommend restoring `SdkLocalConfidentialWrapper` from `32852774` and keeping the three e2e assertions. If the override feels out of place long-term, tie its removal to a TODO against the SDK-70/71 protocol-apps version bump that will align source XOR with the deployed ID.

https://github.com/zama-ai/sdk/blob/3682412add6ec3975ea7bc20795ce879099ce5b9/test/playwright/tests/node/server-bootstrap.node.spec.ts#L65-L75

https://github.com/zama-ai/sdk/blob/3682412add6ec3975ea7bc20795ce879099ce5b9/test/playwright/tests/token-metadata.spec.ts#L1-L22

enitrat (changes requested):
It seems that we're only replacing one hardcoded value by another, and the test verifying that it is the same value is a bit tautologeous.

We already have a bunch of typescript-abi files available at `packages/sdk/src/abi` we could repurpose and fix this issue for good: we _must not_ have hardcoded ABIs and values corresponding to specific contract versions.

Rather, we need a codegen step (e.g. `generate:abi`) that will be called in our `pnpm build` pipeline, that will:
- build the artifacts of the relevant contracts with foundry
- outputs the `abi/contract.ts` files with contract ABIs and relevant INTERFACE_ID 
This way, we'll be sure that the interface ID we have in the SDK code _always_ match those of our contracts, and will be auto-updated with our contracts.

for that, use the Wagmi CLI: https://wagmi.sh/cli/why

ankurdotb:
Do the contracts deployed/being tested return true for `interfaceId = 0xf1f4c25a` still? Where did that come from? I want to understand the root cause of it, because these wrapper ERC-165s are not based on static values but [derived rather from the specific structure of the contract deployed](https://zama-ai.slack.com/archives/C0A2WRRAM7F/p1771585297589219?thread_ts=1771582976.583349&cid=C0A2WRRAM7F).

So if it was off initially, that could point towards an underlying issue.

zama-cremaud:
> It seems that we're only replacing one hardcoded value by another, and the test verifying that it is the same value is a bit tautologeous.
> 
> We already have a bunch of typescript-abi files available at `packages/sdk/src/abi` we could repurpose and fix this issue for good: we _must not_ have hardcoded ABIs and values corresponding to specific contract versions.
> 
> Rather, we need a codegen step (e.g. `generate:abi`) that will be called in our `pnpm build` pipeline, that will:
> 
> * build the artifacts of the relevant contracts with foundry
> * outputs the `abi/contract.ts` files with contract ABIs and relevant INTERFACE_ID
>   This way, we'll be sure that the interface ID we have in the SDK code _always_ match those of our contracts, and will be auto-updated with our contracts.
> 
> for that, use the Wagmi CLI: https://wagmi.sh/cli/why

I agree with the direction. The current fix still relies on a manually maintained interface ID, and the constant equality test does not prevent future drift. 
Now: should we keep this PR as the SDK-113 hotfix and address the generator in a focused follow-up? It's kind of blocking the rest of the "Adapt SDK to protocol-apps breaking changes" process.

enitrat (approved):
> should we keep this PR as the SDK-113 hotfix and address the generator in a focused follow-up

ok for me, please create associated linear ticket

PR #278, opened 2026-04-23
fix(sdk): run delegation check before cache lookup in decryptBalanceAs

ghermet:
## Summary

- **`decryptBalanceAs`** — pre-flight delegation check ran after the cache lookup, so a revoked delegation returned the stale cached plaintext. Moved the check above the cache lookup so revocations are enforced on every call.
- **`batchDecryptBalancesAs`** — had the same shape: it short-circuited with `return results` when every balance resolved from cache, skipping the delegation check. Moved the check ahead of the cache lookups, gated on "at least one non-zero handle" so batches of zero balances still skip the RPC.

## Scope

- [x] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [ ] Docs

## Validation

- `pnpm test:run packages/sdk/src/token` — 260/260 pass. New regression tests:
  - `re-checks delegation on cache hit and throws when revoked` (singular)
  - `runs pre-flight delegation check even when balance is pre-cached` (batch, happy path)
  - `throws DelegationNotFoundError on cache hit when delegation is revoked` (batch, revoked)
- `pnpm typecheck` — clean.
- `pnpm lint` — clean.
- `pnpm format:check` on changed files — clean.

## Tradeoff

The batch method loses an optimization that avoided one RPC round-trip when every balance in the batch was already cached. That optimization was explicitly called out as intentional in the old comment, but it was also the mechanism through which revoked delegations could leak stale data, so the revocation guarantee wins.

### DIFF
--- a/docs/gitbook/src/reference/sdk/delegation.md
+++ b/docs/gitbook/src/reference/sdk/delegation.md
@@ -235,7 +235,7 @@ These are caught **before** submitting a transaction, saving gas and providing a
 | `DelegationSelfNotAllowedError`         | Delegate address equals the connected wallet (`delegate === msg.sender`)      |
 | `DelegationDelegateEqualsContractError` | Delegate address equals the token contract address                            |
 | `DelegationExpiryUnchangedError`        | New expiration date matches the current one (no on-chain change needed)       |
-| `DelegationNotFoundError`               | Attempting to revoke a delegation that was never established (expiry is zero) |
+| `DelegationNotFoundError`               | Revoking a delegation that was never established, or reading via `decryptBalanceAs` / `batchDecryptBalancesAs` when the delegation is missing or has been revoked (expiry is zero) |
 
 ### On-chain revert errors
 
--- a/docs/gitbook/src/reference/sdk/delegation.md
+++ b/docs/gitbook/src/reference/sdk/delegation.md
@@ -229,12 +229,12 @@ Because the ACL contract resets the expiry to `0n` on revocation, `DelegationNot
 
 These are caught **before** submitting a transaction, saving gas and providing actionable messages:
 
-| Error                                   | When                                                                          |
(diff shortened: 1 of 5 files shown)


ankurdotb:
### Code review

Found 2 issues:

1. The `DelegationNotFoundError` row in the delegation error table is scoped only to the revoke path — "Attempting to revoke a delegation that was never established (expiry is zero)". After this PR, the same error also fires from `decryptBalanceAs` / `batchDecryptBalancesAs` on cache hits when a delegation has been revoked. Worth broadening the description alongside this change.

https://github.com/zama-ai/sdk/blob/c0e10c94663909eb6e5f6ab1cbbb56ddd9f62ee1/docs/gitbook/src/reference/sdk/delegation.md#L236-L240

2. Same gap in the error reference page: "Thrown when attempting to revoke a non-existent delegation" — now also thrown from the delegated decrypt path (even on cache hits) once this PR lands.

https://github.com/zama-ai/sdk/blob/c0e10c94663909eb6e5f6ab1cbbb56ddd9f62ee1/docs/gitbook/src/reference/sdk/errors.md#L397-L406

Both are soft suggestions. The gap pre-dates this PR (`#assertDelegationActive` already threw this error on cache-miss paths), and [delegation.md:224](https://github.com/zama-ai/sdk/blob/c0e10c94663909eb6e5f6ab1cbbb56ddd9f62ee1/docs/gitbook/src/reference/sdk/delegation.md#L222-L226) already correctly notes the error covers revoked cases. No blocking issues; core change looks clean.

ankurdotb (approved):

PR #279, opened 2026-04-23
ci(api-report): split api-report PR comment into per-file diffs

ghermet:
## Summary
- Splits the API report PR comment into per-file collapsible sections instead of one monolithic diff
- Makes it easier to review which specific API surface files changed

## Test plan
- [ ] Verify the api-report workflow runs on a PR with API changes and produces per-file collapsible sections
- [ ] Verify PRs without API changes still get the "no changes" comment


ankurdotb (commented):
Dislike the inline scripting as it's an anti-pattern for workflows: That is too much logic to live inline in YAML: it's not unit-testable, not runnable locally, and not type/lint-checked.

So I'm pushing an update to make it a script under `pnpm api-report:*` namespace (similar pattern to `pnpm llm:*` workspace, which solves some of those issue above.

Check it for correctness on what you were trying to achieve.

ankurdotb (approved):

PR #280, opened 2026-04-23
refactor(sdk): remove viem as hard dependency

ghermet:
## Summary
- Remove `viem` from `dependencies`, keep it as an optional `peerDependency` (`>=2`) alongside `ethers` (`>=6`)
- Replace all direct `viem` imports with lightweight local utilities backed by `@noble/hashes`
- Add local `Address` type, `getAddress`, `isAddress`, `isHex`, `toHex` helpers and an ERC-20 ABI module

## Motivation
`viem` is a large dependency that pulls in significant bundle weight. Users who only use `ethers` were forced to install it anyway. This change makes both `viem` and `ethers` peer-only, so consumers pay only for the library they actually use.

## Key changes
| Area | Before | After |
|------|--------|-------|
| `dependencies` | `viem ^2.47` | `@noble/hashes ^2.2` |
| `peerDependencies` | `ethers >=6` | `ethers >=6`, `viem >=2` (both optional) |
| Internal imports | `from "viem"` | `from "../utils/address"` / `"../utils/hex"` |
| New files | — | `utils/address.ts`, `types/ethereum.ts`, `abi/erc20.abi.ts` |

## Test plan
- [ ] Unit tests pass (`pnpm vitest`)
- [ ] Typecheck passes (`pnpm typecheck`)
- [ ] E2E tests with ethers-only consumer
- [ ] E2E tests with viem-only consumer
- [ ] Bundle size comparison before/after

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -12,6 +12,6 @@
   {
     "name": "@zama-fhe/react-sdk (all JS)",
     "path": "packages/react-sdk/dist/**/*.js",
-    "limit": "15 KB"
+    "limit": "7 KB"
   }
 ]
--- a/packages/sdk/etc/sdk-ethers.api.md
+++ b/packages/sdk/etc/sdk-ethers.api.md
@@ -10,23 +10,43 @@ import { Bytes32Hex } from '@zama-fhe/relayer-sdk/bundle';
 import { ContractFunctionArgs } from 'viem';
 import { ContractFunctionName } from 'viem';
 import { ContractFunctionReturnType } from 'viem';
-import { EIP1193EventMap } from 'viem';
-import { EIP1193Events } from 'viem';
-import { EIP1193Provider } from 'viem';
 import { ethers } from 'ethers';
 import { Hex } from 'viem';
 import { KmsDelegatedUserDecryptEIP712Type } from '@zama-fhe/relayer-sdk/bundle';
 import { KmsUserDecryptEIP712Type } from '@zama-fhe/relayer-sdk/bundle';
-import { ProviderConnectInfo } from 'viem';
-import { ProviderMessage } from 'viem';
-import { ProviderRpcError } from 'viem';
 import { Signer } from 'ethers';
 
-export { EIP1193EventMap }
+// @public
+export interface EIP1193EventMap {
+    // Warning: (ae-forgotten-export) The symbol "Address$1" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    accountsChanged: (accounts: Address$1[]) => void;
+    // (undocumented)
+    chainChanged: (chainId: string) => void;
+    // (undocumented)
+    connect: (connectInfo: ProviderConnectInfo) => void;
+    // (undocumented)
+    disconnect: (error: ProviderRpcError) => void;
+    // (undocumented)
+    message: (message: ProviderMessage) => void;
+}
 
-export { EIP1193Events }
+// @public
+export interface EIP1193Events {
+    // (undocumented)
+    on<TEvent extends keyof EIP1193EventMap>(event: TEvent, listener: EIP1193EventMap[TEvent]): void;
+    // (undocumented)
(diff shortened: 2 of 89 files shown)


enitrat:
> viem is a large dependency that pulls in significant bundle weight

can you show me a concrete number on how much lighter the bundle would be ? I don't want to move away from using viem internally. At best it's an installation-only added weight, not a bundle weight, which our bundler should be optimizing properly.

ghermet:
> > viem is a large dependency that pulls in significant bundle weight
> 
> can you show me a concrete number on how much lighter the bundle would be ? I don't want to move away from using viem internally. At best it's an installation-only added weight, not a bundle weight, which our bundler should be optimizing properly.

it's not just a bundle size thing, having a direct deps on viem clashes with using type primitives with the new zama config-> typescript can't resolve types for publicClient / walletClient with 2 versions of viem 

Screenshot 2026-04-23 at 17 38 09

enitrat:
> having a direct deps on viem clashes with using type primitives with the new zama config-> typescript can't resolve types for publicClient / walletClient with 2 versions of viem

wouldn't it be better to solve this with loosening the viem requirement to >=2 ?

ghermet:
> > having a direct deps on viem clashes with using type primitives with the new zama config-> typescript can't resolve types for publicClient / walletClient with 2 versions of viem
> 
> wouldn't it be better to solve this with loosening the viem requirement to >=2 ?



> > having a direct deps on viem clashes with using type primitives with the new zama config-> typescript can't resolve types for publicClient / walletClient with 2 versions of viem
> 
> wouldn't it be better to solve this with loosening the viem requirement to >=2 ?

Could also put `viem: >=2` as a hard dep to dedup versions. We barely use viem in the sdk, is it really worth keeping it as a dep when noble/hashes does most what we need?

enitrat:
yes

ghermet:
> yes

What's the rational for keeping viem as hard deps if i can replace the few functions we use with noble/hashes library they use with this little friction?
Screenshot 2026-04-24 at 12 40 26

enitrat:
The arguments in favor would be:
- it reduces bundle weight (but, we saw that it doesn't)
- it causes dependency versioning conflicts (but, we saw that it's a problem of allowed version ranges more than anything, that can easily be solved)

The cost, on the other hand, is that:
- we re-implement core ethereum primitives that we're not sure are properly tested, instead of relying on a multi-year old production library
- we increase the surface of the code we're responsible of: every edge case in checksumming addresses, ABI types, etc... that we'd need to cover
- it adds unnecessary workload for no clear gains

So it's not clear what our benefits are here, while it's clear what the downsides are

ghermet:
> The arguments in favor would be:
> 
> * it reduces bundle weight (but, we saw that it doesn't)
> * it causes dependency versioning conflicts (but, we saw that it's a problem of allowed version ranges more than anything, that can easily be solved)
> 
> The cost, on the other hand, is that:
> 
> * we re-implement core ethereum primitives that we're not sure are properly tested, instead of relying on a multi-year old production library
> * we increase the surface of the code we're responsible of: every edge case in checksumming addresses, ABI types, etc... that we'd need to cover
> * it adds unnecessary workload for no clear gains
> 
> So it's not clear what our benefits are here, while it's clear what the downsides are

We're not re-implementing anything we're using the same `@noble/hashes` that viem uses internally, just without the viem wrapper around it. Our `getAddress` calls keccak256 from `@noble/hashes/sha3` — the exact same codepath viem takes. The "untested re-implementation" concern doesn't apply when the core crypto is the same audited library.

The real issue isn't bundle weight it's that viem as a hard dep creates a type-level coupling. When a user has viem@2.21 and our SDK pins viem@2.47, TypeScript resolves two incompatible copies and PublicClient/WalletClient types stop unifying. We solved this with >=2 in the peer range, but making it a hard dep re-introduces the duplication risk whenever npm/pnpm hoists differently. Keeping viem as a peer-only dep means: (1) ethers-only users don't install viem at all, (2) viem users get their own version without conflicts, (3) we depend on 1 small audited library (@noble/hashes) instead of 1 large one. The 4 utility functions we wrote are ~50 lines total and tested.

PR #281, opened 2026-04-24
feat: remove auto-signer resolving in hooks

enitrat:
## Summary

Make the account identifier a caller-supplied parameter on read hooks instead of implicitly resolving it from the connected signer. This gives callers full control over which address is queried and eliminates hidden signer-address resolution, so that we can remove required signers for those queries.

## Changes

- `useConfidentialBalance` / `useConfidentialBalances`: add required `account` param
- `useConfidentialIsApproved` (+Suspense): add required `holder` param
- `useUnderlyingAllowance` (+Suspense): add required `owner` param
- Move auto-gating on `owner` into the Layer 2 query-options factories (`confidentialBalanceQueryOptions`, `confidentialBalancesQueryOptions`)
- Remove internal `useSignerAddress` calls from all affected hooks
- Non-suspense variants accept `Address | undefined` and auto-disable the query when unset (wagmi `useBalance` convention)
- Suspense variants require a defined value (no `enabled` on `useSuspenseQuery`)
- Update all test components to pass `useAccount().address` explicitly

### DIFF
--- a/test/test-components/src/unshield-form.tsx
+++ b/test/test-components/src/unshield-form.tsx
@@ -2,6 +2,7 @@
 
 import { useUnshield, useConfidentialBalance, useMetadata } from "@zama-fhe/react-sdk";
 import type { Address } from "@zama-fhe/sdk";
+import { useAccount } from "wagmi";
 
 export function UnshieldForm({
   tokenAddress,
@@ -10,8 +11,9 @@ export function UnshieldForm({
   tokenAddress: Address;
   wrapperAddress?: Address;
 }) {
+  const { address } = useAccount();
   const { data: metadata } = useMetadata(tokenAddress);
-  const { data: balance } = useConfidentialBalance({ tokenAddress });
+  const { data: balance } = useConfidentialBalance({ tokenAddress, account: address });
   const unshield = useUnshield({ tokenAddress, wrapperAddress });
 
   return (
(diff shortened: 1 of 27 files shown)


test/test-components/src/unwrap-manual-form.tsx
```diff
@@ -8,6 +8,7 @@ import {
   useConfidentialBalance,
   useMetadata,
 } from "@zama-fhe/react-sdk";
+import { useAccount } from "wagmi";
```
ghermet, line 11: use the `wagmi/compat.ts` for `connection/account` related things to avoid issues between v2/v3

test/test-components/src/unshield-form.tsx
```diff
@@ -2,6 +2,7 @@
 
 import { useUnshield, useConfidentialBalance, useMetadata } from "@zama-fhe/react-sdk";
 import type { Address } from "@zama-fhe/sdk";
+import { useAccount } from "wagmi";
```
ghermet, line 5: use the wagmi/compat.ts for connection/account related things to avoid issues between v2/v3

test/test-components/src/unwrap-manual-form.tsx
enitrat: this is in the test components, which is ~similar to what a consumer code would look like. I don't think this should end up in consumer code?

ghermet: My bad test is not an issue

ghermet (approved):

PR #282, opened 2026-04-24
feat(sdk): require explicit owner on token reads

enitrat:
Mirror the React hook change (#281) at the SDK layer: remove the implicit signer fallback on trivial `ReadonlyToken`/`Token` reads so callers always supply the owner/holder address.

- `ReadonlyToken.balanceOf(owner)`, `confidentialBalanceOf(owner)`, `allowance(wrapper, owner)`, `batchBalancesOf(tokens, owner)` — owner now required
- `Token.isApproved(spender, holder)` — holder now required
- `#assertConfidentialBalance` internal caller resolves the signer address once and passes it to `balanceOf`
- `confidentialBalance{,s}QueryOptions` keep `account` optional at the factory level (auto-gated via `enabled`) and pass it through to the SDK call
- Tests and `etc/*.api.md` updated accordingly

### DIFF
--- a/packages/sdk/src/token/token.ts
+++ b/packages/sdk/src/token/token.ts
@@ -322,23 +322,19 @@ export class Token extends ReadonlyToken {
    * Check if a spender is an approved operator for a given holder.
    *
    * @param spender - The address to check operator approval for.
-   * @param holder - The token holder address. Defaults to the connected wallet.
+   * @param holder - The token holder address.
    * @returns `true` if the spender is an approved operator for the holder.
    *
    * @example
    * ```ts
-   * if (await token.isApproved("0xSpender")) {
-   *   // spender can call transferFrom on behalf of connected wallet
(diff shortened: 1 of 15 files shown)


packages/sdk/src/token/token.ts
```diff
@@ -322,23 +322,19 @@ export class Token extends ReadonlyToken {
    * Check if a spender is an approved operator for a given holder.
    *
    * @param spender - The address to check operator approval for.
-   * @param holder - The token holder address. Defaults to the connected wallet.
+   * @param holder - The token holder address.
```
ghermet, line 325: holder / owner ? which one is it?

enitrat: the solidity entrypoint decided to use `holder` so we're going to align on this. I made a bigger Linear ticket to address any other disrepancies in the future.

ankurdotb (approved):

PR #283, opened 2026-04-24
feat: optional signer [SDK-109]

enitrat:
## Summary

Make `signer` optional across the SDK so the core is constructible and useful before a wallet exists. Public reads, relayer operations, and token metadata work without a signer. Signer-requiring operations fail at invocation with a typed `SignerRequiredError(operation)`.

### Core SDK

- `ZamaSDKConfig.signer` is optional. When absent, `credentials` and `delegatedCredentials` are `undefined`, no lifecycle subscription is registered.
- Three guard methods — `requireSigner`, `requireCredentials`, `requireDelegatedCredentials` — centralize the check and throw `SignerRequiredError` with the caller's operation name. Every write/sign path in `Token` and `ReadonlyToken` calls through these guards before doing async work.
- New `SignerRequiredError` error class with code `SIGNER_REQUIRED`.

### Identity-change events (replaces lifecycle callbacks)

The previous `SignerLifecycleCallbacks` interface (`onDisconnect`, `onAccountChange`, `onChainChange`) and the `signerLifecycleCallbacks` config option have been removed. Signer adapters now emit a single `SignerIdentityChange` event carrying `{ previous?, next? }` identity snapshots. The SDK subscribes once and fans out to registered listeners via `sdk.onIdentityChange(listener)` after running core cleanup (credential revocation for the previous identity + cache clear).

This replaces the former `#initIdentity` / `#lastAddress` / `#lastChainId` state tracking: cleanup now targets the identity carried in the event rather than SDK-tracked state, eliminating the partial-state window that existed during asynchronous account/chain transitions.

### Query layer

- `isAllowedQueryOptions` no longer requires an `account` argument. It auto-disables when `sdk.credentials` is `undefined` (`staleTime: 0`, `gcTime: 0`), so it naturally idles pre-connect without needing a signer-scope intermediary.
- `invalidateWalletLifecycleQueries` removes `isAllowed` cache entries outright on identity change (in addition to decryption cache), preventing stale `true` reads between disconnect and the next refetch.
- `signerAddressQueryOptions` has been removed; signer address resolution is handled at the React layer.

### React SDK

- `ZamaProviderProps.signer` is optional.
- `useSignerAddress` / `useSignerAddressSuspense` removed. The React layer hooks into `sdk.onIdentityChange` directly to invalidate wallet-lifecycle queries on every identity transition.
- `useIsAllowed` auto-idles when no signer is configured by relying on the query factory's `enabled` flag rather than a separate hook.

### Wagmi integration — `ZamaWagmiProvider`

Introduced `ZamaWagmiProvider` as the single public entry point for wagmi React apps:

```tsx
<WagmiProvider config={wagmiConfig}>
  <ZamaWagmiProvider relayer={relayer} storage={storage}>
    <App />
  </ZamaWagmiProvider>
</WagmiProvider>
```

It reads wagmi's connection state internally and passes `signer={undefined}` to `ZamaProvider` until wagmi reports a concrete wallet identity. This eliminates the "stable disconnected signer" problem: consumers can no longer accidentally pass a `WagmiSigner` that exists while wagmi is disconnected.

Both `WagmiSigner` and `WagmiProvider` (the read-only chain adapter) are now internal — not exported from `@zama-fhe/react-sdk/wagmi`. `ZamaWagmiProvider` constructs both adapters from `useConfig()`, so there is no consumer use case for manual instantiation. Keeping them internal prevents the main footgun and avoids the naming collision between Zama's `WagmiProvider` class and wagmi's own `<WagmiProvider>` component.

### Signer adapters

- `ViemSigner` and `WagmiSigner` now emit `SignerIdentityChange` events instead of the former `onDisconnect`/`onAccountChange`/`onChainChange` callbacks.
- Both throw `SignerRequiredError("getAddress")` instead of `TypeError` when no account is available.

## What's not in this PR

- Operation-boundary error normalization: the React render race (wagmi disconnects synchronously, React re-renders asynchronously) can still surface `SignerRequiredError("getAddress")` instead of `SignerRequiredError(operation)` during the gap. A follow-up should catch adapter-level `SignerRequiredError` in write paths and re-throw with the operation name.
- `requireChainAlignment` now calls `requireSigner` internally, so it throws `SignerRequiredError` before the chain-ID comparison — the call-site guard in each Token method is therefore redundant for chain alignment but retained for clarity at the operation boundary.


ghermet:
isAllowed is inherently a signer question — "does my wallet have a cached session for these contracts?" Without a signer, the answer is always false. You could make the query return false directly when there's no signer instead of disabling:
```ts
  queryFn: () => {
    if (!sdk.credentials) return false;
    return sdk.credentials.isAllowed(contractAddresses);
  },
  enabled: callerEnabled, // no signer check needed
```

enitrat:
> isAllowed is inherently a signer question — "does my wallet have a cached session for these contracts?" Without a signer, the answer is always false. You could make the query return false directly when there's no signer instead of disabling:

I was thinking that having the query in a disabled state might be easier UX/UI-wise, as you could render it in a disabled state instead of having it returning false. Also for coherency with other queries, I think it's good

ghermet:
[AUTOMATED]

## Code Review

**Verdict: No issues found** ✓

Reviewed for bugs, security issues, guideline compliance, and error handling across all 50 changed files (900 additions, 387 deletions).

### What was checked
- **Bug hunting** (2 independent passes): No runtime bugs found. The signer optionality guards (`requireSigner`/`requireCredentials`/`requireDelegatedCredentials`) are consistently applied across all write paths. Query `enabled` gating correctly disables signer-dependent queries when credentials are absent. Identity change handling, disposal, and cleanup are sound.
- **Guideline compliance**: TanStack Query factory pattern followed correctly (user options spread first, factory-critical props override). `"use client"` directives present on all React files. JSDoc conventions followed on new public exports.
- **Error handling**: No silent failures introduced. The `try/finally` pattern in `revokeSession` correctly ensures cache clearing. The `swallow` usage in `#handleIdentityChange` follows established project patterns.

### Architecture notes
- Clean separation: construction without signer works, signer-required ops fail-fast with typed errors
- `ZamaWagmiProvider` correctly centralizes wagmi lifecycle, preventing the "stable disconnected signer" footgun
- Query key simplification (removing `account` from `isAllowed` scope) is safe given `removeQueries` on identity change

### Minor observation (non-blocking)
- `WagmiSigner.getAddress` changed error message from `"Wagmi config is not connected"` to `"Invalid address"` — less descriptive, but since `WagmiSigner` is now internal-only, impact is minimal.

zama-cremaud:
Conflicts to be resolved but otherwise LGTM!

zama-cremaud (dismissed):

zama-cremaud:
# PR Review: feat: optional signer [SDK-109] (#283)

**Date:** 2026-04-27  
**Reviewer:** Automated (Claude)  
**Base:** `prerelease` → `feat/optional-signers`  
**Head SHA:** `c93af9d91be780406ac755434e3fe77709d73d4c`

---

## 1. Requirement Coverage

**Source:** Linear SDK-109 + PR description

| ID | Requirement | Status | Evidence |
|----|-------------|--------|----------|
| R-1 | `signer` optional in `ZamaSDKConfig` | SATISFIED | `zama-sdk.ts:54` (`signer?: GenericSigner`) |
| R-2 | Public reads work without signer | SATISFIED | `optional-signer.test.ts` lines 35–40 |
| R-3 | `createReadonlyToken` / `createToken` work without signer | SATISFIED | `optional-signer.test.ts` lines 42–56 |
| R-4 | Signer-required ops throw `SignerRequiredError` | SATISFIED | `zama-sdk.ts:181–210`, tested via `optional-signer.test.ts` |
| R-5 | `credentials` / `delegatedCredentials` absent when no signer | SATISFIED | `zama-sdk.ts:167–170` |
| R-6 | React provider mounts with `signer={undefined}` | SATISFIED | `provider.tsx:43`, `optional-signer.test.tsx` |
| R-7 | `useIsAllowed` auto-idles when no signer | SATISFIED | `is-allowed.ts:36` (`enabled: ... && sdk.credentials !== undefined`) |
| R-8 | Central guards `requireSigner/Credentials/DelegatedCredentials` | SATISFIED | `zama-sdk.ts:181–210` |
| R-9 | `SignerRequiredError` with `operation` field | SATISFIED | `errors/signer.ts` |
| R-10 | `ZamaWagmiProvider` as single wagmi entry point | SATISFIED | `wagmi/zama-wagmi-provider.tsx`, internal exports |
| R-11 | `signerAddressQueryOptions` / `useSignerAddress` removed | SATISFIED | confirmed deleted |
| R-12 | `isAllowedQueryOptions` no longer requires `account` arg | SATISFIED | `is-allowed.ts:7–16` |
| R-13 | Identity-change clears `isAllowed` cache (not just invalidates) | SATISFIED | `invalidation.ts:99` (`removeQueries`) |

All SDK-109 requirements are satisfied.

---

## 2. New High-Signal Findings

### [NON-BLOCKING] PR description claims signer adapters throw `SignerRequiredError` — they still throw `TypeError`

**Confidence: 95**

The PR description states under "Signer adapters":
> "Both throw `SignerRequiredError("getAddress")` instead of `TypeError` when no account is available."

The actual code does not match this:

**`WagmiSigner.getAddress()` (`wagmi-signer.ts:53`):**
```ts
// before:
throw new TypeError("Wagmi config is not connected");
// after (this PR):
throw new TypeError("Invalid address");
```

**`ViemSigner.getAddress()` (`viem-signer.ts:57–59`):**
```ts
if (!this.#walletClient.account) {
  throw new TypeError("WalletClient has no account");
}
```

Neither adapter throws `SignerRequiredError`. The PR description is inaccurate about this point. The note under "What's not in this PR" addresses only the *operation name propagation* race, not the missing type change.

Additionally, the error message change `"Wagmi config is not connected"` → `"Invalid address"` is a regression in debuggability. The new message loses the actionable context about wagmi state (even though `WagmiSigner` is now internal, it still surfaces in stack traces).

**Impact:** When the wagmi disconnect→React re-render race occurs, callers catch `TypeError` instead of the documented `SignerRequiredError`. This breaks `instanceof SignerRequiredError` guards in application code.

**Suggested fix:**
```ts
// wagmi-signer.ts
async getAddress(): Promise<Address> {
  const account = getConnection(this.#config);
  if (!account?.address) {
    throw new SignerRequiredError("getAddress");
  }
  return account.address;
}
```

---

### [NON-BLOCKING] `ZamaWagmiProvider` — new `ZamaSDK` instance created on every connect/disconnect

**Confidence: 85**

In `zama-wagmi-provider.tsx`:
```tsx
const signer = useMemo(
  () => (hasSigner ? new WagmiSigner({ config: wagmiConfig }) : undefined),
  [hasSigner, wagmiConfig],
);
```

`hasSigner` is a boolean. When it toggles (`false → true` on connect or `true → false` on disconnect), a new `WagmiSigner` instance is created. `ZamaProvider` then creates a new `ZamaSDK` (its `useMemo` depends on `signer`). This means:
- The decrypt cache is destroyed on every disconnect
- The credentials manager is re-created on every connect
- All `onIdentityChange` subscriptions are re-registered

For most apps this is fine (the old SDK disposes cleanly via the `useEffect(() => () => sdk.dispose(), [sdk])` cleanup). However, apps that experience frequent reconnects (e.g., mobile wallets that reconnect on wake) will pay the full SDK construction cost repeatedly.

This is an intentional design choice (the old `ZamaSDK` had the same behavior when `signer` reference changed), but it's worth documenting explicitly in the `ZamaWagmiProvider` JSDoc so consumers are aware.

---

### [NON-BLOCKING] `hasSignerIdentity` admits `"connecting"` status without a test

**Confidence: 70**

```ts
function hasSignerIdentity(connection: WagmiConnection): boolean {
  return (
    connection.status !== "disconnected" &&
    Boolean(connection.address) &&
    connection.chainId !== undefined
  );
}
```

This returns `true` for `"connecting"` status if `address` and `chainId` happen to be set. In practice, wagmi's `"connecting"` phase typically has `address: undefined`, so the `Boolean(connection.address)` guard prevents it. But there's no test for `"connecting"` with a non-null address.

The existing test suite only covers `"reconnecting"` (signer expected) and `"disconnected"` (no signer). A test for `"connecting"` with no address and `"connecting"` with an address would clarify the intended behavior and prevent regressions if wagmi changes its state machine.

---

## 3. Blocked Context

- No linked GitHub issues in this PR beyond the Linear ticket tag in the title.
- Linear SDK-109 comments: empty (0 comments). No additional context from discussion.
- All required sources were accessible.

---

## 4. Decision

**APPROVE** (with non-blocking comments)

The core optional-signer implementation is solid and complete:
- The `requireSigner`/`requireCredentials`/`requireDelegatedCredentials` guard pattern is clean.
- `ZamaWagmiProvider` correctly solves the disconnected-signer footgun.
- Query layer changes (`isAllowed` disabling, cache invalidation on identity change) are correct.
- Test coverage for the new paths is comprehensive.

The only actionable concerns are the signer adapter error type discrepancy (PR description inaccuracy + minor behavioral gap in the disconnect race window) and two low-confidence observations about SDK instance lifecycle and `"connecting"` status coverage. None are blocking.

packages/sdk/src/errors/signer.ts
```diff
@@ -0,0 +1,33 @@
...
+ * }
+ * ```
+ */
+export class SignerRequiredError extends ZamaError {
+  readonly operation: string;
+
+  constructor(operation: string, options?: ErrorOptions) {
+    super(
+      ZamaErrorCode.SignerRequired,
+      `Cannot ${operation} without a signer. Configure one via ZamaSDKConfig.signer or <ZamaProvider config={createConfig({ signer: ... })}>.`,
```
ghermet, line 27: Does the operation matter here? the root problem is the missing signer on the zama provider

packages/sdk/src/query/__tests__/confidential-balances.test.ts
```diff
@@ -52,6 +53,19 @@ describe("confidentialBalancesQueryOptions", () => {
     expect(options.enabled).toBe(false);
   });
 
+  test("enabled is false when signer-backed credentials are absent", ({
+    relayer,
+    provider,
+    storage,
+  }) => {
+    const sdk = new ZamaSDK({ relayer, provider, storage });
```
ghermet:
```suggestion
  test("enabled is false when signer-backed credentials are absent", ({
    sdk
  }) => {
```

packages/sdk/src/errors/signer.ts
enitrat: it's reusing the same principle as the others utils written where you pass the op name for tracking. not specifically for or against

ghermet: Not a biggie for now

ghermet (approved):

PR #284, opened 2026-04-24
refactor(sdk): replace signer lifecycle callbacks with identity-change events

enitrat:
Replace `SignerLifecycleCallbacks` (a bag of `onDisconnect`, `onAccountChange`, `onChainChange`) with a single `SignerIdentityListener` that receives `{ previous?, next? }` identity snapshots.

The old design required the SDK to shadow-track wallet identity (`#lastAddress`, `#lastChainId`) by probing the signer at construction and during transitions. This created race conditions (`getAddress()` throwing mid-disconnect, events firing before `#identityReady` resolved) and blocked making the signer optional (SDK-109), since an absent signer cannot be probed.

The fix pushes identity reporting to the signer adapter, which already knows both sides of every transition, and makes the SDK a passive consumer. By moving the concern of lifecycle tracking to the signer rather than internal to the SDK, we can make the signer in ZamaSDK optional trivially.

`ZamaSDK` maintains one `signer.subscribe` call, which, on identity change, revokes session signatures for `change.previous`, clears only that previous requester's decrypt cache, and fans out to registered listeners via `sdk.onIdentityChange()`. Initial identity discovery seeds transition state without revoking sessions or clearing cache.

The React provider now registers query invalidation as an identity listener.

Internal details:
- EIP-1193 subscriptions reconcile observed address + chain state in one place, so late probes after unsubscribe cannot emit and reconnects do not reuse stale chain IDs.
- `WagmiSigner.subscribe` seeds the current connected identity using the same transition logic as `watchConnection`.
- `revokeSession()` remains live-signer based; docs note to call it before disconnecting or rely on lifecycle cleanup.

Closes SDK-122


packages/sdk/src/signer/__tests__/eip1193-subscribe.test.ts
```diff
@@ -0,0 +1,386 @@
...
+      provider.emit("accountsChanged", [ADDR_A]);
+      expect(onIdentityChange).not.toHaveBeenCalled();
+
+      provider.emit("chainChanged", "0x7a69");
+      expect(onIdentityChange).toHaveBeenCalledOnce();
+      expect(onIdentityChange).toHaveBeenCalledWith({
+        previous: undefined,
+        next: { address: ADDR_A, chainId: CHAIN_31337 },
+      });
+    },
```
ghermet, line 94: When does `chainChanged` fires exactly? Does it fire when you connect your wallet?

enitrat: from what I understand of the spec, not on connection, but rather when changing the chain inside the wallet

ghermet: So `onIdentityChange` doesn't kick in if you change account on the same chain?

ghermet (approved):

packages/sdk/src/signer/__tests__/eip1193-subscribe.test.ts
enitrat: it does in the handleAccountChange code path

PR #287, opened 2026-04-27
feat(sdk): add sdk.encrypt() top-level entrypoint [SDK-124]

ghermet:
## Summary

- Promote FHE encryption to a first-class `sdk.encrypt()` method alongside `sdk.userDecrypt` / `sdk.publicDecrypt`
- Centralize `EncryptStart`/`End`/`Error` event emission and error wrapping in the SDK method (single source of truth)
- Migrate Token write paths (`confidentialTransfer`, `confidentialTransferFrom`, `unwrap`) and the `encryptMutationOptions` query factory to delegate to `sdk.encrypt()`

Closes SDK-124

## Test plan

- [x] 5 new tests for `sdk.encrypt()` (delegation, events, error wrapping, ZamaError passthrough)
- [x] Updated 3 existing token tests for unified error message
- [x] Full SDK test suite passes (1388 tests)
- [x] Full react-sdk test suite passes (194 tests)
- [x] API reports regenerated


enitrat:
you lost some tests while merging prerelease in; can you restore them?

packages/sdk/src/token/__tests__/token.test.ts
enitrat: these tests are testing internals of the sdk.encrypt now. you should only keep token-level tests and cleanup what is no longer relevant now

zama-cremaud:
## NOT BLOCKING ; but worth mentioning

### [HIGH CONFIDENCE: 88] Behavioral regression — `tokenAddress` dropped from EncryptStart/End/Error events

**File**: `packages/sdk/src/zama-sdk.ts:532,534,540`

**What changed:**  
Before this PR, Token write paths called `this.emit({ type: ZamaSDKEvents.EncryptStart })` (and `End`/`Error`) via `Token.emit()`, which delegates to `sdk.emitEvent(input, this.address)` — explicitly setting `tokenAddress` on the event (`readonly-token.ts:713-714`).

After this PR, those same events are emitted from `ZamaSDK.encrypt()` via:
```ts
this.emitEvent({ type: ZamaSDKEvents.EncryptStart });
```
with no `tokenAddress` argument. `tokenAddress` is `undefined` on all three encrypt events when triggered from any `Token` operation.

**Confirmed with git**:  
```
# Before (2c8ee72a):
140: this.emit({ type: ZamaSDKEvents.EncryptStart });  // tokenAddress = this.address
```
```
# After (1991b7eb):
// emitted inside sdk.encrypt() — tokenAddress = undefined
```

**Impact:**  
Any `onEvent` consumer that uses `tokenAddress` to route or filter `EncryptStart/End/Error` events will silently lose that context. This is particularly relevant for multi-token UIs where the same SDK instance manages several tokens.

**Fix direction:**  
Since `EncryptParams.contractAddress` is always the target contract (token address for transfers/unwrap, wrapper address for `unwrap`), it could be forwarded:
```ts
// zama-sdk.ts — in encrypt()
this.emitEvent({ type: ZamaSDKEvents.EncryptStart }, params.contractAddress);
// ...
this.emitEvent({ type: ZamaSDKEvents.EncryptEnd, durationMs: ... }, params.contractAddress);
this.emitEvent({ type: ZamaSDKEvents.EncryptError, ... }, params.contractAddress);
```
This restores parity with the old behavior. It also makes direct callers of `sdk.encrypt()` (e.g. via `encryptMutationOptions`) automatically get correct event context.

**Mitigating factor:**  
No existing test asserts `tokenAddress` on encrypt events (only on decrypt events via `decryptBalanceAs` — see `events.test.ts:279`). The PR's comment at `events.test.ts:90` establishes a pattern of "SDK-level pipelines are token-agnostic," which the author may be intentionally extending here.

**Verdict**: Should be addressed or explicitly documented as an intentional behavioral change.

---

### [MEDIUM CONFIDENCE: 72] Undocumented side-effect: `encryptMutationOptions` now emits lifecycle events

**File**: `packages/sdk/src/query/encrypt.ts`

Before this PR, `encryptMutationOptions.mutationFn` called `sdk.relayer.encrypt(params)` directly — **no events were emitted**.

After this PR, it calls `sdk.encrypt(params)` — **EncryptStart/End/Error events now fire** whenever a mutation using this factory is invoked.

This is **likely the desired behavior** (it was arguably a bug before), but:
1. The PR description doesn't call it out.
2. No test verifies that the mutation factory now emits events.
3. Consumers who built their own event emission layer assuming the factory was silent may see double-emission.

**Evidence**: `query/encrypt.ts:10` + `zama-sdk.ts:532-543`.

**Verdict**: Low risk in practice, but warrants a note in the changelog or PR description for transparency.

ghermet:
> you lost some tests while merging prerelease in; can you restore them?

Can you clarify which tests you are talking about? I can't see any in the git diff

enitrat:
> Can you clarify which tests you are talking about? I can't see any in the git diff

it was showing in the merge commit diff but seems good now

packages/sdk/src/query/__tests__/encrypt.test.ts
```diff
@@ -19,4 +20,25 @@ describe("encryptMutationOptions", () => {
 
     expect(relayer.encrypt).toHaveBeenCalledWith(params);
   });
+
+  test("emits encrypt lifecycle events", async ({ signer, relayer, storage }) => {
```
enitrat: I don't think this belong in query tests: looks more like this is an `sdk.encrypt` test

packages/sdk/src/zama-sdk.ts
```diff
@@ -555,6 +567,58 @@ export class ZamaSDK {
     }
   }
 
+  /**
+   * Encrypt one or more plaintext values into FHE ciphertexts.
+   *
+   * This is the canonical encryption entrypoint — all internal write paths
+   * (Token transfers, unshield) and the query-layer mutation factory delegate
+   * here. Event lifecycle (`EncryptStart`/`End`/`Error`) is emitted in exactly
+   * one place.
```
enitrat:
```suggestion
```

enitrat (dismissed):

enitrat (approved):

PR #297, opened 2026-04-28
docs: update for signer/provider split [SDK-104]

enitrat:
## Summary

Updates all Gitbook documentation to reflect the signer/provider split from SDK-34:

- **New reference pages**: `GenericProvider`, `ViemProvider`, `EthersProvider` — documenting the read-only provider interface and its adapters
- **Removed**: `WagmiSigner.md` (internal; replaced by `createConfig` from `@zama-fhe/react-sdk/wagmi`)
- **Updated configuration guide**: `createConfig` pattern with per-library tabs (wagmi, viem, ethers, custom, web extensions, Node.js)
- **Updated all code snippets**: explicit `account` param on balance hooks, `sdk.requireSigner()` pattern, `sdk.provider.waitForTransactionReceipt()` calls
- **Added missing API docs**: `SignerRequiredError`, `sdk.onIdentityChange()`
- **Fixed**: `useConfidentialBalances` `account` param falsely claimed a signer fallback default

Closes SDK-104

## Test plan

- [ ] Verify Gitbook renders correctly (links, tabs, code blocks)
- [ ] Spot-check code snippets compile against current SDK exports
- [ ] Confirm no references to removed `WagmiSigner` or old single-signer constructor pattern remain


docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -216,22 +217,24 @@ After encryption, pass the handles and proof to your custom FHE contract:
...
     const { handles, inputProof } = await encrypt.mutateAsync({
       values: [{ value: 1000n, type: "euint64" }],
       contractAddress: "0xYourContract",
-      userAddress: await sdk.signer.getAddress(),
+      userAddress: address!,
     });
 
     // 2. Call your contract with the encrypted data
-    await sdk.signer.writeContract({
+    await sdk.signer!.writeContract({
```
ghermet, line 237: What do you think about gating signer on the sdk with 
```ts
get signer(){
 if(!this.#signer) throw new SignerRequiredError()
 return this.#signer
}

ghermet (approved):

docs/gitbook/src/guides/encrypt-decrypt.md
enitrat: a getter that throws might be surprising, i'm not a fan. ideally i'd like to leverage the typesystem more to have "progressive capabilities" between a sdk with a provider which has a subset S of methods, then you attach a signer and get the full set. maybe later...

PR #302, opened 2026-04-29
fix(examples): add Turnkey × Zama confidential tokens example [SDK-76]

zama-cremaud:
## What                                                                                                                                                                                                         
                                                                                                                                                                                                                  
  Adds `examples/turnkey-community-example/` — a working Next.js app integrating
  Turnkey embedded wallets with the Zama SDK for confidential ERC-7984 token operations                                                                                                                           
  on Ethereum Sepolia (and mainnet).                                                                                                                                                                              
                                                                                                                                                                                                                  
  This is the runnable companion to the Turnkey community-examples MDX guide being                                                                                                                                
  prepared for submission to `tkhq/community-examples`.                                                                                                                                                           
                                                                                                                                                                                                                  
  ## What it covers
                                                                                                                                                                                                                  
  - **Shield** — wrap public ERC-20 into a confidential ERC-7984 wrapper                                                                                                                                          
  - **Confidential balance decrypt** — read & decrypt an encrypted on-chain balance
  - **Confidential transfer** — FHE-encrypted transfer between addresses                                                                                                                                          
  - **Unshield** — two-phase unwrap back to the public ERC-20                                                                                                                                                     
                                                                                                                                                                                                                  
  ## Key integration points                                                                                                                                                                                       
                                                                                                                                                                                                                  
  - `TurnkeyProvider` + `useTurnkey()` from `@turnkey/react-wallet-kit`                                                                                                                                           
  - `createAccount()` + viem clients from `@turnkey/viem`
  - `TurnkeyViemSigner` implementing `GenericSigner` (wraps `ViemSigner` pattern)                                                                                                                                 
  - `ZamaProvider` + hooks from `@zama-fhe/react-sdk`                                                                                                                                                             
  - `/api/relayer/*` Next.js proxy (CORS + server-side API key)                                                                                                                                                   
                                                                                                                                                                                                                  
  ## How to run                                                                                                                                                                                                   
                                                                                                                                                                                                                  
  ```bash                                                                                                                                                                                                         
  cp .env.example .env.local   # fill in Turnkey org ID + auth proxy config
  npm install                                                                                                                                                                                                     
  npm run dev


examples/turnkey-community-example/src/app/page.tsx
```diff
@@ -0,0 +1,814 @@
...
+        params: [walletAddress, "latest"],
+        id: 1,
+      }),
+    })
+      .then((r) => r.json() as Promise<{ result?: string }>)
+      .then((data) => {
+        if (data.result) setEthBalance(BigInt(data.result));
+      })
+      .catch(console.error);
+  }, [walletAddress]);
```
ghermet: Prefer using a query from `@tanstack/query` instead for fetching inside a `useEffect`

examples/turnkey-community-example/src/app/page.tsx
```diff
@@ -0,0 +1,814 @@
...
+  useEffect(() => {
+    const initial = window.setTimeout(() => {
+      void fetchPublicBalance();
+    }, 0);
+    const id = setInterval(fetchPublicBalance, 10_000);
+    return () => {
+      window.clearTimeout(initial);
+      clearInterval(id);
+    };
+  }, [fetchPublicBalance]);
```
ghermet: Prefer using a query with a `refetchInterval` from `@tanstack/query` instead for fetching inside a useEffect

examples/react-turnkey-wallet/src/app/page.tsx
```diff
@@ -0,0 +1,808 @@
...
+  return `${addr.slice(0, 6)}…${addr.slice(-4)}`;
+}
+
+function txLink(hash: string): string {
+  return `${explorerUrl}/tx/${hash}`;
+}
+
+// ── Main page ─────────────────────────────────────────────────────────────────
+
+export default function Home() {
```
ghermet, line 9: This page could use being split up into smaller components with their own hooks, It would help people follow along.

ghermet:
# PR Review Summary

## Critical Issues (3)

### 1. SSRF risk in relayer proxy
**`src/app/api/relayer/[...path]/route.ts:~17`**

The proxy joins user-controlled path segments into the URL without origin validation. An attacker could craft paths that redirect the proxy to arbitrary hosts.

**Fix:** Validate `target.origin === expected.origin` after constructing the URL:
```ts
const target = new URL(`${RELAYER_BASE}/${path.join("/")}`);
const expected = new URL(RELAYER_BASE);
if (target.origin !== expected.origin) {
  return Response.json({ error: "Forbidden" }, { status: 403 });
}
```

### 2. `ZAMA_RELAYER_API_KEY` defined in `.env.example` but never forwarded
**`route.ts:~21-24`**

The proxy never reads `process.env.ZAMA_RELAYER_API_KEY` or attaches auth headers. Mainnet deployments will silently get 401/403 from the relayer.

**Fix:** Read the env var and forward it as an `Authorization` header when present.

### 3. Stale `#ensureAllowance` bug comment
**`page.tsx`, `preApproveShield` block comment**

States the SDK doesn't await the approval receipt before wrapping, but this was fixed in commit `4b744c18`. The workaround code works but the rationale is now false.

**Fix:** Either remove the manual approval workaround entirely and let the SDK handle it, or rewrite the comment to reflect the actual reason (e.g., distinct "Approving…" / "Shielding…" UX states).

---

## Important Issues (6)

### 4. `.catch(console.error)` on login and wallet creation
**`page.tsx:~73,85`**

If `handleLogin()` or `createEmbeddedWallet()` rejects, the error goes to console only. Users see nothing and will click repeatedly.

### 5. `createEmbeddedWallet` clears `initError` then has no catch
**`providers.tsx:~218-230`**

Error is wiped at the top with `setInitError(null)`, then if the call throws, `finally` resets `isCreatingWallet` but error state is never restored.

### 6. Empty catch on `loadPendingUnshield` silently wipes recovery state
**`page.tsx:~149-152`**

If IndexedDB fails, pending unshield hash is set to `null` with no logging. User loses the ability to resume an interrupted two-phase unshield (potential fund-loss scenario).

### 7. First-decrypt not gated behind user click
**`page.tsx:~159`**

`useConfidentialBalance` enables automatically when a pair is selected. The react-sdk `AGENTS.md` requires an explicit user click to gate the first EIP-712 blind-sign.

### 8. `package-lock.json` in a pnpm monorepo
The root uses pnpm. Shipping an npm `package-lock.json` is confusing and will cause conflicts. Should be deleted or gitignored.

### 9. `hashMessage`/`signTypedData` interaction with Turnkey API
**`providers.tsx:~91-113`**

The signing logic is subtle. `signTypedData` uses `serializeTypedData` + `PAYLOAD_ENCODING_EIP712` + `HASH_FUNCTION_NO_OP`. Needs manual verification that this produces valid EIP-712 signatures with Turnkey's API.

---

## Suggestions (10)

10. **Store `publicClient` in state instead of `receiptWaiter` function** — Simplifies providers.tsx and enables using `publicClient.getBalance()` instead of raw JSON-RPC fetch for ETH balance.
11. **Merge `chain-config.ts` into `config.ts`** — Two files for 10 lines of config is unnecessary indirection.
12. **Collapse 4 identical route handlers into one** — `GET`, `POST`, `PUT`, `DELETE` can be `export { handler as GET, ... }`.
13. **Use viem's `zeroAddress`** instead of hand-rolled `ZERO_ADDRESS` constant.
14. **Remove trivial `login` wrapper** — `useCallback(async () => { await handleLogin() }, ...)` adds nothing.
15. **Deduplicate state resets in `syncSigner`** — Three branches repeat the same 3-4 setter calls.
16. **ETH balance query ignores errors and treats missing `result` as `0n`** — Should check `response.ok`, JSON-RPC error field, and display query error state.
17. **README says Node.js 20+** but monorepo requires Node 22+.
18. **`.env.example` says "likely required on mainnet"** — Too vague; will age poorly.
19. **README project structure omits `src/lib/config.ts`** — All imports use `@/lib/config`, not `chain-config`.

---

## Strengths

- **Clean SDK integration** — Correctly uses `useShield`, `useConfidentialTransfer`, `useUnshield`, `useResumeUnshield`, `useListPairs`. No handle leakage — follows "clear-text in, clear-text out".
- **USDT-style approval reset** — `preApproveShield` correctly resets non-zero allowance before setting a new one.
- **Pending unshield persistence** — Two-phase unshield with IndexedDB persistence and resume UI is a non-trivial UX detail done well.
- **Separate IndexedDB for sessions** — `sessionDBStorage` avoids a real key-collision issue.
- **Well-structured single-file page** — Despite ~500 lines, `page.tsx` is well-organized with clear component boundaries. Right call for a learning example.
- **Excellent `.env.example` documentation** — Inline comments clearly explain each variable and what happens when you switch chains.

---

## Recommended Action

1. Fix critical issues 1–3 before merge (SSRF, missing API key forwarding, stale comment)
2. Address important issues 4–7 (error handling on user-facing actions, first-decrypt gating)
3. Delete `package-lock.json` (issue 8)
4. Consider suggestions for polish — particularly 10–12 which reduce code complexity

zama-cremaud:
All comments from review above have been addressed.

examples/turnkey-community-example/src/app/favicon.ico
ankurdotb, line 1: What's this binary file? Is it meant to be checked in or should it be gitignored?

ankurdotb (changes requested):
Name the folder "react-turnkey-wallet" or similar to stick to convention. Looks fine otherwise as long as other code review comments addressed.

examples/turnkey-community-example/src/app/favicon.ico
ghermet: It's the icon for the tab in your browser

zama-cremaud: Yeah, I agree this might not be very useful to have on the repo though, removing it

ankurdotb: Ahhhhh okay I was reviewing on mobile, and it didn't show me the file name, just that there was a binary file. 😂  This is fine to keep or ditch.

examples/react-turnkey-wallet/src/app/api/relayer/[...path]/route.ts
Comment (ankurdotb) [inline examples/react-turnkey-wallet/src/app/api/relayer/[...path]/route.ts:1]: Should be a literal `[...path]` folder name?

Comment (zama-cremaud) [reply examples/react-turnkey-wallet/src/app/api/relayer/[...path]/route.ts:1]: Yes, this is intentional. In Next.js App Router, catch-all segments use the literal [...segment] folder convention. So src/app/api/relayer/[...path]/route.ts is the expected structure for a catch-all route handler.

ankurdotb (approved):

PR #303, opened 2026-04-30
refactor(credentials): replace credentials model with keypair vault + permission store [SDK-134]

enitrat:
## Summary

Refactor the credential subsystem from two monolithic classes (`CredentialsManager` + `DelegatedCredentialsManager`, both ~700 LoC each, fully duplicated) into one `CredentialService` facade backed by two narrowly scoped collaborators: a `KeypairVault` (identity-scoped, chain-independent) and a `PermissionStore` (1-to-many, scoped by `signer × chain × delegator`). Direct and delegated permits now share storage and code paths, distinguished only by scope key — the duplicated class is gone.

Net diff: **+2,892 / −4,333**, removing ~1,400 LoC of duplicated logic and ~1,600 LoC of tests against it, while adding focused per-component tests.

## Why

- `CredentialsManager` and `DelegatedCredentialsManager` shared 80% of their code (sign/store/lookup/revoke) but were independently maintained. Bug fixes had to be ported twice.
- Lifetime mismatch: keypairs (identity-scoped, survive chain switches) were stored alongside permits (chain-scoped, invalidated by chain change), so chain switches needlessly threw away keypairs.
- The session/TTL machinery (one expiry timestamp covering the whole "session") didn't match the underlying primitive — every signed EIP-712 permit already carries its own `(startTimestamp, durationDays)`. Result was a layer of expiry-tracking that didn't reflect what was actually on-disk.
- `delegatedUserDecrypt` cached plaintext indefinitely as long as the delegate's signed permit was live, even after the delegator had revoked the on-chain delegation. Cache could leak plaintext the chain had already withdrawn permission for.

### New: `packages/sdk/src/credentials/`

| File | Role |
| --- | --- |
| `credential-service.ts` | Single facade. `allow / isAllowed / revokePermits / clearCredentials / handleIdentityChange`. Held by `ZamaSDK`. |
| `keypair-vault.ts` | One ML-KEM keypair per signer, persisted under `keypair:<signer>`. Survives chain switches. Dedupes concurrent `getOrCreate`. |
| `permission-store.ts` | 1-to-many permit store keyed by `(signer, chainId, delegatorAddress)`. Direct permits live under `delegator === signer`; delegated permits under `delegator !== signer`. Per-signer scope index allows `clearAllForSigner` to cascade across chains/delegators without enumerating storage keys. |
| `permission-coverage.ts` | Pure: classify a request as `covered` vs `missing`, chunk uncovered into ≤10 contracts per permit. |
| `permissions.ts` | Pure: `upsertPermission`, `pruneUnusablePermissions`, `deletePermitsTouchingContracts`. No I/O. |
| `delegation-check.ts` | New on-chain ACL read. `findRevokedDelegations` is consulted by `delegatedUserDecrypt` before serving cached plaintext. |
| `decrypt-permit.ts` | Pure: resolve a `Permission` from a `CredentialBundle` for the relayer call. |
| `schemas.ts` / `types.ts` / `storage-keys.ts` / `utils.ts` | Zod schemas, type contracts, deterministic storage keys, `normalizeAddresses` / `coversContracts` helpers. |

### Deleted

| File | Replaced by |
| --- | --- |
| `credentials-manager.ts` (~214 LoC) | `CredentialService` (direct path) |
| `credentials-manager-base.ts` (~484 LoC) | Split across `KeypairVault`, `PermissionStore`, `permission-coverage`, `permissions` |
| `delegated-credentials-manager.ts` (~216 LoC) | `CredentialService` (delegated path — same code, scope flag) |
| `credential-crypto.ts`, `credential-validation.ts`, `session-signatures.ts` | Schemas / `permissions.ts` / removed |
| `__tests__/credentials-manager.test.ts` (1,396 LoC), `delegated-credentials-manager.test.ts` (210 LoC), `session-ttl.test.ts` (319 LoC) | New focused tests on each new module + lifecycle integration tests on `ZamaSDK` |

### Public API changes (breaking)

**`ZamaSDK`**
- `sdk.revoke(contracts?)` → `sdk.revokePermits(contracts?)` (per-scope, direct-decrypt only)
- `sdk.revokeSession()` → `sdk.clearCredentials()` (full wipe: keypair + every permit referencing it across chains and delegators)
- `sdk.isAllowed([])` now returns `true` (vacuous truth) — was `false` before. Query options also short-circuit when `contractAddresses.length === 0`.

**Exports (`packages/sdk/src/index.ts`)**
- Removed: `CredentialsManager`, `CredentialsManagerConfig`, `DelegatedCredentialsManager`, `DelegatedCredentialsManagerConfig`, `StoredCredentials`, `DelegatedStoredCredentials`
- Added (types only): `CredentialBundle`, `Keypair`, `KeypairGenerator`, `Permission`, `PermitFactory`, `PermitSigner`, `StoredKeypair`. The service itself is internal — callers go through `ZamaSDK`.

**Events (`sdk-events.ts`)**
- Removed: `CredentialsExpired`, `CredentialsCorrupted`, `SessionExpired` — they didn't fire anywhere meaningful in the new model (per-permit TTL replaces the session concept).

**React hooks**
- `useRevoke` → `useRevokePermits`
- `useRevokeSession` → `useClearCredentials`

**Queries (`packages/sdk/src/query/`)**
- `revoke.ts` / `revoke-session.ts` removed → `revoke-permits.ts` / `clear-credentials.ts`

### On-chain delegation freshness check

`ZamaSDK.delegatedUserDecrypt` now calls `findRevokedDelegations` before serving cached plaintext. For each contract it reads `ACL.delegationExpiry(delegator, delegate, contract)`; if `0n` or `expiry <= block.timestamp`, the cache entry for `(account, contract, handle)` is dropped via the new `DecryptCache.delete(...)` and the relayer is consulted fresh (which performs the canonical on-chain check). This closes the leak where a delegator's on-chain revocation was invisible to the cache.

## Areas that need attention

- **`CredentialService.allow()` concurrency dedup** (`credential-service.ts:161-183`): in-flight requests share a promise keyed by `signer:chainId:delegator:contracts...`. Order of `contracts` in the key matters — `normalizeAddresses` is supposed to make this deterministic, please verify it sorts (it currently does `getAddress` only — confirm if intentional or if dedup will miss reordered duplicates).
- **`isAllowed([]) === true` flip**: deliberate (vacuous truth, matches set-coverage semantics) but it is a behavior change. Worth a sanity check that nothing in the React-SDK relied on the previous `false`.
- **`DecryptCache.delete` and the index write queue** (`decrypt-cache.ts:79-104`): the new method joins `#indexWriteQueue` to stay consistent with concurrent `set()`. Confirm the queue chain doesn't keep references and that `delete` returning before the queued op finishes is acceptable in the delegated-decrypt loop (it `await`s the queue tail, so should be fine, but the pattern is subtle).
- **No storage migration**: storage keys changed shape (`keypair:<signer>` and `permits:<signer>:<chainId>:<delegator>` via `storage-keys.ts`). Existing users on the previous build will see `null` reads on first load and the SDK will mint a fresh keypair / re-prompt for permits. Acceptable while the package is still `3.0.0-alpha.*`, but call out in CHANGELOG.
- **`findRevokedDelegations` cost**: one ACL read per contract per `delegatedUserDecrypt` call, in parallel via `Promise.all`. For multi-contract requests this adds an RPC fan-out before any cache hit can be served. Consider whether to memoize per `(delegator, delegate, contract)` for the lifetime of the SDK or rely on provider-level batching.
- **`PermissionStore.#untrackScope` swallows errors silently** (`permission-store.ts:182-193`): warning is logged but `clearAllForSigner` won't notice if the index ends up stale. Acceptable since the index is a hint, not a source of truth, but worth a second look.
- **Coverage gaps flagged by CI**: `credential-service.ts` 92.5% (lines 233-237 — persist failure path), `permission-store.ts` 92% (cleanup branches), `decrypt-permit.ts` 84.6% (the "no permit covers" `throw` paths). Decide whether to backfill or accept.

## Test plan

- [x] `pnpm test` — unit + lifecycle (zama-sdk-credentials-lifecycle) green
- [x] Coverage: 91.91% lines, 85.47% branches (CI gate at 80%)
- [ ] Manual smoke in `test/test-components` — direct decrypt, delegated decrypt, chain switch, account change, revoke, clear all
- [ ] Confirm storage migration story in CHANGELOG before un-WIPing


packages/react-sdk/src/provider.tsx
```diff
@@ -49,9 +49,9 @@ export function ZamaProvider({ children, config }: ZamaProviderProps) {
         provider: config.provider,
         signer: config.signer,
         storage: config.storage,
-        sessionStorage: config.sessionStorage,
+        permitStorage: config.permitStorage,
         keypairTTL: config.keypairTTL,
-        sessionTTL: config.sessionTTL,
+        permitDuration: config.permitDuration,
```
ghermet: Wouldn't this be `permitTTL` if we follow the current nomenclature?

packages/sdk/src/__tests__/chain-alignment.test.ts
```diff
@@ -76,6 +76,43 @@ describe("requireChainAlignment", () => {
...
+  it("sdk.allow throws ChainMismatchError before signing credentials", async ({
+    relayer,
+    provider,
+    signer,
+    storage,
+    tokenAddress,
+  }) => {
+    vi.mocked(signer.getChainId).mockResolvedValue(1);
+    vi.mocked(provider.getChainId).mockResolvedValue(11155111);
+    const sdk = new ZamaSDK({ relayer, provider, signer, storage });
```
ghermet: use `sdk` or `createSDK` for overrides from the fixtures

packages/sdk/src/config/types.ts
```diff
@@ -58,12 +58,12 @@ export interface ZamaConfigBase<TChains extends AtLeastOneChain = AtLeastOneChai
   relayers: { [K in TChains[number]["id"]]: RelayerConfig };
   /** Credential storage. Default: IndexedDB in browser, memory in Node. */
   storage?: GenericStorage;
-  /** Session storage. Default: IndexedDB in browser, memory in Node. */
-  sessionStorage?: GenericStorage;
+  /** Optional dedicated storage for permits. Defaults to `storage`. */
```
ghermet, line 61: if you want to make it default to `storage` you need to address the key collision issue currently present

enitrat: what issue ?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -0,0 +1,382 @@
...
+  get currentIdentity(): SignerIdentity | undefined {
+    return this.#identity;
+  }
+
+  async #signPermit(input: {
+    chunk: Address[];
+    keypair: StoredKeypair;
+    signerAddress: Address;
+    chainId: number;
+    delegatorAddress: Address;
```
ghermet: Why don't we use `delegatorAddress` instead of `delegator`?
``` ts
async #signPermit(input: {
    chunk: Address[];
    keypair: StoredKeypair;
    signerAddress: Address;
    chainId: number;
    delegatorAddress?: Address
  }): Promise<Permission> {
    const startTimestamp = Math.floor(Date.now() / 1000);
    const durationDays = this.#permitDurationDays;

    try {
      const eip712 = input.delegatorAddress
        ? await this.#permitFactory.createDelegatedUserDecryptEIP712(
            input.keypair.publicKey,
            input.chunk,
            input.delegatorAddress,
            startTimestamp,
            durationDays,
          )
        : await this.#permitFactory.createEIP712(
            input.keypair.publicKey,
            input.chunk,
            startTimestamp,
            durationDays,
          );

      const signature = await this.#permitSigner.signTypedData(eip712);

      return {
        keypairPublicKey: input.keypair.publicKey,
        signerAddress: input.signerAddress,
        delegatorAddress: input.delegatorAddress,
        chainId: input.chainId,
        signedContractAddresses: input.chunk,
        signature,
        startTimestamp,
        durationDays,
      };
    } catch (error) {
      if (error instanceof ZamaError) {
        throw error;
      }
      return wrapSigningError(error, "Credential signing failed");
    }
  }
}`

File: packages/sdk/src/credentials/credential-service.ts [hunk 5]
```diff
@@ -0,0 +1,382 @@
...
+    const ttl = (() => {
+      const value = KeypairTTLSchema.parse(config.keypairTTL ?? DEFAULT_KEYPAIR_TTL_SECONDS);
+      if (value > MAX_KEYPAIR_TTL_SECONDS) {
+        // oxlint-disable-next-line no-console
+        console.warn(
+          `[zama-sdk] keypairTTL (${value}s) exceeds the fhevm maximum of 365 days (${MAX_KEYPAIR_TTL_SECONDS}s); capping to ${MAX_KEYPAIR_TTL_SECONDS}s.`,
+        );
+        return MAX_KEYPAIR_TTL_SECONDS;
+      }
+      return value;
```
ghermet: Shouldn't this be done in `KeypairTTLSchema.parse`?

packages/sdk/src/events/sdk-events.ts
```diff
@@ -5,17 +5,6 @@ import type { ClearValueType, Handle } from "../relayer/relayer-sdk.types";
...
-  CredentialsLoading: "credentials:loading",
-  CredentialsCached: "credentials:cached",
-  CredentialsExpired: "credentials:expired",
-  CredentialsCreating: "credentials:creating",
-  CredentialsCreated: "credentials:created",
-  CredentialsRevoked: "credentials:revoked",
-  CredentialsPersistFailed: "credentials:persist_failed",
-  CredentialsAllowed: "credentials:allowed",
-  CredentialsCorrupted: "credentials:corrupted",
-  SessionExpired: "session:expired",
```
ghermet, line 18: Why did you remove events ?

packages/sdk/src/__tests__/chain-alignment.test.ts
enitrat: ✅

packages/react-sdk/src/provider.tsx
enitrat: ✅

packages/sdk/src/credentials/credential-service.ts
enitrat: yes. gated `isDelegated` on `signerAddress !== delegatorAddress` instead

packages/sdk/src/events/sdk-events.ts
enitrat: I removed the events because I didn't want to add them not knowing if they were still 100% relevant and matching. I am not so sure whether we should expose credentials lifecycle events _yet_, and if we do so I want this to be addressed properly inside the observability epic. I'd rather ship less and add later (non-breaking) than ship something not thought properly (overcooked)

zama-cremaud:
### Finding A — `sdk.revokePermits([])` silently clears the decrypt cache without revoking any permits
**Confidence: 75 | Severity: Medium**

In `ZamaSDK.revokePermits()` (`zama-sdk.ts:800-809`):

```ts
async revokePermits(contracts?: Address[]): Promise<void> {
  const signer = this.requireSigner("revokePermits");
  const service = this.#requireCredentialService("revokePermits");
  const signerAddress = await signer.getAddress();
  try {
    await service.revokePermits(contracts);
  } finally {
    // ← this always runs, even when no permits were revoked
    await swallow("clear decrypt cache", () => this.cache.clearForRequester(signerAddress));
  }
}
```

In `CredentialService.revokePermits()` (`credential-service.ts:264-281`):

```ts
async revokePermits(contracts?: Address[]): Promise<void> {
  ...
  const normalized = normalizeAddresses(contracts);
  if (normalized.length === 0) {
    return;  // ← no-op: no storage change
  }
  ...
}
```

**Effect:** `sdk.revokePermits([])` is a no-op on the permit store but still triggers `cache.clearForRequester(signerAddress)`. The caller's decrypt cache is wiped even though no permits changed. The next decrypt call will miss the cache and hit the relayer again.

**Who can hit this?** Any caller of `sdk.revokePermits([])` directly, or the React `useRevokePermits().mutate([])` path. The current mutation options pass through the array as-is (`revokePermitsMutationOptions`).

**Suggested fix:**

```ts
// Option A: guard the cache clear
try {
  await service.revokePermits(contracts);
} finally {
  if (contracts === undefined || (contracts.length > 0)) {
    await swallow("clear decrypt cache", () => this.cache.clearForRequester(signerAddress));
  }
}

// Option B: normalize early and avoid the ambiguity entirely
// (also ensures consistent service-level and SDK-level behavior for [])
```

---

### Finding B — `normalizeAddresses` in `allowDedupKey` DOES sort: open question in PR description is already resolved

The PR description says under "Areas that need attention":

> `normalizeAddresses` is supposed to make this deterministic, please verify it sorts (it currently does `getAddress` only — confirm if intentional or if dedup will miss reordered duplicates).

This is **incorrect**: `utils.ts:8` is:

```ts
return [...new Set(addresses.map((address) => getAddress(address)))].toSorted();
```

`.toSorted()` is present. The dedup key built in `allowDedupKey` (`credential-service.ts:370-382`) passes the already-normalized (sorted, checksummed) list, so the key IS deterministic for any ordering of the same contract set. No action needed here — just confirming the concern is already addressed by the implementation.

---

### Finding C — `PermissionStore.append` + `#trackScope` are not atomic; concurrent allow() for different scopes under the same signer can lose an index entry
**Confidence: 65 | Severity: Low (data hygiene, not security)**

`PermissionStore.#trackScope` (`permission-store.ts:166-174`) does a read-modify-write on the per-signer scope index:

```ts
async #trackScope(scope: PermissionScope): Promise<void> {
  const indexKey = await PermissionStore.indexKey(scope.signerAddress);
  const scopeKey = await PermissionStore.scopeKey(scope);
  const list = await this.#readIndex(indexKey);
  if (!list.includes(scopeKey)) {
    list.push(scopeKey);
    await this.#storage.set(indexKey, list);  // ← non-atomic
  }
}
```

**Scenario that loses an entry:**

1. `allow([tokenA])` (direct scope) and `allow([tokenB], delegatorAddr)` (delegated scope) run concurrently.
2. Both call `#store.append(...)` with different scopes but the same `signerAddress`.
3. Both read `index = []` before either write completes.
4. Scope A writes `[scopeKeyA]`; scope B writes `[scopeKeyB]`.
5. Last write wins — one scope key is lost from the index.

**Effect:** When `clearAllForSigner(signerAddress)` is called later (on account disconnect), the scope whose key was dropped from the index is NOT cleaned up. Its permits remain in storage, orphaned. This does not enable unauthorized decryption (the permits are scoped to the now-disconnected signer), but it's a storage leak.

**Note:** This is a known limitation in single-writer storage models without compare-and-set. The PR already acknowledges a similar issue in `#untrackScope`. Flagging for awareness; the index is documented as a "hint, not a source of truth".

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -0,0 +1,348 @@
...
+  #identity: SignerIdentity | undefined;
+
+  /**
+   * @throws {ZodError} if `keypairTTL` or `permitTTL` is not a positive integer,
+   *   or if `keypairTTL` exceeds the fhevm ACL maximum of 365 days.
+   * @throws {ConfigurationError} if `permitTTL` (days) exceeds `keypairTTL / 86400`.
+   */
+  constructor(config: CredentialServiceConfig) {
+    const ttl = KeypairTTLSchema.parse(config.keypairTTL ?? DEFAULT_KEYPAIR_TTL_SECONDS);
+    const permitTTL = PermitTTLSchema.parse(config.permitTTL ?? DEFAULT_PERMIT_DURATION_DAYS);
```
ankurdotb: [AUTOMATED] **Unenforced cross-check.** The constructor JSDoc above (`credential-service.ts:64`) claims:

> `@throws {ConfigurationError} if permitTTL (days) exceeds keypairTTL / 86400.`

But the constructor only does two independent `*Schema.parse(...)` calls — there is no comparison. SDK-134 §"Locked public surface" specifies `permitTTL` as `// days, default 30, maximum = keypairTTL`, so the bound is part of the locked contract and currently silently violated.

Two paths:

- **Implement the check** (preferred — spec-faithful):
  ```ts
  if (permitTTL * 86400 > ttl) {
    throw new ConfigurationError(
      `permitTTL (${permitTTL} days) exceeds keypairTTL (${ttl}s ≈ ${Math.floor(ttl/86400)} days)`,
    );
  }
  ```
- **Strike the claim** from the JSDoc and amend SDK-134 to drop the bound.

Reason this matters: `keypairPublicKey` is the cache-validity key for permits in `pruneUnusablePermissions`. A permit that outlives its keypair is automatically pruned at next `listUsableAndPrune`, so the bound isn't *security*-load-bearing — it's a UX guarantee that the user never signs a permit they can't use.

Confidence: 95/100

packages/sdk/src/zama-sdk.ts
```diff
@@ -766,26 +783,43 @@ export class ZamaSDK {
   }
 
   /**
-   * Revoke the session signature for the current signer without requiring
-   * contract addresses. Uses the tracked identity when available (safe during
-   * account switches), falling back to querying the signer directly.
+   * Wipe FHE permits for the current signer.
```
ankurdotb, line 1019: [AUTOMATED] **Asymmetry between no-arg and with-arg should be called out explicitly.** The two bullets describe the behavior correctly, but a reader scanning the docstring will likely miss that they operate on **different scopes**:

- `revokePermits()` → cross-chain, cross-delegator (everything for this signer)
- `revokePermits([A])` → current chain, direct only (delegated permits untouched)

Suggest one summary line up top, e.g.:

```
/**
 * Wipe FHE permits for the current signer.
 *
 * Note: scope differs between the two forms — no-arg is signer-wide, the
 * with-arg form is restricted to the current chain's direct-decrypt scope.
```

The behavior itself is fine (confirmed during review), this is just a docstring sharpening so consumers don't read the two bullets as parallel.

Confidence: 88/100

packages/sdk/src/zama-sdk.ts
```diff
@@ -372,27 +342,63 @@ export class ZamaSDK {
...
+   * Pure store lookup: are stored permits sufficient to cover `contracts`?
+   * No wallet prompt, no keypair generation. Returns `false` when no signer
+   * is configured.
+   */
+  async isAllowed(contracts: Address[]): Promise<boolean> {
+    if (!this.#credentialService) {
+      return false;
+    }
+    return this.#credentialService.isAllowed(contracts);
+  }
```
ankurdotb, line 403: [AUTOMATED] **Heads-up: `isAllowed([])` now returns `true` (was `false`).**

Vacuous-truth is the right answer for set-coverage semantics, and SDK-134 doesn't constrain the empty-list case explicitly, so this is defensible. Internally we're safe:

- `Token.isAllowed()` / `ReadonlyToken.isAllowed()` always pass exactly one address.
- `useIsAllowed` short-circuits on `length === 0` before hitting `sdk.isAllowed`.
- `BalancesCard` examples and `session-panel.tsx` only pass populated lists.

No action needed in this PR. Worth one sentence in the CHANGELOG / migration notes for external consumers — anyone using the previous `false` as a proxy for "nothing connected yet" will see a behavior flip on first install.

Confidence: 82/100

ankurdotb (commented):
[AUTOMATED]

## Architectural review — SDK-134 / SDK-89

Review focus per the author's request: structural/architectural fit against the locked specs, and breaking changes — not a line-by-line code review.

**Verdict: ships the intended architecture cleanly.** Three components + facade match SDK-134 verbatim (`KeypairVault` identity-scoped, `PermissionStore` 1-to-many on `(signer, chain, delegator)`, `CredentialService` facade). Locked internal interfaces (`KeypairGenerator`, `PermitFactory`, `PermitSigner`) match. Delegation collapsed to a `delegatorAddress` axis — no parallel manager. PBKDF2-from-permit-signature wrapping is gone. `isAllowed` is pure. Multiple permits per scope, chunked at ≤10, partial-success preserved. SDK-89 acceptance criteria addressable from this PR all hit (per-page-reload persistence, identity-scoped keypair surviving chain switches, no fake at-rest encryption, unified delegated path).

### Items I'm flagging at line level

1. **Unenforced `permitTTL ≤ keypairTTL/86400` cross-check** (`credential-service.ts:62-68`) — JSDoc claims it throws, the code doesn't. Real bug, please fix or strike the claim.
2. **`revokePermits` docstring** (`zama-sdk.ts:786-797`) — describes both cases correctly, but doesn't underline that no-arg and with-arg operate on **different scopes**. Worth a one-line callout so readers don't miss it.
3. **`isAllowed([])` flips `false → true`** (`zama-sdk.ts:383`) — heads-up note for external consumers; nothing inside this monorepo is affected.
4. **Per-signer keypair scoping baked into `KeypairVault`** (`keypair-vault.ts:14-19`) — no escape hatch for shared-tenant WaaS / B2B2C operators. Internal interface gets locked downstream by SDK-136/137 — flag for a sibling sub-issue under SDK-89 *before* that lock-in.
5. **`findRevokedDelegations` RPC fan-out** (`delegation-check.ts:60-71`) — runs on every `delegatedUserDecrypt` before any cache hit can be served. Optional optimisation: short-lived memo per `(delegator, delegate, contract)`.

### Items confirmed during review (not posting)

- `permitDuration` → `permitTTL` rename: agreed in Slack.
- `revokePermits()` no-arg cross-chain vs `revokePermits([A])` current-chain-direct-only asymmetry: accepted.
- `findRevokedDelegations` being beyond the SDK-134 scope table: accepted in-scope here, closes a real cache-leak.
- Spec wording for `revokePermits(contracts)` ("remove from `signedContractAddresses`"): not implementable — implementation is right, **SDK-134 spec wording will be amended in Linear** to match the drop-the-whole-permit semantics.

### Out of band

- New ticket to file under SDK-89 for the **shared ML-KEM keypair scope (B2B2C / WaaS opt-in)** — see comment 4. Justification material is in the late-April Slack thread.

packages/sdk/src/credentials/keypair-vault.ts
```diff
@@ -0,0 +1,112 @@
...
+  /** Keypair lifetime in seconds. */
+  ttl: number;
+}
+
+/**
+ * Identity-scoped, chain-independent vault for ML-KEM keypairs.
+ *
+ * One keypair per signer address; the keypair survives chain switches and
+ * permit revocations. Storage entries are keyed only by the signer address.
+ */
```
ankurdotb, line 21: [AUTOMATED] **Per-signer scoping is baked in here, with no escape hatch.** `getOrCreate(signerAddress)` keys solely on the signer's EVM address (`storage-keys.ts:keypairStorageKey`). This is the right default for separately-tenanted deployments (browser/mobile, treasury) per the per-wallet defense-in-depth argument — no concern for this PR.

The gap is **shared-tenant WaaS / B2B2C operators** (1000 client wallets behind one operator key store): per-wallet scoping derives no defense-in-depth benefit, costs 1000× key generation + storage, and there's no escape hatch. The Slack thread (Mathieu / Arik / Ankur, late April) converged on "configurable, default per-wallet, opt-in shared".

This PR doesn't preclude shared-keypair, but the locked `CredentialService.allow(contracts, delegator?)` interface is going to start being depended on by SDK-136 and SDK-137 — adding a `keypairScope` axis after that point will churn it.

Filed as [SDK-142](https://linear.app/zama/issue/SDK-142) (sibling sub-issue under SDK-89): *opt-in shared ML-KEM keypair scope for shared-tenant operators (B2B2C / WaaS)*. Two outcomes are both fine:

- Accepted → spec the `keypairScope?: string` field on `CredentialService.allow()` now.
- Rejected → close as won't-fix with the rationale recorded.

Not a blocker for this PR. Flagging so the question is on the record.

Confidence: 85/100

packages/sdk/src/credentials/delegation-check.ts
```diff
@@ -0,0 +1,79 @@
...
+  contractAddresses: readonly Address[];
+  delegatorAddress: Address;
+  delegateAddress: Address;
+}): Promise<Set<Address>> {
+  const { provider, relayer, contractAddresses, delegatorAddress, delegateAddress } = args;
+  if (contractAddresses.length === 0) {
+    return new Set();
+  }
+  const aclAddress = await relayer.getAclAddress();
+  const results = await Promise.all(
```
ankurdotb, line 60: [AUTOMATED] **Optional optimisation.** This `Promise.all` runs an `ACL.delegationExpiry` read per contract on every `delegatedUserDecrypt`, **before any cache hit can be served**. For multi-contract delegated decrypts this adds a meaningful RPC fan-out latency on the cache-warm path.

The correctness story is solid (this closes the cache-leak on on-chain revocation, which is a real problem and is the right fix), so I'm not blocking on it. Two follow-ups worth considering:

1. **Short-lived memo** per `(delegator, delegate, contractAddress)` keyed on chain block height bucket (e.g. cache for ~12s = ~1 block on most EVM chains). Trades freshness window for fewer RPCs; bounded by the same chain-time semantics `isDelegationActive` already uses.
2. **Provider-level multicall** if available — would collapse N reads into one round-trip without changing freshness semantics.

Neither is required to land. Ticket-worthy if/when latency on this path shows up in real-world telemetry.

Confidence: 80/100

packages/sdk/src/config/types.ts
ghermet: nervermind you actually fixed it by prefixing  keys with `permit` and `keypair`

packages/sdk/src/events/sdk-events.ts
ghermet: Okay we'll revisit this topic soon then this is very important to reach production grade

packages/sdk/src/credentials/credential-service.ts
enitrat: valid catch but I decided to drop that requirement instead—updated issue

packages/sdk/src/zama-sdk.ts
enitrat: fwiw: the high level hooks exposed don't have any behavioral change as passing an empty list disables it naturally now

packages/sdk/src/credentials/delegation-check.ts
enitrat: not sure it's worth doing here. multicall might, but it's another optimization concern

packages/sdk/src/zama-sdk.ts
enitrat: documented already

packages/sdk/src/events/sdk-events.ts
enitrat: fwiw it's noted as an epic to address in future sprints already

test/test-components/src/session-panel.tsx
```diff
@@ -57,12 +57,12 @@ export function SessionPanel({ tokenAddresses }: { tokenAddresses: [Address, ...
           </p>
         )}
         {revokePermits.isSuccess && (
-          <p className="text-zama-success" data-testid="revoke-permits-success">
-            Permits revoked successfully
+          <p className="text-zama-success" data-testid="revoke-session-success">
+            Session revoked successfully
```
ghermet: is this intended?

test/test-components/src/session-panel.tsx
```diff
@@ -57,12 +57,12 @@ export function SessionPanel({ tokenAddresses }: { tokenAddresses: [Address, ...
...
           </p>
         )}
         {revokePermits.isError && (
-          <p className="text-zama-error" data-testid="revoke-permits-error">
+          <p className="text-zama-error" data-testid="revoke-session-error">
```
ghermet: Is this intended?

test/test-components/src/session-panel.tsx
enitrat: not sure what rev you're reading but that's a "permit" now (from our last convo with devx)

packages/sdk/package.json
```diff
@@ -160,7 +160,8 @@
   },
   "dependencies": {
     "@zama-fhe/relayer-sdk": "~0.4.2",
-    "viem": ">=2"
+    "viem": ">=2",
+    "zod": "^4.3.6"
```
ghermet: not everyone is on zod 4.3.6, I would make it flexible like viem should they have a version of their own.
```json
"zod": ">=4"

enitrat: done

ghermet (approved):

PR #306, opened 2026-05-04
refactor: new signer system to fix SSR issues

enitrat:
## Context

`ZamaSDK` was crashing Next.js pages during SSR when a wagmi signer was wired in but no wallet was connected — the normal pre-hydration state. Root chain: the constructor fired `void this.#credentialService.initialize()`, which `await`ed `permitSigner.getAddress()`, which converted wagmi's safe `disconnected` state into a `SignerRequiredError` throw. The `void` swallowed the rejection into an unhandled promise; Next.js surfaced it as a render-time error.

The same throw-on-disconnect chain broke `useIsAllowed`: its `enabled` guard was a structural `signer !== undefined` check, so the queryFn fired before wagmi's post-hydration reconnect resolved, hit `SignerRequiredError`, and entered TanStack's retry backoff. Consumers saw a stuck `isLoading: true` for several seconds.

## Motivation

`ZamaSDK` conflated two distinct facts:

1. A signer adapter is configured.
2. A wallet account is currently connected and usable.

That's wrong for wagmi (and SSR generally), where the adapter is intentionally present from app startup while the wallet is disconnected or reconnecting. Every safe read path in the SDK and React layer needs to gate on (2), not (3). The existing `getAddress()` / `getChainId()` API forced the same throwing semantics on safe reads (where disconnect should mean "not ready") and action paths (where disconnect correctly throws).

We also needed `{ address, chainId }` as one unit — credentials, permits, relayer chain state, cache scoping, and query keys are all wallet-account-scoped, and reading the two halves separately invited cross-call account-change races.

## Architectural decisions

1. **New `WalletAccount` primitive.** Replaces `SignerIdentity` everywhere: `{ address, chainId }` as one snapshot. Credential service, query keys, and React hooks all consume this single unit.

2. **`GenericSigner` redesigned around wallet-account readiness.** Drops `getAddress()` / `getChainId()`. Adds:
   - `walletAccount: WalletAccountStore` — a synchronous observable (`getSnapshot` / `subscribe`) that returns `undefined` when not ready and never throws.
   - `requireWalletAccount(operation)` — sync, non-prompting; throws `WalletNotConnectedError` if no account.
   - `refreshWalletAccount?()` — optional hook for adapters whose initial account resolves asynchronously (ethers direct signer).
   - `dispose?()` — clean up provider event listeners.

3. **Sync `requireWalletAccount`, not async.** Hot read paths (render-time `enabled` gates, `isAllowed`) stay synchronous. The retry-once-after-refresh dance lives in `ZamaSDK.requireAlignedWalletAccount`, so action callers transparently get the async-resolution path without paying a microtask in the common case.

4. **Lifecycle is subscription-driven, not constructor-driven.** No more eager `void initialize()`. The SDK subscribes to `signer.walletAccount` and warms credentials / clears stale caches in response to transitions. Built-in stores synchronously re-emit the current snapshot on subscribe, so already-connected signers still warm immediately.

5. **Two error classes for two distinct states.** `SignerNotConfiguredError` (no adapter wired up — fatal config error) vs. `WalletNotConnectedError` (adapter present, no account — transient). Disambiguates what was previously fused in `SignerRequiredError`.

6. **Query keys include `walletAccount`.** Signer-gated queries (`isAllowed`, `confidentialBalance`, `userDecrypt`, `delegatedUserDecrypt`) key on `{ walletAddress, walletChainId }`, and their `enabled` flag gates on `walletAccount !== undefined` rather than structural signer presence. Removes the retry-window race by construction.

7. **Public `MutableWalletAccountStore` for custom signer adapters.** External `GenericSigner` authors call `createWalletAccountStore()` (returns `MutableWalletAccountStore`) and push transitions via `setSnapshot()`. Consumers see the read-only `WalletAccountStore` shape.

## Changes overview

- `packages/sdk/src/types/signer.ts` — new `WalletAccount`, `WalletAccountChange`, `WalletAccountListener`, `WalletAccountStore`; `GenericSigner` redesigned.
- `packages/sdk/src/signer/wallet-account-store.ts` (new) — `MutableWalletAccountStore` + `createWalletAccountStore` factory.
- `packages/sdk/src/signer/eip1193-subscribe.ts` — refactored to translate EIP-1193 events into `WalletAccountChange` transitions; shared by viem / ethers adapters.
- `packages/sdk/src/{viem,ethers}/...-signer.ts` and `packages/react-sdk/src/wagmi/wagmi-signer.ts` — adapters reimplement against the new contract; wagmi reuses the public `createWalletAccountStore`.
- `packages/sdk/src/zama-sdk.ts` — drops constructor-time credential warmup; subscribes to `signer.walletAccount`; `requireAlignedWalletAccount` returns the resolved `WalletAccount` for action-path reuse; `.catch()` on the subscription handler as defense-in-depth against the original failure mode.
- `packages/sdk/src/credentials/credential-service.ts` — `allow` / `revokePermits` / `clearCredentials` use `requireWalletAccount`; `isAllowed` uses the safe snapshot and returns `false` when disconnected.
- `packages/sdk/src/errors/signer.ts` — `WalletNotConnectedError`, `WalletAccountNotReadyError` (under shared `SignerRequiredError` base).
- `packages/sdk/src/query/*` — query factories take a `SignerQueryContext` with `walletAccount`; query keys scoped accordingly.
- `packages/react-sdk/src/utils/wallet-account.ts` (new) — `useWalletAccount(sdk)` via `useSyncExternalStore`. Every signer-gated hook reads it.
- `packages/react-sdk/src/provider.tsx` — invalidates wallet-lifecycle queries on account transitions.
- Tests + API reports + docs updated. Custom-signer page now compiles and explains the lifecycle contract.


enitrat (commented):
—

zama-cremaud (commented):
[AUTOMATED]

## Code Review

Thorough refactor of the signer lifecycle — the move to a non-throwing observable `WalletAccount` is the right shape and addresses the original SSR root cause cleanly. Four findings below, all on logic / error-handling correctness in the new lifecycle code. None are blockers; the `dispose()` one is the most material because it interacts with React StrictMode and `useMemo` re-creation.

Reviewed against the new `GenericSigner` contract, the per-adapter rewrites (viem / ethers / wagmi), the subscription-driven lifecycle in `ZamaSDK`, the new error class hierarchy, and the query-key changes. Compatibility-wise this is purely a single-package breaking change (3.0.0-alpha pre-release), no protocol-cross-service impact.

packages/sdk/src/zama-sdk.ts
```diff
@@ -817,7 +835,8 @@ export class ZamaSDK {
   dispose(): void {
     this.#unsubscribeSigner?.();
     this.#unsubscribeSigner = undefined;
-    this.#identityListeners.clear();
+    this.#walletAccountListeners.clear();
+    this.signer?.dispose?.();
```
zama-cremaud: [AUTOMATED] **`dispose()` now destroys the user-owned signer, contradicting its own docstring.**

The added `this.signer?.dispose?.()` makes `dispose()` terminal for the signer, but the docstring two lines above states it should be **non-terminal** (“Call this when the SDK instance is being replaced but the relayer is shared, e.g. React provider remount in Strict Mode”). The signer adapters' `dispose()` is destructive: `ViemSigner` / `EthersSigner` call `#unsubscribeProvider()` and remove the EIP-1193 listeners; `WagmiSigner` detaches `watchConnection`. After dispose the `MutableWalletAccountStore` keeps subscribers but never receives another `setSnapshot()` — the signer is frozen.

Realistic failure modes (signer is user-owned and survives SDK churn):
- **React 18 StrictMode dev double-render**: mount → unmount calls `sdk.dispose()` → `signer.dispose()` → remount with a new SDK that holds the same dead signer. Wallet events stop reaching the SDK.
- **`config` object identity changes**: the provider's `useMemo([config])` recreates the SDK; cleanup disposes the old SDK *and* the user-owned signer; the new SDK then holds the dead signer for the rest of the page.

Provider cleanup confirms the path: `packages/react-sdk/src/provider.tsx:75` does `useEffect(() => () => sdk.dispose(), [sdk])`. The existing `dispose calls signer.dispose` test only asserts the call, not its consequence (`packages/sdk/src/__tests__/zama-sdk.test.ts:107-114`).

**Suggested fix**: move `this.signer?.dispose?.()` from `dispose()` to `terminate()` — `terminate()` already calls `dispose()` first, so the signer teardown becomes the terminal-only step that matches both docstrings.

```ts
dispose(): void {
  this.#unsubscribeSigner?.();
  this.#unsubscribeSigner = undefined;
  this.#walletAccountListeners.clear();
  // signer is caller-owned; do not tear it down here
}

terminate(): void {
  this.dispose();
  this.signer?.dispose?.();
  this.relayer.terminate();
}
```

Confidence: 95/100

packages/sdk/src/zama-sdk.ts
```diff
@@ -247,8 +265,10 @@ export class ZamaSDK {
       }
     }
     await Promise.all(
-      Array.from(this.#identityListeners, (listener) =>
-        swallow("identity listener", () => listener(change)),
+      Array.from(this.#walletAccountListeners, (listener) =>
```
zama-cremaud, line 263: [AUTOMATED] **Listener fan-out is silently skipped when `relayer.switchChain` throws — the React layer never learns the wallet changed.**

The `return` at line 264 above sits before this `Array.from(this.#walletAccountListeners, ...)` fan-out. `relayer.switchChain` throws `ConfigurationError` whenever the new `chainId` isn't in the dApp's configured chains (`packages/sdk/src/relayer/relayer-dispatcher.ts:113-120`) — a realistic case when a user switches their wallet to an unsupported network.

When this happens:
- Credential service `handleWalletAccountChange` and `cache.clearForRequester(previousAccount.address)` did run before the throw — SDK-internal cache is fine.
- But `#walletAccountListeners` is the **only** hook the React provider uses (`provider.tsx:63-71` registers `invalidateWalletLifecycleQueries(queryClient)` via `sdk.onWalletAccountChange`). Skipping it leaves stale `decryption.*` and `isAllowed.*` query caches scoped to the previous (account, chain), with only a `console.warn` for diagnostics.

The existing test at `zama-sdk.test.ts:232-263` (“does not notify listeners when relayer chain switching fails”) locks the current behavior in but does not justify it — it was added in PR #283 and only codifies what the code does today.

**Suggested fix**: notify listeners regardless of switchChain success. If the chain isn't supported, the UI still needs to know the wallet transitioned. Either always fan-out, or fan-out with `next: undefined` so consumers can render an “unsupported chain” state. The `console.warn` for the dispatcher remains as the diagnostic for the dev.

Confidence: 90/100

packages/sdk/src/ethers/ethers-signer.ts
```diff
@@ -38,36 +44,84 @@ export type EthersSignerConfig = { ethereum: EIP1193Provider } | { signer: Signe
...
-      this.#signerPromise = browserProvider.getSigner();
+      this.#browserProvider = new BrowserProvider(config.ethereum);
       this.#eip1193 = config.ethereum;
+      this.#unsubscribeProvider = eip1193Subscribe({
+        provider: config.ethereum,
+        getInitialWalletAccount: () => this.#loadBrowserWalletAccount(),
+        onWalletAccountChange: ({ next }) => {
+          this.walletAccount.setSnapshot(next);
+        },
+      });
```
zama-cremaud, line 65: [AUTOMATED] **Browser branch never sets `#resolvingInitialAccount = true`, so the not-ready→refresh→retry mechanism in `requireAlignedWalletAccount` silently fails to engage.**

The direct-signer branch at lines 67-72 sets `#resolvingInitialAccount = true` before `void this.refreshWalletAccount().catch(...)`. The browser branch at 57-66 doesn't, even though `eip1193Subscribe` schedules `getInitialWalletAccount()` asynchronously (`Promise.resolve().then(...)` plus two `eth_accounts` / `eth_chainId` round-trips, `eip1193-subscribe.ts:116-135`).

During that window, `requireWalletAccount` sees `account === undefined` *and* `#resolvingInitialAccount === false`, so it throws `WalletNotConnectedError` — not `WalletAccountNotReadyError`. `ZamaSDK.requireAlignedWalletAccount` (`zama-sdk.ts:216-227`) only auto-retries on `WalletAccountNotReadyError`, so the browser path silently bypasses the recovery dance the direct-signer path benefits from.

Window is narrow (one microtask + two RPC round-trips) but real — any consumer constructing `new EthersSigner({ ethereum })` and immediately invoking a signer-required SDK call sees a confusing `WalletNotConnectedError` despite a connected wallet. ViemSigner is exempt because it resolves the initial account synchronously.

**Suggested fix**: set the flag in the browser branch and clear it on both initial-load and event-driven snapshot updates:

```ts
this.#resolvingInitialAccount = true;
this.#unsubscribeProvider = eip1193Subscribe({
  provider: config.ethereum,
  getInitialWalletAccount: async () => {
    try {
      return await this.#loadBrowserWalletAccount();
    } finally {
      this.#resolvingInitialAccount = false;
    }
  },
  onWalletAccountChange: ({ next }) => {
    this.#resolvingInitialAccount = false;
    this.walletAccount.setSnapshot(next);
  },
});
```

Clearing in `onWalletAccountChange` matters because the `eventVersion` guard in `eip1193Subscribe` allows a real wallet event to preempt the initial loader.

Confidence: 85/100

packages/sdk/src/ethers/ethers-signer.ts
```diff
@@ -38,36 +44,84 @@ export type EthersSignerConfig = { ethereum: EIP1193Provider } | { signer: Signe
...
     } else {
-      this.#signerPromise = Promise.resolve(config.signer);
+      this.#directSigner = config.signer;
+      this.#unsubscribeProvider = () => {};
+      this.#resolvingInitialAccount = true;
+      void this.refreshWalletAccount().catch(() => undefined);
```
zama-cremaud: [AUTOMATED] **`Nit:` constructor `.catch(() => undefined)` discards initial-resolution errors with no log — inconsistent with the rest of the codebase.**

If the underlying ethers signer's `getAddress()` or `provider.getNetwork()` throws (RPC down, locked key, custom provider failure), this swallow returns silently, `#resolvingInitialAccount` flips to `false` in `#refreshFromSigner`'s finally, and every subsequent call throws `WalletNotConnectedError` with no upstream context for the developer.

The sibling browser path — same scenario, identical “initial identity load” failure shape — is logged at `signer/eip1193-subscribe.ts:131-134`: `console.warn("[zama-sdk] initial identity load failed:", error)`. The shared `swallow()` utility uses the same convention. This direct-signer branch is the only async-init signer constructor that violates it.

**Suggested fix**: replace with the existing convention, e.g.

```ts
void this.refreshWalletAccount().catch((error) => {
  // oxlint-disable-next-line no-console
  console.warn("[zama-sdk] initial wallet account resolution failed:", error);
});
```

or delegate to `swallow("initial wallet account resolution", () => this.refreshWalletAccount())`. No tests assert log-absence, so this is purely additive.

Confidence: 85/100

packages/sdk/src/zama-sdk.ts
enitrat: hmm interesting thanks. moving in terminate makes sense

packages/sdk/src/zama-sdk.ts
enitrat: removed the try/catch around relayer change altogether, shouldnt have been there

packages/sdk/src/ethers/ethers-signer.ts
enitrat: fixed by moving ownership of the flag to the WalletAccount store

zama-cremaud (approved):

docs/gitbook/src/reference/sdk/GenericSigner.md
```diff
@@ -51,41 +57,90 @@ const sdk = new ZamaSDK(config);
...
     /* ... */
   }
+
   async writeContract(config) {
     /* ... */
   }
+
+  dispose() {
+    this.#unsubscribe();
+  }
```
ghermet: If all `dispose` does is call the underlying `unsubscribe` function we probably should remove it and use `unsubscribe` or rename `unsubscribe` to `dispose`

docs/gitbook/src/reference/sdk/GenericSigner.md
```diff
@@ -51,41 +57,90 @@ const sdk = new ZamaSDK(config);
...
 
 ```ts
-import type { GenericSigner } from "@zama-fhe/sdk";
+import {
+  createWalletAccountStore,
+  WalletNotConnectedError,
+  type GenericSigner,
+} from "@zama-fhe/sdk";
 
 class MySigner implements GenericSigner {
```
ghermet: `MySigner` is eligible for the `Disposable` interface with the subscription system:
```ts
class MySigner extends GenericSigner,Disposable {

 [Symbol.dispose]() {
  this.dispose()
 }
}
```

```ts
using mySigner = new MySigner()
```

docs/gitbook/src/reference/sdk/GenericSigner.md
```diff
@@ -51,41 +57,90 @@ const sdk = new ZamaSDK(config);
...
-  async getAddress() {
-    /* ... */
+
+  requireWalletAccount(operation: string) {
+    const account = this.walletAccount.getSnapshot();
+    if (!account) {
+      throw new WalletNotConnectedError(operation);
+    }
+    return account;
   }
```
ghermet: This could be a `BaseSigner` class people extend with `signTypedData` and `writeContact` instead of having to write this boilerplate code

docs/gitbook/src/reference/sdk/GenericSigner.md
```diff
@@ -17,11 +17,17 @@ import type { GenericSigner } from "@zama-fhe/sdk";
...
   signTypedData(typedData: EIP712TypedData): Promise<Hex>;
   writeContract(config: WriteContractConfig): Promise<Hex>;
-  subscribe?(onIdentityChange: SignerIdentityListener): () => void;
+  dispose?(): void;
+}
+
+interface WalletAccountStore {
+  getSnapshot(): WalletAccount | undefined;
+  subscribe(onWalletAccountChange: WalletAccountListener): () => void;
 }
```
ghermet, line 31: You can use [valtio](https://valtio.dev/docs/introduction/getting-started) for the wallet store

packages/react-sdk/src/relayer/use-user-decrypt.ts
```diff
@@ -15,7 +16,8 @@ export function useUserDecrypt(
   options?: Omit<UseQueryOptions<DecryptResult>, "queryKey" | "queryFn">,
 ) {
   const sdk = useZamaSDK();
-  const queryOpts = userDecryptQueryOptions(sdk, config);
+  const walletAccount = useWalletAccount(sdk);
```
ghermet, line 19: why would `useWalletAccount` not use `useZamaSDK` internally?

packages/sdk/src/signer/wallet-account-store.ts
```diff
@@ -0,0 +1,88 @@
...
+  #snapshot: WalletAccount | undefined;
+  #resolved: boolean;
+
+  constructor(initial?: WalletAccount) {
+    this.#snapshot = initial;
+    this.#resolved = initial !== undefined;
+  }
+
+  getSnapshot(): WalletAccount | undefined {
+    return this.#snapshot;
```
ghermet, line 34: why not make `snapshot` public instead

packages/react-sdk/src/utils/wallet-account.ts
```diff
@@ -0,0 +1,12 @@
+"use client";
+
+import { useSyncExternalStore } from "react";
+import type { WalletAccount, ZamaSDK } from "@zama-fhe/sdk";
+
+export function useWalletAccount(sdk: ZamaSDK): WalletAccount | undefined {
+  return useSyncExternalStore(
+    (listener) => sdk.onWalletAccountChange(listener),
+    () => sdk.signer?.walletAccount.getSnapshot(),
```
ghermet, line 9: Wouldn't a missing signer client side be considered a silent failure?

packages/react-sdk/src/wagmi/wagmi-signer.ts
```diff
@@ -39,22 +40,29 @@ export interface WagmiSignerConfig {
  * @param signerConfig - {@link WagmiSignerConfig} with wagmi config
  */
 export class WagmiSigner implements GenericSigner {
+  readonly walletAccount: MutableWalletAccountStore;
   readonly #config: Config;
+  readonly #unsubscribeConnection: () => void;
+  #disposed = false;
```
ghermet: prefer the `Disposable` [interface](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management) for this

packages/react-sdk/src/utils/wallet-account.ts
enitrat: I don't believe so: no signer | signer disconnected means you can't get account-related data in both cases. If I start the ZamaSDK with no signer attached I dont want signer-related hooks to be enabled

docs/gitbook/src/reference/sdk/GenericSigner.md
enitrat: isn't `dispose` suppose to signal that the object is done and wont be used anymore? idk whether to keep or not, looks appropriate as a teardown method

docs/gitbook/src/reference/sdk/GenericSigner.md
enitrat: done

docs/gitbook/src/reference/sdk/GenericSigner.md
enitrat: done

packages/sdk/src/signer/wallet-account-store.ts
enitrat: making this public means you could write it which would bypass our change detections

packages/react-sdk/src/relayer/use-user-decrypt.ts
enitrat: no strong reason to have it inside, it would duplicate context reads when the instance is already available outside

PR #320, opened 2026-05-05
feat(sdk): detect ERC-1363 support and route Token.shield() accordingly [SDK-145]

ghermet:
## Summary

- `Token.shield()` now uses ERC-1363 `transferAndCall` (single tx, no approval) when the underlying ERC-20 advertises `IERC1363` via ERC-165 `supportsInterface`, and falls back to the legacy `approve` + `wrap` two-tx path otherwise — routing is decided by introspection only, with no runtime catch-and-fallback
- New `Token.isPayable()` public method (ERC-165 probe with per-instance caching; failures cache as `false` so non-payable tokens never re-probe)
- New `ShieldSubmittedEvent.shieldPath` and `TransactionErrorEvent.shieldPath` telemetry so integrators can observe which path ran or failed
- `TestERC1363` mock Solidity contract + deploy-script update for on-chain e2e coverage

Closes SDK-145

## Test plan

- [x] Unit tests in `shield.test.ts` cover routing, ERC-1363 data encoding (self-shield empty `0x` vs raw 20-byte recipient), `isPayable()` caching, `ShieldSubmitted` / `TransactionError` events, approval-strategy interaction, and query-mutation passthrough
- [x] E2E tests in `erc1363-detection.node.spec.ts` validate detection against a real Anvil deployment (ERC-1363 token, standard ERC-20, non-ERC-165 contract)
- [x] All 1642 SDK tests pass, 0 regressions
- [ ] Run `pnpm e2e:test:node` after fresh Anvil deploy to validate e2e detection tests


packages/sdk/src/token/__tests__/callbacks.test.ts
```diff
@@ -170,69 +170,8 @@ describe("Unshield callbacks (P4)", () => {
...
-      .mockResolvedValueOnce(1000n);
-
-    const recipient = "0x8b8b8b8b8B8B8b8B8B8b8b8b8b8B8B8B8B8b8B8b" as Address;
-    await token.shield(100n, { to: recipient });
-
-    expect(signer.writeContract).toHaveBeenCalled();
-  });
-});
+// Shield callback tests live in shield.test.ts so all shield routing,
+// data-encoding, and callback semantics stay co-located.
```
enitrat:
```suggestion
```

packages/sdk/src/types/token.ts
```diff
@@ -6,11 +6,61 @@ export interface TransferOptions extends TransferCallbacks {
...
+ *   revert during `transferAndCall` falls back to `approveAndWrap`; user
+ *   rejections and RPC failures do not (they propagate so the caller never
+ *   sees a second wallet popup or risks a duplicate broadcast).
+ * - `"transferAndCall"`: force the single-tx path. Throws
+ *   {@link ERC1363NotSupportedError} when the underlying does not advertise
+ *   ERC-1363 support, and never falls back on revert.
+ * - `"approveAndWrap"`: skip detection and run the legacy two-tx path
+ *   (`approve` then `wrap`). Honours `approvalStrategy`.
+ */
+export type ShieldStrategy = "auto" | "transferAndCall" | "approveAndWrap";
```
enitrat: I don't think it's desirable to have this as an option. Could we make it "auto" all the time?

packages/sdk/src/types/token.ts
enitrat, line 1: nit: most of these types I think should live under token/types

packages/sdk/src/types/token.ts
```diff
@@ -6,11 +6,61 @@ export interface TransferOptions extends TransferCallbacks {
...
+ *   `onTransferReceived` callback receives the underlying tokens and mints
+ *   confidential tokens to `from` (self-shield, `data === "0x"`) or to the
+ *   address encoded in `data` (shield-to-other, raw 20-byte address). No
+ *   ERC-20 approval is needed.
+ * - `"approveAndWrap"`: two transactions. The user calls `underlying.approve`
+ *   (skipped when `approvalStrategy === "skip"` or the existing allowance is
+ *   sufficient) and then `wrapper.wrap(recipient, amount)`. Used when the
+ *   underlying does not implement ERC-1363, or when the caller forced
+ *   `shieldStrategy: "approveAndWrap"`.
+ */
```
enitrat, line 28:
```suggestion
```

packages/sdk/src/token/token.ts
```diff
@@ -348,6 +409,108 @@ export class Token extends ReadonlyToken {
...
+        // Transaction was submitted on-chain — commit to this path (no fallback)
+        this.emit({
+          type: ZamaSDKEvents.ShieldSubmitted,
+          txHash,
+          shieldPath: "transferAndCall",
+        });
+        safeCallback(() => options?.onShieldSubmitted?.(txHash));
+        try {
+          const receipt = await this.sdk.provider.waitForTransactionReceipt(txHash);
+          return { txHash, receipt };
```
enitrat: can you self-contain this logic inside `submitTransferAndCall` (or `shieldViaTransferAndCall` for consistency) and  make sure that `shieldViaTransferAndCall` and `shieldViaApproveAndWrap` follow the same model of submit -> emit -> wait for receipt -> return

also: I don't want to catch transaction errors to try one way or another, can't we do introspection on the supported interface to route properly

packages/sdk/src/types/token.ts
ghermet: That was part of the [specs](https://linear.app/zama/issue/SDK-145/sdk-detect-erc-1363-support-and-route-tokenshield-accordingly), If you know your token is payable or not it is useful. auto mode is great for unknown tokens.

packages/sdk/src/types/token.ts
ghermet: This is useful documentation about the different paths

packages/sdk/src/token/token.ts
ghermet: Done.

- Renamed `#submitTransferAndCall` → `#shieldViaTransferAndCall`. Both `#shieldViaTransferAndCall` and `#shieldViaApproveAndWrap` now follow the same pattern: submit → emit `ShieldSubmitted` → callback → `waitForTransactionReceipt` → return. `shield()` becomes a clean dispatcher with no fallback.
- Removed the runtime catch-and-fallback. Routing is decided entirely by `isPayable()` (ERC-165 introspection on the underlying via `supportsInterface`). If a token advertises ERC-1363 but `transferAndCall` reverts at runtime, we throw — the user can opt out with `shieldStrategy: "approveAndWrap"`. This also means `isContractCallError` is no longer needed in `token.ts`.
- Updated tests: removed the "auto + runtime revert: falls back" assertion (now asserts throws + no second wallet popup), removed the cache-poisoning test (moot without fallback), and updated event tests to match.

All 1646 SDK tests pass.

packages/sdk/src/token/__tests__/callbacks.test.ts
ghermet: Removed

packages/sdk/src/types/token.ts
ghermet: I trimmed the doc to the minimum

packages/sdk/src/types/token.ts
enitrat: just to avoid one RPC call?

packages/sdk/src/types/token.ts
enitrat: it's written three times in the same file that's why

packages/sdk/src/token/token.ts
```diff
@@ -91,6 +97,47 @@ export class Token extends ReadonlyToken {
...
+    if (this.#isPayable !== null) {
+      return this.#isPayable;
+    }
+    try {
+      const underlying = await this.#getUnderlying();
+      this.#isPayable = await this.sdk.provider.readContract(isPayableTokenContract(underlying));
+      return this.#isPayable;
+    } catch (error) {
+      this.emit({
+        type: ZamaSDKEvents.TransactionError,
```
enitrat: transaction error on a read call is weird

ghermet: Agreed. Removed the emit in `isPayable()` and dropped `"isPayable"` from `TransactionErrorOperation`. Probe failures still get cached as `false` so we don't re-probe on every shield. JSDoc updated.

packages/sdk/src/types/token.ts
enitrat: I think the right way to type that would be to have a single option in ShieldParams to `forceApproveAndWrap`. "auto" mode should not bleed into the type system, we should either call `transferAndCall` if supported, if not just default to `approveAndWrap`. I am not sure we should expose a `ERC1363NotSupportedError` error either, I would prefer to have it completely transparent. 

Regarding the introspection problem (might not register as an ERC1363 supported contract) I don't think we want to force through...

ghermet: One could argue that the shield strategy overhead  is not worth dealing with for saving a rpc call

ghermet: Done. Removed `ShieldStrategy` (and the `shieldStrategy` option). Routing is now always determined by ERC-165 introspection on the underlying — `transferAndCall` if `supportsInterface(IERC1363)` returns true, `approveAndWrap` otherwise.

Knock-on cleanup:
- Deleted `ERC1363NotSupportedError` and the `ERC1363NotSupported` `ZamaErrorCode` (only thrown when the user forced `transferAndCall`)
- Removed `ShieldStrategy` from public exports (`@zama-fhe/sdk`, `@zama-fhe/sdk/query`)
- `ShieldPath` is now a literal union (`"transferAndCall" | "approveAndWrap"`) instead of `Exclude<ShieldStrategy, "auto">`
- Dropped the now-stale "explicit transferAndCall/approveAndWrap" tests; kept the auto-mode coverage
- Regenerated `etc/*.api.md`

All 1642 SDK tests pass.

packages/sdk/src/events/sdk-events.ts
```diff
@@ -84,17 +85,35 @@ export interface DecryptErrorEvent extends BaseEvent {
...
 export interface TransactionErrorEvent extends BaseEvent {
   type: typeof ZamaSDKEvents.TransactionError;
-  /** Which write operation failed. */
-  operation: string;
+  /** Which SDK operation failed. */
+  operation: TransactionErrorOperation;
   /** The error that caused the transaction to fail. */
   error: Error;
+  /** For shield errors: which execution path failed. Always set when `operation === "shield"`. */
+  shieldPath?: ShieldPath;
```
enitrat: shield-specific attribute in a shared, generic `TransactionErrorEvent` does not look right

ghermet: yeah I agree its not great but what would be the alternative? I thought we could encode this in the operation as follows: `shield:transferAndCall` >

enitrat: do you even need that info in the error considering that you have the `ShieldSubmittedEvent` carrying that data?

enitrat: otherwise you'd need to make a superclass of TransactionError just for Shielding

ghermet: You need to know which shield path failed, `ShieldSubmittedEvent` doesn't account for failures

enitrat (approved):

PR #323, opened 2026-05-06
refactor: split Token into Token + WrappedToken hierarchy [SDK-96]

zama-cremaud:
## Summary

Restructures the high-level token API to mirror the ERC-7984 contract architecture.

- **`Token`** — base confidential token (reads + transfer + operator + delegated decrypt)
- **`WrappedToken extends Token`** — adds wrapper-specific ops: shield, unshield, allowance, unwrap, finalizeUnwrap, etc.
- **`ReadonlyToken` is removed** — reads live on `Token`. `createToken(addr)` is now single-arg; new `createWrappedToken(addr)` returns a `WrappedToken` (the wrapper IS the confidential token).
- **React hooks aligned**: `useToken(address)` / `useWrappedToken(address)`; `useReadonlyToken` removed; single-arg / `{ address, ... }` shapes unified across all hooks (see commit body of `b26cadb5` for the full list).
- **Mutation factories** updated: shield/unshield/unwrap/finalize/approveUnderlying/resumeUnshield take a `WrappedToken`; `delegateDecryption` / `revokeDelegation` take `(sdk, contractAddress)` matching the SDK-95 SDK-level primitives.
- Test harness pages/forms updated; `wrapperAddress` props dropped (defaults to `tokenAddress`).
- Docs: `ReadonlyToken.md` / `useReadonlyToken.md` deleted; placeholder pages added for `WrappedToken.md` / `useWrappedToken.md`.
- API reports regenerated.

## Breaking changes

- `ReadonlyToken`, `createReadonlyToken`, and the two-argument `createToken(addr, wrapper)` are removed.
- `useReadonlyToken` is removed.
- Several React hooks change shape (single-arg vs. `{ address, ... }`); object-form callers should rename `tokenAddress` → `address` and drop the now-unused `wrapperAddress`.

Migration: for wrapper operations, use `createWrappedToken(wrapperAddr)` / `useWrappedToken(wrapperAddr)`.

## Stack / context

Builds on top of [SDK-95 (#307)](https://github.com/zama-ai/sdk/pull/307), now squash-merged into `prerelease`. Reconciled with the SDK-134 credentials refactor that landed in parallel: `signer.requireWalletAccount(op).address` replaces `signer.getAddress()`; `isFatalBatchError` replaces `isSessionError`; `confidentialBalanceQueryOptions` keeps SDK-134's signer-scoped third arg.

## Commits (ordered for review)

1. `feat(sdk)!: split Token into Token + WrappedToken hierarchy`
2. `feat(react-sdk)!: align hooks with Token/WrappedToken split`
3. `test(harness): align test pages/forms with single-address API`
4. `docs(gitbook): replace ReadonlyToken with WrappedToken stubs`
5. `chore(api-report): regenerate after Token/WrappedToken split`

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm test:run` — **1480 passed, 5 skipped, 0 failed**
- [x] `pnpm api-report` — regenerated, committed
- [ ] CI green
- [ ] Manual smoke test of the test harness pages (test-nextjs / test-vite)


packages/sdk/src/utils/safe-callback.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Invoke a callback inside a try/catch so a throwing listener can never
+ * break the surrounding flow (e.g. an unwrap that has already been
+ * submitted on-chain).
+ */
+export function safeCallback(fn: () => void): void {
```
enitrat: we have `swallow` for that already

ghermet (changes requested):
> Automated multi-agent review (code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer). Findings are filtered to high-confidence items; please treat them as starting points, not verdicts.

# PR #323 Review Summary — Token / WrappedToken Split

**Stats:** 172 files, +3474 / −5928

---

## 🔴 Critical (must fix before merge)

| # | Source | Issue | Location |
|---|---|---|---|
| C1 | tests | **`integration.test.ts` is broken** — `TypeError: token.unwrap is not a function`. The fixture still binds `token = new Token(...)`, but `shield`/`unwrap`/`unshield`/`finalizeUnwrap` live on `WrappedToken`. Vitest unit config excludes it (CI hides this), but `vitest.integration.config.ts` includes it. | `packages/sdk/src/token/__tests__/integration.test.ts:14,104,110,153,177` |
| C2 | code-review | **`ZamaSDK.cache` made `public`** so `Token` can reach `this.sdk.cache.get(...)`. API extractor raises `ae-forgotten-export: CachingService`. Violates the "no handle leakage; SDK cache owns two-phase optimization" rule. | `packages/sdk/src/zama-sdk.ts:55-56,68`; `packages/sdk/src/token/token.ts:229,397`; `packages/sdk/etc/sdk.api.md:20029` |
| C3 | code-review | **`useUnderlyingAllowance` passes the same address twice** as `tokenAddress` and `wrapperAddress` — stale two-address query-key shape under the new single-address hook surface (a "silent migration shim"). | `packages/react-sdk/src/shield/use-underlying-allowance.ts:41-46,73-77`; `packages/sdk/src/query/query-keys.ts:91-104` |
| C4 | silent-failures | **`WrappedToken.isPayable()` silently routes to `approve+wrap` on ANY error**, including `#getUnderlying()` rejections. A transient RPC blip silently extracts an extra approve tx and wallet popup from the user for an ERC-1363 token that should have one-tx-shielded. | `packages/sdk/src/token/wrapped-token.ts:74-90` |
| C5 | silent-failures | **`AccountService.#handleWalletAccountChange` fire-and-forgets relayer chain switch** — `void swallow("switch relayer chain", () => this.#relayer.switchChain(nextChainId))`. If the switch fails, subsequent decrypts run against the previous chain's relayer and return stale plaintexts. Violates "RelayerNative parity". | `packages/sdk/src/services/account-service.ts:102-123` |
| C6 | silent-failures | **`isFatalBatchError` is too narrow** — missing `ChainMismatchError`, `WalletAccountNotReadyError`, `SignerNotConfiguredError`. A session-level chain switch mid-batch produces "N of M tokens failed" results where every error is identical and meaningless. | `packages/sdk/src/errors/fatal-batch.ts:14-20` |
| C7 | comments | **`Token.md` documents a non-existent `handle?` parameter** for `decryptBalanceAs`. Readers will pass it and silently lose the value. | `docs/gitbook/src/reference/sdk/Token.md:52` |
| C8 | comments | **`unshield-tokens.md` examples use `sdk.createToken("0xEncryptedERC20").unshield(...)`** — a TypeScript error after the split (`shield-tokens.md` was updated, this guide was missed). Same applies to `.unshieldAll()` and `.resumeUnshield()`. | `docs/gitbook/src/guides/unshield-tokens.md:35-37,83,107` |

---

## 🟡 Important (should fix)

**Architecture / types**
- Public constructors `new Token(...)` / `new WrappedToken(...)` bypass the SDK factories (`zama-sdk.ts:248-250`); no eager ERC-165 check inside `createWrappedToken` — plain ERC-20 addresses can be wrapped and only revert at first `shield()`.
- `underlyingAllowanceQueryOptions(sdk, tokenAddress, config)` is the **only** wrapper-related query that takes a bare address rather than `(token: WrappedToken)` — inconsistent with `shieldMutationOptions`, `unshieldMutationOptions`, etc. (`packages/sdk/etc/sdk-query.api.md:844`)
- React hook surface re-erases wrapper-ness to a string: `useShield`/`useUnshield`/`useUnwrap`/etc. take `address: Address`, internally call `useWrappedToken(address)` — type discipline established in `@zama-fhe/sdk/query` doesn't survive the trip to react-sdk.

**Error handling**
- `WrappedToken.#waitAndFinalizeUnshield` does **not** emit `ZamaSDKEvents.TransactionError` on failure (all other write paths do). Worst case: phase-1 receipt parse fails after on-chain unwrap succeeded — user loses the in-flight unshield with no event and a generic error message that doesn't mention `resumeUnshield(hash)`. (`wrapped-token.ts:557-596`)
- `assertConfidentialBalance` docstring promises "no surprise EIP-712 popup" but the implementation can trigger one when credentials aren't cached. (`token.ts:686-689,693-717`)
- `swallow()` uses `console.warn`, not `logError` → invisible to Sentry. Used in 7+ load-bearing sites including chain switch and cache cleanup. (`packages/sdk/src/utils/swallow.ts`)
- `Token.batchBalancesOf` total-failure aggregation throws only the *first* error; the rest are dropped, making debugging brutal. (`token.ts:321-325`)

**Tests**
- No assertion that `WrappedToken instanceof Token` (or that inherited `balanceOf` works on a `WrappedToken`). Central design claim of the PR is unguarded against future refactors.
- `ZamaSDK.createWrappedToken` has no direct factory test (matching the three `createToken` cases).
- The three direct unit tests for `Token.decryptBalanceAs` from the deleted `delegation.test.ts:117-164` (zero-handle short-circuit, happy path, "no value returned" error) were not migrated.
- `assertSameSdk` cross-SDK guard tests deleted with no replacement.
- New hooks `useToken` / `useWrappedToken` have **no** dedicated tests.

**Code review**
- `Token.confidentialTransferFrom` uses `requireChainAlignment`, but `confidentialTransfer` uses `requireAlignedWalletAccount` — inconsistent pre-flight error story. (`token.ts:502-503` vs `:566-567`)
- `Token` class JSDoc says delegation lives on `Token`, but `delegateDecryption`/`revokeDelegation`/etc. were removed from the class surface and now live only on `ZamaSDK`. The PR description and `token.ts:70-81` overstate the class's surface.

**Comments / JSDoc**
- `wrapped-token.ts:310,348` — `Orchestrates: unshield → wait → ... finalize` is self-referential (impl calls `unwrap`/`unwrapAll`).
- `@throws` blocks dropped during migration across 8 methods (`confidentialTransferFrom`, `setOperator`, `unshield`, `resumeUnshield`, `unwrap`, `unwrapAll`, `finalizeUnwrap`, `approveUnderlying`).
- `use-confidential-transfer.ts:42` references "optional wrapper" leftover from the two-address model.

---

## 🟢 Suggestions

- Resolve or remove the self-flagged `TODO: code smell` at `token.ts:369` — refactor static batch methods to `(sdk, tokens, options)` so `cache`/`sdk` can be made private (folds in C2).
- Consider `protected` on `ZamaSDK.cache` and `Token.sdk` (or narrow the published type via `@internal`).
- `Token.isWrapper` uses `Promise.all` → rejects on first probe failure rather than returning `false`. Consider `Promise.allSettled` so it behaves like a probe.
- `unwrapAll` zero-balance guard throws `DecryptionFailedError` — error type is misleading. Use `InsufficientConfidentialBalanceError` or new `ZeroBalanceError`. (`wrapped-token.ts:460-462`)
- Section-banner comments (`// METADATA`, `// BALANCES`, `// SHIELD (ERC-20 → confidential)`, …) violate "don't explain WHAT" guidance and add no info beyond the JSDoc on each method.
- Remove the trivial restatement docstrings (`/** Read the token name from the contract. */`).

---

## ✅ Strengths

- **No legacy shims, no compat re-exports.** `ReadonlyToken` / `createReadonlyToken` / `useReadonlyToken` removed cleanly — no `ReadonlyToken` references survive in `.api.md`.
- **Constructor consistency**: both classes use `(sdk, address)`; no separate `wrapperAddress` field anywhere; address normalization preserved in query keys.
- **Type-system-enforced separation in `@zama-fhe/sdk/query`**: `shieldMutationOptions`, `unshieldMutationOptions`, `finalizeUnwrapMutationOptions`, … are typed `token: WrappedToken`. A plain `Token` instance fails at compile time.
- **`Token` and `WrappedToken` class-level JSDoc** clearly states the split; "the wrapper IS the confidential token" kills the two-address mental model.
- **`WrappedToken.isPayable` JSDoc was updated** to match the new "don't cache failures" behavior — textbook good behavior + comment migration.
- **Load-bearing rationale preserved**: e.g. `// ERC7984ERC20Wrapper.onTransferReceived decodes the recipient via 'address(bytes20(data))'` correctly migrated to `wrapped-token.ts:182-185`; the sequential-write rationale on `#batchDelegationOp` preserved.
- **Test fixtures factorized well**: `mockToken` and `mockWrappedToken` share `createMockTokenInternal`.
- **Shield routing rule honored**: no per-call routing flag on `ShieldOptions`; path chosen via ERC-165 introspection only.
- **`chain-alignment.test.ts` migration is complete**: `shield`/`unwrap` go through `createWrappedToken`, `delegateDecryption` moved onto SDK, `batchBalancesOf` stayed on `Token`.
- **Type-design ratings: 4 / 4 / 4 / 3** (Encapsulation, Invariants, Usefulness, Enforcement). Clear improvement over the prior optional-`wrapper` model.

---

## Recommended Action Plan

1. **Block merge on C1–C8.** C1 is a hard test regression; C2 is a CI-visible API contract break; C3 is a query-key shape bug; C4–C6 are silent-failure paths that mislead users; C7–C8 are docs that don't compile.
2. **Important issues** in priority order: emit `TransactionError` from `#waitAndFinalizeUnshield`; fix `assertConfidentialBalance` doc-vs-impl drift; route `swallow` through `logError`; add the missing tests (`instanceof Token`, `createWrappedToken` factory, `decryptBalanceAs` direct, hook memoization).
3. **Re-run** `pnpm test:run` **and** `pnpm test:integration` (the integration run is what catches C1) before re-requesting review.

packages/sdk/src/zama-sdk.ts
```diff
@@ -52,7 +52,8 @@ export class ZamaSDK {
   readonly #registryTTL: number;
   readonly #onEvent: ZamaSDKEventListener;
   readonly #walletAccountListeners = new Set<WalletAccountListener>();
-  readonly #cache: CachingService;
+  /** Persistent cache for decrypted FHE plaintext values, scoped by (requester, contract, handle). */
+  readonly cache: CachingService;
```
enitrat: ⚠️ this got promoted to public visibility

packages/sdk/src/token/token.ts
```diff
@@ -1,116 +1,478 @@
...
+    const handle = await this.readConfidentialBalanceOf(normalizedAccount);
+    if (isZeroHandle(handle)) {
+      return 0n;
+    }
+
+    // Pre-flight delegation check — must run before cache lookup so that
+    // revoked delegations are caught even when a stale cached value exists.
+    await this.#assertDelegationActive(normalizedDelegator);
+
+    const cached = await this.sdk.cache.get(normalizedAccount, this.address, handle);
```
enitrat: looks like this is calling cache internally. you should be going through `sdk.delegatedUserDecrypt`

packages/sdk/src/token/token.ts
```diff
@@ -1,116 +1,478 @@
...
+   *
+   * @example
+   * ```ts
+   * const balances = await Token.batchDecryptBalancesAs(tokens, {
+   *   delegatorAddress: "0xDelegator",
+   *   onError: (err, addr) => { console.error(addr, err); return 0n; },
+   * });
+   * ```
+   */
+  static async batchDecryptBalancesAs(
```
enitrat, line 343: regression: should delegate to
  `sdk.delegatedBatchDecryptHandlesAs`, which batches of all handles in a single relayer call. no cache logic should be done there. see current version in prerelease (the inner of this function should not have changed at all

packages/sdk/src/token/token.ts
```diff
@@ -910,98 +716,59 @@ export class Token extends ReadonlyToken {
...
-    const account = await this.sdk.requireAlignedWalletAccount("approveUnderlying");
-    const userAddress = getAddress(account.address);
-    const allowance = await this.sdk.provider.readContract(
-      allowanceContract(underlying, userAddress, this.wrapper),
-    );
+  // PRIVATE HELPERS
 
-    if (allowance >= amount) {
-      return;
+  async #assertDelegationActive(delegatorAddress: Address): Promise<void> {
```
enitrat: no longer belongs in token file. property of the delegation service. shouldn't need to be invoked here if properly routing through the right SDK functions

packages/sdk/src/token/wrapped-token.ts
```diff
@@ -0,0 +1,640 @@
...
+  async isPayable(): Promise<boolean> {
+    if (this.#isPayable !== null) {
+      return this.#isPayable;
+    }
+    try {
+      const underlying = await this.#getUnderlying();
+      this.#isPayable = await this.sdk.provider.readContract(isPayableTokenContract(underlying));
+      return this.#isPayable;
+    } catch (error) {
+      // eslint-disable-next-line no-console
```
enitrat: this used to cache the result to `false` before, erasing what was in there

enitrat (commented):
some regressions observed, globally this PR should not be a lot more than taking existing functions and moving them around files

packages/sdk/src/token/token.ts
```diff
@@ -1,116 +1,422 @@
...
+   * const balances = await Token.batchDecryptBalancesAs(tokens, {
+   *   delegatorAddress: "0xDelegator",
+   *   onError: (err, addr) => { console.error(addr, err); return 0n; },
+   * });
+   * ```
    */
-  async isPayable(): Promise<boolean> {
-    if (this.#isPayable !== null) {
-      return this.#isPayable;
+  static async batchDecryptBalancesAs(
```
enitrat, line 343: needs alignment with original readonly-token method from `prerelease`, notably on filtering out  zero handles before sending relayer requests

enitrat: and setting concurrency to 10 as well (just copy-paste function from prerelease branch)

packages/sdk/src/token/wrapped-token.ts
```diff
@@ -0,0 +1,633 @@
...
+   * @example
+   * ```ts
+   * await wrappedToken.approveUnderlying(); // max approval
+   * await wrappedToken.approveUnderlying(1000n); // exact amount
+   * ```
+   */
+  async approveUnderlying(amount?: bigint): Promise<TransactionResult> {
+    const signer = this.sdk.requireSigner("approveUnderlying");
+    await this.sdk.requireChainAlignment("approveUnderlying");
+    const underlying = await this.#getUnderlying();
```
enitrat, line 266: update with copy-paste from `prerelease`'s approveUnderlying

enitrat:
one last thing from the last merge commit from main is to use the new `requireAlignedWalletAccount` standalone function  inside shield, unwrap, unwrapAll, #ensureAllowance, confidentialTransfer and confidentialTransferFrom

- `const account = await requireAlignedWalletAccount("shield", this.sdk.signer, this.sdk.provider);` 

and we will be good to merge

enitrat (approved):

ghermet (approved):

PR #325, opened 2026-05-06
chore(sdk): centralize validation with Zod schemas at public boundaries

enitrat:
## Context

Validation was scattered: manual `assertObject`/`assertStringProp` in the artifact cache, inline `typeof` checks in pending-unshield, ad-hoc refinements duplicated between credential modules, and numeric config values (`keypairTTL`, `permitTTL`, `registryTTL`) validated inside the `ZamaSDK` constructor rather than at the config-creation boundary.

### Where Zod belongs — and where it doesn't

We added Zod at **trust boundaries where data crosses from `unknown` to typed**: storage reads (`GenericStorage.get()`), network responses (relayer manifest), and deserialized JSON (cached artifacts, persisted credentials). These are the places where the type system can't help — the data is genuinely untyped at runtime.

We deliberately **did not** add Zod validation of TypeScript-typed function arguments flowing between internal modules. The type system already provides that guarantee; runtime Zod there adds cost without proportional safety.

### Branded `ZamaConfig` — validate once, trust everywhere

`ZamaConfig` is now a branded type that can only be produced by `createConfig()` (or adapter factories). The `ZamaSDK` constructor accepts only `ZamaConfig` — no loose-input union, no re-validation. TTLs and defaults are resolved once at `createConfig()` time; internal code trusts the result.

`ZamaSDKConfig` (the old loose constructor input) is removed from the public API. `buildZamaConfig` is also removed from exports (it was `@internal`). All config construction goes through `createConfig()` or adapter-specific factories.

### Registry addresses derived from chains

`registryAddresses` is no longer a separate input on the config. The SDK constructor derives it from `config.chains` at point of use — single source of truth. Users who need custom registry addresses (e.g. Hardhat) set `registryAddress` on the chain definition.

## What changed

**Artifact cache** (`fhe-artifact-cache.ts`) — Replaced 4+ copies of read→assert→fallback→delete with `#readCachedEntry` / `#parseCachedEntry` / `#deleteCorruptEntry` helpers backed by Zod schemas (`CachedPublicKeySchema`, `CachedPublicParamsSchema`, `ManifestSchema`). This is the highest-value change: genuinely untyped data from `GenericStorage.get()` validated properly.

**Shared primitives** (`schemas/primitives.ts`) — `hex`, `evmAddress`, `checksummedAddress`, `positiveSeconds`, `positiveDays`, `chainId` — reusable Zod atoms imported by credential schemas and config validation.

**Credential storage schemas** (`credentials/schemas.ts`) — `StoredKeypairSchema`, `PermissionSchema` validate data coming out of storage. Same trust-boundary principle as the artifact cache.

**Config validation** (`config/build.ts`, `validation.ts`) — `parseConfiguration()` wraps `safeParse` → `ConfigurationError` with `prettifyError`. TTLs validated once in `buildZamaConfig`; `ZamaConfig` fields narrowed from `number | undefined` to `number`. `web()` and `node()` transport factories validate their numeric options at the factory boundary.

**Pending unshield** (`token/pending-unshield.ts`) — `PendingUnshieldRequestSchema` replaces manual normalization for storage-persisted unshield state.

**Constructor simplification** (`zama-sdk.ts`) — Constructor is now a simple field assignment from a trusted, branded `ZamaConfig`. No validation, no normalization, no default resolution. ~80 lines removed.

## Test plan

- [x] `tsc --noEmit` clean on both `@zama-fhe/sdk` and `@zama-fhe/react-sdk`
- [x] 1599 tests pass across all projects
- [ ] CI green
- [ ] API reports regenerated


packages/sdk/src/chains/types.ts
```diff
@@ -7,9 +7,11 @@ import type { Address, EIP1193Provider, Hex } from "viem";
  *
  * All built-in presets (`mainnet`, `sepolia`, `hardhat`, `hoodi`)
  * are `FheChain` objects exported from `@zama-fhe/sdk/chains`.
+ *
+ * Validated at runtime by {@link FheChainSchema} at the config boundary.
  */
-export interface FheChain<TId extends number = number> {
-  readonly id: TId;
```
ghermet, line 12: Doesn't this remove typecheck at build time? This is what wagmi uses in their sdk

enitrat: reverted

packages/sdk/src/zama-sdk.ts
```diff
@@ -134,39 +130,36 @@ export class ZamaSDK {
   readonly #credentialService: CredentialService | undefined;
   #unsubscribeSigner?: () => void;
 
-  constructor(config: ZamaSDKConfig) {
+  constructor(rawConfig: ZamaSDKConfig) {
+    const config = parseZamaSDKConfig(rawConfig);
```
ghermet: Do we need a dedicated function to call `ZamaSDKConfig.parse(rawconfig)`? Don't we use the schemas only ?

packages/sdk/src/config/schema.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function parseZamaConfigBase<T extends ZamaConfigBase>(params: T): ParsedZamaConfigBase<T> {
+  const parsed = PublicConfigDataSchema.parse(params);
+  return { ...params, ...parsed } as ParsedZamaConfigBase<T>;
+}
+
+export function parseZamaSDKConfig(config: ZamaSDKConfig): ParsedZamaSDKConfig {
+  const parsed = SDKConfigDataSchema.parse(config);
+  return { ...config, ...parsed } as ParsedZamaSDKConfig;
+}
```
ghermet: isn't `parsed` like the same as `params` if it doesn't throw?

enitrat: so the idea here is that not all params could be properly parsed & checks so you'd return a partial check of what you can validate. I'll put more thought into it and figure out whether we have a better way of handling those input params

enitrat: in f98b02f I removed a bit the complexity between the different, duplicated types for zama config. I also removed the awkward override for registryAddresses that wasnt supposed to be there (its just reading data from configured chain)

packages/sdk/src/config/web.ts
```diff
@@ -16,6 +23,9 @@ import type { WebRelayerConfig, WebRelayerOptions } from "./types";
  * ```
  */
 export function web(options?: WebRelayerOptions): WebRelayerConfig {
+  if (options !== undefined) {
+    parseConfiguration(WebRelayerOptionsSchema, options);
```
ghermet, line 27: I think I can make a decorator for this at some point

ghermet (approved):

PR #331, opened 2026-05-06
refactor: fan-out decryption/delegation/caches into services [SDK-161]

enitrat:
## Context

`ZamaSDK` had grown into a 1,100-line god class mixing three distinct domains — decrypt pipeline, delegation lifecycle, and plaintext caching — that change for different reasons. The decrypt pipeline was copy-pasted between `userDecrypt` and `delegatedUserDecrypt`; delegation pre-flight checks were reimplemented verbatim in `Token`; cache ownership was ambiguous enough that both `ReadonlyToken` and `ZamaSDK` checked the cache for the same call.

## Motivation

Extract focused services so each domain gets a single owner, eliminating duplication and the drift that comes with it. The resulting modules are small enough to unit-test in isolation with lightweight mocks. `ZamaSDK` itself becomes a thin orchestrator (~700 lines, down from ~1,100).

## Summary of changes

**Three new services** under `src/services/`:

- **`DecryptionService`** — owns the full decrypt pipeline (normalize, filter zeros, cache partition, credential resolution, group-by-contract, pLimit dispatch, cache write-back, event emission). `userDecrypt` and `delegatedUserDecrypt` are now two entry points into one shared internal pipeline instead of two parallel 130-line methods.
- **`DelegationService`** — owns delegate/revoke/isDelegated/expiry validation. Consolidates logic that was previously scattered across `ZamaSDK`, `Token` (full reimplementations), `ReadonlyToken.#assertDelegationActive`, and the standalone `delegation-check.ts` utility. That utility is deleted.
- **`CachingService`** — extracted from the former `DecryptCache` class; now injected into `DecryptionService` rather than exposed as a public `sdk.cache` property. Eliminates the double-cache-check issue where `ReadonlyToken` and `ZamaSDK` both queried the cache for the same call.

**`ZamaSDK`** is now a thin orchestrator: it constructs and wires the services, and its public methods delegate through. ~400 lines removed.

**`Token.delegateDecryption` / `Token.revokeDelegation`** collapsed from full reimplementations (~80 and ~55 lines) to one-liner delegations to `sdk.delegateDecryption` / `sdk.revokeDelegation`.

**`ReadonlyToken.batchDecryptBalancesAs`** simplified: delegates to `sdk.delegatedBatchDecryptHandlesAs` instead of orchestrating cache lookups, delegation pre-flight, and per-handle decrypt itself.

**Tests refocused**: service-level unit tests added for `DecryptionService`, `DelegationService`, and `CachingService`. The monolithic `zama-sdk.test.ts` is trimmed to integration-level coverage, and the delegation tests in `token/__tests__/delegation.test.ts` are simplified since the logic now lives in one place.

**Net diff**: ~2,800 lines added, ~3,950 removed (−1,150 net).

Closes SDK-161


ghermet (approved):
Don't forget to update the docs as the cache is no longer public on the API

PR #345, opened 2026-05-12
feat(sdk): offline signing `prepare` for out-of-process custody (SDK-75)

ghermet:
## Summary

Adds the offline-signing surface for SDK-75 — a signer-less `prepare` step that builds an RLP-encoded unsigned transaction for institutional custody, HSM ceremonies, and policy-engine workflows where signing runs **out-of-process** and cannot happen synchronously in a single Promise.

Closes [SDK-75](https://linear.app/zama/issue/SDK-75).

**Atomic call sites are unchanged.** Online signers keep their `writeContract` path and existing flows work identically. The new surface is a *parallel* route: the SDK builds unsigned bytes, hands them off, and steps out of the lifecycle.

## The model: the SDK prepares, the custodian signs *and* broadcasts

The core decision on this branch: **the SDK's only job in the deferred path is to build the unsigned transaction.** It does not sign it, does not broadcast it, and does not track its receipt. A custodian (MPC, HSM, policy engine) signs and submits the bytes through its own channel, out-of-band.

This is why the surface collapsed to a single method:

```ts
sdk.offline.prepare(request, options?)   // → PreparedFor<K>  (RLP unsigned tx + { kind, from })
```

Earlier revisions of this branch shipped `sign`, `broadcast`, `resume`, and `refresh` on this namespace. All were removed — every one of them assumed the SDK re-enters the lifecycle after signing, which real custody flows don't do. Keeping only `prepare` removes the payload↔hash binding problem entirely: there's no hash for the SDK to verify because the SDK never sees the signed bytes.

### `sdk.offline.prepare`

Backed by `OfflineSigningService` (owned by `ZamaSDK`). Signer-optional — works with **no configured signer**, the canonical shape for cross-process custody. Encryption, public-decrypt round-trips (for `FinalizeUnwrap`), and on-chain balance reads (for `UnwrapAll`) all happen during `prepare`; the result is frozen bytes.

Returns `PreparedFor<K>` (narrows `PreparedTransaction` by `kind`):

```ts
interface PreparedTransaction {
  readonly kind: TransactionKind;  // classifies the request; not recoverable from the bytes
  readonly from: Address;          // which key/wallet the custodian signs with; absent from an unsigned EIP-1559 tx
  readonly unsignedTx: Hex;        // RLP-encoded EIP-1559 unsigned tx
}
```

`chainId`, `nonce`, `to`, `value`, calldata, and gas/fee bounds are **not** carried as fields — they're all encoded in `unsignedTx` and recovered by RLP-decoding it. Only `from` and `kind` (which the bytes can't yield) travel alongside. All three fields are JSON-safe, so a `PreparedTransaction` `JSON.stringify`'s and ships across a process boundary as-is — no `bigint` landmines.

### `GenericProvider.prepareTransaction` (new, required)

```ts
prepareTransaction({ from, calldata, nonce?, gasLimit?, maxFeePerGas?, maxPriorityFeePerGas? }): Promise<Hex>
```

Builds the RLP-encoded EIP-1559 unsigned tx: `encodeFunctionData` + `estimateGas` + `getTransactionCount("pending")` + `estimateFeesPerGas` + `serializeTransaction` (viem), the analogous `populateTransaction` + `unsignedSerialized` pipeline (ethers v6). Wired in viem, ethers, and wagmi adapters.

- **`"pending"` nonce tag** keeps parallel custodian queues (multiple in-flight approvals against one wallet) from colliding on a stale `"latest"` count. Optional `nonce`/fee/gas overrides let a custodian with its own nonce/fee manager pin values at prepare time — the signed payload is frozen once exported, so pin generous fee bounds for long-latency ceremonies.
- **ethers overload fix** (`EthersProvider.prepareTransaction`): resolve overloaded ABI entries by name + arity. Bare `iface.encodeFunctionData(name, args)` throws on ERC-7984's `confidentialTransfer(address,bytes32)` vs `confidentialTransfer(address,bytes32,bytes)` because ethers' overload disambiguation treats the trailing value as a potential overrides object.

### Eleven transaction kinds wired

`ConfidentialTransfer`, `ConfidentialTransferFrom`, `SetOperator`, `Unwrap`, `UnwrapAll`, `FinalizeUnwrap`, `ApproveUnderlying`, `Wrap`, `TransferAndCall`, `DelegateDecryption`, `RevokeDelegation`. `FinalizeUnwrap` does the `publicDecrypt` round-trip during `prepare`; `UnwrapAll` reads the on-chain balance handle instead of encrypting; `SetOperator` makes `until` a **required** field (the atomic path's relative "now + 1h" default would silently expire mid-ceremony once the payload is frozen, and a far-future sentinel would grant a de-facto permanent operator — so the caller must state the expiry), while `DelegateDecryption` with an omitted expiry bakes a `uint64`-max value.

### DFNS integration test

`packages/sdk/src/services/__tests__/dfns.integration.test.ts` — real DFNS + real Sepolia + real FHEVM relayer, exercising the canonical flow: signer-less `prepare` → DFNS signs the bytes under policy approval → **DFNS broadcasts**. Env-gated via zod, skips cleanly without creds. `pnpm test:integration` runs it. Satisfies SDK-75 acceptance criterion #7 (reference example against a real custodian environment).

## How callers use it — cross-process deferred (the canonical custodian pattern)

Web initiator runs the SDK **without a signer**. A back-end signer service holds the custodian's API client. The Zama SDK is not on the back-end — it only deals with bytes.

```ts
// Web initiator (signer-absent)
const sdk = new ZamaSDK(createConfig({ chains, relayer, provider }));

const prepared = await sdk.offline.prepare({
  kind: "ConfidentialTransfer",
  from: userAddress,
  token: tokenAddress, to: recipient, amount: 1000n,
});

// Custodian signs AND broadcasts (policy approval happens here); owns the tx hash:
await fetch("/api/sign-and-broadcast", {
  method: "POST",
  body: prepared.unsignedTx, 
});
```

Because the custodian broadcasts, **it owns receipt-tracking and cache refresh** for that submission — the SDK doesn't join the tx lifecycle. Air-gapped tooling is a special case with manual byte transfer (the back-end queue becomes a USB stick); pin generous fee bounds at `prepare` time since the exported payload is frozen once signed.

> This pattern authorizes **transactions**. Authorizing **decryption** out-of-process (the EIP-712 permit path) is not available yet — see below.

## Decryption today: the signer-boundary model

The offline `prepare` path covers **transactions only**. To authorize decryption, wrap the custodian's API in a `BaseSigner` subclass whose async `signTypedData` routes the permit envelope to the custody / policy engine and awaits it, then call `sdk.permits.grantPermit(contracts)`. That works whenever the signature can be awaited inside a callback (the DFNS pattern):

```ts
class DfnsSigner extends BaseSigner {
  async signTypedData(typedData: EIP712TypedData): Promise<Hex> {
    const { id } = await this.dfns.wallets.generateSignature({
      walletId: this.walletId, body: { kind: "Eip712", ...reshapeForDfns(typedData) },
    });
    return (await this.awaitApproval(id)).signature as Hex;   // blocks until DFNS policy approves
  }
}

const sdk = new ZamaSDK(createConfig({ chains, provider, relayer, signer: new DfnsSigner(...) }));
await sdk.permits.grantPermit([token]);         // one blind-sign, routed through DFNS approval
const balance = await token.balanceOf(owner);   // decrypts using the cached permit
```

A true out-of-process permit **decompose** (`preparePermit` → sign elsewhere → `registerPermit`) is **not** available — it's blocked one layer down: `@fhevm/sdk` (0.13.2) `signDecryptionPermit` bundles build + sign behind a required `signer` and exposes no "build unsigned envelope" action. Tracked as SDK-75 follow-up.

## Deliberately not in this PR

- **`sign` / `broadcast` / `resume` / `refresh`.** Removed — see "The model" above. The SDK never handles signed bytes in the deferred path.
- **`signTransaction` signer capability.** Removed. `GenericSigner` keeps `signTypedData` (required, for permits) and `writeContract` (required, atomic path); there is no in-process transaction-signing capability, because deferred signing is external by definition — the deferred path is signer-less (it uses the provider directly), so it needs no signer method of its own.
- **`GenericProvider.sendRawTransaction`.** Removed from the interface — with no SDK-side broadcast, nothing calls it.
- **Offline React hooks and TanStack Query wrappers.** No `usePrepare`, no `prepareMutationOptions` — prototyped and removed, no consumers. Offline signing ships on the core SDK only.
- **`WrappedToken.prepareShield` / `ShieldPlan` planner.** Dropped. Compose a deferred shield from the discrete `ApproveUnderlying` + `Wrap` kinds (or the single `TransferAndCall` kind for ERC-1363 underlyings) at the call site.
- **Atomic auto-routing** for offline signers. Scoped out — atomic stays unchanged for existing callers.
- **Custodian guide / deferred shield-and-unshield examples.**

## Test plan

- [x] `pnpm typecheck` — clean across the workspace
- [x] `pnpm test:run` — green (1203 unit tests in the SDK package post-refactor)
- [x] `pnpm build` — clean (rolldown; `@internal` boundary verified)
- [x] `pnpm api-report` — regenerated (`sdk.api.md`, `sdk-viem`, `sdk-ethers`, `sdk-query`) to reflect the removed `sendRawTransaction` / `broadcast` / error types and the `call → calldata` rename
- [x] `pnpm size` — under budget
- [x] DFNS integration suite passes end-to-end on Sepolia (prepare → sign → DFNS broadcast)
- [ ] Atomic path regression on a Sepolia browser dApp (existing examples)

## Risks

- `GenericProvider` gains one new required method (`prepareTransaction`). Custom adapter implementers (none known in the wild) must implement it; viem / ethers / wagmi are wired here.
- `GenericSigner.writeContract` stays **required**. The deferred path is signer-less — it drives the provider directly — so it needs no signer capability of its own, and the atomic path still relies on `writeContract`. Third-party adapters typed via `implements GenericSigner` are unaffected.
- The deferred surface has no shipped real-world callers yet, so the `sdk.offline.*` namespace is validated by tests + the DFNS integration run, not production traffic.


ghermet:
Blocked by lack of EIP712 creation method on fhevm/sdk

packages/react-sdk/src/wagmi/wagmi-provider.ts
```diff
@@ -81,4 +92,84 @@ export class WagmiProvider implements GenericProvider {
...
+      args: call.args as readonly unknown[],
+    });
+    // Wrap estimateGas — pre-flight revert is the high-value failure mode.
+    // Skip the network round-trips entirely when the caller supplied
+    // overrides — useful for custodians with their own nonce/fee managers.
+    const chainIdPromise = publicClient.getChainId();
+    const noncePromise =
+      args.nonce !== undefined
+        ? Promise.resolve(args.nonce)
+        : publicClient.getTransactionCount({ address: from });
```
ghermet: [AUTOMATED] **Bug — nonce read uses `"latest"`, not `"pending"`.** `getTransactionCount({ address: from })` defaults to `blockTag: "latest"`, diverging from both sibling adapters (`viem-provider.ts:110`, `ethers-provider.ts:179` use `"pending"`) and the `GenericProvider.prepareTransaction` interface JSDoc you added in this PR (`types/provider.ts:77`): *"Nonce reads use the `"pending"` block tag so parallel custodian queues … don't collide on a stale `"latest"` count."*

A custodian preparing a second tx for the same wallet before the first mines reads a stale nonce; since the nonce is baked into the frozen RLP, `broadcast` fails "nonce too low" or replaces the pending tx. The wagmi test stubs `getTransactionCount` but never asserts its argument, so this is uncaught.

```ts
: publicClient.getTransactionCount({ address: from, blockTag: "pending" });
```

Confidence: 88/100

packages/sdk/src/types/offline-signing.ts
```diff
@@ -0,0 +1,266 @@
...
+export interface SetOperatorRequest {
+  /** Discriminator tag. */
+  readonly kind: "SetOperator";
+  /** Tx-sender wallet address (the token holder granting the operator). */
+  readonly from: Address;
+  /** Confidential token contract address. */
+  readonly token: Address;
+  /** Operator address to approve or revoke. */
+  readonly operator: Address;
+  /** Unix timestamp the approval expires at; omit for permanent. */
```
ghermet: [AUTOMATED] **Bug / doc mismatch — `SetOperator` with omitted `until` is not "permanent".** This field is documented *"omit for permanent"*, but `setOperatorContract` (`contracts/confidential-wrapper.ts:174`) defaults an omitted `until` to `Math.floor(Date.now()/1000) + 3600` — baked at **prepare** time into a payload the offline pipeline explicitly freezes for long / air-gapped ceremonies. If sign→broadcast exceeds ~1h, the operator grant lands already-expired (silent no-op); under 1h it's a 1-hour grant, never "permanent." Note `DelegateDecryption` (line 175) handles this correctly with `uint64.max`.

**Fix:** for the offline path map an omitted `until` → a far-future / `uint64.max` sentinel like `DelegateDecryption`, or require an explicit `until`; at minimum correct this JSDoc to state the real "defaults to now + 1h" behavior.

Confidence: 80/100

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,508 @@
...
+    } as PreparedFor<K>;
+  }
+
+  // ── sign ────────────────────────────────────────────────────────────────
+
+  /**
+   * Sign a prepared transaction with the configured signer and return
+   * RLP-encoded signed bytes. Pair with {@link broadcast}.
+   *
+   * @throws {@link SignerCapabilityError} when the configured signer has no
```
ghermet: [AUTOMATED] **Nit (convention) — `@throws` link precedes the condition.** Per the repo convention, `{@link}` placed right after `@throws ` sits in the type slot and renders as literal text; the condition goes first and the link at the end. The public counterpart already does it right (`namespaces/offline.ts:104-106`: `@throws if signer lacks 'signTransaction'. {@link SignerCapabilityError}`). Internal-only doc, so low severity — worth matching the sibling on both this and the `SigningFailedError` line below.

Confidence: 92/100

ghermet (commented):
[AUTOMATED]

## Code Review — offline signing (SDK-75)

Reviewed the full diff with parallel bug-hunter / guideline / error-handling / protocol-compatibility passes, then validated every finding against the checked-out code. The pipeline is solid — `FinalizeUnwrap`'s `publicDecrypt` round-trip, `ConfidentialTransferFrom` owner-binding, the ethers overload-by-arity fix, the `keccak256(signedTx)` hash binding, and the `assertWriteContract` guarding of every internal call site all check out.

3 inline findings below (1 bug, 1 bug/doc mismatch, 1 nit).

**FYI, no code change needed:**
- The `GenericProvider` (+2 required methods) and `GenericSigner.writeContract` (required→optional) breaks are source-breaking only for third-party *structural* implementers, are on the unreleased alpha line, and are already documented in your Risks section. ✓
- **Stale PR description:** it still documents `WrappedToken.prepareShield` / `ShieldPlan` (dropped in `e5c52a8c7`, no longer in `src/`), and its "deliberately not in this PR" list claims `"pending"`-block-tag nonce reads are a follow-up — but viem + ethers already implement them (only wagmi doesn't; see finding #1).
- Below threshold (glance only): `#assertMatchesConfiguredSigner` skips the address/chain check when a signer is configured but its wallet snapshot isn't ready yet.

docs/gitbook/src/changelog/alpha.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Alpha
-description: Unreleased changes on the prerelease (alpha) line — not yet in a stable release.
+description: Unreleased changes on the prerelease (alpha) line — not yet in a stable release. Currently the offline-signing pipeline for institutional custody.
```
zmalatrax: Is it required to mention the current unreleased features in the Alpha description? The sections are self-sufficient, no?

packages/sdk/etc/sdk-query.api.md
```diff
@@ -1365,6 +1373,8 @@ export class ZamaSDK {
     readonly delegations: Delegations;
     dispose(): void;
     encrypt(params: EncryptParams, options?: Pick<FhevmRelayerOptions, "signal" | "timeout">): Promise<EncryptResult>;
+    // Warning: (ae-forgotten-export) The symbol "Offline" needs to be exported by the entry point index.d.ts
```
zmalatrax: ? We might need to re-export Offline namespace on query/index.ts ?

packages/sdk/src/namespaces/offline.ts
```diff
@@ -0,0 +1,47 @@
+import type {
+  OfflineSigningOptions,
+  OfflineSigningService,
+} from "../services/offline-signing-service";
```
zmalatrax: Following the rename of the class from OfflineSigning to Offline, internals should drop the `Signing` part as well, renaming the file as well

packages/sdk/src/services/offline-service.ts
zmalatrax, line 1: https://github.com/zama-ai/sdk/pull/345/changes#r3703365739

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  #buildSetOperator(request: SetOperatorRequest): ReturnType<typeof setOperatorContract> {
+    // Offline payloads are frozen at prepare time, so an omitted `until` must
+    // mean "permanent" (uint48 max) — never a relative default like the atomic
+    // path's now + 1h, which would silently expire mid-ceremony. Mirrors the
+    // MAX_UINT64 sentinel `#buildDelegateDecryption` uses for the same reason.
+    return setOperatorContract(
+      request.token,
+      getAddress(request.operator),
+      request.until ?? MAX_UINT48,
```
zmalatrax: Should it really be made permanent if omitted? It looks like not-so-good security hygiene
We could make it a required field
Or a reasonable (and expirable) default, e.g. 1 week or 1 month ; It should be long-enough for proper offline signing

packages/sdk/src/types/provider.ts
```diff
@@ -36,4 +39,42 @@ export interface GenericProvider {
...
+   * `encodeFunctionData` + `estimateGas` + `getTransactionCount` +
+   * `estimateFeesPerGas` + `serializeTransaction` (viem), the analogous
+   * `Contract.populateTransaction` + `Transaction.unsignedSerialized`
+   * pipeline (ethers v6).
+   *
+   * Nonce reads use the `"pending"` block tag so parallel custodian queues
+   * with multiple in-flight approvals against the same wallet don't collide
+   * on a stale `"latest"` count. Pass `nonce` to bypass the read entirely
+   * (e.g. when the custodian assigns nonces itself).
+   */
```
zmalatrax: If 2 parallel offline prepared tx are queued, but none of them are broadcasted (i.e. multiple in-flight approvals against the same wallet), reading the "pending" nonce or the "latest" one will be identical as no tx are broadcasted (hence none in a mempool/sequencer)

packages/sdk/src/types/signer.ts
zmalatrax, line 80: `writeContract` is still required in GenericSigner, while it should have been made optional (as per the docs update)

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+    return confidentialTransferContract(request.token, request.to, handle, inputProof);
+  }
+
+  async #buildConfidentialTransferFrom(
+    request: ConfidentialTransferFromRequest,
+  ): Promise<ReturnType<typeof confidentialTransferFromContract>> {
+    const { encryptedValues, inputProof } = await this.#encryption.encryptValues({
+      values: [{ value: request.amount, type: "euint64" }],
+      contractAddress: request.token,
+      userAddress: getAddress(request.owner),
```
zmalatrax: Should it be the `request.from` address? (i.e. the tx sender) I haven't found e2e test on this (where owner != from)
cf. https://github.com/zama-ai/fhevm/blob/07fb05fb75f0aa6cea934088640ddb4539d0b1b9/host-contracts/lib/Impl.sol#L688

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+  }
+
+  #buildTransferAndCall(
+    request: TransferAndCallRequest,
+  ): ReturnType<typeof transferAndCallContract> {
+    return transferAndCallContract(
+      request.underlying,
+      request.wrapper,
+      request.amount,
+      request.recipientData,
```
zmalatrax, line 300: The wrapper callback checks the recipient data length, and truncates to the first 20 bytes if not exactly 20 bytes: https://github.com/zama-ai/protocol-apps/blob/66d110ec7aa3af485f28090281d6fd341c774875/contracts/confidential-wrapper/contracts/extensions/ERC7984ERC20WrapperUpgradeable.sol#L84

We should add validation of recipientData, forwarding if it's 20-byte long, or truncate to `0x` otherwise (to avoid having addresses ABI-encoded on 32 bytes being truncated, potentially minting the shielded funds to a garbage address)

packages/sdk/src/signer/base-signer.ts
```diff
@@ -1,26 +1,21 @@
...
 /**
- * Abstract base class that implements the shared {@link GenericSigner}
- * boilerplate: wallet-account store, `requireWalletAccount`, idempotent
- * `dispose` / `Disposable`. Subclasses provide `signTypedData`,
- * `writeContract`, and optionally override `onDispose` for cleanup.
+ * Abstract base class that supplies the wallet-account / dispose
+ * boilerplate every signer adapter needs: an observable wallet-account
+ * store, `requireWalletAccount`, and an idempotent `dispose` /
+ * `Disposable`. Subclasses provide `signTypedData` and at least one
+ * tx-signing capability (`writeContract` and/or `signTransaction`) — the
```
zmalatrax: stale docs, `signTransaction` has been removed?

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+   * `sdk.permits.grantPermit`, which signs with the configured signer
+   * (including an out-of-process custody signer).
+   */
+  async prepare<K extends TransactionKind>(
+    request: Extract<PrepareTransactionRequest, { kind: K }>,
+    options?: OfflineSigningOptions,
+  ): Promise<PreparedFor<K>> {
+    const from = getAddress(request.from);
+    const calldata = await this.#buildCalldata(request, from);
+    const unsignedTx = await this.#provider.prepareTransaction({
```
zmalatrax: Shouldn't we add a guard on chain alignment between router and provider, similar to the classical flow (requireChainAlignment)?

Something like
```typescript
if (await this.#provider.getChainId() !== this.#router.chain.id) throw new ChainMismatchError(...)
```

packages/sdk/src/viem/viem-provider.ts
```diff
@@ -60,4 +72,72 @@ export class ViemProvider implements GenericProvider {
...
+                { cause: error },
+              );
+            }));
+    const feesPromise =
+      args.maxFeePerGas !== undefined && args.maxPriorityFeePerGas !== undefined
+        ? Promise.resolve({
+            maxFeePerGas: args.maxFeePerGas,
+            maxPriorityFeePerGas: args.maxPriorityFeePerGas,
+          })
+        : this.#publicClient.estimateFeesPerGas();
```
zmalatrax: if none or only `maxFeePerGas` or `maxPriorityFeePerGas` is provided, gas estimation is performed ; with a low `maxFeePerGas`, `maxPriorityFeePerGas` can be estimated higher, which will throw on serializeTransaction (`TipAboveFeeCapError`): https://github.com/wevm/viem/blob/88d68aa3896c88930c84387ca4145653a7c9615f/src/celo/serializers.ts#L105-L110

Same for ethers/wagmi
I'd enforce providing none or both (or clamp maxPriorityFeePerGas to minimum between estimated and provided maxFeePerGas.. but imo enforcing none or both is better)

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+
+/**
+ * Offline-signing pipeline. Separates `prepare` and `broadcast` for
+ * institutional custody and policy-engine workflows where signing runs
+ * out-of-process and cannot happen synchronously in a single Promise.
+ *
+ * Atomic call sites ({@link Token.confidentialTransfer}, etc.) keep their
+ * `signer.writeContract` path; this service is the parallel route where the
+ * caller signs `prepared.unsignedTx` externally (HSM, custody API, policy
+ * engine) and feeds the signed bytes back through {@link broadcast}.
```
zmalatrax: stale mention of`broadcast` method, it has been removed from offline namespace?

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+    this.#encryption = config.encryption;
+  }
+
+  // ── prepare ─────────────────────────────────────────────────────────────
+
+  /**
+   * Build the offline-signing payload for the given transaction request:
+   * an RLP-encoded unsigned transaction the caller signs externally (an HSM,
+   * custody API, or any out-of-process signer) and feeds back through
+   * {@link broadcast}.
```
zmalatrax: https://github.com/zama-ai/sdk/pull/345/changes#r3703888189

packages/sdk/src/services/offline-signing-service.ts
```diff
@@ -0,0 +1,316 @@
...
+ * The nonce and fees are part of the RLP that gets signed, so a custodian
+ * that signs the bytes you hand it (rather than assembling the tx itself)
+ * cannot fill or change them afterwards — own the `nonce` here when signing
+ * through such a path. For long-latency ceremonies (air-gapped HSM signing,
+ * slow multi-party approval) the signed payload is frozen once exported, so
+ * pin generous fee bounds up front rather than expecting to re-stamp later.
+ */
+export interface OfflineSigningOptions {
+  /** Optional {@link AbortSignal} to cancel the in-flight chain reads. */
+  readonly signal?: AbortSignal;
```
zmalatrax: This field is not wired anywhere (GenericProvider.prepareTransaction, #buildCalldata..), should it be kept?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -257,6 +277,7 @@ export class CredentialService {
    * @throws if the underlying storage delete fails.
    */
   async revokeTransportKeyPair(scopeId: string): Promise<void> {
+    this.#requireSigner("revokeTransportKeyPair");
```
zmalatrax: I don't think that revoking the transport key pair requires a signer, it's storage ops only ; docs mentions "operator-level action"
I'd drop the guard

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -296,6 +317,7 @@ export class CredentialService {
    * @throws if no scope is configured, or `scopeId` doesn't match it. {@link ConfigurationError}
    */
   async warmTransportKeyPairScope(scopeId: string): Promise<void> {
+    this.#requireSigner("warmTransportKeyPairScope");
```
zmalatrax: I'd drop the guard, docs mention 'operator-level, no wallet needed'
Similar to https://github.com/zama-ai/sdk/pull/345/changes#r3703987018

packages/sdk/src/viem/__tests__/viem.test.ts
```diff
@@ -304,6 +304,74 @@ describe("ViemProvider", () => {
...
+            abi: balanceOfAbi,
+            functionName: "balanceOf",
+            args: [userAddress],
+          },
+        });
+
+        expect(publicClient.getTransactionCount).toHaveBeenCalledWith({
+          address: userAddress,
+          blockTag: "pending",
+        });
```
zmalatrax, line 367: No assertions on the produced tx bytes, tests should parse and assert the decoded tx fields against their expected values

packages/sdk/src/viem/__tests__/viem.test.ts
```diff
@@ -304,6 +304,74 @@ describe("ViemProvider", () => {
...
+          calldata: {
+            address: tokenAddress,
+            abi: balanceOfAbi,
+            functionName: "balanceOf",
+            args: [userAddress],
+          },
+          nonce: 42,
+        });
+
+        expect(publicClient.getTransactionCount).not.toHaveBeenCalled();
```
zmalatrax, line 388: https://github.com/zama-ai/sdk/pull/345/changes#r3704112269

packages/sdk/src/ethers/__tests__/ethers.test.ts
```diff
@@ -527,10 +527,73 @@ describe("EthersProvider", () => {
...
+        from: userAddress,
+        calldata: {
+          address: tokenAddress,
+          abi: balanceOfAbi,
+          functionName: "balanceOf",
+          args: [userAddress],
+        },
+      });
+
+      expect(mockProvider.getTransactionCount).toHaveBeenCalledWith(userAddress, "pending");
```
zmalatrax, line 598: https://github.com/zama-ai/sdk/pull/345/changes#r3704112269

packages/sdk/src/ethers/__tests__/ethers.test.ts
```diff
@@ -527,10 +527,73 @@ describe("EthersProvider", () => {
...
+        calldata: {
+          address: tokenAddress,
+          abi: balanceOfAbi,
+          functionName: "balanceOf",
+          args: [userAddress],
+        },
+        nonce: 42,
+      });
+
+      expect(mockProvider.getTransactionCount).not.toHaveBeenCalled();
```
zmalatrax, line 619: https://github.com/zama-ai/sdk/pull/345/changes#r3704112269

packages/sdk/src/services/__tests__/dfns.integration.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        token: env.TOKEN_ADDRESS,
+        to: env.RECIPIENT_ADDRESS,
+        amount: 1n * 10n ** 6n,
+      });
+      expect(prepared.unsignedTx).toMatch(/^0x[0-9a-f]+$/i);
+
+      // The custodian owns signing AND broadcast — the SDK's job ended at
+      // `prepare`. A real tx hash back proves the full cross-process round-trip.
+      const { txHash } = await signAndBroadcast(prepared.unsignedTx);
+      expect(isHash(txHash)).toBe(true);
```
zmalatrax, line 189: a tx hash could be returned even though the tx reverted on chain, no?
I'd wait for the corresponding receipt and verify it succeeded

packages/sdk/src/viem/viem-provider.ts
```diff
@@ -60,4 +72,72 @@ export class ViemProvider implements GenericProvider {
...
+    const noncePromise =
+      args.nonce !== undefined
+        ? Promise.resolve(args.nonce)
+        : this.#publicClient.getTransactionCount({ address: from, blockTag: "pending" });
+    const gasPromise =
+      args.gasLimit !== undefined
+        ? Promise.resolve(args.gasLimit)
+        : (calldata.gas ??
+          this.#publicClient
+            .estimateGas({ account: from, to: calldata.address, data, value: calldata.value ?? 0n })
```
zmalatrax, line 110: Reflecting on the PR body:
> WrappedToken.prepareShield / ShieldPlan planner. Dropped. Compose a deferred shield from the discrete ApproveUnderlying + Wrap kinds (or the single TransferAndCall kind for ERC-1363 underlyings) at the call site.

If a user does `prepare(ApproveUnderlying)` and `prepare(Wrap)`, as long as the approval has not been mined on-chain, the `prepare(Wrap)` gas estimation will fail ; Providing a `gasLimit` allows skipping this gas estimation, though the nonce resolution will provide the same nonce as the `prepare(ApproveUnderlying)` one unless it has been broadcasted already (cf. this [review comment](https://github.com/zama-ai/sdk/pull/345/changes#r3703532907)), or set manually
--> It's a bit tricky to make an offline shielding operation and the docs does not reflect it

zmalatrax (changes requested):

packages/sdk/etc/sdk-query.api.md
ghermet: Done — `query/index.ts` now re-exports the namespace (`export type { … Offline } from "../namespaces"`) plus `OfflineOptions` from the service.

packages/sdk/src/namespaces/offline.ts
ghermet: Done — renamed throughout: `OfflineSigningService` → `OfflineService`, file `offline-signing-service.ts` → `offline-service.ts`, options `OfflineSigningOptions` → `OfflineOptions`. The public namespace class is `Offline`.

packages/sdk/src/services/offline-signing-service.ts
ghermet: Agreed — made `until` a **required** field on the offline `SetOperator` request rather than defaulting it. Both alternatives are unsafe for a frozen payload: the atomic path's relative `now + 1h` default would silently expire mid-ceremony, and a far-future sentinel would grant a de-facto permanent operator. So the caller must state the expiry explicitly. See `offline-service.ts` `#buildSetOperator`.

packages/sdk/src/types/provider.ts
ghermet: Correct — `"pending"` only diverges from `"latest"` once an earlier tx has actually been broadcast; several payloads prepared before any broadcast all read the same count. Kept `"pending"` as the default (right once a queue starts draining) but documented the limitation and the remedy — assign the nonces yourself — on `OfflineOptions.nonce` and in the changelog hint (**Dependent txs must pin `nonce` and `gasLimit`**).

packages/sdk/src/types/signer.ts
ghermet: Decision reversed the other way: `writeContract` stays **required** on `GenericSigner`. The deferred path is signer-less (it drives the provider directly), so it needs no signer method of its own, and the atomic path still relies on `writeContract`. The code was already required (`types/signer.ts`); the stale "optional" wording was only in the PR description, now corrected.

packages/sdk/src/services/offline-signing-service.ts
ghermet: Right — it is the tx sender. The encrypted-input proof binds to `msg.sender` (fhevm `verifyInput`), which for `transferFrom` is the operator == the `from` wallet that signs, **not** the `owner` being debited — so encryption uses `from` while `owner` is passed through as the distinct `confidentialTransferFrom(owner, …)` arg. Added unit coverage with `owner != from`: `offline-service.test.ts` — "ConfidentialTransferFrom encrypts under the tx-sender (`from`), not the `owner`" and "args are [owner, to, handle, inputProof]".

packages/sdk/src/services/offline-signing-service.ts
ghermet: Good catch — added validation on `TransferAndCall.recipientData`: it must be omitted, `0x` (self-shield to the sender), or a raw 20-byte address; anything else (e.g. a 32-byte ABI-encoded address) is rejected with a `ConfigurationError` explaining the wrapper's first-20-bytes truncation footgun, rather than silently minting to a garbage address. See `offline-service.ts` `#buildTransferAndCall`.

packages/sdk/src/signer/base-signer.ts
ghermet: Done — dropped the `signTransaction` mention; the class doc now states subclasses provide `signTypedData` and `writeContract` only.

packages/sdk/src/services/offline-signing-service.ts
ghermet: Done — added a chain-alignment guard at the top of `prepare`. Offline is signer-less so it can't reuse `requireChainAlignment`; it compares `provider.getChainId()` against `router.chain.id` directly and throws `ConfigurationError` on mismatch (the prepared tx's contract addresses and encrypted inputs are bound to the configured chain).

packages/sdk/src/viem/viem-provider.ts
ghermet: Done — enforced none-or-both, at the type level rather than at runtime. `GenericProvider.prepareTransaction`'s arg (and `OfflineOptions`) is now a union: either both `maxFeePerGas`/`maxPriorityFeePerGas` are `never`, or both are required. A partial pair is a compile error across viem/ethers/wagmi.

packages/sdk/src/services/offline-signing-service.ts
ghermet: Done — removed. There is no `broadcast` method on the offline namespace; the SDK never handles signed bytes in the deferred path (the caller signs and publishes out-of-process), and the service doc reflects that.

packages/sdk/src/services/offline-signing-service.ts
ghermet: Removed — that was the `signal?: AbortSignal` field, which was never threaded into `prepareTransaction`/`#buildCalldata`. `OfflineOptions` now carries only fields that actually flow through (`nonce`, `gasLimit`, and the `maxFeePerGas`/`maxPriorityFeePerGas` pair).

packages/sdk/src/credentials/credential-service.ts
ghermet: Agreed — dropped the signer guard. `revokeTransportKeyPair` is a storage-only, operator-level action (deletes the shared key pair from local storage, no connected wallet needed). Only the scope-config guards remain.

packages/sdk/src/credentials/credential-service.ts
ghermet: Same — dropped the guard on `warmTransportKeyPairScope` (storage-only, operator-level, no wallet needed), matching `revokeTransportKeyPair`.

packages/sdk/src/viem/__tests__/viem.test.ts
ghermet: Done — the prepare tests now `parseTransaction(unsignedTx)` and assert every decoded field against its expected value: `type`, `chainId`, `nonce`, `to`, `value`, `gas`, `maxFeePerGas`, `maxPriorityFeePerGas`, and `data`.

packages/sdk/src/ethers/__tests__/ethers.test.ts
ghermet: Done here too — `ethers.test.ts` now `parseTransaction(unsignedTx)` and asserts the decoded `type`/`chainId`/`nonce`/`to`/`value`/`gas`/`maxFeePerGas`/`maxPriorityFeePerGas`/`data`.

packages/sdk/src/services/__tests__/dfns.integration.test.ts
ghermet: Agreed — a returned hash does not imply on-chain success. The integration test now waits for the receipt on the underlying provider and asserts `status === 1` (the SDK's `GenericProvider` receipt only carries `logs`, so it goes to the raw provider for `status`).

packages/sdk/src/viem/viem-provider.ts
ghermet: Documented in the changelog offline section: multi-step flows compose the discrete kinds (`ApproveUnderlying` then `Wrap`, or a single `TransferAndCall` for ERC-1363 underlyings), plus a **Dependent txs must pin `nonce` and `gasLimit`** warning — preparing the `Wrap` before the `ApproveUnderlying` is broadcast means gas estimation reverts (allowance not yet on-chain) and the `"pending"` nonce is identical for both, so pin an incrementing `nonce` + explicit `gasLimit`, or broadcast-and-confirm each tx before preparing the next.

docs/gitbook/src/changelog/alpha.md
ghermet: The intro is the fixed stage-gate banner every alpha page carries — it flags the release stage (unreleased / will be retitled and folded on release), not the specific features. It is feature-agnostic on purpose, so the per-feature sections below stay self-sufficient. Kept as-is for consistency with the changelog convention.

packages/sdk/etc/sdk-query.api.md
```diff
@@ -574,6 +581,28 @@ export interface MutationFactoryOptions<TMutationKey extends readonly unknown[],
     onSuccess?: (data: TData, variables: TVariables, onMutateResult: TOnMutateResult, context: MutationFunctionContext) => void;
 }
 
+// @public
+export class Offline {
+    // Warning: (ae-forgotten-export) The symbol "TransactionKind" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "PrepareTransactionRequest" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "PreparedFor" needs to be exported by the entry point index.d.ts
```
zmalatrax: `Offline.prepare` parameters and return types (PrepareTransactionRequest, PreparedFor, TransactionKind) were not exported along with the class...

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -277,7 +277,8 @@ export class CredentialService {
    * @throws if the underlying storage delete fails.
    */
   async revokeTransportKeyPair(scopeId: string): Promise<void> {
-    this.#requireSigner("revokeTransportKeyPair");
+    // Storage-only, operator-level action: deletes the shared key pair from
+    // local storage and requires no connected wallet — so no signer guard.
```
zmalatrax: I don't think that this comment is necessary, it's already covered by the JSDoc?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -317,7 +318,8 @@ export class CredentialService {
    * @throws if no scope is configured, or `scopeId` doesn't match it. {@link ConfigurationError}
    */
   async warmTransportKeyPairScope(scopeId: string): Promise<void> {
-    this.#requireSigner("warmTransportKeyPairScope");
+    // Storage-only, operator-level action (pre-warm counterpart of
+    // revokeTransportKeyPair): no connected wallet, so no signer guard.
```
zmalatrax: I don't think that this comment is necessary, it's already covered by the JSDoc?

packages/sdk/src/ethers/__tests__/ethers.test.ts
```diff
@@ -593,6 +600,39 @@ describe("EthersProvider", () => {
...
+
+      const decoded = parseTransaction(unsignedTx);
+      expect(decoded.type).toBe("eip1559");
+      expect(decoded.chainId).toBe(1); // from getNetwork
+      expect(decoded.nonce).toBe(7); // from getTransactionCount
+      expect(getAddress(decoded.to!)).toBe(getAddress(tokenAddress));
+      expect(decoded.value ?? 0n).toBe(0n);
+      expect(decoded.gas).toBe(21000n); // from estimateGas
+      expect(decoded.maxFeePerGas).toBe(500n); // pinned override wins over estimate
+      expect(decoded.maxPriorityFeePerGas).toBe(2n);
```
zmalatrax: Not a big fan of magic numbers for expected values

packages/sdk/src/ethers/__tests__/ethers.test.ts
```diff
@@ -593,6 +600,39 @@ describe("EthersProvider", () => {
...
+        maxFeePerGas: 500n,
+        maxPriorityFeePerGas: 2n,
+      });
+
+      const decoded = parseTransaction(unsignedTx);
+      expect(decoded.type).toBe("eip1559");
+      expect(decoded.chainId).toBe(1); // from getNetwork
+      expect(decoded.nonce).toBe(7); // from getTransactionCount
+      expect(getAddress(decoded.to!)).toBe(getAddress(tokenAddress));
+      expect(decoded.value ?? 0n).toBe(0n);
```
zmalatrax: defaulting the decoded value to the expected one is sketchy

packages/sdk/src/types/offline-signing.ts
```diff
@@ -51,8 +61,8 @@ export interface SetOperatorRequest {
   readonly token: Address;
   /** Operator address to approve or revoke. */
   readonly operator: Address;
-  /** Unix timestamp the approval expires at; omit for permanent (uint48 max). */
-  readonly until?: number;
+  /** Unix timestamp (seconds) the approval expires at. Required — no implicit default. */
```
zmalatrax:
```suggestion
  /** Unix timestamp (seconds) the approval expires at. */
```

packages/sdk/src/viem/__tests__/viem.test.ts
```diff
@@ -371,6 +372,42 @@ describe("ViemProvider", () => {
...
+
+        const decoded = parseTransaction(unsignedTx);
+        expect(decoded.type).toBe("eip1559");
+        expect(decoded.chainId).toBe(1); // from getChainId
+        expect(decoded.nonce).toBe(7); // from getTransactionCount
+        expect(getAddress(decoded.to!)).toBe(getAddress(tokenAddress));
+        expect(decoded.value ?? 0n).toBe(0n);
+        expect(decoded.gas).toBe(21000n); // from estimateGas
+        expect(decoded.maxFeePerGas).toBe(500n); // pinned override wins over estimate
+        expect(decoded.maxPriorityFeePerGas).toBe(2n);
```
zmalatrax: https://github.com/zama-ai/sdk/pull/345/changes/68124ccdf8abaa1d8ef75542cd3b68e1be4b40bc..74501e7feafe00143e55bf771f3d04f18e442d5b#r3710733591

packages/react-sdk/src/wagmi/wagmi-provider.ts
```diff
@@ -81,4 +92,85 @@ export class WagmiProvider implements GenericProvider {
...
+          maxPriorityFeePerGas?: never;
+        }
+      | {
+          from: Address;
+          calldata: WriteContractConfig<TAbi, TFunctionName, TArgs>;
+          nonce?: number;
+          gasLimit?: bigint;
+          maxFeePerGas: bigint;
+          maxPriorityFeePerGas: bigint;
+        },
```
zmalatrax: Not a big fan of the union type, and current code for none-or-both fee pair is only at the type level (a guard was introduced and then removed in your last 2 commits)

What about a field `fees?: { maxFeePerGas: bigint, maxPriorityFeePerGas: bigint }` ?
It simplifies the 3 adapters to `args.fees ?? await estimateFeesPerGas()`, and runtime guard simply is the validation of the two members of the `fees` object

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -257,6 +277,8 @@ export class CredentialService {
    * @throws if the underlying storage delete fails.
    */
   async revokeTransportKeyPair(scopeId: string): Promise<void> {
```
zmalatrax, line 279: The `permits.ts` functions methods using this service methods are still documenting throw error if no signer is configured `* @throws if no signer is configured. {@link SignerNotConfiguredError}`

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -296,6 +318,8 @@ export class CredentialService {
    * @throws if no scope is configured, or `scopeId` doesn't match it. {@link ConfigurationError}
    */
   async warmTransportKeyPairScope(scopeId: string): Promise<void> {
```
zmalatrax, line 318: https://github.com/zama-ai/sdk/pull/345/changes#r3711002138

packages/sdk/src/signer/base-signer.ts
```diff
@@ -1,26 +1,20 @@
...
+ * boilerplate every signer adapter needs: an observable wallet-account
+ * store, `requireWalletAccount`, and an idempotent `dispose` /
+ * `Disposable`. Subclasses provide `signTypedData` and `writeContract` — the
+ * resulting concrete class automatically satisfies {@link GenericSigner}.
  *
  * Using this class is optional — implementing {@link GenericSigner} directly
  * with {@link createWalletAccountStore} remains fully supported.
  */
-export abstract class BaseSigner implements GenericSigner, Disposable {
+export abstract class BaseSigner implements Disposable {
```
zmalatrax: if we keep writeContract required (cf. https://github.com/zama-ai/sdk/pull/345#discussion_r3705403815), then we should reintroduce `immplements GenericSigner`

Currently, typecheck fails with things like:

```typescript
class DfnsSigner extends BaseSigner {
  async signTypedData(_typedData: EIP712TypedData): Promise<Hex> {
    return "0x";
  }
}

const signer: GenericSigner = new DfnsSigner();
```

`Property 'writeContract' is missing in type 'DfnsSigner' but required in type 'GenericSigner'.
signer.ts(80, 3): 'writeContract' is declared here.`

packages/sdk/src/services/offline-service.ts
ghermet: Done — the file is `offline-service.ts` and the internals dropped `Signing` (`OfflineService`, `OfflineServiceConfig`, `OfflineOptions`); the public namespace class is `Offline`.

packages/sdk/src/services/offline-signing-service.ts
ghermet: Done — no `broadcast` mention remains in the renamed `offline-service.ts`; the SDK never handles the signed bytes on the deferred path.

packages/sdk/src/viem/__tests__/viem.test.ts
ghermet: The decoded-field assertions live in the dedicated "serializes an EIP-1559 tx whose decoded fields match chain state + overrides" test just below (`parseTransaction` + a per-field assert for type/chainId/nonce/to/value/gas/fees/data). This case is deliberately narrow — it only asserts `getTransactionCount` isn't called when the nonce is pinned.

packages/sdk/src/ethers/__tests__/ethers.test.ts
ghermet: Same as the viem side: the per-field decode assertions are in the dedicated "serializes an EIP-1559 tx whose decoded fields match…" test below. This case only asserts `getTransactionCount` isn't called when the nonce is pinned.

packages/sdk/etc/sdk-query.api.md
ghermet: Done — `query/index.ts` now re-exports `PrepareTransactionRequest`, `PreparedFor`, and `TransactionKind`. The `ae-forgotten-export` warnings on `Offline.prepare` are cleared and the query API report is regenerated (aa8f7009a).

packages/sdk/src/credentials/credential-service.ts
ghermet: Removed.

packages/sdk/src/credentials/credential-service.ts
ghermet: Removed.

packages/sdk/src/ethers/__tests__/ethers.test.ts
ghermet: Done — the decoded fields now assert against named chain-state constants (`CHAIN_ID`/`NONCE`/`GAS`/`ESTIMATED_MAX_*`) shared with the mock, and the pinned overrides are named distinctly (`OVERRIDE_MAX_FEE`/`OVERRIDE_MAX_PRIORITY`) so "override wins over estimate" is observable rather than a re-typed literal.

packages/sdk/src/ethers/__tests__/ethers.test.ts
ghermet: Done — dropped the `?? 0n`. The calldata now carries a non-zero `value` (`TX_VALUE`) and the test asserts `decoded.value === TX_VALUE` exactly. (viem's `parseTransaction` returns `undefined` for a zero value, which is exactly what the default was masking.)

packages/sdk/src/types/offline-signing.ts
ghermet: Applied — the required-ness is already carried by the non-optional field type and the interface-level doc, so the shorter line is enough.

packages/sdk/src/viem/__tests__/viem.test.ts
ghermet: Same fix applied to the viem test — named chain-state constants + a real non-zero `value` assertion, no `?? 0n`.

packages/react-sdk/src/wagmi/wagmi-provider.ts
ghermet: Agreed — done (aa8f7009a). Replaced the union with `fees?: { maxFeePerGas: bigint; maxPriorityFeePerGas: bigint }` on `GenericProvider.prepareTransaction` and `OfflineOptions`. All-or-nothing is now enforced by the object shape (no `?: never`, no runtime guard), and each of the three adapters collapses to `const fees = args.fees ?? estimate…`.

packages/sdk/src/credentials/credential-service.ts
ghermet: Done — dropped the `@throws … SignerNotConfiguredError` and the "a signer must still be configured" prose from `permits.revokeTransportKeyPair` and `warmTransportKeyPairScope`. Both are storage-only, operator-level ops that go through the signer-less path in `CredentialService` (the `CredentialService` is always constructed regardless of signer), so neither actually throws on a missing signer.

packages/sdk/src/credentials/credential-service.ts
ghermet: Same — fixed in the same change (see the reply on the sibling thread).

packages/sdk/src/signer/base-signer.ts
ghermet: Done (cad115670) — restored `implements GenericSigner` on `BaseSigner` and added `writeContract` as an abstract member. A subclass now gets the error at its own definition ("non-abstract class does not implement inherited abstract member `writeContract`") rather than only at a later `GenericSigner` assignment. The real adapters (`ViemSigner`, `EthersSigner`, `WagmiSigner`) already implement it, so no behavior change for them.

packages/sdk/src/services/offline-service.ts
```diff
@@ -0,0 +1,370 @@
...
+    );
+  }
+
+  #buildSetOperator(request: SetOperatorRequest): ReturnType<typeof setOperatorContract> {
+    // `until` is a required field on the offline request: the atomic path's
+    // relative default (now + 1h) would silently expire mid-ceremony once the
+    // payload is frozen, and defaulting to a far-future sentinel would grant a
+    // de-facto permanent operator — both are unacceptable for a frozen offline
+    // payload, so the caller must state the expiry explicitly.
+    return setOperatorContract(request.token, getAddress(request.operator), request.until);
```
zmalatrax: `until` is only required at the type level, what about adding a runtime guard similar to recipientData?

```typescript
if (typeof request.until !== "number") {
  throw new ConfigurationError("...");
}
```

But it's not the best pattern as it scatters validation everywhere, we could consider introducing zod schemas for `PrepareTransactionRequest`, there are already some declared (config, wrapper-registry, caching..) ; it can be revisited in another PR, as a follow-up ; especially if we want to merge a first version of offline signing asap*

packages/react-sdk/src/wagmi/wagmi-provider.ts
ghermet: Correction to my earlier reply — you were right that the runtime guard was the other half of this. Added it in 6f28f2348: each adapter (viem, ethers, wagmi) now `assertBigint`s both `fees` members before use, so a JS caller passing a non-bigint leg is rejected rather than silently serialized. Also guarded `SetOperator.until` with `assertNumber` on the offline builder (required field a JS caller could omit). Each of the four suites gets a rejects-non-bigint / rejects-missing test.

packages/sdk/src/services/offline-service.ts
```diff
@@ -0,0 +1,370 @@
...
+      recipientData,
+    );
+  }
+
+  #buildDelegateDecryption(
+    request: DelegateDecryptionRequest,
+  ): ReturnType<typeof delegateForUserDecryptionContract> {
+    const expDate = request.expirationDate
+      ? BigInt(Math.floor(request.expirationDate.getTime() / 1000))
+      : MAX_UINT64;
```
zmalatrax, line 329: Missing the 1-hour-in-the-future expirationDate check ; done in existing delegation-service: https://github.com/zama-ai/sdk/blob/d4114dbd326075510c824a7510c8121cd0d2828d/packages/sdk/src/services/delegation-service.ts#L80-L84

zmalatrax: It actually also misses the two other validation checks: https://github.com/zama-ai/sdk/blob/d4114dbd326075510c824a7510c8121cd0d2828d/packages/sdk/src/services/delegation-service.ts#L90-L100

ghermet: Done (cd0933072) — `#buildDelegateDecryption` now mirrors the atomic guard: `expirationDate` under 1h from now throws `DelegationExpirationTooSoonError`. Same bound and message as `delegation-service`. Added a rejects-too-soon test (and future-dated the existing timestamp test, whose 2023 date is now in the past). Kept the 1h bound for parity rather than a stricter offline-specific one, though the effective margin is tighter here since the payload is signed/broadcast later.

packages/sdk/src/schemas/offline.ts
```diff
@@ -0,0 +1,365 @@
...
+ * future. The offline payload is frozen at prepare time and signed/broadcast
+ * later, so an expiry under 1h out risks landing already-expired mid-ceremony;
+ * set a far-future timestamp for an effectively permanent grant.
+ */
+const setOperatorUntil = unixSeconds.check(
+  z.refine(
+    (until) => until >= Math.floor(Date.now() / 1000) + ONE_HOUR_SECONDS,
+    "SetOperator.until must be at least 1 hour in the future (unix seconds)",
+  ),
+);
```
zmalatrax: Drop the refine, there should not be minimum on the `until` operator as otherwise users cannot revoke an operator.

To revoke an operator, user has to pass `0 < until < Date.now()`, and there is no "minimum duration" here, the atomic setOperator defaults to 1h for conviniency ; `until: z.int().nonnegative()`

packages/sdk/src/schemas/offline.ts
```diff
@@ -0,0 +1,365 @@
...
+const ONE_HOUR_MS = ONE_HOUR_SECONDS * 1000;
+
+//MARK: Schemas
+
+/** {@link ConfidentialTransferRequest} schema. */
+export const confidentialTransferRequest = z.object({
+  kind: z.literal("ConfidentialTransfer"),
+  from: evmAddress,
+  token: evmAddress,
+  to: evmAddress,
```
zmalatrax: addresses are not checksummed anymore, we should use `checksummedAddress` instead of evmAddress

packages/sdk/src/schemas/offline.ts
```diff
@@ -0,0 +1,365 @@
...
+      z
+        .date()
+        .check(
+          z.refine(
+            (d) => d.getTime() >= Date.now() + ONE_HOUR_MS,
+            "Expiration date must be at least 1 hour in the future",
+          ),
+        ),
+    ),
+  }),
```
zmalatrax: Actually, this check should be in the builder `#buildDelegateDecryption`, zod throws `ConfigurationError` when we'd like more granular errors (`DelegationExpirationTooSoonError`)

Also, it still misses the other check on userAddress/contractAddress thrown by the existing atomic path: `DelegationSelfNotAllowedError` and `DelegationDelegateEqualsContractError`

packages/sdk/src/schemas/offline.ts
```diff
@@ -0,0 +1,365 @@
...
+]);
+
+/**
+ * EIP-1559 fees. `maxFeePerGas` and `maxPriorityFeePerGas` live in one
+ * object so they can only be supplied together — pinning a cap (the total fee)
+ * while the tip is estimated can produce a tip above the cap and fail
+ * serialization. Both legs are validated as `bigint` so a JS caller (no
+ * compiler) can't slip a `number` into the frozen, signed payload.
+ */
+export const fees = z.object({ maxFeePerGas: z.bigint(), maxPriorityFeePerGas: z.bigint() });
```
zmalatrax:
```suggestion
export const fees = z.strictObject({ maxFeePerGas: z.bigint(), maxPriorityFeePerGas: z.bigint() });
```

packages/sdk/src/utils/assertions.ts
```diff
@@ -19,6 +21,12 @@ export function assertString(value: unknown, context: string): asserts value is
   }
 }
 
+export function assertNumber(value: unknown, context: string): asserts value is number {
```
zmalatrax: dead code

packages/sdk/etc/sdk.api.md
```diff
@@ -342,6 +343,11 @@ export function approveContract(tokenAddress: Address, spender: Address, value:
     readonly args: readonly [`0x${string}`, bigint];
 };
 
+// Warning: (ae-forgotten-export) The symbol "approveUnderlyingRequest" needs to be exported by the entry point index.d.ts
```
zmalatrax: The zod migration made the offline types opaque in the API reports: 13 new `ae-forgotten-export` warnings here, and every request/option type now renders as `z.infer<typeof …>`, so the report shows no fields for the whole offline surface. `pnpm api-report` can no longer catch a breaking change to it (add/remove/retype a field: zero report diff).

We already have the convention for this in the repo: `credentials/__tests__/schemas.test-d.ts` keeps hand-written interfaces as the public types and pins them to the internal zod schemas with vitest type tests, precisely so the schemas can stay `@internal` and out of the shipped `.d.ts`. Please apply the same pattern here:

1. Restore hand-written interfaces in `types/offline-signing.ts` as the public types (the pre-`32015117` ones, adapted to the current surface: `PrepareOptions` with the grouped `fees` object, etc.). Keep the schemas in `schemas/offline.ts` unexported/`@internal`.
2. Pin each schema to its interface in a `types/__tests__/offline-signing.test-d.ts`, mirroring the credentials file:
   ```ts
   expectTypeOf<z.input<typeof setOperatorRequest>>().toEqualTypeOf<SetOperatorRequest>();
   ```
   Pin against `z.input` (not `z.infer`): `delegateDecryptionRequest` transforms `Date → bigint` on parse, so input and output diverge, and the public type must describe what callers pass. The vitest `typecheck` project already picks up `*.test-d.ts`, and since `tsconfig` includes them, plain `pnpm typecheck` enforces the pins too. Drift in either direction is a compile error (verified: retyping `until` in the schema fails both vitest typecheck and `tsc` with a readable per-field message).
3. Give the grouped fee pair a named `PrepareFees` interface (an inline object type would itself become a forgotten export), and re-export the 11 request interfaces + `PreparedTransaction` from `query/index.ts` — the union becoming a real type otherwise re-creates the warnings on the `/query` report.
4. Regenerate the reports. Expected result (verified locally on this branch): `sdk.api.md` 13 → 0 warnings, `sdk-query.api.md` 3 → 0 (this also clears that report's pre-existing ones), field-level shapes back for all offline types, `pnpm api-report:check` green, full suite green.

Please don't resolve the warnings by exporting the schema constants instead: that leaks zod values into the public API and the report still shows no fields.

Refs: input/output divergence under transforms — https://zod.dev/basics#inferring-types ; vitest type testing — https://vitest.dev/guide/testing-types

packages/sdk/src/schemas/offline.ts
ghermet: Moved into the builder in 0f2cbcf80. `#buildDelegateDecryption` now mirrors the atomic delegation-service path and throws the domain-specific errors: `DelegationExpirationTooSoonError` (expiry <1h out), `DelegationSelfNotAllowedError` (delegate === from), and `DelegationDelegateEqualsContractError` (delegate === contract). The zod schema keeps only shape validation. Added rejects-self and rejects-delegate==contract tests.

packages/sdk/src/utils/assertions.ts
ghermet: Removed in 35f7a6171 — `assertNumber` had no remaining callers once `until` validation moved to the zod schema.

packages/sdk/src/schemas/offline.ts
ghermet: Confirmed — `until` is `unixSeconds` (`z.int().check(z.nonnegative())`), no minimum-duration floor, so a positive `until` already in the past revokes. Also dropped the stale offline-service test that still asserted the old ≥1h floor (0f2cbcf80) and added one asserting a past `until` prepares successfully.

packages/sdk/src/schemas/offline.ts
ghermet: Already `checksummedAddress` for every address field in `schemas/offline.ts` (from/token/to/owner/operator/underlying/spender/wrapper/aclAddress/contractAddress/delegateAddress). No `evmAddress` remains on the offline surface.

packages/sdk/src/schemas/offline.ts
ghermet: Applied — `fees` is now `z.strictObject({ maxFeePerGas: z.bigint(), maxPriorityFeePerGas: z.bigint() })`.

packages/sdk/etc/sdk.api.md
ghermet: Done (d2737825c), following the credentials `schemas.test-d.ts` pattern:

1. Restored hand-written interfaces in `types/offline-signing.ts` as the public types (`PrepareOptions` with the grouped `fees` object, the 11 requests, `PreparedTransaction`, etc.). The zod schemas in `schemas/offline.ts` are now `@internal` and stripped from the shipped `.d.ts`.
2. `types/__tests__/offline-signing.test-d.ts` pins each schema to its interface against `z.input` (not `z.infer`) — `delegateDecryptionRequest` transforms `Date → bigint`, so the public type describes the caller-facing input. Enforced by both vitest typecheck and plain `tsc`.
3. `PrepareFees` is a named interface; the 11 request interfaces + `PreparedTransaction` are re-exported from `/query` so the union rendering there stays warning-free.
4. Regenerated reports: `sdk.api.md` 13 → 0 `ae-forgotten-export`, `sdk-query.api.md` 3 → 0, field-level shapes back for the whole offline surface. `api-report:check` green, full suite green (sdk 1223 + react-sdk 239).

Schema constants stay unexported from the public API (no zod values leak). One aside: rolldown-plugin-dts choked on a `<` inside a single-line field JSDoc, so I reworded that one line.

packages/sdk/src/services/__tests__/offline-service.test.ts
```diff
@@ -0,0 +1,784 @@
...
+    await expect(
+      sdk.offline.prepare({
+        kind: "DelegateDecryption",
+        from: userAddress,
+        aclAddress: TOKEN,
+        contractAddress: RECIPIENT,
+        delegateAddress: DELEGATE,
+        expirationDate: new Date(Date.now() + 60_000),
+      }),
+    ).rejects.toThrow(/at least 1 hour in the future/);
```
zmalatrax: test should assert on the error `DelegationExpirationTooSoonError` rather than the error message

ghermet: good catch

packages/sdk/src/schemas/offline.ts
```diff
@@ -0,0 +1,249 @@
...
+ *
+ * The ≥1h-in-the-future expiry check and the self/contract guards live in the
+ * `#buildDelegateDecryption` builder (they throw domain-specific delegation
+ * errors this schema can't).
+ * @internal
+ */
+export const delegateDecryptionRequest = z.object({
+  kind: z.literal("DelegateDecryption"),
+  from: checksummedAddress,
+  aclAddress: checksummedAddress,
```
zmalatrax: Thinking through it, I don't see the point of letting the user supply the `aclAddress` on `delegateDecryptionRequest` and `revokeDelegationRequest`
We can drop the field and recover it from `router.chain.aclContractAddress` in `#buildDelegateDecryption` as done in the atomic path (the already released one)

zmalatrax (dismissed):
lgtm, ty!
I haven't checked that all the doc files were updated (besides the modified files in the PR)

zmalatrax (approved):

PR #347, opened 2026-05-12
chore: update rolldown to 1.0.0

ghermet:
## Summary
- Bump `rolldown` from `1.0.0-rc.7` / `1.0.0-rc.17` to stable `1.0.0` and `rolldown-plugin-dts` from `^0.24.1` to `^0.25.0` across the root, `@zama-fhe/sdk`, and `@zama-fhe/react-sdk` package.json files — aligning all three workspaces on the same stable release.
- Pass `--no-error-on-unmatched-pattern` to the lint-staged `oxfmt` invocation so commits that touch only `pnpm-lock.yaml` plus `package.json` (which `.oxfmtrc.json` ignores) no longer fail the pre-commit hook.

## Test plan
- [x] `pnpm install` succeeds, lockfile regenerated
- [x] `pnpm -F @zama-fhe/sdk build` succeeds on rolldown 1.0.0
- [x] `pnpm -F @zama-fhe/react-sdk build` succeeds on rolldown 1.0.0
- [x] `pnpm typecheck` passes (pre-commit hook)
- [ ] CI green


package.json
```diff
@@ -105,8 +105,7 @@
       "oxlint --fix",
       "ast-grep scan --config sgconfig.yml"
     ],
-    "*.{ts,tsx,js,mjs,cjs,yaml,yml}": "oxfmt",
-    "!(package-lock).json": "oxfmt"
+    "*.{ts,tsx,js,mjs,cjs,json,yaml,yml}": "oxfmt --no-error-on-unmatched-pattern"
```
enitrat, line 108: what's that for? Im not sure this is desired? sounds like a hack from opus?

ghermet: `--no-error-on-unmatched-pattern` is to prevent ofmt from churning on files like `pnpm-lock.yaml` or `package-lock.json`

enitrat: why don't we want to autocheck them?

ghermet: they are auto formatted on install and they are not handled by oxfmt. this is why we had `"!(package-lock).json": "oxfmt"` to omit the package lock from json files. the alternative is to do the same for yaml with `   "!(pnpm-lock).yaml": "oxfmt"`

enitrat (approved):

PR #353, opened 2026-05-13
refactor: factorize tx submission

enitrat:
## Context

Every SDK call that submits an on-chain write followed the same pattern: `requireSigner` → `writeContract` → emit `<Op>Submitted` → wait for receipt, with a `try/catch` that emits `TransactionError` and rethrows as a typed `ZamaError`. That block was duplicated across `Token`, `WrappedToken`, and `DelegationService`, and the operation-string → submitted-event-type mapping lived inline at each call site.

## Motivation

Centralize this pipeline so:
- The submitted/error event contract is enforced in one place (impossible to forget the `TransactionError` emit, or get the event payload shape wrong).
- The set of valid `TransactionOperation` strings is derived from the dispatch table — no hand-maintained union that can drift from the actual call sites.
- Adding a new write operation is a one-row change in a single table, not five edits across three files.

## High-level overview

- **New `utils/submit-transaction.ts` helper** owns the entire submit pipeline: write → emit submitted → wait for receipt, or emit `TransactionError` and rethrow. Precedence on failure is documented in its JSDoc: `ZamaError` rethrown as-is → `mapError` → `errorClass ?? TransactionRevertedError` with a single generic `"Transaction failed during ${operation}"` message.
- **Single source of truth for operation → submitted-event payloads.** `transactionOperationMetadata` (in `events/sdk-events.ts`) maps every `TransactionOperation` to the exact event shape it produces, guarded by `satisfies` for exhaustiveness. `TransactionOperation` itself is derived from this table via `keyof`, replacing the old hand-maintained `TransactionErrorOperation` union.
- **Token / WrappedToken / DelegationService** now route every write through the helper. Token exposes a thin protected `submitTransaction` wrapper so subclasses (e.g. `WrappedToken`) can call it without re-importing the free function.
- **Observable behavior changes** (worth noting since events are public API; SDK is alpha so all acceptable):
  - `DelegationService` now emits `TransactionError` on delegate/revoke failures (was previously silent — the `try/catch` only rethrew). Brings delegation in line with every other write path.
  - `unwrapAll` error events now carry `operation: "unwrapAll"` instead of the previous `"unwrap"` (the success path was already distinct; this aligns the error path).
  - New `operation` discriminator `"approveUnderlying:reset"` is publicly visible on both `TransactionErrorEvent.operation` and as `step: "reset"` on `ApproveUnderlyingSubmittedEvent`. The pre-flight allowance-reset transaction was previously silent; it now emits like any other approval, with `step: "reset" | "approve"` distinguishing the two.
  - `TransactionErrorOperation` is renamed to `TransactionOperation` and broadened with `"approveUnderlying:reset"` and `"unwrapAll"`. Exhaustive-switch consumers must add those cases.
- **Tests** updated to match: error-message-coupled assertions relaxed to `code` matches (the per-operation strings are gone), and new coverage added for the `step: "reset" | "approve"` discriminator on `ApproveUnderlyingSubmitted` events.

### DIFF
--- a/packages/sdk/src/token/wrapped-token.ts
+++ b/packages/sdk/src/token/wrapped-token.ts
@@ -15,7 +15,6 @@ import { findUnwrapRequested } from "../events/onchain-events";
 import { ZamaSDKEvents } from "../events/sdk-events";
 import type { Handle } from "../relayer/relayer-sdk.types";
 import {
-  ApprovalFailedError,
   DecryptionFailedError,
   ERC20ReadFailedError,
   EncryptionFailedError,
@@ -121,8 +120,7 @@ export class WrappedToken extends Token {
    * @returns The transaction hash and mined receipt.
    * @throws {@link ChainMismatchError} if signer and provider are on different chains.
    * @throws {@link InsufficientERC20BalanceError} if the ERC-20 balance is less than `amount`.
-   * @throws {@link ApprovalFailedError} if the ERC-20 approval step fails (approveAndWrap path).
-   * @throws {@link TransactionRevertedError} if the shield transaction reverts.
+   * @throws {@link TransactionRevertedError} if the ERC-20 approval or shield transaction reverts.
    *
    * @example
    * ```ts
@@ -235,7 +233,6 @@ export class WrappedToken extends Token {
         await this.submitTransaction({
           operation: "approveUnderlying:reset",
           config: approveContract(underlying, this.address, 0n),
-          errorClass: ApprovalFailedError,
         });
       }
     }
@@ -243,7 +240,6 @@ export class WrappedToken extends Token {
     return this.submitTransaction({
       operation: "approveUnderlying",
       config: approveContract(underlying, this.address, approvalAmount),
-      errorClass: ApprovalFailedError,
     });
   }
 
@@ -519,7 +515,6 @@ export class WrappedToken extends Token {
       await this.submitTransaction({
         operation: "approveUnderlying:reset",
         config: approveContract(underlying, this.address, 0n),
-        errorClass: ApprovalFailedError,
       });
     }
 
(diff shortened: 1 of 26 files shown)


packages/sdk/src/token/wrapped-token.ts
```diff
@@ -268,36 +226,25 @@ export class WrappedToken extends Token {
...
-        error: toError(error),
-      });
-      if (error instanceof ZamaError) {
-        throw error;
+
+      if (currentAllowance > 0n) {
+        await this.submitTransaction({
+          operation: "approveUnderlying:reset",
+          config: approveContract(underlying, this.address, 0n),
+          errorClass: ApprovalFailedError,
```
ghermet: Would we be better off handling errors outside submitTransaction?

packages/sdk/src/token/wrapped-token.ts
```diff
@@ -268,36 +226,25 @@ export class WrappedToken extends Token {
...
+        });
```
ghermet, line 236:
```suggestion
try {
        await this.submitTransaction({
          operation: "approveUnderlying:reset",
          config: approveContract(underlying, this.address, 0n),
        });
      } catch (error) {
       throw new ApprovalFailedError("ERC-20 approval failed", { cause: error })
      }
```

ghermet: This way you'd get the transaction reverted error piped as the root cause of the failed approval error

packages/sdk/src/utils/submit-transaction.ts
```diff
@@ -0,0 +1,71 @@
...
+      error: toError(error),
+    });
+    if (error instanceof ZamaError) {
+      throw error;
+    }
+    const mapped = mapError?.(error);
+    if (mapped) {
+      throw mapped;
+    }
+    const ErrorCtor = errorClass ?? TransactionRevertedError;
```
ghermet: This one should probably be a `TransactionRevertedError` at all time, you can compose it with `cause` in situ as shown above with the approval

ghermet (commented):
good refactor 💯 , see error handling suggestion for composing with `cause`

enitrat:
removed error mapping in the util, removed approval-specific error instead, for specific errors like delegation, built them with cause like suggested

ghermet (approved):

PR #354, opened 2026-05-13
refactor(sdk): carve ZamaSDK into domain namespaces [SDK-169]

ghermet:
**Linear:** [SDK-169](https://linear.app/zama/issue/SDK-169/restructure-zamasdk-into-domain-sub-clients)

## Summary

- Carve the flat `ZamaSDK` surface into three domain **namespaces** exposed as readonly properties — `sdk.permits`, `sdk.delegations`, `sdk.decryption`. `sdk.encrypt(...)`, `sdk.registry`, lifecycle, and Token factories (`createToken` / `createWrappedToken`) stay at the root.
- Each namespace is a thin public class (`Permits`, `Delegations`, `Decryption`) that owns SDK-level guards (chain alignment, signer requirement, empty-array short-circuit, decrypt-cache invalidation) and delegates the actual work to the existing internal `*Service` classes. Guards use a shared `assertNonNullable` + `SignerNotConfiguredError({ cause })` pattern.
- Public verbs align with the domain language: `permits` grant/revoke/has, `delegations` delegateDecryption/revokeDelegation/isActive/getExpiry, `decryption` userDecrypt/delegatedDecrypt/publicDecrypt. `Permit` (and `DelegationPermit`) is the noun.
- React hook signatures unchanged in shape; names and folders re-aligned with the new SDK surface (`useGrantPermit`, `useHasPermit`, `useDelegatedDecrypt`, `permits/`, `delegations/`, `decrypt/`).
- Rename `tokenAddress` → `contractAddress` in `DelegationStatusQueryConfig` and `UseDelegationStatusConfig` for consistency with the `Delegations` namespace (which uses `contractAddress` throughout — the ACL delegates on contracts, not just tokens).
- Split delegation docs: new how-to guide at `guides/delegated-decryption.md`, rewrote `reference/sdk/delegation.md` as a pure `Delegations` namespace API reference.
- Added full JSDoc to `sdk.decryption.delegatedBatchDecrypt`.

## Rename map (before → after)

| Before | After |
|---|---|
| `sdk.allow(c)` | `sdk.permits.grantPermit(c)` |
| `sdk.allowAs(d, c)` | `sdk.permits.grantDelegationPermit(d, c)` |
| `sdk.isAllowed(c)` | `sdk.permits.hasPermit(c)` |
| `sdk.isAllowedAs(d, c)` | `sdk.permits.hasDelegationPermit(d, c)` |
| `sdk.revokePermits(c?)` | `sdk.permits.revokePermits(c?)` |
| `sdk.clearCredentials()` | `sdk.permits.clear()` |
| `sdk.delegateDecryption({...})` | `sdk.delegations.delegateDecryption({...})` |
| `sdk.revokeDelegation({...})` | `sdk.delegations.revokeDelegation({...})` |
| `sdk.isDelegated({...})` | `sdk.delegations.isActive({...})` *(semantic shift — see below)* |
| `sdk.getDelegationExpiry({...})` | `sdk.delegations.getExpiry({...})` |
| `sdk.userDecrypt(h)` | `sdk.decryption.userDecrypt(h)` |
| `sdk.delegatedUserDecrypt(h, d, a?)` | `sdk.decryption.delegatedDecrypt(h, d, a?)` |
| `sdk.publicDecrypt(h)` | `sdk.decryption.publicDecrypt(h)` |
| `sdk.encrypt(p)` | `sdk.encrypt(p)` *(unchanged)* |
| `sdk.createToken(a)` | `sdk.createToken(a)` *(unchanged — root factory)* |
| `sdk.createWrappedToken(a)` | `sdk.createWrappedToken(a)` *(unchanged — root factory)* |
| `sdk.registry` | `sdk.registry` *(unchanged)* |

**`isDelegated` → `isActive`** is the one row that's more than a structural move — flag this explicitly in the CHANGELOG for `3.0.0-alpha.38`.

### React hook renames

| Before | After |
|---|---|
| `useAllow` | `useGrantPermit` |
| `useIsAllowed` | `useHasPermit` |
| `useRevokePermits` | `useRevokePermits` *(unchanged)* |
| `useClearCredentials` | `useClearCredentials` *(unchanged)* |
| `useDelegatedUserDecrypt` | `useDelegatedDecrypt` |

Folders moved to mirror the SDK: `authorization/` → `permits/`, `delegation/` → `delegations/`, decrypt hooks consolidated under `decrypt/`.

### Query / config renames

| Before | After |
|---|---|
| `DelegationStatusQueryConfig.tokenAddress` | `DelegationStatusQueryConfig.contractAddress` |
| `UseDelegationStatusConfig.tokenAddress` | `UseDelegationStatusConfig.contractAddress` |
| `zamaQueryKeys.delegationStatus.token()` | `zamaQueryKeys.delegationStatus.contract()` |
| Query key shape `{ tokenAddress }` | `{ contractAddress }` |

### Documentation

| Change | Details |
|---|---|
| New guide | `docs/gitbook/src/guides/delegated-decryption.md` — how-to guide for the delegation flow |
| Rewritten reference | `docs/gitbook/src/reference/sdk/delegation.md` — pure `Delegations` namespace API reference |
| Updated reference | `docs/gitbook/src/reference/react/useDelegationStatus.md` — `tokenAddress` → `contractAddress` |
| JSDoc | `sdk.decryption.delegatedBatchDecrypt` — full params, throws, example |

## Design refinements vs. the original spec

- **Namespaces, not "sub-clients".** Per @enitrat's review: `client` in a TS SDK implies something self-instantiable (à la a Viem client). These accessors only exist as properties on a parent `ZamaSDK` instance, so framing them as namespaces is more accurate. Reflected in class names (`Permits` / `Delegations` / `Decryption`, no `Client` suffix), folder name (`packages/sdk/src/namespaces/`), and JSDoc / user-facing docs throughout.
- **Decryption method names are verb-suffixed.** `userDecrypt` / `delegatedDecrypt` / `publicDecrypt` instead of `user` / `delegated` / `public`. The bare nouns collided with the reserved Zama-ecosystem phrases "user decrypt / public decrypt / delegated user decrypt"; the suffixed form keeps the verb attached at every call site (`sdk.decryption.userDecrypt(...)`).
- **Token factories stay at the root.** An initial `sdk.tokens` namespace was prototyped but dropped — it owned no state, no guards, no orchestration. `Token` is a first-class primitive, so the two factories live as root methods (`sdk.createToken`, `sdk.createWrappedToken`).
- **Permit verbs** chosen so the noun reads clearly: `grantPermit` / `revokePermits` / `hasPermit`. `DelegationPermit` is a compound noun (permit-that-references-a-delegation), distinct from `Delegation` itself.
- **`decryption.delegatedDecrypt`** (not `.delegatedUserDecrypt`) — `delegated` is the mode; `user` / `public` are the siblings. The internal `DecryptionService.delegatedUserDecrypt` keeps its name to mirror the relayer protocol verb.
- **`Token.#requireSigner`** is a private fail-fast on `Token` and `WrappedToken` — not a public `ZamaSDK.requireSigner`. Called once at the entry of each mutating method and again inside `submitTransaction` (defensive, cheap).

## What's not in this PR

- **Backwards-compat aliases.** Hard break — alpha is the breakage window.
- **Restructuring `Token` / `WrappedToken`.** Their flat method surface is a deliberate ERC-20-mirroring choice; not a god class.
- **Lazy/getter-based namespace construction.** Eager construction in the constructor; service instantiation is cheap.
- **`examples/example-hoodi` and `examples/react-ethers` ShieldCard source migration.** These two `.tsx` files pin published SDK versions that don't expose the new surface yet. Their README/WALKTHROUGH snippets *are* migrated forward-compatibly in this PR; the source carries a `TODO(SDK-170)` comment and will be migrated once `3.0.0-alpha.38` is published.

## Test plan

- [x] `pnpm typecheck` — clean across 6 packages
- [x] `pnpm test:run` — 1567 passing, 5 skipped, 0 failed
- [x] `pnpm lint` — 0 warnings, 0 errors
- [x] Playwright node e2e suite migrated to the namespace API (commit `cbc7079e`)
- [ ] Manual smoke test against an example app after rebase onto whichever commit lands `3.0.0-alpha.38`
- [ ] Full Playwright e2e (`pnpm e2e:test`) — please confirm in CI

## Reviewer notes

- The rename map is the source of truth for the mechanical part. Skim that first.
- The three namespace classes in `packages/sdk/src/namespaces/` are tiny — each is < 200 LOC. Their `__tests__/` files contain the new behavioural assertions.
- `packages/sdk/src/zama-sdk.ts` shrunk substantially; that's the meat of the surface change.
- React hook diffs are mostly one-line internal reroutes plus the rename moves; nothing to design-review there.
- Two merges from `prerelease` are in the history (`fa456716`, `61041db2`) — they pull in `3.0.0-alpha.38` and the `submitTransaction` factorization (#353), which is rewired through `Token.#requireSigner`.


packages/sdk/src/token/wrapped-token.ts
```diff
@@ -214,6 +228,7 @@ export class WrappedToken extends Token {
    * ```
    */
   async approveUnderlying(amount?: bigint): Promise<TransactionResult> {
+    this.#requireSigner("approveUnderlying");
```
enitrat, line 231: covered by line below, no?

enitrat:
Overall API looks good to me. However,  I have a few comments to make on the naming system we use: 
- In the global terminology we have inside the zama ecosystem we have user decrypt, public decrypt, delegated user decrypt which are like reserved words: I am afraid that if we just go for `sdk.decryption.user` then it's not very clear what "user" means in this case, so I would like to think of a better naming (i'm ok shipping in this state but I dont want a release to be done with this naming).
- I personally understand a client as something that can be self-instanciated and interacted with on its own. For example, when you're using Viem and you create the Viem client, then you get the Viem client objects on which you can perform operations. But for the system we're putting in place here, you can't really create and interact with an isolated client, so I just want to make sure that the word client doesn't not really surface inside the docs or something we expose to the user, but rather I'd prefer framing it like a namespace.

packages/sdk/src/namespaces/__tests__/permits.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      await sdk.permits.grantDelegationPermit(DELEGATOR, [CONTRACT_A]);
+      // The underlying CredentialService signs with the delegator scope — the
+      // signature must include the delegator address as the delegatorAddress dimension.
+      // We assert it was called; deeper invariants live in credential-service tests.
+      expect(signer.signTypedData).toHaveBeenCalled();
+    });
+  });
+
+  describe("happy paths", () => {
+    it("allow triggers a wallet signature for uncached contracts", async ({ sdk, signer }) => {
```
enitrat: is that a responsibility in the client?

packages/sdk/src/namespaces/__tests__/permits.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    });
+
+    it("revoke after allow clears the decrypt cache for the signer", async ({
+      sdk,
+      relayer,
+      handle,
+    }) => {
+      const handles = [{ handle, contractAddress: CONTRACT_A }];
+      await sdk.decrypt.user(handles);
+      expect(relayer.userDecrypt).toHaveBeenCalledOnce();
```
enitrat, line 114: relayer calls should not be asserted here. not responsibility of this component

enitrat (commented):
following these moves, ensure unit tests are testing their single unit of concerns and not anything underlying that is responsibility of other unit tests

packages/sdk/src/namespaces/__tests__/permits.test.ts
ghermet: I've refactored the tests to use mock services piped into client creators

packages/sdk/src/namespaces/__tests__/permits.test.ts
ghermet: Yes indeed, now we test the correct service method has been called instead

enitrat (dismissed):

enitrat (dismissed):

enitrat (approved):

PR #368, opened 2026-05-19
ci: upgrade pnpm to v11.1.2

ghermet:
## Summary
- Bumps `packageManager` to `pnpm@11.1.2` and `engines.pnpm` to `>=11`.
- Moves the per-package build-script decision into `pnpm-workspace.yaml` (`allowBuilds`), which pnpm 11 now treats as the canonical location and re-injects on every install until each entry is `true` or `false`:
  - `@ast-grep/cli: true` (unchanged — already approved)
  - `keccak: false`, `sharp: false` (intentionally not built; ship prebuilt binaries and were never built under pnpm 10's ignored-builds warning either)
- Drops the redundant `pnpm.onlyBuiltDependencies` entry from `package.json` since `allowBuilds` supersedes it on pnpm 11.

The workflows already use `pnpm/action-setup@v6.0.3`, which auto-detects pnpm version from `packageManager` — no workflow changes required.

`lockfileVersion` is unchanged (9.0); the lockfile diff is purely pnpm 11 format compaction.

## Why
pnpm 11 turns the previously-warning "ignored build scripts" condition into a hard failure (`ERR_PNPM_IGNORED_BUILDS`) before running commands like `pnpm typecheck`, so the per-package decision has to be recorded explicitly.

## Test plan
- [x] `pnpm install` (CI=true) — clean install on pnpm 11.1.2
- [x] `pnpm typecheck` across all 6 packages
- [x] `pnpm lint` (oxlint + ast-grep)
- [x] `pnpm test:run` — 1609 passed, 5 skipped
- [x] `pnpm --filter @zama-fhe/sdk --filter @zama-fhe/react-sdk build`
- [ ] CI green on the PR (workflows pick up the bumped `packageManager` automatically via `pnpm/action-setup@v6.0.3`)


package.json
```diff
@@ -109,9 +109,6 @@
     "*.{ts,tsx,js,mjs,cjs,json,yaml,yml}": "oxfmt --no-error-on-unmatched-pattern"
   },
   "pnpm": {
```
enitrat: [AUTOMATED] 🔴 **Critical: `pnpm.overrides` silently lost**

[pnpm 11 no longer reads the `pnpm` field from `package.json`](https://pnpm.io/migration). The PR correctly removes `onlyBuiltDependencies` but leaves the `overrides` block behind — it's now dead config.

The lockfile diff confirms it: the top-level `overrides:` section was **dropped entirely** from `pnpm-lock.yaml`. These security/compat pins are no longer applied:
- `minimatch: 10.2.4`
- `next: >=16.2.5`
- `@walletconnect/ethereum-provider: ^2.23.7`
- `lodash: >=4.18.0`
- `lodash-es: >=4.18.0`

**Fix:** Move overrides to `pnpm-workspace.yaml` and delete the entire `"pnpm"` field from `package.json`:

```yaml
# pnpm-workspace.yaml
overrides:
  minimatch: "10.2.4"
  next: ">=16.2.5"
  "@walletconnect/ethereum-provider": "^2.23.7"
  lodash: ">=4.18.0"
  lodash-es: ">=4.18.0"
```

Ref: [pnpm v10→v11 migration — Configuration Relocation](https://pnpm.io/migration), [pnpm 11.0 release notes — Removed Config Sources](https://pnpm.io/blog/releases/11.0)

pnpm-workspace.yaml
```diff
@@ -3,6 +3,11 @@ packages:
   - "tools/*"
   - "test/*"
 
+allowBuilds:
```
enitrat, line 6: [AUTOMATED] 🟡 **`allowBuilds` is correct — but overrides need to land here too**

The `allowBuilds` migration is well done. However, the `overrides` that lived in `package.json#pnpm.overrides` must also be moved here — pnpm 11 reads all config from `pnpm-workspace.yaml`, not from the `pnpm` field in `package.json`.

Add an `overrides:` key to this file (see comment on `package.json` for the full block).

pnpm-workspace.yaml
```diff
@@ -3,6 +3,11 @@ packages:
...
+  "@ast-grep/cli": true
+  keccak: false
```
enitrat, line 8: [AUTOMATED] ℹ️ **Nit: new pnpm 11 features available**

Non-blocking — a few pnpm 11 defaults that benefit this repo for free:
- **`verifyDepsBeforeRun: install`** (default) — auto-installs deps before scripts if out of date.
- **`blockExoticSubdeps: true`** (default) — additional supply-chain protection.
- **[`pnpm ci`](https://pnpm.io/blog/releases/11.0)** — new clean-install command. CI workflows could optionally migrate from `pnpm install --frozen-lockfile` to `pnpm ci` for brevity, but not required.

enitrat (dismissed):
[AUTOMATED]

## pnpm 11 Migration Review

### What the PR does right

- **`allowBuilds` migration** — correctly replaces `pnpm.onlyBuiltDependencies` with `allowBuilds` map in `pnpm-workspace.yaml`.
- **`packageManager` and `engines.pnpm`** bumped correctly.
- **No CI workflow changes needed** — `pnpm/action-setup@v6.0.3` auto-detects version from `packageManager`. Confirmed across all 9 workflow files.
- **Lockfile** — version stays 9.0; the diff is purely pnpm 11 format compaction.
- **`minimumReleaseAge`** already in `pnpm-workspace.yaml` (4320), unaffected by new default (1440).

### Verified — NOT issues

- **`NPM_CONFIG_PROVENANCE: true`** in `release.yml` — consumed by `npm publish`, not pnpm. Unaffected by pnpm 11's drop of `npm_config_*` for pnpm commands.
- **No `.npmrc`** to migrate.
- **No legacy settings** (`allowNonAppliedPatches`, `ignorePatchFailures`, `auditConfig.ignoreCves`, etc.).
- **Script name shadowing** (`clean`, `setup`, `deploy`, `rebuild` now override pnpm builtins in v11) — repo only has `setup:claude` (colon-namespaced, safe).
- **`pnpm add -g npm@^11`** in release workflows — pnpm 11 isolates global installs, but `pnpm bin -g` should return the correct path. The existing `npm --version` assertion guards against breakage.

Sources: [pnpm v10→v11 migration](https://pnpm.io/migration) · [pnpm 11.0 release notes](https://pnpm.io/blog/releases/11.0)

.github/workflows/api-report.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: curl -fsSL https://github.com/AikidoSec/safe-chain/releases/latest/download/install-safe-chain.sh | sh -s -- --ci
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile --ignore-scripts
```
ankurdotb, line 48: I would keep it as frozen-lockfile for better understandability, I think that's the pnpm native version of this. `ci` is an alias to map back to what `npm ci` does which I think is functionally same as frozen lockfile. No need to change this.

ankurdotb (changes requested):
Agree with Mathieu's review that exotic deps and other pnpm specific security features might be good to include here.

.github/workflows/api-report.yml
enitrat: "The command runs pnpm clean followed by pnpm install --frozen-lockfile. Designed for CI/CD environments where reproducible builds are critical" it'd rather keep it

ghermet: @ankurdotb See the [documentation](https://pnpm.io/continuous-integration) for v11,

> When pnpm detects that it is running in CI, it switches to frozen-lockfile mode automatically

it probably checks `process.env.CI===true` before appending `--frozen-lockfile`

ghermet: That being said, there seems to be a new interesting behavior linked to `pnpm install`:

> Since v11, pnpm also fails on incompatible lockfiles in CI — if the lockfile was written by a newer pnpm major version, the install will error out instead of silently rewriting it

ghermet: according to their docs for v11 we could drop both `--frozen-lockfile` and `pnpm ci`

ankurdotb (approved):

PR #370, opened 2026-05-19
chore: dedupe dependencies and fix test-component a11y lint

ghermet:
> Supersedes #369 (auto-closed by branch rename).

## Summary

- Run `pnpm dedupe` to collapse duplicated `viem` (2.48.4/2.49.3), `@tanstack/react-query` (5.100.5/5.100.10), `react` (19.2.5/19.2.6), `zustand` (5.0.0/5.0.3/5.0.13) and other multi-version installs — fixes a viem cross-version structural type mismatch surfacing in `test/test-vite` typecheck.
- Patch/minor upgrades to dev dependencies (`@ast-grep/cli`, `@semantic-release/*`, `@types/node`, `@vitest/*`, `oxlint`, `react`, `react-dom`, …) plus `@zama-fhe/relayer-sdk` to `~0.4.3` and `@tanstack/query-core` to `^5.100.10` in `packages/sdk`. Drops `rolldown` / `rolldown-plugin-dts` from per-package devDependencies.
- Allow `_default` in `no-underscore-dangle` — `oxlint@1.65` started flagging zod/mini's `z._default(...)` API.
- Add `aria-label` to bare `<input>` elements across nine `test-components` forms to satisfy `jsx-a11y/control-has-associated-label` (was 14 errors, now 0).

### Consumer-facing change

- **`@zama-fhe/react-sdk` `viem` peerDependency widened from `^2.47.0` to `>=2`**, aligning with `@zama-fhe/sdk` which already declares `viem: ">=2"`. The react-sdk's direct viem surface is just `getAddress` (stable since 2.0); everything else is wagmi-mediated. This is a deliberate consistency change, not dedupe fallout — callers on older viem 2.x releases are now formally supported.

## Test plan

- [x] `pnpm lint` — 0 warnings, 0 errors
- [x] `pnpm typecheck` — all 6 packages pass
- [ ] `pnpm test:run`
- [ ] `pnpm --filter @zama-fhe/playwright e2e:test` (the touched test-component forms are exercised by the e2e suite)


enitrat (dismissed):

zama-cremaud:
One question -> `viem` peerDependency on `@zama-fhe/react-sdk` widened from `^2.47.0` to `>=2`

`packages/react-sdk/package.json` line ~67:
```diff
-    "viem": "^2.47.0",
+    "viem": ">=2",
```

The PR description ties the dedupe to a viem cross-version structural-type mismatch (2.48.4 vs 2.49.3), which justifies dedupe. But the peerDep *relaxation* is a separate, consumer-facing change: it now claims the react-sdk works on every viem 2.x release back to 2.0.

This may be intentional alignment with `packages/sdk` (which already uses`viem: ">=2"`), but it's worth confirming:
- Does the react-sdk actually work against viem 2.0–2.46? `wagmi-signer.ts` uses `getAddress` (stable), but TS users on older viem may hit type-level breakage — particularly around `EIP712TypedData`, which the file already has a workaround comment about.
- If unsure, `^2.0.0` or `>=2.0.0 <3` is fine; the issue is the implicit contract, not the syntax.

Suggestion: either keep `^2.47.0` (or whatever minimum is actually tested), or document in CHANGELOG that the floor was lowered to `>=2`.

enitrat (approved):

ghermet:
Good catch — it was a deliberate alignment with `@zama-fhe/sdk` (which is already `viem: ">=2"`), but I never called it out, so it read like dedupe fallout. You're right about the implicit contract widening.

On the technical floor: react-sdk's direct viem surface is just `getAddress` (stable since 2.0); everything else goes through `wagmi/actions`. The `EIP712TypedData` workaround comment in `wagmi-signer.ts` is actually about a type from `@zama-fhe/sdk`, not viem — viem only sees the cast input — so it's not a viem-version concern.

I've updated the PR description to document the widening explicitly as a consumer-facing change. CHANGELOG entry will follow from semantic-release on the next prerelease cut.

PR #372, opened 2026-05-20
feat(credentials): widen permits on overlap (SDK-136)

ghermet:
## Summary

Implements [SDK-136](https://linear.app/zama/issue/SDK-136/featcredentials-widen-permits-on-overlap). When `CredentialService.grantPermit` (called `allow()` in the issue) resolves an uncovered set, prefer widening an existing in-scope permit to creating a new one — provided the union stays within the 10-contract cap.

- A user with permit `[A,B]` requesting `[A,B,C]` now gets **one** wallet prompt and ends up with a single permit covering all three, instead of accumulating a separate `[C]` permit.
- When `signed ∪ uncovered > 10`, behaviour falls back to the existing chunked-by-10 path with no regression.
- Wallet rejection during widening throws before any store mutation, so the original permit survives untouched.
- Candidate selection is deterministic: largest overlap with `requested`, ties broken by most-recent `startTimestamp`.

No public API changes. `grantPermit`'s signature and return shape are unchanged.

## Changes

- `packages/sdk/src/credentials/permissions.ts` — pure helpers `sortedUnion` and `findPermitToWiden`.
- `packages/sdk/src/credentials/permission-store.ts` — new `replace(scope, oldSignature, newPermission)` method, atomic via single `storage.set`. Falls back to append-equivalent semantics when the signature isn't found.
- `packages/sdk/src/credentials/credential-service.ts` — widen-or-chunk branch in `grantPermit`. Updated JSDoc documents the new algorithm.

## Test plan

- [x] Unit tests for `sortedUnion`, `findPermitToWiden` (empty/no-fit/single-feasible/largest-overlap/tie-break by recency).
- [x] Unit tests for `PermissionStore.replace` (swap-by-signature, unknown-signature → append).
- [x] Integration tests for `grantPermit` widening: happy path, union-exceeds-cap fallback, wallet rejection rollback, fresh `startTimestamp`.
- [x] Full credentials suite passes (44/44).
- [x] No diff against `packages/sdk/src/credentials/index.ts` (public API unchanged).


packages/sdk/src/credentials/__tests__/credential-service.test.ts
```diff
@@ -208,3 +217,100 @@ describe("CredentialService delegator-scope isolation", () => {
...
+    const raw = await storage.get<Permission[] | null>(DIRECT_SCOPE_KEY);
+
+    expect(raw).not.toBeNull();
+    expect(raw).toHaveLength(1);
+    expect(raw![0]!.signedContractAddresses).toEqual([
+      checksum(ADDRS[0]!),
+      checksum(ADDRS[1]!),
+      checksum(ADDRS[2]!),
+    ]);
+  });
```
enitrat: i would not inspect raw storage here.
"falls back to chunking when the union exceeds the cap" I don't think this cares about any internal layout of the signed contracts

--> can be applied to multiple tests here, `DIRECT_SCOPE_KEY` can be deleted as well

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -64,10 +64,16 @@ export class CredentialService {
...
+   * occurs. Otherwise the service prefers widening: when an existing in-scope
+   * permit's `signedContractAddresses ∪ uncovered` still fits the 10-contract
+   * cap, that permit is re-signed for the union (one wallet prompt) and
+   * atomically replaced in the store. Candidate selection is deterministic —
+   * largest overlap with `contracts`, ties broken by most-recent
+   * `startTimestamp`. When no candidate fits, the uncovered subset is chunked
+   * into groups of ≤10 and one permit per chunk is signed sequentially. Each
+   * signed permit (widened or chunked) is persisted before the next prompt;
+   * newly signed permits are returned in-memory even when best-effort
+   * persistence fails.
```
enitrat: can you compact to max like 2 lines, no one will ever read that otherwise

enitrat (dismissed):
lgtm, can improve test a bit and make better jsdoc

packages/sdk/src/credentials/__tests__/credential-service.test.ts
ghermet: Refactored — the widening tests now assert through public observables only (signing count, `hasPermit`, `revokePermits` cascade, and time-advancing for the TTL case). Dropped `DIRECT_SCOPE_KEY` and the `storage` fixture from those tests. The cascade gives us the widening-vs-chunking distinction without inspecting the stored `Permission` shape.

packages/sdk/src/credentials/credential-service.ts
ghermet: Compacted to 3 lines covering the API contract (prompt minimization + empty-list warm). Widening/chunking specifics live in the implementation and `permissions.ts` unit tests.

enitrat (approved):

PR #378, opened 2026-05-23
fix: SSR warmup + wrong-chain keypair warmup

enitrat:
## Summary

Fixes two regressions introduced by eager keypair warming in `@zama-fhe/sdk@3.0.0-alpha.41`:

1. **SSR crash** — `[zama-sdk] warm keypair failed: ReferenceError: Worker is not defined` during Next.js SSR.
2. **Wrong chain** — warmup dispatched against `chains[0]` (the dispatcher default) instead of the connected wallet's chain, causing spurious mainnet 403s in local/Anvil environments.

See `docs/agents/warm-keypair-regression-investigation.md` for the full investigation.

## What changed

- `CredentialService.handleWalletAccountChange` no longer eagerly warms a keypair. Removing the call removes the SSR `new Worker(...)` path entirely from SDK construction.
- `LifecycleService.#handleWalletAccountChange` now `await`s `switchChain(next.chainId)` **first**, before credential cleanup, cache clear, and listener fan-out. By the time external listeners see the change, the dispatcher is on the wallet chain.
- `ZamaProvider` warms the keypair from a client-only `useEffect`:
  - `typeof window === "undefined"` guard skips SSR.
  - Calls `sdk.permits.warmKeypair(account)` on mount with the current snapshot, and on every `onWalletAccountChange` after lifecycle handling.
  - Errors are swallowed — warmup is a latency optimization, the first real `grantPermit`/`userDecrypt` lazily generates the keypair and surfaces actionable errors.
- Public API addition: `sdk.permits.warmKeypair(account?: WalletAccount)` — optional latency primitive. Reuses the existing `WalletAccount` type.

## What was explicitly **not** added

An earlier draft threaded `chainId` explicitly through `RelayerDispatcher.generateKeypair / createEIP712 / createDelegatedUserDecryptEIP712` and through `KeypairVault.getOrCreate`. That plumbing was dropped after review: with `switchChain` reordered first in lifecycle and `walletAccount.subscribe` replaying synchronously into the same handler, the dispatcher is always on the right chain by the time warmup runs. The simpler API is sufficient.

## Test plan

- [x] `pnpm test:run packages/sdk packages/react-sdk` — 1508 pass, 5 skipped
- [x] `pnpm typecheck` — clean across all workspaces
- [x] New unit tests:
  - `provider.test.tsx` — warms current wallet keypair from client effect; warms next wallet keypair after lifecycle handling
  - `zama-sdk-credentials-lifecycle.test.ts` — constructing the SDK does not warm keypairs
  - `credential-service.test.ts` — does not warm during wallet lifecycle cleanup; `warmKeypair` populates the vault
  - `lifecycle-service.test.ts` — switches relayer before cleanup and listeners
- [ ] Manual SSR verification in a Next.js consumer app (no `Worker is not defined` warnings on render)
- [ ] Manual wallet-connect flow in a multi-chain app with `[mainnet, sepolia, anvil]` + cleartext anvil relayer (no spurious mainnet 403)


packages/sdk/src/namespaces/permits.ts
```diff
@@ -105,6 +105,51 @@ export class Permits {
...
+   * (`SignerNotConfiguredError`, `WalletAccountNotReadyError`,
+   * `ChainMismatchError`) are swallowed silently — they are expected transient
+   * states for a latency primitive. Pass an explicit `address` to skip the
+   * alignment check entirely; this is the form the React provider uses.
+   *
+   * @param address - Signer address to warm. When omitted, the connected,
+   *   provider-aligned wallet account's address is used (silent on
+   *   misalignment).
+   */
+  async warmKeypair(address?: Address): Promise<void> {
```
enitrat: hmm should we just make this not take an address as arg? I can't think of a use case to take an address that would not be the connected's signer address. this removes all the internal complexity +   reduces test surface + in providers.tsx we can just call it without args

zama-cremaud (approved):

PR #383, opened 2026-05-26
fix(sdk-189): warm keypair through wallet chain, not chains[0]

zama-cremaud:
## Summary

Fixes two regressions in the credential warmup that fires when a wallet connects:

1. **Wrong chain.** `CredentialService` warmed the keypair via `RelayerDispatcher.generateKeypair()`, which routes to the dispatcher's *active* chain. The dispatcher initialises its active chain to `chains[0]`, so a wallet connected to (say) Anvil while `chains[0]` is mainnet would warm against the mainnet relayer — surfacing as a 403 in environments where mainnet credentials aren't configured. The lifecycle ordering made this worse: `switchChain` was a fire-and-forget `void swallow(...)` *after* the credential handler, so it couldn't even correct the active chain before warmup.

2. **SSR noise.** During Next.js SSR (which constructs `ZamaSDK` server-side because `"use client"` doesn't prevent SSR), the lazy Web Worker creation surfaced as a noisy `"Worker is not defined"` warning. The warmup should silently no-op in non-browser environments.

Linear: [SDK-189](https://linear.app/zama/issue/SDK-189/bug-warm-keypair-regression-investigation)

## What changed

| Concern | Change |
|---|---|
| Chain-explicit warmup | `RelayerDispatcher.generateKeypair({ chainId? })` peeks at a specific chain's relayer without mutating active state. `RelayerSDK` interface unchanged. |
| Vault threading | `KeypairVault` generator now accepts `chainId?`; `getOrCreate(addr, chainId?)` forwards it. Dedup key stays signer-address only (keypair is signer-scoped, chain-independent). |
| Credential service | `handleWalletAccountChange(prev?, next?)` takes full `WalletAccount` and forwards `next.chainId` to the vault. `WorkerUnavailableError` from warmup is treated as expected and silenced; other errors still warn. |
| Lifecycle ordering | `switchChain` is now `await`ed **first**, before the credential handler. Downstream listeners observe a consistent dispatcher state. |
| SSR-safe Worker | New `WorkerUnavailableError` (public export) thrown by `RelayerWorkerClient.createWorker()` when `typeof Worker === "undefined"`. |

Public API impact: `WorkerUnavailableError` and `ZamaErrorCode.WorkerUnavailable` are new public exports. `RelayerDispatcher.generateKeypair` accepts an optional `{ chainId? }` argument (interface-compatible widening, non-breaking). `etc/sdk.api.md` regenerated.

## Test plan

- [x] `pnpm typecheck` — green
- [x] `pnpm lint` (oxlint + ast-grep) — 0 warnings, 0 errors
- [x] `pnpm test:run` — 1489 passed, 5 skipped, 0 failures
- [x] `pnpm api-report:check:sdk` — green
- [x] Lock-in tests reproduce both regressions:
  - `relayer-dispatcher.test.ts` — explicit `chainId` routes correctly without mutating active chain
  - `credential-service.test.ts` — warmup forwards `next.chainId` instead of dispatcher's active chain; `WorkerUnavailableError` silenced while other errors still warn
  - `lifecycle-service.test.ts` — `switchChain` awaited before credential handler
  - `worker.client.test.ts` — `createWorker()` throws `WorkerUnavailableError` when `Worker` is undefined

## Out-of-scope (intentionally)

- `ZamaProvider`'s `useMemo(() => new ZamaSDK(...))` reconstruction on every `config` reference change — separate concern.
- Hoisting `walletAccount.subscribe` out of the `ZamaSDK` ctor into a `start()` lifecycle method — architectural cleanup beyond this fix.
- User-initiated calls (e.g. `grantPermit`) during SSR — the ticket scope is the eager warmup, not blanket SSR support.

### DIFF
--- a/packages/sdk/etc/sdk-query.api.md
+++ b/packages/sdk/etc/sdk-query.api.md
@@ -1260,9 +1260,9 @@ export const ZamaSDKEvents: {
 
 // Warnings were encountered during analysis:
 //
-// dist/esm/types-CC0Mp18i.d.ts:637:3 - (ae-forgotten-export) The symbol "FheChain" needs to be exported by the entry point index.d.ts
-// dist/esm/types-CC0Mp18i.d.ts:638:3 - (ae-forgotten-export) The symbol "RelayerDispatcher" needs to be exported by the entry point index.d.ts
-// dist/esm/types-CC0Mp18i.d.ts:639:3 - (ae-forgotten-export) The symbol "GenericProvider" needs to be exported by the entry point index.d.ts
+// dist/esm/types-CHE3pJ-P.d.ts:648:3 - (ae-forgotten-export) The symbol "FheChain" needs to be exported by the entry point index.d.ts
+// dist/esm/types-CHE3pJ-P.d.ts:649:3 - (ae-forgotten-export) The symbol "RelayerDispatcher" needs to be exported by the entry point index.d.ts
+// dist/esm/types-CHE3pJ-P.d.ts:650:3 - (ae-forgotten-export) The symbol "GenericProvider" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/packages/sdk/etc/sdk.api.md
+++ b/packages/sdk/etc/sdk.api.md
@@ -13083,8 +13083,9 @@ export class RelayerDispatcher implements RelayerSDK, Disposable {
     delegatedUserDecrypt(params: DelegatedUserDecryptParams): Promise<Readonly<Record<EncryptedValue, ClearValue>>>;
     // (undocumented)
     encrypt(params: EncryptParams): Promise<EncryptResult>;
-    // (undocumented)
-    generateKeypair(): Promise<KeypairType<Hex>>;
+    generateKeypair(options?: {
+        chainId?: number;
+    }): Promise<KeypairType<Hex>>;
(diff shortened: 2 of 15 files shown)


enitrat (commented):
can you give me your thoughts on this?

- I don't like so much introducing `chainId` requirements in the relayer dispatcher and everything downstream, especially when you can freely pass chainIds that are not corresponding to any supported chain
- if you explicitly switch the relayer chain properly, do you even need to pass explicit chainIds everywhere?
- how about removing automatic credentials warmup  and have it only be done as part of the provider's `useEffect`? This way, you're sure it only runs client side

zama-cremaud:
> can you give me your thoughts on this?
> 
> * I don't like so much introducing `chainId` requirements in the relayer dispatcher and everything downstream, especially when you can freely pass chainIds that are not corresponding to any supported chain
> * if you explicitly switch the relayer chain properly, do you even need to pass explicit chainIds everywhere?
> * how about removing automatic credentials warmup  and have it only be done as part of the provider's `useEffect`? This way, you're sure it only runs client side

I see this it what you have done in #383. You are right, your approach is less invasive and less "over engineer" compare to this PR.

enitrat:
fwiw i did not give lots of thoughts into 383 so it might not be optimal either. i just want to make sure we don't fall into other issues like this in the future

PR #391, opened 2026-06-03
fix(react-sdk): accept enabled option and empty contract list in useHasPermit

zama-cremaud:
## What

`useHasPermit` (`@zama-fhe/react-sdk`) now:

- accepts a second `options` argument (TanStack query options), so callers can pass `{ enabled: false }` to mount the hook in an idle state — no work, no signature while disabled;
- tolerates an **empty** `contractAddresses` list as a no-op (query disabled, `data` stays `undefined`) instead of a type/runtime error.

This follows the package convention already used by `useConfidentialIsOperator`, `useConfidentialBalance`, etc. (2nd-arg `options`, `enabled` combined). The SDK-level factory (`hasPermitQueryOptions`) already respected `query.enabled` and disabled on empty lists at runtime — the gap was purely the React hook's API surface and the non-empty tuple type.

## Why

Closes [SDK-190](https://linear.app/zama/issue/SDK-190). Rules of Hooks force `useHasPermit` to be called unconditionally even when there's nothing to check yet (e.g. a chain with no confidential tokens). Consumers had to pass a dummy `[zeroAddress]` probe and discard the result. They can now call the hook with a possibly-empty list directly.

## Design note

Empty list → **disabled / `data: undefined`** (no-op), not "enabled returning `true`". This keeps the hook aligned with the rest of the package (missing input → disabled query) and removes the ugly probe; the residual `length === 0 ? true` decision stays with the consumer where it belongs.

## Changes

- `packages/sdk/src/query/has-permit.ts` — `contractAddresses` type `[Address, ...Address[]]` → `Address[]` (+ JSDoc)
- `packages/react-sdk/src/permits/use-has-permit.ts` — loosen type, add `options?` arg, combine `enabled`, JSDoc
- tests: +1 SDK factory (empty list disabled), +2 React (`enabled: false` with signer; empty list disabled)
- `docs/gitbook/src/reference/react/useHasPermit.md` — document `options`/`enabled` + empty-list no-op
- `llms-full.txt` — regenerated by pre-commit hook from the docs corpus

No changeset file: repo uses semantic-release / conventional commits.

## Verification

- ✅ typecheck (`@zama-fhe/sdk`, `@zama-fhe/react-sdk`)
- ✅ tests: 10 SDK factory, 4 React `useHasPermit`
- ✅ oxlint + oxfmt clean

### DIFF
--- a/docs/gitbook/src/reference/react/useHasPermit.md
+++ b/docs/gitbook/src/reference/react/useHasPermit.md
@@ -85,6 +85,26 @@ const { data: hasPermit } = useHasPermit({
 });
 ```
 
+An empty list is a no-op: the query is disabled and `data` stays `undefined`. This lets you call the hook unconditionally (as the Rules of Hooks require) even when there is nothing to check yet — no dummy probe address needed.
+
+```tsx
+// Safe to call with a list that may be empty.
+const { data: hasPermit } = useHasPermit({ contractAddresses });
+```
+
+## options
+
+`Omit<UseQueryOptions<boolean>, "queryKey" | "queryFn">` — **optional**
+
+Standard React Query options, forwarded to the underlying query. Pass `{ enabled: false }` to mount the hook in an idle state — it performs no work and triggers no signature while disabled.
+
+```tsx
+const { data: hasPermit } = useHasPermit(
+  { contractAddresses: ["0xContractA"] },
+  { enabled: isReady }, // hold the check until the app is ready
+);
+```
+
 {% hint style="warning" %}
 **You must gate decrypt queries yourself.** `useUserDecrypt` does not automatically wait for permits — if you call it before `useGrantPermit`, the user sees an unexpected wallet popup. Use `useHasPermit` to conditionally enable the decrypt query via `{ enabled: !!hasPermit }` as the second argument, or conditionally render the decrypt component only when `hasPermit` is `true`.
 {% endhint %}
--- a/docs/gitbook/src/reference/react/useHasPermit.md
+++ b/docs/gitbook/src/reference/react/useHasPermit.md
@@ -85,25 +85,13 @@ const { data: hasPermit } = useHasPermit({
 });
 ```
 
-An empty list is a no-op: the query is disabled and `data` stays `undefined`. This lets you call the hook unconditionally (as the Rules of Hooks require) even when there is nothing to check yet — no dummy probe address needed.
-
-```tsx
-// Safe to call with a list that may be empty.
-const { data: hasPermit } = useHasPermit({ contractAddresses });
-```
+An empty list is a no-op: the query is disabled and `data` stays `undefined`, so you can call the hook unconditionally even when there is nothing to check yet.
 
 ## options
 
 `Omit<UseQueryOptions<boolean>, "queryKey" | "queryFn">` — **optional**
 
-Standard React Query options, forwarded to the underlying query. Pass `{ enabled: false }` to mount the hook in an idle state — it performs no work and triggers no signature while disabled.
-
-```tsx
-const { data: hasPermit } = useHasPermit(
-  { contractAddresses: ["0xContractA"] },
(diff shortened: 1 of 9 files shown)


docs/gitbook/src/reference/react/useHasPermit.md
```diff
@@ -85,6 +85,26 @@ const { data: hasPermit } = useHasPermit({
 });
 ```
 
+An empty list is a no-op: the query is disabled and `data` stays `undefined`. This lets you call the hook unconditionally (as the Rules of Hooks require) even when there is nothing to check yet — no dummy probe address needed.
+
+```tsx
+// Safe to call with a list that may be empty.
+const { data: hasPermit } = useHasPermit({ contractAddresses });
+```
```
enitrat: consider not making the docs too noisy


```suggestion
```

examples/react-viem/WALKTHROUGH.md
```diff
@@ -337,8 +337,8 @@ function handleDecrypt() {
 `BalancesCard`.
 
 Token-dependent balance and authorization hooks are inside `SelectedTokenPanel`, so
-`useHasPermit` always receives a non-empty contract tuple and `useConfidentialBalance`
-always receives a real token address plus an explicit owner account.
+`useHasPermit` always receives the selected token's address (never an empty list) and
```
enitrat: is that relevant?

packages/react-sdk/src/permits/use-has-permit.ts
```diff
@@ -1,35 +1,54 @@
...
 import { useZamaSDK } from "../provider";
 import { useWalletAccount } from "../utils/wallet-account";
 
 /** Configuration for {@link useHasPermit}. */
 export interface UseHasPermitConfig {
-  /** Contract addresses to check credentials against (at least one required). */
-  contractAddresses: [Address, ...Address[]];
+  /**
+   * Contract addresses to check credentials against. An empty list disables the
+   * query (it is a no-op rather than a type or runtime error).
```
enitrat:
```suggestion
```

query behavior should not be documented at the type level

packages/react-sdk/src/permits/use-has-permit.ts
```diff
@@ -1,35 +1,54 @@
...
- *   every entry in `contractAddresses`. The query auto-disables when no signer is configured
- *   (`data` stays `undefined`, `status` stays `"pending"`).
- * @throws if the query runs without a signer configured (the `enabled` guard normally
- *   prevents this; only reachable if the caller forces `query: { enabled: true }`). {@link SignerNotConfiguredError}
+ *   every entry in `contractAddresses`. The query auto-disables when no signer is
+ *   configured, when `contractAddresses` is empty, or when `options.enabled` is
+ *   `false` (`data` stays `undefined`, `status` stays `"pending"`).
+ * @remarks The internal signer/empty-list guard is combined with `options.enabled`
+ *   via `&&`, so it always wins: passing `enabled: true` cannot re-enable the query
+ *   while no signer is configured or `contractAddresses` is empty.
```
enitrat: new claude models seem to be VERY noisy about comments and documentation for nothing, we can considerably trim them

packages/sdk/src/query/has-permit.ts
```diff
@@ -6,8 +6,11 @@ import type { SignerQueryContext } from "./signer-query-context";
 import { filterQueryOptions } from "./utils";
 
 export interface HasPermitQueryConfig {
-  /** Contract addresses to check credentials against. */
-  contractAddresses: [Address, ...Address[]];
+  /**
+   * Contract addresses to check credentials against. An empty list disables the
+   * query (it is a no-op rather than a type or runtime error).
+   */
```
enitrat: same

enitrat (dismissed):
ok but comments are noisy, we should not let them explode

enitrat (approved):

PR #394, opened 2026-06-04
refactor(sdk): return encrypt handles and proof as contract-ready hex [SDK-200]

zama-cremaud:
## What

`sdk.encrypt` / `useEncrypt` now return `EncryptResult` as
`{ handles: EncryptedValue[]; inputProof: Hex }` (`0x`-prefixed hex) instead of
`{ handles: Uint8Array[]; inputProof: Uint8Array }`.

Closes SDK-200.

## Why

Every caller passing an encrypt result into a viem/wagmi contract write had to
re-encode it by hand — `handles.map(bytesToHex)` + `bytesToHex(inputProof)` —
duplicated across 5+ sites. The hex form is now the contract-ready default, and
it's symmetric with `userDecrypt`, which already consumes `EncryptedValue` (hex).

## Changes

- **Type**: `EncryptResult` is now `{ handles: EncryptedValue[]; inputProof: Hex }`.
- **Backends**: `relayer-web` / `relayer-node` convert with `toHex` right after the
  worker boundary. The worker still exchanges `Uint8Array` (zero-copy transfer
  preserved). `relayer-cleartext` already produced hex internally — dropped its
  `toBytes` round-trip.
- **Contract builders** (`confidentialTransferContract`, `confidentialTransferFromContract`,
  `unwrapContract`) and their viem/ethers write wrappers now take `EncryptedValue`/`Hex`
  and forward without conversion.
- **Docs**: encrypt guide and `useEncrypt` reference updated (no more `bytesToHex`,
  output documented as contract-ready).

## Heads-up for consumers (Super App)

This changes the shape of `EncryptResult`. When bumping to this alpha, remove the
manual `bytesToHex(handles[0])` / `bytesToHex(inputProof)` conversions at encrypt
sites (e.g. `use-create-swap-intent-flow.ts`, `unshield-steps.ts`,
`confidential-transfer-and-call-steps.ts`, `use-confidential-transfer-flow.ts`) —
handles and proof are now `0x`-hex and go straight into `writeContract`.

## Verification

- `tsc --noEmit` (sdk + react-sdk): clean
- vitest: 1382 passed, 5 skipped
- `oxlint`, `oxfmt --check`, `ast-grep scan`: clean


enitrat (dismissed):
nice!

packages/sdk/src/viem/contracts.ts
```diff
@@ -56,13 +56,13 @@ export function writeConfidentialTransferContract(
   client: WalletClient,
   tokenAddress: Address,
   to: Address,
-  handle: Uint8Array,
-  inputProof: Uint8Array,
+  encryptedAmount: EncryptedValue,
```
ankurdotb, line 59: [AUTOMATED] **Nit (pre-existing, widened by this PR):** api-extractor flags `(ae-forgotten-export) The symbol "EncryptedValue" needs to be exported by the entry point` for both `@zama-fhe/sdk/viem` and `@zama-fhe/sdk/ethers`. `writeConfidentialTransferContract` / `writeUnwrapContract` take `EncryptedValue` in their public signatures, but it's only exported from the main `@zama-fhe/sdk` entry — so a TS consumer of the viem/ethers entry can't name the type. It still type-checks via a `0x…` literal (`EncryptedValue = \`0x${string}\``), so non-blocking. Re-exporting `EncryptedValue` (and `Hex`) from `viem/index.ts` + `ethers/index.ts` would clear the warning and improve ergonomics.

Confidence: 85/100

packages/sdk/src/relayer/relayer-sdk.types.ts
```diff
@@ -75,12 +74,15 @@ export interface RelayerWebConfig {
...
 
-/** Result from encryption operation. Alias for {@link InputProofBytesType}. */
-export type EncryptResult = InputProofBytesType;
-
 /** Canonical SDK type for an encrypted value — a `bytes32` ciphertext reference. Alias for {@link Bytes32Hex}. */
 export type EncryptedValue = Bytes32Hex;
 
+/** Result from encryption — contract-ready hex encrypted values and input proof. */
+export type EncryptResult = {
+  encryptedValues: EncryptedValue[];
```
ankurdotb, line 82: [AUTOMATED] **FYI:** `EncryptResult` (`{ encryptedValues: EncryptedValue[]; inputProof: Hex }`, hex) and `EncryptResponseData` (`= InputProofBytesType` → `{ handles: Uint8Array[]; inputProof: Uint8Array }`, bytes) are both `@public` with similar names but intentionally different shapes — the contract-ready public result vs. the zero-copy worker-boundary payload. The `handles` → `encryptedValues` rename now also distinguishes them at the field level. No action needed; flagging so the two aren't accidentally unified later.

ankurdotb (commented):
[AUTOMATED] **Scope note**

Beyond SDK-200's `Uint8Array` → hex change, this PR now also renames the public encrypt API for glossary alignment (`EncryptedValue` / "encrypted value" is the canonical external term; `handle` stays a secondary FHE.sol alias):

- `EncryptResult.handles` → `encryptedValues`
- `confidentialTransfer{,From}Contract` param `handle` → `encryptedAmount` (matches the Solidity ABI arg name; `unwrapContract` already used it)

Both are folded into the existing breaking change so consumers update once. The worker boundary still exchanges `Uint8Array` handles (legacy relayer-sdk shape, zero-copy) — only the public hex result is renamed.

**Super App migration:** in addition to dropping the `bytesToHex(...)` conversions, rename the destructured `handles` → `encryptedValues` at encrypt sites.

Two non-blocking nits inline.

ankurdotb:
[AUTOMATED]

## Follow-up changes pushed to this branch

Refinements layered on the original SDK-200 change (encrypt result → contract-ready hex). All green: typecheck (7 projects), 1493 tests, oxlint, oxfmt, api-report, llms.

**1. Terminology: `handles` → `encryptedValues` (glossary alignment)**
`EncryptResult.handles` → `encryptedValues`; `confidentialTransfer{,From}Contract` param `handle` → `encryptedAmount` (matches the Solidity ABI; `unwrapContract` already used it). Folds the terminology into the existing breaking change so consumers update once. `EncryptedValue`/`Hex` are now also re-exported from the viem/ethers entry points (clears the api-extractor forgotten-export warning).

**2. Realistic, single-source mock values**
The encrypt-result mock now uses a realistic 32-byte `bytes32` value + a proof-shaped blob, sourced once from `VALID_ENCRYPTED_VALUE` / `VALID_INPUT_PROOF` in `test-fixtures/constants.ts` and consumed via fixtures — so the mock and every assertion cannot drift (previously a 3-byte `0x010203` placeholder).

**3. `VALID_HANDLE` → `VALID_ENCRYPTED_VALUE`** (shared test constant), typed `EncryptedValue`.

**Deliberately deferred:** renaming the `handle` test *fixture* → `encryptedValue` repo-wide. It is entangled with local `handle` variables, an ABI parameter name (`bytes32 handle`), and ~20 test descriptions across the decrypt-side suite — that is the separate decrypt-side terminology migration, for its own PR. "handle" remains a valid FHE.sol alias meanwhile.

ankurdotb (approved):

PR #406, opened 2026-06-12
build(api-report): keep committed API reports semantic-only

ankurdotb:
## Summary

Removes the volatile, non-semantic content from the committed API reports — the source of the api-report churn surfaced by the rolldown 1.0.0 → 1.1.0 bump (#401).

### Problem

api-extractor routes `ae-forgotten-export` warnings into the committed `.api.md` files (`addToApiReportFile: true`). Some of those warnings embed **rolldown chunk content-hashes and line numbers**:

```
// dist/esm/index-Cq1_BuG_.d.ts:19573:5 - (ae-forgotten-export) The symbol "DecryptionService" needs to be exported ...
```

Any build-tool bump (new chunk hash) or unrelated source shift (new line numbers) then churns the reports with diffs that look like API changes but aren't.

These warnings are also a false positive for this repo: api-extractor analyses each entry point as if it were a standalone package, but the sub-path entries (`@zama-fhe/sdk/query`, `/viem`, `/ethers`, `/node`, `/web`) are not — they cannot be used without the root `@zama-fhe/sdk`, so the "needs to be exported" symbols are already reachable from the root entry.

### Change

**`api-extractor.base.json`** — `ae-forgotten-export: { logLevel: none, addToApiReportFile: false }`.

The committed reports are now a pure snapshot of the public API surface, with **no public-surface changes** in this PR — every report diff is warning-line removal only; `react-sdk.api.md` is untouched.

api-extractor has no "warn but pass" mode for this rule in check mode (verified on-branch): in-file routing (status quo) carries the volatile coordinates; console-at-`warning` fails `api-report:check` outright. So the rule is silenced (`logLevel: none`) and the leak inventory + the per-symbol export-or-hide audit is tracked in **SDK-209**, whose done-state re-enables it as `logLevel: error`.

### Why not resolve the leaks instead? (evaluated, deferred to SDK-209)

Driving `ae-forgotten-export` to zero via `stripInternal` was attempted and **breaks the build**: 7 cross-module `@internal` cascades, including genuinely public-facing surface (`NodeWorkerPool` referenced by the public `RelayerNodeConfig.pool`, the alignment helpers used by every write path, `buildZamaConfig` used by the viem/ethers config factories). Resolving those is a deliberate API-design audit, not a churn fix.

### Validation

- Zero `ae-forgotten-export` / warning footers across all committed reports
- Report diffs are warning-removal only — **no added/changed public declarations** in any of the 6 reports; `react-sdk.api.md` untouched
- `pnpm api-report:check` exit 0 (direct exit-code capture; all 8 extractor runs clean)
- `pnpm typecheck` across all workspaces
- `test-nextjs` + `test-vite` production builds
- `npm pack` + scratch-consumer typecheck (`--strict --skipLibCheck false`): published types self-contained

### Sequencing

Independent of #402. Whichever lands second needs a `pnpm api-report` regen-sync (the footers stop existing once this merges, so the rolldown hash churn from #401 won't recur).

### DIFF
--- a/packages/sdk/src/query/index.ts
+++ b/packages/sdk/src/query/index.ts
@@ -3,13 +3,6 @@ export type { StrippedQueryOptionKeys } from "./utils";
 export type { QueryFactoryOptions, MutationFactoryOptions } from "./factory-types";
 export { zamaQueryKeys } from "./query-keys";
 
-// Re-export public types referenced by this entry point's query/mutation option
-// signatures so they resolve from `@zama-fhe/sdk/query` directly (already public
-// via the root entry; this avoids api-extractor ae-forgotten-export warnings).
-export type { FheChain } from "../chains/types";
-export type { GenericProvider } from "../types";
-export type { RelayerDispatcher } from "../relayer/relayer-dispatcher";
-
 export {
   invalidateAfterApproveUnderlying,
   invalidateAfterSetOperator,
--- a/packages/sdk/src/query/index.ts
+++ b/packages/sdk/src/query/index.ts
@@ -3,6 +3,13 @@ export type { StrippedQueryOptionKeys } from "./utils";
 export type { QueryFactoryOptions, MutationFactoryOptions } from "./factory-types";
 export { zamaQueryKeys } from "./query-keys";
 
+// Re-export public types referenced by this entry point's query/mutation option
+// signatures so they resolve from `@zama-fhe/sdk/query` directly (already public
+// via the root entry; this avoids api-extractor ae-forgotten-export warnings).
+export type { FheChain } from "../chains/types";
+export type { GenericProvider } from "../types";
+export type { RelayerDispatcher } from "../relayer/relayer-dispatcher";
+
 export {
   invalidateAfterApproveUnderlying,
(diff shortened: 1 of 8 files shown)


packages/sdk/src/query/index.ts
```diff
@@ -3,6 +3,13 @@ export type { StrippedQueryOptionKeys } from "./utils";
 export type { QueryFactoryOptions, MutationFactoryOptions } from "./factory-types";
 export { zamaQueryKeys } from "./query-keys";
 
+// Re-export public types referenced by this entry point's query/mutation option
+// signatures so they resolve from `@zama-fhe/sdk/query` directly (already public
+// via the root entry; this avoids api-extractor ae-forgotten-export warnings).
```
enitrat: it's not really something we want though. api-extractor assumes that this is a standalone package but it's not, you cannot use `sdk/query` without the rest of the SDK, so those types are accessible anyway. I would not make them reachable from here

ankurdotb: Going with your suggestion

ghermet (approved):

PR #409, opened 2026-06-14
feat(sdk): EventService — multi-listener typed event bus

ghermet:
## Summary
- Introduces `EventService`: a multi-listener, typed event bus with `on(type, listener)`, `once(type, listener)`, catch-all `subscribe(listener)`, and `AbortSignal` lifetime binding. Exposed as `sdk.events`.
- Retires the `emitEvent` callback that was threaded into every service constructor (SDK-198). Services now compose the bus directly; `Token` and test fixtures route through `sdk.events.emit`; the deprecated `ZamaSDK.emitEvent` shim is removed.
- `emit` is synchronous and isolates SDK ops from listener throws — re-throws on a fresh microtask so errors surface via `window.onerror` / `uncaughtException` / Sentry instead of being silenced.
- `on` / `once` / `subscribe` return a plain `() => void` unsubscribe function. Lifetime can also be tied to an `AbortSignal` via the `{ signal }` option.
- Listener types are exported as `TypedZamaSDKEventListener<K>` and `ZamaSDKEventListener` (with `TypedListener` / `AnyListener` re-export aliases for back-compat).
- Test fixtures expose `eventService` so tests subscribe to the real bus instead of mocking captured callbacks.

Closes SDK-181, SDK-198.

## Test plan
- [x] \`pnpm typecheck\` — clean
- [x] \`pnpm lint\` — clean
- [x] \`pnpm test:run\` — 1623 passing, 5 skipped
- [ ] Manual: hook an example app's \`sdk.events.on(EncryptStart, ...)\` and verify it fires
- [ ] Manual: throw inside a listener, confirm the SDK call still resolves and the error surfaces to the platform error pipeline


ankurdotb:
Cancelled pending re-doing/structuring this task to be Clear Signing unrelated.

PR #412, opened 2026-06-15
fix(sdk): use x-api-key for the Zama-hosted relayer; keep & document all three auth methods [SDK-210]

ghermet:
## Summary

The Zama-hosted relayer (behind Cloudflare) only accepts the API key in the `x-api-key` header — `Authorization: Bearer` and cookie auth are rejected at the edge. The original approach here was to remove `BearerToken` from the SDK; we've since decided to **keep all three auth methods and disambiguate by deployment** instead, because the relayer-sdk supports all three and self-hosted / proxied setups legitimately use cookie and Bearer.

- **Keeps `RelayerAuth = ApiKeyHeader | ApiKeyCookie | BearerToken`** — no type narrowing; all three stay reachable.
- **`react-turnkey-wallet` proxy example** switched from `Authorization: Bearer …` → `x-api-key` (it targets the hosted relayer, where only the header works).
- **Docs disambiguate by deployment** (authentication, node-js-backend, relayer-api-keys guides + `FheChain.auth` / `RelayerAuth` JSDoc):
  - Zama-hosted relayer → `x-api-key` header (required)
  - Behind your own proxy → `ApiKeyCookie` for the SDK→proxy hop
  - Self-hosted relayer → `BearerToken` if your auth layer expects it
- Corrects the Bearer example to use `token` (the relayer-sdk `BearerToken` field), not `value`.

Existing `ApiKeyHeader` users (every other example) are unaffected.

## Related — error surfacing (fixed upstream in relayer-sdk 0.4.4)

While reviewing we found relayer-sdk **0.4.3** never surfaced relayer/Cloudflare/Kong error bodies: it hardcoded the 401 message and treated 403 as an unknown status. **`@zama-fhe/relayer-sdk@0.4.4` (just released) fixes this** — the 401 handler reads & surfaces the body message, and a new 403 handler surfaces the edge/gateway (Cloudflare/Kong) message, handling both `{ error: { message, label } }` and flat `{ message, label }` bodies. Inheriting it is just a dependency bump (`~0.4.3` → `~0.4.4`); **not done in this PR**. Write-up: https://app.notion.com/p/3815a7358d5e8193847fef0674b084ed

Closes SDK-210.

## Test plan
- [ ] Regenerate `pnpm api-report` (`sdk.api.md` — `RelayerAuth` now includes `BearerToken`) and `pnpm llm:build` (llms corpus) — **pending** (earlier run reflected the now-reverted narrowing)
- [ ] `pnpm typecheck` / `pnpm test` / `pnpm lint`
- [ ] Smoke-test against the live mainnet relayer with a real key


ankurdotb:
Setting the PR to draft, pending confirmation on whether Bearer tokens are actually supported or not.

zama-cremaud:
Two small notes (non-blocking):

- Nit (docs) - docs/gitbook/src/guides/authentication.md & node-js-backend.md: "both of which send the API key under the x-api-key name" is true only for the defaults. ApiKeyHeader.header / ApiKeyCookie.cookie are optional overrides in the relayer-sdk types, so a caller can pick a different name. Fine to keep the simplification, but "by default under x-api-key" would be strictly accurate. Low priority.


- Test plan: the only unchecked item is the live-relayer smoke test (needs deployed creds) — the acceptance criterion from the ticket. Worth landing only after that, or noting it as a follow-up, since the wire behaviour isn't covered by unit tests.

ankurdotb:
@ghermet

**Change request — add tests for the auth scenarios now that relayer-sdk 0.4.4 surfaces error bodies.** Guillaume, could you add these two tiers? I prototyped + verified them locally (deterministic tier passes 16/16; the live file loads and skips cleanly), so this should be mostly lift-and-drop.

### Gotchas I hit (to save you time)
- `wrapDecryptError` keys off a **numeric `statusCode`** on the error. The keyurl/V1 path carries the status only in the message string → you get **`DecryptionFailedError`** with the body in `error.cause.message` (matches your manual run). The V2 path that does carry `statusCode` → `RelayerRequestFailedError`. So **assert on the message chain, not the error class.**
- `createConfig` from `@zama-fhe/sdk/viem` requires `walletClient` (it's not optional). Public decrypt never signs, but you still must pass one — a throwaway account works.
- Lint: avoid `!= null` (eqeqeq) and `String(unknownValue)` (no-base-to-string) in the message-chain helper — the version below is clean.

### Tier 1 — deterministic (offline, runs in normal CI). Append to `packages/sdk/src/errors/__tests__/decrypt.test.ts`:
```ts
describe("relayer auth errors surface the relayer/Cloudflare/Kong message", () => {
  test("403 without a numeric statusCode surfaces the body via cause", () => {
    const relayerError = new Error(
      "HTTP error! status: 403 Unauthorized. Missing or invalid Zama API Key",
    );
    const wrapped = wrapDecryptError(relayerError, "Public decryption failed");
    expect(wrapped).toBeInstanceOf(DecryptionFailedError);
    const cause = (wrapped as { cause?: unknown }).cause as Error;
    expect(cause.message).toMatch(/zama api key/i);
    expect(cause.message).not.toMatch(/unexpected response status/i); // regression guard
  });

  test("403 with a numeric statusCode preserves the code and the body message", () => {
    const relayerError = Object.assign(
      new Error("This server requires a valid Zama API Key in the x-api-key header"),
      { statusCode: 403 },
    );
    const wrapped = wrapDecryptError(relayerError, "Public decryption failed");
    expect(wrapped).toBeInstanceOf(RelayerRequestFailedError);
    expect((wrapped as RelayerRequestFailedError).statusCode).toBe(403);
    expect(wrapped.message).toContain("x-api-key");
  });

  test("401 preserves the code and the body message", () => {
    const relayerError = Object.assign(
      new Error("Unauthorized, missing or invalid Zama API Key."),
      { statusCode: 401 },
    );
    const wrapped = wrapDecryptError(relayerError, "Public decryption failed");
    expect(wrapped).toBeInstanceOf(RelayerRequestFailedError);
    expect((wrapped as RelayerRequestFailedError).statusCode).toBe(401);
    expect(wrapped.message).toMatch(/zama api key/i);
  });
});
```

### Tier 2 — live mainnet (off by default; the "check against the real API" part). New file `packages/sdk/src/relayer/__tests__/relayer-auth.integration.test.ts`:
```ts
import { createPublicClient, createWalletClient, http } from "viem";
import { privateKeyToAccount } from "viem/accounts";
import { mainnet as viemMainnet } from "viem/chains";
import { describe, expect, it } from "vitest";
import type { Auth } from "@zama-fhe/relayer-sdk/bundle";
import { type EncryptedValue, MemoryStorage, ZamaSDK } from "@zama-fhe/sdk";
import { mainnet, type FheChain } from "@zama-fhe/sdk/chains";
import { node } from "@zama-fhe/sdk/node";
import { createConfig } from "@zama-fhe/sdk/viem";

// Live tests against the hosted mainnet relayer. OFF by default so the offline
// `pnpm test:integration` run stays green; also needs the built worker bundle:
//   pnpm build:sdk
//   ENABLE_RELAYER_AUTH_TESTS=true pnpm test:integration
// Negative cases need NO real key; the positive case needs ZAMA_RELAYER_API_KEY.
const RUN = process.env.ENABLE_RELAYER_AUTH_TESTS === "true";

const HANDLE =
  "0x83c5dae9465be5dc53e7582b5cd5aaa35360b19cc7ff00000000000000010500" as EncryptedValue;

// Public decrypt never signs, but the viem config requires a wallet client.
const THROWAWAY_KEY =
  "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" as const;

function makeSdk(auth?: Auth): ZamaSDK {
  const transport = http(process.env.MAINNET_RPC_URL);
  const publicClient = createPublicClient({ chain: viemMainnet, transport });
  const walletClient = createWalletClient({
    account: privateKeyToAccount(THROWAWAY_KEY),
    chain: viemMainnet,
    transport,
  });
  const chain = { ...mainnet, ...(auth ? { auth } : {}) } as const satisfies FheChain;
  return new ZamaSDK(
    createConfig({
      chains: [chain],
      publicClient,
      walletClient,
      storage: new MemoryStorage(),
      relayers: { [chain.id]: node() },
    }),
  );
}

function messageChain(error: unknown): string {
  const parts: string[] = [];
  let current: unknown = error;
  for (let depth = 0; depth < 5 && current !== null && current !== undefined; depth++) {
    if (current instanceof Error) {
      parts.push(current.message);
      current = current.cause;
    } else {
      parts.push(typeof current === "string" ? current : JSON.stringify(current));
      break;
    }
  }
  return parts.join(" | ");
}

describe.skipIf(!RUN)("relayer auth against the hosted mainnet relayer (live)", () => {
  const rejectedCases: Array<readonly [string, Auth | undefined]> = [
    ["ApiKeyCookie", { __type: "ApiKeyCookie", value: "not-a-real-key" }],
    ["BearerToken", { __type: "BearerToken", token: "not-a-real-key" }],
    ["no auth", undefined],
    ["ApiKeyHeader with a bogus key", { __type: "ApiKeyHeader", value: "not-a-real-key" }],
  ];

  it.each(rejectedCases)(
    "rejects %s and surfaces the relayer's error message",
    async (_label, auth) => {
      using sdk = makeSdk(auth);
      let caught: unknown;
      try {
        await sdk.decryption.decryptPublicValues([HANDLE]);
      } catch (error) {
        caught = error;
      }
      expect(caught, "expected the hosted relayer to reject this request").toBeDefined();
      const message = messageChain(caught);
      expect(message).toMatch(/x-api-key|api key|unauthorized|forbidden/i); // 0.4.4 surfaces the body
      expect(message).not.toMatch(/unexpected response status/i);          // not the pre-0.4.4 generic
    },
  );

  it.skipIf(!process.env.ZAMA_RELAYER_API_KEY)(
    "accepts a valid ApiKeyHeader and returns a clear value",
    async () => {
      using sdk = makeSdk({ __type: "ApiKeyHeader", value: process.env.ZAMA_RELAYER_API_KEY! });
      const result = await sdk.decryption.decryptPublicValues([HANDLE]);
      expect(result.clearValues).toHaveProperty(HANDLE);
    },
  );
});
```

This matches your manual 4-scenario run: `ApiKeyHeader`/proxy pass, `ApiKeyCookie`/`BearerToken`/no-auth are rejected (403) with the surfaced message. Note `test:integration` isn't in CI, so the live tier won't gate merges — it's a `build:sdk` + network on-demand check. Happy to pair on it if useful.

zama-cremaud (dismissed):

ankurdotb:
I'm adding a negative scenario integration test, because I'd like to understand whether it broke our side or from the infra side.

zama-cremaud (dismissed):

ankurdotb (approved):

PR #418, opened 2026-06-16
fix(wrappers): sync wrapper ABI with protocol-apps wrapper (unwrapRequestId shape) [SDK-213]

ghermet:
## Summary

Realign the SDK's wrapper ABI with what `zama-ai/protocol-apps@main` actually deploys — the `unwrapRequestId` unshield shape used by every chain we target (Hoodi, Ingen, BNB, testnet, mainnet) via the shared `ConfidentialTokenWrappersRegistry` and `ConfidentialWrapper` (V1 / V2 / V3) proxies. Drops dead legacy entries that no deployed wrapper emits.

- Rename `burntAmountCleartext` → `unwrapAmountCleartext` on `finalizeUnwrap`, matching `ERC7984ERC20WrapperUpgradeable.sol:143`.
- Drop the orphan `UnwrappedStarted` event (decoder, type, exports, tests, reference doc) — absent from `protocol-apps/main`.
- Drop the orphan `UnauthorizedFinalizeUnwrapCaller(euint64 burntAmount, ...)` error — last `burntAmount` site in source; confirmed absent from compiled wrapper ABI.

`FinalizeUnwrapParams.burnAmount` is **kept** as a deliberate legacy alias for resuming unshields persisted by older SDK versions that didn't record `unwrapRequestId` — see `packages/sdk/src/query/finalize-unwrap.ts:11-17`.

## Why every deployed chain is on this shape

Hoodi, Ingen, and BNB cleartext demos don't ship their own wrapper Solidity. They deploy through the same `ConfidentialTokenWrappersRegistry` as testnet/mainnet, which deploys `ConfidentialWrapper` (V1 / V2 / V3) from `protocol-apps`. All three variants inherit `ERC7984ERC20WrapperUpgradeable` and use:

```sol
function finalizeUnwrap(bytes32 unwrapRequestId, uint64 unwrapAmountCleartext, bytes decryptionProof)
emit UnwrapRequested(to, unwrapRequestId, unwrapAmount_)
emit UnwrapFinalized(to, unwrapRequestId, unwrapAmount_, unwrapAmountCleartext)
```

V3 overrides `finalizeUnwrap` for a denylist check but `super`-calls with the same signature. No legacy amount-handle shape, no `burntAmount` field, no `UnwrappedStarted` event exists anywhere in `protocol-apps/main`.

## Out of scope (follow-up candidates)

[...]

PR #421, opened 2026-06-16
refactor(sdk): sync Wrap event ABI with protocol-apps (Tier 1) [SDK-216]

ghermet:
## Summary

- Verified against compiled artifact (`contracts/out/ERC7984ERC20WrapperUpgradeable.sol/ERC7984ERC20WrapperUpgradeable.json`): the deployed wrapper emits `Wrap(address indexed to, uint256 roundedAmount, euint64 encryptedWrappedAmount)`. The legacy `Wrapped(to, amountIn)` event does not exist in `protocol-apps/main` — no back-compat needed.
- `packages/sdk/src/abi/encrypted.abi.ts` — replace `Wrapped(to, amountIn)` with canonical `Wrap(to, roundedAmount, encryptedWrappedAmount)`.
- `packages/sdk/src/abi/wrapper.abi.ts` — add the missing `Wrap` event (the file had no wrap-side event at all), slotted between `UnwrapRequested` and `Upgraded` to match compiled-artifact order.

This is **Tier 1** scope from [SDK-216](https://linear.app/zama/issue/SDK-216) — ABI correctness fix only. Tier 2 (rename `decodeWrapped` / `findWrapped` / `Topics.Wrapped` → `decodeWrap` / `findWrap` / `Topics.Wrap`, expose the new fields) is a public-API change deferred to a follow-up PR.

## Why this hasn't caused user-visible bugs

The SDK does not actually use the ABI for wrap-event decoding — `decodeWrapped` in `events/onchain-events.ts` uses a hardcoded topic string (`Topics.Wrapped = eventTopic("Wrapped(address,uint256)")`), so it's decoupled from the ABI const. Today the decoder matches nothing on chain (deployed wrapper emits `Wrap`, decoder looks for `Wrapped`), and the public-side amount is the value the caller already passed. So the bug was invisible — but it was a bug.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test:run packages/sdk/src/{abi,__tests__,events,contracts}` — 205 tests pass
- [x] `forge build --force` compiles clean (68 files)
- [ ] CI green

## Follow-up

Tier 2 (decoder rename + new field exposure) tracked under [SDK-216](https://linear.app/zama/issue/SDK-216) — to be opened as a separate PR after this lands.

### DIFF
--- a/packages/sdk/etc/sdk.api.md
+++ b/packages/sdk/etc/sdk.api.md
@@ -1171,10 +1171,15 @@ export function confidentialBalanceOfContract(tokenAddress: Address, userAddress
         }, {
             readonly indexed: false;
             readonly internalType: "uint256";
-            readonly name: "amountIn";
+            readonly name: "roundedAmount";
             readonly type: "uint256";
+        }, {
+            readonly indexed: false;
+            readonly internalType: "euint64";
+            readonly name: "encryptedWrappedAmount";
+            readonly type: "bytes32";
         }];
-        readonly name: "Wrapped";
+        readonly name: "Wrap";
         readonly type: "event";
     }, {
         readonly inputs: readonly [];
@@ -2492,10 +2497,15 @@ export function confidentialTotalSupplyContract(tokenAddress: Address): {
         }, {
             readonly indexed: false;
             readonly internalType: "uint256";
-            readonly name: "amountIn";
+            readonly name: "roundedAmount";
             readonly type: "uint256";
+        }, {
+            readonly indexed: false;
+            readonly internalType: "euint64";
+            readonly name: "encryptedWrappedAmount";
+            readonly type: "bytes32";
         }];
-        readonly name: "Wrapped";
+        readonly name: "Wrap";
         readonly type: "event";
     }, {
         readonly inputs: readonly [];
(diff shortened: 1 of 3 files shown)


ankurdotb (commented):
[AUTOMATED] ## Code Review — SDK-216

**Verdict: Approve (Tier 1).**

Tier 1 ABI change is exact-correct vs the deployed wrapper: `Wrap(address indexed to, uint256 roundedAmount, euint64 encryptedWrappedAmount)`; no `Wrapped` event exists in any protocol-apps wrapper (V1/V2/V3 on `main`). `wrapper.abi.ts` previously had no wrap-side event; the added entry's shape matches.

Re open question #1 (is the `Wrapped` entry stale, or was it once emitted?): verified against `protocol-apps@main` — no wrapper variant ever emitted `Wrapped(address,uint256)`; only `Wrap`. So the legacy decoder in `events/onchain-events.ts` is decoding a phantom event, and its "retained for backward compatibility with older deployments" comments are inaccurate (real only for pre-SDK-61 Zaiffer contracts, if any survive).

Recommend greenlighting Tier 2 to add `decodeWrap` / `Topics.Wrap` and **drop** the `Wrapped` decoder rather than keep it — today the SDK decodes zero wrap events on chain. Suggest merging after #418.

ghermet:
Superseded by #438 — the canonical `Wrap(to, roundedAmount, encryptedWrappedAmount)` event is included in the regenerated `encrypted.abi.ts` / `wrapper.abi.ts` from the `ConfidentialWrapperV3` Forge artifact. Tier 2 (decoder rename + new field exposure) remains tracked under SDK-216.

PR #423, opened 2026-06-16
feat(sdk): add confidentialTransferAndCall to Token methods [SDK-168]

ghermet:
## Summary

Closes [SDK-168](https://linear.app/zama/issue/SDK-168/feat-add-transferandcall-to-base-token-methods).

Adds `Token.confidentialTransferAndCall` and `Token.confidentialTransferFromAndCall` on the ERC-7984 confidential-token surface so callers can invoke a recipient's receiver hook (`onConfidentialTransferReceived`) in a single transaction. The caller crafts the opaque `data` payload; the SDK does not encode, validate, or interpret it.

This is the **confidential-token level** (ERC-7984), distinct from the shielding `transferAndCall` work in SDK-144/145 (ERC-1363 on the underlying).

## Changes

- `packages/sdk/src/contracts/encrypted.ts` — `confidentialTransferAndCallContract` and `confidentialTransferFromAndCallContract` pure builders.
- `packages/sdk/src/token/token.ts` — `Token.confidentialTransferAndCall(to, amount, data, options?)` and `Token.confidentialTransferFromAndCall(from, to, amount, data, callbacks?)`, mirroring the non-`AndCall` variants' balance validation, encryption, and callback semantics.
- `packages/sdk/src/events/sdk-events.ts` — new `transferAndCall` / `transferFromAndCall` operations (reuse `TransferSubmitted` / `TransferFromSubmitted` event shapes for clean error attribution).
- `packages/sdk/src/query/transfer-and-call.ts` + `transfer-from-and-call.ts` — TanStack mutation options factories.
- `packages/react-sdk/src/transfer/use-confidential-transfer-and-call.ts` + `use-confidential-transfer-from-and-call.ts` — hooks (with optimistic balance for the sender variant).

## Tests

- Unit tests in `packages/sdk/src/contracts/__tests__/contracts.test.ts` and `packages/sdk/src/token/__tests__/token.test.ts` (10 new tests) — `data` forwarding, balance validation, error mapping.
- Browser Playwright e2e (`test/playwright/tests/transfer-and-call.spec.ts`) — drives `useConfidentialTransferAndCall` against OpenZeppelin's `ERC7984ReceiverMock`, deployed via `contracts/script/Deploy.s.sol`. Asserts both the success path (`abi.encode(uint8(1))` → confidential balance decreases) and a revert path (`abi.encode(uint8(2))` → `InvalidInput` surfaces as a transfer error). Passes on Vite and Next.js.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm test:run` — 1504 passing (10 new unit tests)
- [x] `pnpm e2e:test:vite --grep transferAndCall` — 2/2 pass
- [x] `pnpm e2e:test:nextjs --grep transferAndCall` — 2/2 pass


test/test-components/src/transfer-and-call-form.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const transferAndCall = useConfidentialTransferAndCall({ address: tokenAddress });
+
+  return (
+    <form
+      action={(formData) => {
+        transferAndCall.mutate({
+          to: formData.get("recipient") as Address,
+          amount: BigInt(formData.get("amount") as string),
+          data: formData.get("data") as Hex,
+          skipBalanceCheck: true,
```
ankurdotb, line 28: [AUTOMATED] This demo hardcodes `skipBalanceCheck: true`, but it already decrypts and displays the balance just above (`useConfidentialBalance`) and is a plain EOA flow — so it models the wrong default. For a real deposit (e.g. confidential balance 40, deposit 10) the SDK's default decrypt-and-compare catches an over-deposit with a clean `InsufficientConfidentialBalanceError` before spending gas on a tx the receiver would revert. `skipBalanceCheck` is documented as the smart-wallet escape hatch (token.ts:509). This is the test harness rather than `examples/`, so non-blocking — but if there's no E2E reason for it, drop the flag (let it default to `false`); if there is, add a one-line comment so integrators don't copy it blindly.

Confidence: 80/100

packages/react-sdk/src/transfer/use-confidential-transfer-and-call.ts
```diff
@@ -0,0 +1,113 @@
...
+  const queryClient = useQueryClient();
+
+  // Internal mutation uses `any` for TContext because optimistic mode wraps
+  // the caller's context in OptimisticMutateContext; the public return type
+  // is cast back to the caller's TContext.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return useMutation<TransactionResult, Error, ConfidentialTransferAndCallParams, any>({
+    ...confidentialTransferAndCallMutationOptions(token),
+    ...options,
+    onMutate: config.optimistic
```
ankurdotb, line 78: [AUTOMATED] `useConfidentialTransferAndCall` re-implements the full optimistic `onMutate`/`onError`/`onSuccess` machinery (subtract, snapshot, rollback-on-error, caller-context unwrapping) — the most regression-prone code in the hook — but there's no hook-level test for it. The sibling `useConfidentialTransfer` has a dedicated suite covering optimistic subtract, rollback-on-error, and try/finally resilience. The SDK-level `token.test.ts` can't catch a regression here because this logic lives only in the hook layer. Please add a `use-confidential-transfer-and-call.test.tsx` mirroring that suite.

Confidence: 90/100

packages/sdk/src/token/__tests__/token.test.ts
```diff
@@ -233,6 +233,131 @@ describe("Token", () => {
...
+
+      expect(relayer.encrypt).toHaveBeenCalledWith({
+        values: [{ value: 100n, type: "euint64" }],
+        contractAddress: tokenAddress,
+        userAddress,
+      });
+      expect(signer.writeContract).toHaveBeenCalledWith(
+        expect.objectContaining({
+          functionName: "confidentialTransferAndCall",
+          args: expect.arrayContaining([DATA]),
```
ankurdotb: [AUTOMATED] `args: expect.arrayContaining([DATA])` only checks that `data` appears somewhere in the args — it won't fail if `to` / `encryptedAmount` / `inputProof` / `data` are reordered, or if `to` is wrong, so it doesn't lock the token method's arg ordering at this layer. (Same at line 335 for the `…FromAndCall` variant.) Exact ordering is pinned in `contracts.test.ts`, but tightening these to the full ordered args (`toEqual([...])`) would let the token-method test catch a wrong-order regression on its own.

Confidence: 85/100

ankurdotb (changes requested):
[AUTOMATED]
## Code Review — confidentialTransferAndCall / confidentialTransferFromAndCall

Strong, well-designed addition. A multi-pass sweep (including validating against the live vault-deposit flow this enables) found **no runtime bugs and no convention violations** — contract-builder arg order matches the ABI overloads, encryption `userAddress` is per-variant correct, the `…FromAndCall` balance-check omission is intentional (mirrors `confidentialTransferFrom`), and the opaque-`data` / no-helper-utils decision is right (the payload is the receiver contract's ABI, not the token's).

Requesting changes on the docs deliverable before this closes SDK-168, plus minor inline items:

1. **Docs (blocking for SDK-168).** The ticket scopes *"promote the pattern in the docs once available."* This PR touches no GitBook docs. Please add a "transfer into a contract (receiver hook)" section to `docs/gitbook/src/guides/transfer-privately.md`, modeled on the real use case (depositing into a confidential vault): `token.confidentialTransferAndCall(vault, amount, data)` + the `useConfidentialTransferAndCall` hook, how to build `data` with `encodeAbiParameters`, the balance-check behavior (and how it contrasts with the `skipBalanceCheck` smart-wallet escape hatch), and a note that with a queued/batched vault a confirmed transaction does not mean the deposit has settled.
2. See inline comments: the demo's `skipBalanceCheck` default, and two test-coverage items.

Otherwise this is merge-ready — nice work.

ghermet:
Addressed the review and merged `prerelease` to clear the conflict. Summary:

**1. Docs (blocking).** Added a **"Transfer into a contract (receiver hook)"** section to `docs/gitbook/src/guides/transfer-privately.md`, modeled on a confidential-vault deposit: `token.confidentialTransferAndCall(vault, amount, data)` + the `useConfidentialTransferAndCall` hook, building `data` with `encodeAbiParameters`, the default decrypt-and-compare balance check vs the `skipBalanceCheck` smart-wallet escape hatch, the `confidentialTransferFromAndCall` operator variant, and a warning that a confirmed tx ≠ a settled deposit for queued/batched vaults.

**2. Demo `skipBalanceCheck`.** Kept the flag (the sibling `transfer-form` / `unshield-form` harness forms set it too — it's a deliberate harness convention so the E2E flow drives only the on-chain tx) but added a comment explaining the rationale and pointing integrators at the `false` default.

**3. Hook test.** Added `use-confidential-transfer-and-call.test.tsx` mirroring the `useConfidentialTransfer` suite — optimistic subtract, rollback-on-error, try/finally resilience, raw/optimistic context unwrapping, cache invalidation, and error propagation. Also mocked `confidentialTransferAndCall` / `confidentialTransferFromAndCall` on the token fixtures.

**4. Arg ordering.** Tightened `token.test.ts` for both `…AndCall` variants from `arrayContaining([DATA])` to the full ordered `toEqual([...])`, so a wrong-order regression fails at the token-method layer too.

**Merge note:** the textual merge missed a semantic conflict — `prerelease` removed the `encryptedAbi` binding, so the two `…AndCall` contract builders no longer compiled. Repointed them at the consolidated `confidentialWrapperAbi`, which already carries both overloads.

Local verification: typecheck, `oxlint` + `ast-grep`, `forge build`, `api-report:check`, `llm:check`, and the full vitest suite (1555 passing) all green.

packages/sdk/src/token/token.ts
```diff
@@ -617,6 +622,143 @@ export class Token {
...
+    if (!skipBalanceCheck) {
+      await this.assertConfidentialBalance(amount);
+    }
+
+    const { encryptedValues, inputProof } = await this.sdk.encrypt({
+      values: [{ value: amount, type: "euint64" }],
+      contractAddress: this.address,
+      userAddress: getAddress(account.address),
+    });
+    void swallow("transferAndCall: onEncryptComplete", () => onEncryptComplete?.());
```
ankurdotb: [AUTOMATED] **Nit (non-blocking):** this `swallow(...)` — and the one at line 742 in `confidentialTransferFromAndCall` — doesn't pass `this.sdk.logger`, unlike `confidentialTransfer` (line 550) and `confidentialTransferFrom` (line 602). So if a caller's `onEncryptComplete` throws, it's swallowed with no `warn` log in the AndCall variants only. Looks like fallout from the `LoggerService` change (SDK-232) merging in after these methods were written. 2-line parity fix: add `this.sdk.logger` as the third arg in both spots.

packages/react-sdk/src/transfer/use-confidential-transfer-from-and-call.ts
```diff
@@ -0,0 +1,58 @@
...
+    ConfidentialTransferFromAndCallParams,
+    Address
+  >,
+) {
+  const token = useToken(address);
+
+  return useMutation<TransactionResult, Error, ConfidentialTransferFromAndCallParams, Address>({
+    ...confidentialTransferFromAndCallMutationOptions(token),
+    ...options,
+    onSuccess: (data, variables, onMutateResult, context) => {
```
ankurdotb, line 53: [AUTOMATED] **Nit (non-blocking):** `useConfidentialTransferFromAndCall` has no hook-level test, while the AndCall variant got a full suite and the sibling `useConfidentialTransferFrom` has a small one. This `onSuccess` override (caller-forwarding + `invalidateAfterTransfer`) is currently untested at the hook layer. A ~2-test suite mirroring `use-confidential-transfer-from.test.tsx` would restore parity.

ankurdotb (dismissed):
[AUTOMATED]
## Re-review — all prior points addressed ✅

Thanks — the three items from the earlier review are resolved cleanly:

- **Docs (the SDK-168 close-gate):** the new "Transfer into a contract (receiver hook)" section in `transfer-privately.md` is exactly right — vault-deposit example, `encodeAbiParameters` for the payload, balance-check guidance (with `skipBalanceCheck` framed as the smart-wallet escape hatch), and the "a confirmed tx means the hook ran, not that the deposit settled" caveat for queued/batched vaults.
- **Hook tests:** the 652-line `use-confidential-transfer-and-call.test.tsx` genuinely exercises the bug-prone paths — optimistic subtract, rollback-on-error, try/finally resilience, caller-context unwrapping, and cache invalidation. It would fail on a regression.
- **Assertions:** `token.test.ts` now pins the full ordered args instead of `arrayContaining`.

Also nice: the demo's `skipBalanceCheck: true` now carries the explanatory comment.

I also re-verified the builder↔generated-ABI alignment (the ABI is now generated from the V3 wrapper artifacts), barrel exports, and the token-method logic — all clean.

Approving. Two non-blocking Nits inline (fix-if-convenient).

ankurdotb (approved):
[AUTOMATED]
## Re-review — both Nits resolved ✅

- **`swallow` logger parity:** all four calls now pass `this.sdk.logger`, so `transferAndCall` / `transferFromAndCall` match `transfer` / `transferFrom` — a throwing `onEncryptComplete` is now logged at `warn` in the AndCall variants too.
- **`useConfidentialTransferFromAndCall` hook test:** new suite covering default state, balance-cache invalidation (right token invalidated, others untouched), and `onSuccess` forwarding — mirrors the `useConfidentialTransferFrom` suite.

Re-approving (my earlier approval was dismissed by the new pushes). Nothing else outstanding from my side.

PR #422, opened 2026-06-16
refactor(sdk): add ConfidentialWrapperV3 denylist ABI surface [SDK-215]

ghermet:
## Summary

Adds the `ConfidentialWrapperV3` denylist surface to `packages/sdk/src/abi/{encrypted,wrapper}.abi.ts`. Entries are hand-mirrored from canonical [`ConfidentialWrapperV3.sol`](https://github.com/zama-ai/protocol-apps/blob/main/contracts/confidential-wrapper/contracts/upgrades/ConfidentialWrapperV3.sol) on `protocol-apps/main` because the locally pinned soldeer snapshot (`protocol-apps-wrapper@3bd308fb7cb1`) predates V3.

**Functions** (5): `reinitializeV3`, `blockUser`, `unblockUser`, `isBlocked`, `getUnderlyingDenyListSelector`
**Events** (2): `UserBlocked`, `UserUnblocked`
**Errors** (6): `BlockedUser`, `UserAlreadyBlocked`, `UserAlreadyUnblocked`, `UnderlyingDenyListCallFailed`, `InvalidUnderlyingDenyListResponse`, `UnderlyingDenyListedAddress`

This is **ABI-only** scope from [SDK-215](https://linear.app/zama/issue/SDK-215). No SDK method surface, no `Token.isBlocked()` helper, no event decoders in `events/onchain-events.ts`. Those are open product questions deferred to follow-ups; this PR gets the raw ABI in so viem/ethers consumers can already decode V3 reverts and events.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm test:run packages/sdk/src/{abi,__tests__,events,contracts}` — 205 tests pass
- [ ] CI green
- [ ] Once the soldeer pin is bumped past V3, regenerate ABIs and diff against this hand-authored set to confirm parity

## Follow-ups

Tracked in [SDK-215](https://linear.app/zama/issue/SDK-215):
- Expose `blockUser` / `unblockUser` / `isBlocked` as `Token` methods? (product decision — owner-only admin actions, may not belong in the user-facing SDK)
- Add `UserBlocked` / `UserUnblocked` decoders to `events/onchain-events.ts` if consumers want to subscribe to denylist changes
- Glossary entry for the three underlying-denylist passthrough errors (USDT-style sanctioned-address surface)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #435, opened 2026-06-16
chore(contracts): bump soldeer wrapper to ConfidentialWrapperV3 and deploy V3 in tests [SDK-220]

ghermet:
## Summary

The wrapper deployed on mainnet and Sepolia is `ConfidentialWrapperV3` on `zama-ai/protocol-apps@main`, but the SDK's soldeer pin `protocol-apps-wrapper@3bd308fb7cb1` was V2-era and had no V3 / no denylist surface — so `contracts/script/Deploy.s.sol` couldn't exercise V3, and [SDK-215](https://linear.app/zama/issue/SDK-215) (#422) had to hand-mirror the denylist ABI.

- Bump `protocol-apps-wrapper` pin to `71611c624ddc` (latest on `protocol-apps@main`; no release tags exist, `main` is the deployment). Registry pin (`0647c5fd9e41`) untouched — no upstream change since April.
- `contracts/script/Deploy.s.sol`: deploy `ConfidentialWrapperV3` as the UUPS implementation and call `reinitializeV3([], bytes4(0), false)` on each proxy so denylist admin (`blockUser`/`unblockUser`/`isBlocked`) and V3 enforcement on transfer/wrap/unwrap are active in the test harness.
- Verified the compiled V3 ABI against the hand-mirrored entries added in [SDK-215](https://linear.app/zama/issue/SDK-215) (#422) — exact match on all 5 functions, 2 events, 6 errors. No reconciliation diff needed; #422 can land as-is.
- Refresh `contracts/VERSIONS.md`: new pin, V3 denylist section, corrected base properties (added missing `Wrap` event, fixed `inferredTotalSupply` accessor description).

Closes [SDK-220](https://linear.app/zama/issue/SDK-220).

## Test plan

- [x] `forge build` clean (V3 + V2 + base wrapper compile together)
- [x] `bash test/playwright/start-anvil.sh 8545` deploys cleanly end-to-end
- [x] On the live anvil proxy, V3 surface verified via `cast`:
  - `isBlocked(0x…01) → false` initially
  - `getUnderlyingDenyListSelector() → (false, 0x00000000)`
  - `blockUser(0x…99)` succeeds (owner = deployer)
  - `isBlocked(0x…99) → true` after blocking
- [x] `pnpm typecheck` clean
- [ ] CI green

## Follow-ups

- [SDK-216](https://linear.app/zama/issue/SDK-216): add the 3-field `Wrap` event to `wrapper.abi.ts` (now verifiable against the local compiled artifact).
- [SDK-223](https://linear.app/zama/issue/SDK-223) (blocked by this): generate ABIs from compiled artifacts instead of hand-maintaining.

### DIFF
--- a/contracts/script/Deploy.s.sol
+++ b/contracts/script/Deploy.s.sol
@@ -6,6 +6,8 @@ import {console} from "forge-std/console.sol";
 import {TestERC20} from "../src/mocks/Erc20Mintable.sol";
 import {TestERC1363} from "../src/mocks/Erc1363Mintable.sol";
 import {ConfidentialWrapper} from "protocol-apps-wrapper/contracts/confidential-wrapper/contracts/ConfidentialWrapper.sol";
+import {ConfidentialWrapperV3} from
+    "protocol-apps-wrapper/contracts/confidential-wrapper/contracts/upgrades/ConfidentialWrapperV3.sol";
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import {ERC1967Proxy} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
 import {ConfidentialTokenWrappersRegistry} from "protocol-apps-registry/contracts/confidential-token-wrappers-registry/contracts/ConfidentialTokenWrappersRegistry.sol";
@@ -14,8 +16,10 @@ contract Deploy is Script {
     function run() external {
         vm.startBroadcast();
 
-        // 1. Deploy ConfidentialWrapper implementation (for UUPS proxies)
-        ConfidentialWrapper wrapperImpl = new ConfidentialWrapper();
+        // 1. Deploy ConfidentialWrapperV3 implementation (for UUPS proxies).
+        // V3 matches what's live on mainnet and sepolia (zama-ai/protocol-apps@main),
+        // and exposes the owner-controlled denylist surface (blockUser/unblockUser/isBlocked).
+        ConfidentialWrapperV3 wrapperImpl = new ConfidentialWrapperV3();
         console.log("WrapperImpl:", address(wrapperImpl));
 
         // 2. Deploy test ERC20 tokens
@@ -26,12 +30,12 @@ contract Deploy is Script {
         console.log("USDT:", address(usdt));
 
         // 3. Deploy wrapper proxies directly
-        ConfidentialWrapper cUSDC = _deployWrapper(
+        ConfidentialWrapperV3 cUSDC = _deployWrapper(
             address(wrapperImpl), "Confidential ERC20 Token", "cERC20", IERC20(address(usdc))
         );
         console.log("cUSDC:", address(cUSDC));
 
-        ConfidentialWrapper cUSDT = _deployWrapper(
+        ConfidentialWrapperV3 cUSDT = _deployWrapper(
             address(wrapperImpl), "Confidential Tether USD", "cUSDT", IERC20(address(usdt))
         );
         console.log("cUSDT:", address(cUSDT));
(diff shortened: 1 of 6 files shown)


ankurdotb (dismissed):
[AUTOMATED] Superseded by the review below (re-posted with the Deploy.s.sol nit anchored as an inline line comment). See review 4513883937.

contracts/script/Deploy.s.sol
```diff
@@ -70,6 +74,17 @@ contract Deploy is Script {
         registry.registerConfidentialToken(address(erc1363Token), address(cERC1363));
         console.log("WrappersRegistry:", address(registry));
 
+        // 7. Bring each proxy to V3 storage so denylist admin (blockUser/unblockUser/isBlocked)
```
ankurdotb, line 77: [AUTOMATED] **Required (nit):** duplicate `// 7.` step number. This reinit block is `// 7.`, and line 90 (`// 7. Write deployments.json`) is also labelled `// 7.`. Renumber — this block → `7`, the deployments.json write → `8`.

ankurdotb (changes requested):
[AUTOMATED] ## Code Review — SDK-220

**Verdict: Request changes.** The substance is correct (verified below) — but two stale artifacts introduced by this pin bump should be fixed first. One is inline on `Deploy.s.sol`; the other is here because its file isn't part of this PR's diff, so GitHub can't anchor an inline comment to it.

### Required — stale pin label in the version smoke test
`packages/sdk/src/contracts/__tests__/contracts.test.ts` (not in this PR's diff):
- **Line 216** (comment) and **line 220** (`describe` title) still name `protocol-apps@3bd308fb7cb1`. This PR moves the pin to `71611c624ddc`, so the *version* smoke test now labels the wrong commit. The assertions still pass (the V1 base ABI is byte-identical across the two commits), so it's label-only: update both occurrences `3bd308fb7cb1` → `71611c624ddc`.

### Verified clean (for the record)
- **Pin bump is purely additive** — base wrapper byte-identical at `71611c624ddc` (`finalizeUnwrap` / `Wrap` / `UnwrapRequested` / interfaceId `0x1f1c62b2` unchanged), so the hand-maintained ABIs in #418/#419/#421/#422 stay valid.
- **`Deploy.s.sol` init sequence valid** — `initialize` (v1) → `reinitializeV3` (`reinitializer(3)`) is sound under OZ Initializable 5.1.0 (v2's reinitializer is an empty no-op); `reinitializeV3([], 0x0, false)` has no revert path; empty denylist ⇒ V3 overrides are a no-op pass-through ⇒ existing E2E unaffected.
- **Deterministic addresses preserved** — the `reinitializeV3` calls are deferred past all `CREATE`s, so nonces and the checked-in `deployments.json` (consumed by test-nextjs/test-vite) stay valid.
- **`VERSIONS.md`** accurate, stage-gate language, no internal ticket refs; `forge build` clean; registry pin correctly untouched.

### Merge note
Land this before/with #422 — it's what makes #422's denylist ABI artifact-verifiable and gives the harness a real V3. Contracts-only, so no conflict with the ABI-family PRs.

ghermet:
Fixed the stale pin label in 3a5bff4d — both occurrences in `packages/sdk/src/contracts/__tests__/contracts.test.ts` (line 216 comment + line 220 `describe` title) updated `3bd308fb7cb1` → `71611c624ddc`. All 27 contract tests still pass (assertions were ABI-identical across the two commits, as you noted). Grepped the rest of the repo for the old SHA: no other occurrences.

contracts/script/Deploy.s.sol
ghermet: Fixed in f0e276df — the deployments.json write step is now `// 8.`.

ankurdotb (approved):
[AUTOMATED] ## Code Review — SDK-220

Both required nits are fixed — approving.

- **`contracts/script/Deploy.s.sol`** — step numbers are now sequential: reinit block = `// 7.`, deployments.json write = `// 8.` ✓
- **`packages/sdk/src/contracts/__tests__/contracts.test.ts`** — pin label updated to `71611c624ddc` on both lines 216 and 220 ✓

All prior analysis stands: pin bump is additive (base wrapper ABI byte-identical), V3 init sequence valid under OZ Initializable 5.1.0, deterministic addresses preserved, `VERSIONS.md` accurate.

PR #438, opened 2026-06-17
chore(sdk): generate ABIs from compiled artifacts [SDK-223]

ghermet:
## Summary

- Replaces hand-maintained `packages/sdk/src/abi/{encrypted,wrapper}.abi.ts` with output generated from the deployed `ConfidentialWrapperV3` Forge artifact (`contracts/out/ConfidentialWrapperV3.sol/ConfidentialWrapperV3.json`).
- Adds `pnpm abi:build` (regenerate) and `pnpm abi:check` (build → vitest → `git diff --exit-code`), mirroring the `llm:check` / `api-report:check` shape — commit locally, CI verifies, no write-back to the tree.
- New `.github/workflows/abi.yml` runs on PRs touching `contracts/`, `packages/sdk/src/abi/`, or `scripts/abi/`. Reuses the existing Forge build cache and then runs `pnpm abi:check`.
- Closes [SDK-223](https://linear.app/zama/issue/SDK-223).

Stacked on top of #435 (SDK-220). Retarget to `prerelease` once that lands.

## Test plan

- [x] `pnpm abi:check` — build is deterministic and tree stays clean
- [x] `pnpm typecheck` — generated ABIs still narrow correctly across SDK + React SDK
- [ ] CI green on `ABI Check` workflow
- [ ] After SDK-216 / SDK-215 / SDK-221 hand-edits made the ABIs drift, this regeneration must remain in lockstep — verify by inspecting the new `*.abi.ts` files for the V3 surface (denylist methods, 3-field `Wrap` event, no `CannotReceiveEthForTokenWrap`)


.github/workflows/abi.yml
```diff
@@ -0,0 +1,104 @@
...
+    branches: [main, prerelease]
+    paths:
+      - "contracts/src/**"
+      - "contracts/foundry.toml"
+      - "contracts/lib/forge-fhevm/src/**"
+      - "contracts/lib/forge-fhevm/foundry.toml"
+      - "packages/sdk/src/abi/**"
+      - "scripts/abi/**"
+      - "package.json"
+      - "pnpm-lock.yaml"
```
ankurdotb, line 16: [AUTOMATED] `contracts/soldeer.lock` is missing from `paths:`. The Forge artifacts that `abi:build` reads come from the Soldeer-managed `protocol-apps-wrapper` dependency, not from `contracts/src/`. A soldeer pin bump (like the one in SDK-220/#435 that makes this PR's V3 output correct) changes `soldeer.lock` without touching any `.sol` in `contracts/src/`, so this workflow never fires on the exact change that shifts the ABI surface.

Add `- "contracts/soldeer.lock"` here. Same fix needed in both `hashFiles(...)` cache keys (lines 60 and 86) — otherwise the cached `contracts/out/` from the old dependency is restored and all three `forge build` steps are skipped, letting `abi:check` validate against a stale artifact.

scripts/abi/check.mjs
```diff
@@ -0,0 +1,37 @@
...
+    name: "verify-clean",
+    command: "git",
+    args: ["diff", "--exit-code", "--", ...generatedFiles],
+  },
+];
+
+for (const step of steps) {
+  console.log(`\n▶ abi:check [${step.name}] ${step.command} ${step.args.join(" ")}`);
+  const result = spawnSync(step.command, step.args, { stdio: "inherit" });
+  if (result.status !== 0) {
```
ankurdotb: [AUTOMATED] When `spawnSync` fails to launch the child process (e.g. `pnpm` not on PATH, `EACCES`), `result.status` is `null` and `result.error` is the OS-level `Error` — but `result.error` is never checked or printed. The script exits 1 correctly via `null ?? 1`, but the only output is `✖ abi:check failed at [build]` with no indication the process never started. In CI that's indistinguishable from a genuine ABI mismatch.

Fix: check `result.error` before the status check:
```js
if (result.error) {
  console.error(`\n✖ abi:check failed at [${step.name}]: ${result.error.message}`);
  process.exit(1);
}
```

ankurdotb (changes requested):
[AUTOMATED] ## Code Review — SDK-223

**Verdict: Request changes.** Two required fixes before merging.

### Required

**1. `soldeer.lock` missing from CI trigger paths and cache key** — `.github/workflows/abi.yml`

The Forge artifacts come from the Soldeer-managed `protocol-apps-wrapper` dependency (pulled via `contracts/soldeer.lock`), not from anything in `contracts/src/`. The `paths:` trigger does not include `contracts/soldeer.lock`, so a soldeer pin bump — precisely the kind of change that bumps the ABI surface (as SDK-220 just did) — will not trigger this workflow at all. The `hashFiles(...)` cache keys have the same gap: a dependency-rev bump with no `.sol` changes in `contracts/src/` hits the same cache, restores stale `contracts/out/`, skips all three guarded `forge build` steps, and lets `pnpm abi:check` validate against the old artifact. The check reports clean (or diffs against wrong data) and the whole invariant is silently broken. Fix: add `contracts/soldeer.lock` to the `paths:` block and to both `hashFiles(...)` cache keys (lines 60 and 86).

**2. `check.mjs` swallows spawn errors** — `scripts/abi/check.mjs:25-26`

When `spawnSync` fails to launch the child process (e.g. `pnpm` not on PATH in a fresh CI runner, or a permissions error), `result.status` is `null` and `result.error` is the OS-level `Error` — but `result.error` is never inspected or printed. The script exits 1 via `null ?? 1`, which is correct, but the only output is `✖ abi:check failed at [build]` with no indication that the process never started. In CI that's indistinguishable from a genuine ABI mismatch, sending the developer on a debugging detour. Fix: check `result.error` before the status check and print `result.error.message`.

### Verified clean
- ABI surface changes are correct for V3: `Wrap(address,uint256,bytes32)`, full denylist surface, `UnwrapRequested`/`UnwrapFinalized`, `unwrapAmountCleartext` — all match `ConfidentialWrapperV3` on `protocol-apps@main`.
- `build.mjs` fails loudly on missing/malformed artifacts (`ENOENT` → named error, missing `.abi` → `TypeError`).
- `git diff --exit-code` in `verify-clean` correctly detects drift against tracked files in CI.
- `pnpm abi:check` (not `node scripts/abi/check.mjs`) — namespace convention followed.
- `abi:build`/`abi:check` pattern correctly mirrors `llm:build`/`llm:check`.
- Tests are non-trivial (real round-trip ABI structural checks against the compiled artifact).
- No ticket refs in `.md` files.

### Note on #418/#419/#421/#422
This PR's generated output covers what those four ABI-family PRs were fixing manually (Wrap event, denylist surface, param renames, removed orphans). Once this lands, those PRs are superseded for their ABI-file portions and can be closed.

ghermet:
Thanks for the careful review. Both required fixes addressed, plus a couple of follow-ups while I was in the file.

### Required fixes

**1. `soldeer.lock` in trigger paths and cache key** — adeedc13

Added `contracts/soldeer.lock` AND `contracts/lib/forge-fhevm/soldeer.lock` to the `paths:` block and to both `hashFiles(...)` cache keys (restore + save). The second lockfile pins forge-fhevm's own Soldeer deps; the IERC165/IERC1363 generated ABIs added in 1d2f6cce / 3c25e1db come from the OZ pin under that lockfile, so a bump there would silently hit the cache and skip the rebuild for the same reason.

**2. `check.mjs` swallows spawn errors** — adeedc13

Added an explicit `result.error` branch before the status check that prints `✖ abi:check failed at [<step>]: could not spawn \`<cmd>\`: <ENOENT msg>` and exits 1. Verified the `result.error.code === "ENOENT"` shape with a smoke test against a missing binary.

### Follow-ups in the same area

**3. Generate erc165Abi and erc1363Abi from artifacts** — 1d2f6cce, 3c25e1db

Hand-written `erc165.abi.ts` and `erc1363.abi.ts` were the obvious gap remaining after this PR's pattern landed. Both now generated from the compiled OZ artifacts (`IERC165.sol/IERC165.json`, `IERC1363.sol/IERC1363.json`). ERC-1363 expands from a single `transferAndCall` slice to the full 15-entry IERC1363 surface; viem disambiguates the 3-arg overload from the args tuple, so `transferAndCallContract` needed no call-site change.

**4. Collapse `encrypted`/`wrapper` into `confidential-wrapper`** — 02cba91a

Following the comment in `build.mjs` (\"kept as two exports purely for SDK-internal import compatibility\"), merged the two duplicate ABI files (sliced from the same `ConfidentialWrapperV3` artifact) into a single `confidentialWrapperAbi`, and likewise merged the two contract helper modules. The barrel re-exports the same 13 helpers — no public surface change.

**5. Cleanup of check.mjs error messages** — 8aeb9ddc, 2a774ad0

Dropped the duplicated "✖ abi:check failed at [step]" banner line (already present via the leading `▶ abi:check [step]` log or the child's own stderr), folded it into the verify-clean dedicated message. Also replaced a top-level `return` with `process.exit(0)` for ESM correctness.

### Public API impact

Verified `git diff origin/prerelease..HEAD -- packages/*/etc/*.api.md`:

- Top-level export set is identical (sorted diff of `+export Name` vs `-export Name` is empty).
- Function signatures unchanged.
- `functionName` and `args` discriminants on every `*Contract` helper match exactly.
- The 18.4 K-line churn in `sdk.api.md` is purely the Forge-emitted ABI key order (`type, name, inputs, outputs, stateMutability, internalType`) vs the previous hand-written order. TS object types are structural, so consumer types resolve identically.

re #418/#419/#421/#422 — agreed, those are superseded for their ABI-file portions and can close once this lands.

scripts/abi/check.mjs
```diff
@@ -0,0 +1,40 @@
...
+    command: "git",
+    args: ["diff", "--exit-code", "--", ...generatedFiles],
+  },
+];
+
+for (const step of steps) {
+  console.log(`\n▶ abi:check [${step.name}] ${step.command} ${step.args.join(" ")}`);
+  const result = spawnSync(step.command, step.args, { stdio: "inherit" });
+  if (result.status === 0) {
+    process.exit(0);
```
ankurdotb: [AUTOMATED] **Blocking:** this `process.exit(0)` is inside the `for (step of steps)` loop, so it fires as soon as the first step (`build`) succeeds — which is the normal case. `test` and `verify-clean` then never run, and `verify-clean` (the `git diff --exit-code`) is the entire point of this script. As written, `abi:check` passes green whenever `pnpm abi:build` doesn't crash, so committed-vs-generated drift is never caught. The `✓ abi:check passed` banner on line 40 is also now unreachable.

The commit message ("replace top-level return with process.exit") explains how this slipped in: a top-level `return` is a SyntaxError inside an `.mjs` loop, but the right loop-control keyword here is `continue`, not `process.exit(0)`:

```js
if (result.status === 0) {
  continue;
}
```

That lets all three steps run, makes line 40 reachable again, and exits 0 only after `verify-clean` passes.

Verified with a standalone repro of this exact loop: with drift staged in `verify-clean`, the script still exits 0 today.

Confidence: 98/100

ankurdotb (changes requested):
[AUTOMATED] ## Code Review — SDK-223 (re-review)

Thanks for the follow-ups — both points from the last round are fixed, and the extra cleanup is a real improvement. One new regression in `check.mjs` blocks, but it's a one-line fix.

### Resolved from last review
- **`soldeer.lock` now tracked** — added to `paths:` and both `hashFiles(...)` cache keys, plus `contracts/lib/forge-fhevm/soldeer.lock`. A soldeer pin bump now correctly triggers and busts the cache.
- **`check.mjs` surfaces spawn errors** — `result.error` is now inspected and reported with a useful message.

### Nice improvements beyond the asks
- **Collapsed the duplicated ABI** — `encrypted.abi.ts` + `wrapper.abi.ts` (byte-identical, both `ConfidentialWrapperV3`) → one `confidential-wrapper.abi.ts` (`confidentialWrapperAbi`); `contracts/{encrypted,wrapper}.ts` → `contracts/confidential-wrapper.ts`. Internal-only (these ABI consts were never in the public API report), builders re-pointed cleanly, no lingering imports, test re-pointed. Not a breaking change.
- **`erc165Abi` / `erc1363Abi` now generated** from `IERC165` / `IERC1363` artifacts — content verified (erc165 = `supportsInterface`; erc1363 = the three `*AndCall` methods + data overloads + inherited ERC-20 + `supportsInterface`).
- `check.mjs` derives its file list from `build.mjs` `targets` — single source of truth.

### Blocking (1)
**`scripts/abi/check.mjs:24` — `process.exit(0)` short-circuits the whole check after the first step.** Inline comment below. On a successful `build` (the normal case) the script exits 0 and `test` + `verify-clean` never run — so the drift check, the entire point of this PR, is a no-op that always passes green. Fix is `process.exit(0)` → `continue`.

With that one line fixed this is good to merge. As noted last round, once this lands #418/#419/#421/#422 are superseded for their ABI-file portions and can be closed.

zama-cremaud:
Couple of FYI but approved otherwise!

1 - vitest.abi.config.ts is not in the workflow paths filter
```
.github/workflows/abi.yml
```
The paths trigger lists scripts/abi/**, packages/sdk/src/abi/**, package.json, etc., but not vitest.abi.config.ts (it lives at repo root). A PR that touches only that config file won't run ABI Check. Very low impact (the config rarely changes alone), but adding - "vitest.abi.config.ts" to the paths list would close the gap.

2 - targets coverage is not self-enforcing
```
scripts/abi/build.mjs
```
abi:check only validates the 4 ABI files listed in targets. There is no guard that every file under packages/sdk/src/abi/** is covered by a target, so a future hand-added ABI file there would silently escape the drift check. Today all 4 files are covered, so this is purely a future-proofing observation — e.g. an assertion that the set of *.abi.ts files equals the set of targets[].path would make the contract explicit. Optional.


3 - new Function in the test
```
scripts/abi/__tests__/abi-build.test.mjs
```
Already acknowledged with a comment + oxlint-disable. Acceptable here since the input is repo-controlled source; no action needed.

zama-cremaud (approved):

ankurdotb (approved):
[AUTOMATED] ## Code Review — SDK-223 (re-review)

**Verdict: Approve.** The blocking regression is fixed.

`scripts/abi/check.mjs:24` now uses `continue` instead of `process.exit(0)`, so all three steps run and `verify-clean` (the `git diff --exit-code` drift check) actually executes — the success banner on line 40 is reachable again. Scoped diff since the last review confirms this one-liner is the only change to the PR's files; the intervening `prerelease` merges didn't touch the ABI scripts, generated ABIs, workflow, or contracts.

Everything from the prior rounds stands: soldeer.lock tracked in `paths:` + both cache keys, spawn errors surfaced, the encrypted/wrapper ABI collapse, and erc165/erc1363 generation. Good to merge.

Reminder: once this lands, #418/#419/#421/#422 are superseded for their ABI-file portions and can be closed.

PR #441, opened 2026-06-17
ci(docs): link-checking tooling (docs:check-links + lychee) [SDK-227]

ankurdotb:
Closes SDK-227. Sub-task of SDK-224. **Branch-agnostic link-checking tooling**, split out of the original blended PR so it can be cherry-picked to `main` independently of the content link-fixes.

> **Stacked on #437** (base = `docs/gitbook-relative-links`). Merge #437 (the prerelease link-fix) first; this PR's gate is hard-fail and depends on those fixes being present. After #437 merges I'll retarget this base to `prerelease`.

## What

- **`scripts/docs/check-links.mjs`** (markdown-it) — bans host-absolute internal links, validates relative `.md` targets, resolves `#anchor`s with GitBook's slug rule. Handles inline / reference-style / HTML `<a>`+`` links; ignores links in code by construction.
- **`pnpm docs:check-links`** + focused vitest tests ([scripts/docs/__tests__](scripts/docs/__tests__/check-links.test.mjs)), wired into CI via a new `scripts` vitest project scoped to `scripts/docs/**` (the `scripts/llm` corpus tests keep running via `pnpm llm:check`, so they are not double-run).
- **lychee** external-link job in the docs workflow — **PR-only** (`if: github.event_name == 'pull_request'`, so the release `workflow_call` isn't gated on external-link liveness), `--accept 200..=299,403,429`, [.lycheeignore](.lycheeignore) for placeholder URLs.
- **Renamed** the docs workflow `LLM Docs` → **`Docs`** (it now does link-checking + corpus); added `scripts/docs/**` to its path filter. ⚠️ Update any branch-protection required-check rule keyed on `LLM Docs / …`.
- Convention documented in [docs/agents/gotchas.md](docs/agents/gotchas.md) + [CONTRIBUTING.md](CONTRIBUTING.md).

## Verification

- `pnpm docs:check-links`: ✓ 87 pages. `pnpm exec vitest run --project scripts`, `pnpm lint`, `pnpm format:check`: clean.

## Cherry-pick to main

After SDK-226 (main link-fix) lands, cherry-pick this PR's commit to `main` (the gate needs main's links fixed first).


zama-cremaud:
Couple of "optional / FYI" but approved otherwise.

FYI — duplicate-heading anchors are a latent false-positive. extractHeadingSlugs stores one slug per heading; GitBook (like GitHub) disambiguates repeated headings by appending -1, -2, … (two ## Example → #example, #example-1). A valid link to #example-1 would be reported as "anchor not found." Not currently triggered (PR reports ✓ 87 pages), so it's a latent risk only — worth a one-line code comment near slug/extractHeadingSlugs so a future doc author isn't surprised by a CI failure on a correct link.

scripts/docs/check-links.mjs ~ extractHeadingSlugs / checkTree anchor branch.
FYI — PR description vs. code mismatch on scripts/llm tests. The description says the new vitest project "Also adopts the previously-orphaned scripts/llm tests," but vitest.config.ts includes only scripts/docs/**/*.test.mjs, and the inline comment explicitly states scripts/llm tests run via pnpm llm:check (vitest.llm.config.ts). The code is internally consistent; just update the PR body so reviewers/branch-protection owners aren't misled. (No code change required.)

FYI — .lycheeignore patterns are broad substrings. Entries like infura\.io and example\.com match anywhere in a URL, so they suppress all such links, not only the placeholder samples. That's acceptable for these placeholder hosts, but if a real, link-checkable infura.io URL is ever added it'll be silently skipped. Fine as-is; flagging for awareness.

Optional — orphan gate may surprise on intentionally-unlinked pages. findOrphans treats every .md under src (except SUMMARY.md) as required-in-SUMMARY. This is correct for GitBook publishing, but if the team ever keeps a deliberately-unpublished draft in src/, the gate will hard-fail. Consider documenting an escape hatch (e.g. an ignore list) if that case arises — not needed now.

zama-cremaud (dismissed):

ankurdotb:
Thanks for the review — all four addressed:

1. **Duplicate-heading anchors** — fixed properly rather than just commented: `extractHeadingSlugs` now disambiguates repeated headings GitBook/GitHub-style (`#example`, `#example-1`, `#example-2`, …), so a valid link to `#example-1` resolves. Added a unit test.
2. **PR-description mismatch** — corrected the body; it now states the `scripts` vitest project is scoped to `scripts/docs/**` and that `scripts/llm` tests keep running via `pnpm llm:check` (no double-run).
3. **Broad `.lycheeignore` patterns** — tightened: dropped the broad `infura\.io` (the placeholders all carry `YOUR_KEY`, which is what we match now, so a real infura URL would still be checked) and removed the unused `example\.com`.
4. **Orphan-gate escape hatch** — intentionally **not** adding one (maintainer call): every page under `docs/gitbook/src` is meant to be published, so an orphan is a real error and should hard-fail. Non-published docs (e.g. `docs/agents/`) live outside `src/` and aren't checked; deliberately-unpublished drafts should stay out of `src/` rather than be ignore-listed. Documented this in a code comment near the orphan check.

ankurdotb:
Applied two more fixes on top of the review round:

**Slug rule now matches GitBook exactly (underscores preserved).** I verified against a live GitBook-published page (the SDK reference, server-rendered HTML): `## TOKEN_TOPICS` → `id="token_topics"` / `href="#token_topics"` (underscore kept), zero `token-topics`. Our `slug()` had been converting `_`→`-`, so a future link to an underscore heading (e.g. `event-decoders.md#token_topics`) would have falsely failed. Fixed: keep `[a-z0-9_]`; periods/spaces/etc. still collapse to `-`. Headings are now rendered (markdown-it inline) before slugging so italic markup like `_(static)_` doesn't leak a literal underscore — the existing `Token.md#token-batchdecryptbalancesas-static` link still resolves. Added tests for both.

**Vacuous-pass guard.** `checkTree()` now fails loudly if no pages are found under `docs/gitbook/src` (e.g. run from the wrong CWD) or if `SUMMARY.md` is missing, instead of printing "0 pages checked — all good" and exiting 0.

Together with the earlier round (duplicate-heading dedup, tightened `.lycheeignore`, PR-body fix, documented orphan-gate scope), all review points are addressed.

zama-cremaud (approved):

PR #442, opened 2026-06-17
fix(chains): add ingenTestnet + bscTestnet presets

zama-cremaud:
## What

Promotes two cleartext deployments to first-class `FheChain` presets in `@zama-fhe/sdk/chains`, alongside `mainnet` / `sepolia` / `hoodi`:

| Preset | Chain ID | Network |
| --- | --- | --- |
| `ingenTestnet` | `364301` | InGen Testnet |
| `bscTestnet` | `97` | BNB Smart Chain Testnet (Chapel) |

Until now these chains existed only as inline config inside `example-ingen` / `example-bnb`. This makes the SDK able to talk to them out of the box (registry address auto-resolved), at the same tier as the other presets.

**Naming convention:** network-specific chains that can grow a mainnet sibling carry the `Testnet` suffix (`bscTestnet`, `ingenTestnet`), reserving the bare name for a future mainnet preset. The Ethereum testnets `sepolia` / `hoodi` stay bare since their mainnet is just `mainnet`.

## Why

Follow-up to the SDK-211 (Ingen) / SDK-212 (BNB) wrapper-upgrade verification. Both deployments' wrappers already expose the post-mainnet `IERC7984ERC20Wrapper` interface — verified on-chain:

- `supportsInterface(0x1f1c62b2) == true`, `inferredTotalSupply()` present, legacy `totalSupply()` reverts
- A full shield → transfer → **unshield (unwrap + finalize)** cycle was run end-to-end against the current SDK on **both** chains; the emitted log topics match the canonical `UnwrapRequested(address,bytes32,bytes32)` / `UnwrapFinalized(address,bytes32,bytes32,uint64)` signatures.

The e2e runs also confirmed both deployments use the default mock cleartext signers, so **no private keys are needed in the presets**.

## Changes

- `chains/configs.ts`: add `ingenTestnet` and `bscTestnet`, register both in the `chains` map
- Re-export from `chains`, the root `index`, and the `node` entrypoint
- `DefaultRegistryAddresses`: map both registries so discovery resolves from the connected chain
- Preset enumeration tests (`relayer-utils.test.ts`)
- `configuration.md` network table + regenerated API reports and llms artifacts

## Notes

- The example apps (`example-ingen` / `example-bnb`) are **not** migrated to import these presets here — they're pinned to published SDK versions and will switch over at the next stable release once the presets ship.
- Hoodi is intentionally out of scope for this PR.

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -1466,6 +1466,8 @@ import { sepolia, mainnet, hoodi } from "@zama-fhe/sdk/chains";
 | `bscTestnet`   | `97`       | BNB Smart Chain Testnet |
 | `hardhat`      | `31337`    | Local Hardhat node      |
 
+`anvil` is also exported as an alias for `hardhat` (both target chain ID `31337`), for Foundry users.
+
 ### 2. Pick a relayer
 
 Relayers tell the SDK how to run FHE operations on each chain.
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -1457,14 +1457,14 @@ Import pre-configured chain objects from `@zama-fhe/sdk/chains`. Each chain incl
 import { sepolia, mainnet, hoodi } from "@zama-fhe/sdk/chains";
 ```
 
-| Chain        | Chain ID   | Description         |
-| ------------ | ---------- | ------------------- |
-| `mainnet`    | `1`        | Ethereum Mainnet    |
-| `sepolia`    | `11155111` | Sepolia Testnet     |
-| `hoodi`      | `560048`   | Hoodi Testnet       |
-| `ingen`      | `364301`   | T-Rex InGen Testnet |
-| `bscTestnet` | `97`       | BNB Smart Chain Testnet |
-| `hardhat`    | `31337`    | Local Hardhat node  |
+| Chain          | Chain ID   | Description             |
+| -------------- | ---------- | ----------------------- |
+| `mainnet`      | `1`        | Ethereum Mainnet        |
+| `sepolia`      | `11155111` | Sepolia Testnet         |
+| `hoodi`        | `560048`   | Hoodi Testnet           |
+| `ingenTestnet` | `364301`   | InGen Testnet           |
+| `bscTestnet`   | `97`       | BNB Smart Chain Testnet |
+| `hardhat`      | `31337`    | Local Hardhat node      |
 
 ### 2. Pick a relayer
 
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -1457,12 +1457,14 @@ Import pre-configured chain objects from `@zama-fhe/sdk/chains`. Each chain incl
 import { sepolia, mainnet, hoodi } from "@zama-fhe/sdk/chains";
 ```
 
-| Chain     | Chain ID   | Description        |
-| --------- | ---------- | ------------------ |
-| `mainnet` | `1`        | Ethereum Mainnet   |
-| `sepolia` | `11155111` | Sepolia Testnet    |
-| `hoodi`   | `560048`   | Hoodi Testnet      |
-| `hardhat` | `31337`    | Local Hardhat node |
+| Chain        | Chain ID   | Description         |
+| ------------ | ---------- | ------------------- |
+| `mainnet`    | `1`        | Ethereum Mainnet    |
+| `sepolia`    | `11155111` | Sepolia Testnet     |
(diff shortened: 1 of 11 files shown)


ankurdotb:
nitpick: Ingen is ingen testnet, stick to convention

llms-full.txt
```diff
@@ -1457,12 +1457,14 @@ Import pre-configured chain objects from `@zama-fhe/sdk/chains`. Each chain incl
...
-| `sepolia` | `11155111` | Sepolia Testnet    |
-| `hoodi`   | `560048`   | Hoodi Testnet      |
-| `hardhat` | `31337`    | Local Hardhat node |
+| Chain        | Chain ID   | Description         |
+| ------------ | ---------- | ------------------- |
+| `mainnet`    | `1`        | Ethereum Mainnet    |
+| `sepolia`    | `11155111` | Sepolia Testnet     |
+| `hoodi`      | `560048`   | Hoodi Testnet       |
+| `ingen`      | `364301`   | T-Rex InGen Testnet |
+| `bscTestnet` | `97`       | BNB Smart Chain Testnet |
```
ghermet: why is it called bscTestnet?

ghermet: bsc is both the name for testnet and mainnet I suppose?

ankurdotb: They have a BSC testnet and mainnet as separate chains with chain IDs

ankurdotb (commented):
## Code review — add `ingen` + `bscTestnet` presets

Nice follow-up — graduates the two cleartext deployments into first-class presets. Reviewed the source and correctness looks solid:

- Addresses match the verified `example-bnb` / `example-ingen` configs exactly (registries `0xc0E8B73b…` / `0x7FC3D79E…`, executors, ACL, verifying contracts), `satisfies FheChain` on both, registered in the `chains` map **and** `DefaultRegistryAddresses`, re-exports consistent across the root + node entrypoints, and no private keys needed (default mock signers). Tests cover the ids, executor presence, and the `chains[]` lookups. 👍

Two things before merge:

**1. Rename `ingen` → `ingenTestnet` (change request).** Let's disambiguate testnet vs mainnet from the start, the same way `bscTestnet` does. If an InGen mainnet ever ships we'd otherwise be forced into a breaking rename of the `ingen` export — reserving the bare `ingen` for a future mainnet now avoids that. The rule we're settling on: network-specific chains that can grow a mainnet sibling carry the `Testnet` suffix (`bscTestnet`, `ingenTestnet`); the Ethereum testnets `sepolia` / `hoodi` stay bare since their mainnet is just `mainnet`. This touches the const, the three re-export lines, the `wrappers-registry` import, the preset test, `configuration.md`, and the regenerated API reports + `llms-full.txt`.

**2. CI `lint` is red.** `oxfmt --check` wants `docs/gitbook/src/guides/configuration.md` reformatted — the new rows widened the `Chain` column, so the table needs re-padding. `pnpm format` fixes it.

Minor: in `configuration.md` the description "T-Rex InGen Testnet" → "InGen Testnet" would match the `<Network> Testnet` style of the other rows (folds into the same formatting pass).

Everything else looks good to merge once those land.

llms-full.txt
ankurdotb: Right — BSC mainnet (56) and testnet (97) are separate chains, so `bscTestnet` deliberately keeps the bare `bsc` free for a future mainnet preset. Same logic should apply to `ingen` → `ingenTestnet` (flagged in the review) so we disambiguate from the start rather than rename later.

zama-cremaud:
Thanks both — addressed in `b469cbe`:

1. **`ingen` → `ingenTestnet`** (@ankurdotb's change request). Renamed the const and every reference (three re-export lines, the `wrappers-registry` import + `DefaultRegistryAddresses` key, the preset test, `configuration.md`, and the regenerated API reports + `llms-full.txt`). Same `<Network>Testnet` rationale as `bscTestnet` — reserves the bare `ingen` for a future InGen mainnet.
2. **Red `lint`** — `pnpm format` re-padded the widened `configuration.md` table; `format:check`, `oxlint`, typecheck, and the 102 preset/registry tests are all green locally.
3. **Minor** — InGen row description tightened to "InGen Testnet" to match the `<Network> Testnet` style.

Also rebased onto the latest `prerelease` merge; rebuilt + regenerated artifacts against the new base (no public-API drift from SDK-220). PR title/description updated to reflect `ingenTestnet`.

ankurdotb (approved):

PR #447, opened 2026-06-19
feat(codemod): SDK upgrade codemods (Codemod-OSS / JSSG) for 3.1.0

zama-cremaud:
## What this is

SDK-upgrade **codemods** for the breaking changes in `@zama-fhe/sdk` / `@zama-fhe/react-sdk` **3.1.0**, on the [Codemod OSS](https://codemod.com) engine — a `codemods/sdk-upgrade-v3-1-0` workspace package. It applies the **mechanical** subset deterministically; the non-mechanical tail is left to an opt-in `ai` step + manual review.

- **ast-grep YAML rules** (`rules/*.yml`) for symbol/type renames + config-key changes.
- **JSSG transforms** (`scripts/*.ts`) for structural/context-sensitive rewrites (import-aware `createZamaConfig→createConfig`, config-object→positional, token-field any-shape, `UseZamaConfig` removal).
- Keyed **per breaking release** (`sdk-upgrade-v3-1-0`, O(N)); distributed via the Codemod registry (`npx codemod @zama-fhe/sdk-upgrade-v3-1-0 -t ./src`). No SDK repo / Claude Code dependency for consumers.

Authorship: the codemod engine + transforms are @ghermet's (commits preserved). The commits on top (tests, idempotency, a fix, CI) are the test/hardening pass.

## Added on top (this pass)

- **Rule test coverage** — `codemod jssg test` only ran the JSSG `scripts/`; the 5 `rules/*.yml` had fixtures but **no runner**. Added `test-rules.mjs` (same `tests/<rule>/` convention + idempotency), wired into `test.sh`.
- **Fix surfaced by it** — `use-delegation-status` only matched a single-property object (`{ tokenAddress: $V }`), so realistic `useDelegationStatus({ tokenAddress, delegateAddress })` calls slipped through. Rewrote it as a scoped relational `inside` rule (any object shape / key order; other call sites untouched).
- **JSSG idempotency** — an `idempotent` fixture case per transform (migrated output must round-trip).
- **End-to-end** — `test-e2e.mjs` runs the whole chain in workflow order on a multi-file fixture, asserting convergence + chain-level idempotency.
- **CI** — the package tests ran nowhere (`pnpm test:coverage` is vitest-only); added `.github/workflows/codemod.yml`.

All green: 4 JSSG transforms (+ idempotent cases) + 5 rules + e2e convergence/idempotency.

## Open items for review

- **`.ts` type-position gap (known limitation):** `language: tsx` rules under-cover plain `.ts` files — the engine renames import specifiers but misses type-position occurrences (`function f(h: Handle)` keeps `Handle`). Works in `.tsx`. Fix = `language: typescript` rule variants (or map `.ts`→tsx in engine config). Documented in the package README; e2e fixtures are `.tsx` for this reason.
- **`ai` node gating:** the README says it's off-by-default via `--param ai=true`, but `workflow.yaml` has no node condition — it's a no-op only because no LLM is configured. Consider explicit gating.
- **Release type:** commits use `feat(codemod)` / `fix(codemod)`. Confirm semantic-release does not cut an SDK release for the `codemods/` workspace (scope/path), or adjust types.
- **GA:** re-derive against the final 3.1.0 api-report when it ships.


codemods/sdk-upgrade-v3-1-0/test-rules.mjs
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env node
+// Fixture tests for the ast-grep YAML rules in rules/.
```
ghermet: This is now ran by the test.sh script if I'm not mistaken ? Codemod have their own testing

zama-cremaud: Yes, good point, thanks! Migrated the rule tests to ast-grep's own test harness and dropped the custom test-rules.mjs

package.json
```diff
@@ -84,11 +85,13 @@
     "@vitest/runner": "^4.1.9",
     "@vitest/ui": "^4.1.9",
     "@zama-fhe/relayer-sdk": "0.4.3",
+    "codemod": "1.12.3",
     "conventional-changelog-conventionalcommits": "^9.3.1",
     "ethers": "^6.17.0",
     "fake-indexeddb": "^6.2.5",
     "happy-dom": "^20.10.6",
     "husky": "^9.1.7",
+    "jscodeshift": "^17.0.0",
```
ghermet: Are we sure this is still needed?

package.json
```diff
@@ -23,6 +23,7 @@
     "typecheck": "pnpm run --parallel --recursive '/^typecheck/'",
     "lint": "oxlint && pnpm ast-grep:scan",
     "ast-grep:scan": "ast-grep scan --config sgconfig.yml",
+    "codemod": "codemod workflow run -w codemods/sdk-upgrade-v3-1-0/workflow.yaml",
```
ghermet: Where is this command used? if there will be more codemods maybe we should write `codemod:sdk-upgrade-v3-1-0`?

codemods/sdk-upgrade-v3-1-0/codemod.yaml
```diff
@@ -0,0 +1,25 @@
+schema_version: "1.0"
+
+name: "@zama-fhe/sdk-upgrade-v3-1-0"
```
ghermet: maybe `sdk-migration-v3-1-0` would be industry standard naming?

codemods/sdk-upgrade-v3-1-0/codemod.yaml
```diff
@@ -0,0 +1,25 @@
...
+description: "Mechanical codemods for the breaking changes introduced in @zama-fhe/sdk + @zama-fhe/react-sdk 3.1.0 (renames, config-key changes, structural removals). Assumes a 3.0.x floor (run earlier release codemods first if further behind); idempotent. Provisional until 3.1.0 ships; derived from the v3.0.1 -> v3.1.0-alpha.14 api-report diff."
+author: "Zama"
+license: "BSD-3-Clause-Clear"
+repository: "https://github.com/zama-ai/sdk"
+
+targets:
+  # Keep this aligned with the files matched in workflow.yaml.
+  languages: ["typescript", "tsx"]
+
+keywords: ["zama", "fhe", "sdk", "migration", "upgrade"]
```
ghermet: Can use the same keywords from the package json on top of `migration`

codemods/sdk-migration-v3/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "private": true,
+  "description": "Codemods for the breaking changes introduced in @zama-fhe/sdk 3.1.0 (assumes a 3.0.x floor)",
+  "scripts": {
+    "test": "./test.sh"
+  },
+  "devDependencies": {
+    "codemod": "1.12.3",
+    "@types/node": "^20.0.0",
+    "typescript": "^5.0.0"
+  }
```
ghermet, line 13: I can't remember if these are created when you do `codemod init` or not

codemods/sdk-migration-v3/workflow.yaml
```diff
@@ -0,0 +1,91 @@
...
+          js_file: "scripts/react-hooks-config-object-to-address-first.ts"
+          language: "tsx"
+          include: ["**/*.{ts,tsx}"]
+      - name: "remove UseZamaConfig interface (inline + drop import)"
+        js-ast-grep:
+          js_file: "scripts/use-zama-config-interface-removed.ts"
+          language: "tsx"
+          include: ["**/*.{ts,tsx}"]
+
+  # --- optional AI assisted upgrade: the non-mechanical changes the codemods can't express ---
```
ghermet, line 52: In this case its not opt in anymore

pnpm-workspace.yaml
```diff
@@ -23,6 +24,11 @@ overrides:
   # undici-types restores provenance on a newer release, then remove this override
   # and bump @types/node freely.
   undici-types: "6.19.2"
+  # The codemod CLI -> jscodeshift -> @babel/core wants semver@^6.3.1, but 6.3.1 lost
+  # provenance (npm shipped attested 7.5.x earlier, so no-downgrade reads 6.3.1 as a
+  # downgrade). The 7.x line is attested (SLSA provenance) and API-compatible with
+  # babel's usage, so force semver onto it instead of weakening the trust policy.
+  semver: "^7.0.0"
```
ghermet: This might no longer be useful If we don't need `codeshift` as a dependency

package.json
zama-cremaud: Good catch

codemods/sdk-upgrade-v3-1-0/codemod.yaml
zama-cremaud: Both are valid as far as I am ware but I'm also fine using "migration" if we decide to you use this standard.

codemods/sdk-migration-v3/package.json
zama-cremaud: Not from codemod init scaffolding.

codemods/sdk-upgrade-v3-1-0/codemod.yaml
zama-cremaud: Found that "sdk-migration-v3" is the most appropriate / best reflecting naming for our use case. Let me know if this is Ok for you as well!

.github/workflows/codemod.yml
```diff
@@ -0,0 +1,59 @@
...
+  group: codemod-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  # Safe-chain enforces its own minimum-package-age check independently of
+  # pnpm-workspace.yaml#minimumReleaseAgeExclude — mirror the exemption here.
+  SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS: "@zama-fhe/relayer-sdk"
```
ghermet, line 21: Can you add the MIN_AGE 72 to this like the other workflows pls?

ghermet (approved):

PR #449, opened 2026-06-21
fix(sdk): stop workers logging console.error on handled failures [SDK-230]

ghermet:
## Problem

The Node worker (and the Web worker, which shared the identical pattern) called `console.error("[…Worker] … error:", message)` on **every** operation failure, immediately before `sendError` propagated the same message back to the caller and rejected the returned promise. Because the log fired unconditionally from inside the worker thread, expected/handled failures flooded consumers' error monitoring with noise.

Reported by a customer for the Node SDK; fixed in both workers for consistency.

## Changes

- Removed the 22 redundant `console.error(...)` calls from the per-operation `catch` blocks in `relayer-sdk.node-worker.ts` and `relayer-sdk.worker.ts`. The error still reaches the caller via `sendError` (with `statusCode` preserved on the web worker) — no regression in error reporting. This is option 1 from the issue (default silent), avoiding any new public API surface.
- The two `default:` "unknown request type" cases were **not** redundant — they logged but never propagated, leaving the caller's promise to hang. Replaced each with a `throw`, so the existing outer `catch` rejects the caller. Net improvement, not just removal.

## Verification

- `tsc --noEmit` clean, `oxlint` clean
- Worker tests: 87 passed / 5 skipped
- The `sendError` → `{ success: false }` reject contract (incl. `statusCode`) remains covered by `worker.base-client` / `worker.node-pool` tests

## Done when (from SDK-230)

- [x] Handled worker failures no longer emit `console.error` by default
- [x] The same error is still propagated to the caller and rejects the returned promise
- [x] No configurable/injectable logger added — the error is already surfaced to the caller, so dropping the log is sufficient

Closes SDK-230


packages/sdk/src/worker/relayer-sdk.node-worker.ts
```diff
@@ -434,7 +423,7 @@ async function handleMessage(request: WorkerRequest): Promise<void> {
         await handleGetPublicParams(request);
         break;
       default:
-        console.error("[NodeWorker] Unknown request type:", request.type);
+        throw new Error(`Unknown request type: ${(request as WorkerRequest).type}`);
```
ankurdotb, line 426: [AUTOMATED] FYI — this `default:` case is more than a log removal; it's a behavior fix (and the same change lands in `relayer-sdk.worker.ts`). Previously an unknown request type logged to console and then fell through *without* calling `sendError`, so the caller's promise hung until the ~30s timeout. Throwing here routes it through the surrounding try/catch → `sendError` → the caller rejects promptly with `Unknown request type: …`. Nice catch — worth a line in the PR description so it isn't read as a pure logging change.

Confidence: 95/100

ankurdotb (commented):
[AUTOMATED]

## Code Review — SDK-230 (stage 1)

Reviewed against the Linear ticket and traced into the worker client side (`worker.base-client.ts` `handleResponse` → `reject`). Ran a comprehensive pass — correctness, silent-failure, and test-coverage — and CI is green (10/10). **No blocking issues.** This correctly implements stage 1: the removed `console.error` lines were byte-for-byte the message already propagated via `sendError`, every `catch` still calls `sendError`, and the fix is applied symmetrically across both worker entry points. Error propagation is unchanged.

One inline FYI below, plus one optional nit:

**Nit (pre-existing, optional):** the node-worker top-level catch — `sendError(request.id, request.type, message)` in `relayer-sdk.node-worker.ts` — reads `request.id`/`request.type` directly, whereas `relayer-sdk.worker.ts` uses `request?.id ?? "unknown"` / `request?.type ?? "UNKNOWN"`. Not introduced by this PR and effectively unreachable (the client always posts a well-formed `{ id, type, payload }`), but mirroring the guard would keep the two workers symmetric. Non-blocking.

Confidence: 90/100

packages/sdk/src/worker/relayer-sdk.node-worker.ts
ghermet: Agreed — this is a behavior fix, not just a log removal. It's already called out in the PR description under **Changes**: the two `default:` cases logged but never called `sendError`, so the caller's promise hung until the ~30s timeout; throwing routes them through the surrounding try/catch → `sendError` so the caller rejects promptly with `Unknown request type: …`. Same change in both `relayer-sdk.node-worker.ts` and `relayer-sdk.worker.ts`.

ankurdotb (approved):

PR #451, opened 2026-06-22
refactor(examples): upgrade example apps to @zama-fhe/sdk 3.1.0

zama-cremaud:
## What

Upgrades the internal example apps to `@zama-fhe/sdk` + `@zama-fhe/react-sdk` **3.1.0** (now `latest` on npm), and — for the apps that needed it — migrates hand-rolled primitives back onto the SDK per **[SDK-150]**. One commit per app.

**6 apps covered:** react-wagmi, react-ethers, react-viem, react-turnkey-wallet, example-ingen, example-bnb.
(`example-hoodi` is on 2.2.0, below the codemod's 3.0.x floor and not part of the 3.1.0 upgrade — its SDK-150 shield migration belongs to SDK-152.)

## 3.1.0 upgrade (all apps)

1. Ran the `sdk-upgrade-v3-1-0` codemod (#447) on `<app>/src` (mechanical breaks, deterministic).
2. Bumped pins to `3.1.0`, installed, ran `tsc --noEmit` against the real 3.1.0 types (**all green**).
3. Applied the non-mechanical tail by hand (surfaced by typecheck), removed comments left stale by the rewrites.

Mechanical (codemod): permit-hook renames (`useIsAllowed→useHasPermit`, `useAllow→useGrantPermit`), query-hook `tokenAddress→address`, mutation-hooks config-object → positional `address`.
Manual tail (only where the app used the affected API): `createConfig sessionStorage→permitStorage` (react-wagmi); `sdk.requireSigner(op)→sdk.signer`+guard (ethers/viem/ingen/bnb); `createToken→createWrappedToken`, `DelegationStatusData.isDelegated→.isActive` (ingen/bnb).

## SDK-150 drift migration — hand-rolled approve+wrap → SDK primitives

`example-ingen`, `example-bnb`, and `react-turnkey-wallet` hand-rolled the shield approval (manual allowance read → USDT-style reset → approve via `writeContract` → `shield({approvalStrategy:'skip'})`). Verified against the SDK source (`WrappedToken.#ensureAllowance` / `approveUnderlying`): the SDK performs the allowance check, the **USDT-style reset**, the approval, **ERC-1363 `transferAndCall` routing**, and shield submission. The hand-rolled contract calls were redundant.

Migrated all three to SDK primitives while **keeping the bounded, human-readable spend cap** (the user's full ERC-20 balance — *not* an infinite `MaxUint256` approval): approve via `useApproveUnderlying` (which performs the USDT-style reset internally), then `useShield(..., approvalStrategy:'skip')`. This drops every hand-rolled `approve`/`reset`/`writeContract` (no more USDT gas-revert detection — the SDK handles the reset) and avoids leaving an unbounded standing allowance in code developers copy verbatim.

The rest of the audit (balance, decrypt, transfer, unshield, delegation, wrapper discovery, EIP-712 credential signing) was already clean across all apps — no changes needed there.

**All 6 apps typecheck green against 3.1.0.**

## ⚠️ Do not merge yet

3.1.0 was published `2026-06-22`. The repo's `pnpm-workspace.yaml` sets `minimumReleaseAge: 4320` (3 days) **without excluding `@zama-fhe/sdk`**, so a clean install of 3.1.0 won't pass until ~`2026-06-25`. Validation here used `--config.minimumReleaseAge=0` to typecheck early.


examples/example-bnb/src/components/ShieldCard.tsx
```diff
@@ -29,98 +27,52 @@ export function ShieldCard({
...
+  const errorMessage = error ?? approve.error?.message ?? shield.error?.message ?? null;
   const pendingLabel =
     phase === "approve"
-      ? "Shielding… (1/2 approve)"
-      : phase === "wrap-after-approve"
-        ? "Shielding… (2/2 wrap)"
+      ? "Shielding… (approving)"
+      : phase === "wrap"
+        ? "Shielding… (wrapping)"
         : "Shielding…";
```
ghermet, line 36: Should be able to track the progress of the shielding with events:
```ts
function dispatchEvent(e){
 window.dispatchEvent(new CustomEvent(e.type,{ detail: e })
}
const config = createConfig({ onEvent: dispatchEvent });
const zama = new ZamaSDK(config)
```

```ts
useEffect(()=>{
const ctrl = new AbortController()
window.addEventListener(ZamaSDKEvent.ShieldPhase1, ()=>{},{signal: ctrl.signal})
window.addEventListener(ZamaSDKEvent.ShieldPhase2, ()=>{},{signal: ctrl.signal})

return ()=> ctrl.abort()
},[])

File: examples/example-bnb/src/components/ShieldCard.tsx [hunk 2]
```diff
@@ -29,98 +27,52 @@ export function ShieldCard({
...
-  const [phase, setPhase] = useState<"wrap" | "approve" | "wrap-after-approve">("wrap");
+  const [phase, setPhase] = useState<"prepare" | "approve" | "wrap">("prepare");
+  const [error, setError] = useState<string | null>(null);
 
   const sdk = useZamaSDK();
+  // Spend cap = the user's full ERC-20 balance: a bounded, human-readable allowance (not an
+  // infinite MaxUint256) that still avoids re-approving on every shield. useShield's
+  // approvalStrategy can't express a custom cap, so the approval goes through
+  // useApproveUnderlying — which performs the USDT-style allowance reset — and we then shield
+  // with approvalStrategy "skip".
```
ghermet: Is this a bug in the SDK?

examples/react-turnkey-wallet/src/components/react-turnkey-wallet/shield-card.tsx
```diff
@@ -1,54 +1,71 @@
...
+        { amount: parsed, approvalStrategy: "skip" },
+        {
+          onSuccess: () => {
+            setAmount("");
+            onSuccess();
+          },
         },
-      },
-    );
+      );
```
ghermet: This should ideally be one call, is there something that needs to change on the SDK?
```ts
  await shield.mutateAsync(
        { amount: parsed, approvalStrategy: "skip" },
        {
          onSuccess: () => {
            setAmount("");
            onSuccess();
          },
        },
      );
```

examples/example-bnb/src/components/ShieldCard.tsx
zama-cremaud: This was because of a product decision, not bug in the SDK.
Before, we approved the user's full token balance before shielding, which added extra steps.
I don't think it makes sense to keep this: it's now a single shield() call with "approvalStrategy = exact".
No SDK change needed, this was always possible in one call.

examples/react-turnkey-wallet/src/components/react-turnkey-wallet/shield-card.tsx
zama-cremaud: Same as above

examples/example-bnb/src/components/ShieldCard.tsx
zama-cremaud: This I am not sure, interested in your opinion on this.

For per-card progress I kept the per-call onApprovalSubmitted/onShieldSubmitted callbacks: it's invocation-scoped, so there is no event-correlation/subscription overhead for what is purely local button feedback.
We already use the global onEvent for the right thing; cross-cutting UnshieldPhase1Submitted -> savePendingUnshield persistence

Now, we can move the "shield progress" to onEvent too if you think events to be the canonical progress API makes more sense in this situation.
For this local case the callbacks might be simpler and consistent with the existing split.
Wdyt?

ghermet: We can leave as is for now, we haven't yet decided if events is the canonical API for this.

ghermet (approved):

PR #453, opened 2026-06-22
feat(sdk): configurable, silent-by-default logging via LoggerService [SDK-232]

ghermet:
## Summary

Makes SDK logging configurable and **silent by default** (SDK-232). The SDK no longer emits any console output of its own; handled and expected failures travel only through the rejected promise / typed error the SDK already propagates — never duplicated into the consumer's monitoring (the regression Dfns hit).

A consumer who wants logs supplies their own logger via `createConfig({ logger })`; a consumer who supplies nothing gets nothing.

## How it works

- **Public surface:** `createConfig({ logger })` accepts an optional `GenericLogger` — the minimal four-level interface (`error`/`warn`/`info`/`debug`) that `console`, pino, winston, and OpenTelemetry's `DiagLogger` satisfy directly. No bundled logging library, no imposed format, browser build unaffected.
- **`LoggerService`** (new, in `services/`, exported): the consumer's `GenericLogger` is wrapped **once** in `buildZamaConfig` and threaded through the whole graph — services, worker clients, dispatcher, relayers, and the artifact cache. It is always passed (no optional chaining at call sites), owns the silent-by-default behavior, and centralizes the `[zama-sdk]` message prefix.
- **Single source of truth:** logging is configured only at `createConfig`. The per-relayer `web({ logger })` / `node({ logger })` options were removed.
- **Invariant fix:** worker handled-failure logs (`FAILED`, timeout) demoted `error` → `debug`; worker-level faults stay at `error`. `FheArtifactCache` no longer defaults its logger to `console`.

## Level semantics

| Level | Used for |
| --- | --- |
| `error` | Unexpected internal failures only — never failures already surfaced via a rejection |
| `warn` | Recoverable / degraded conditions (fallback path, retry, swallowed best-effort write) |
| `info` | Coarse lifecycle milestones, used sparingly |
| `debug` | Verbose diagnostics — worker lifecycle, request timing, orchestration |

## Notes

- Builds on the worker-thread `console.error` removal from #449 (SDK-230), now merged into `prerelease`.
- **Breaking (prerelease only):** `logger` removed from `WebRelayerOptions` / `NodePoolOptions`; worker-client config `logger` is now required internally.

## Verification

typecheck ✓ · lint ✓ · format ✓ · 1412 tests pass · API reports + LLM corpus regenerated.

Closes SDK-233. (The SDK-232 parent resolves automatically once SDK-230 + SDK-233 land.)


docs/gitbook/src/guides/configuration.md
```diff
@@ -314,7 +314,31 @@ const config = createConfig({
 
 When done with the SDK, call `sdk.terminate()` to clean up the Web Worker or thread pool.
 
-### 6. (Optional) Choose a storage backend
+### 6. (Optional) Supply a logger
```
ankurdotb: I would do storage before logging (more useful/relevant)

ankurdotb (changes requested):
Also see my other review

packages/react-sdk/src/provider.tsx
```diff
@@ -20,14 +20,14 @@ export interface ZamaProviderProps extends PropsWithChildren {
...
     // Warmup is a latency optimization — the first real permit/decrypt call
     // will lazily retry transport-key-pair generation and surface actionable
-    // errors. We still log so persistent failures (storage corruption, relayer
-    // 4xx during generation) leave a breadcrumb during debugging.
-    // oxlint-disable-next-line no-console
-    console.warn("[zama-sdk] warm transport key pair failed:", error);
+    // errors. We route this through the configured logger (silent by default)
+    // so persistent failures (storage corruption, relayer 4xx during
+    // generation) leave a breadcrumb during debugging.
+    logger.warn("[zama-sdk] warm transport key pair failed", { error });
```
ankurdotb: [AUTOMATED] **Double `[zama-sdk]` prefix.** `LoggerService.warn` already prepends `[zama-sdk]` to every message, so this line makes consumers see `[zama-sdk] [zama-sdk] warm transport key pair failed`. Every other call site passes a bare message (e.g. `lifecycle-service` passes `"wallet account listener"`). Drop the literal prefix here:

```ts
logger.warn("warm transport key pair failed", { error });
```

Confidence: 100/100

packages/sdk/src/index.ts
```diff
@@ -52,6 +52,7 @@ export type {
 } from "./relayer/relayer-sdk.types";
 export type { KmsDelegatedUserDecryptEIP712Type as KmsDelegatedDecryptEIP712Type } from "@zama-fhe/relayer-sdk/bundle";
 export type { GenericLogger } from "./worker/worker.types";
+export { LoggerService } from "./services/logger-service";
```
ankurdotb: [AUTOMATED] **Consider not exporting `LoggerService` from the public surface.** Consumers supply a `GenericLogger` (already exported as a type, the line above) and never construct a `LoggerService` — it's an internal wrapper built in `buildZamaConfig`. It's exported only because a few public type positions name it: `ZamaConfig.logger`, `RelayerConfig.createRelayer`/`createWorker`, the `RelayerDispatcher` constructor, and `CleartextRelayerConfig.createRelayer`. For consistency with how the resolved `ZamaConfig` already exposes `signer: GenericSigner` and `storage: GenericStorage` (interfaces, not wrappers), retype those positions `LoggerService` → `GenericLogger` (sound, since `LoggerService implements GenericLogger`) and drop this export, keeping the class internal. Getting the public shape right is cheaper now on the alpha line than removing the export later (itself a breaking change).

Confidence: 90/100

ankurdotb (changes requested):
[AUTOMATED]

## Code Review — SDK-232 logging / LoggerService

Reviewed locally with a full multi-agent sweep (2 bug hunters + guideline/architecture + error-handling + tests + public-API) and traced the threading graph end to end. The design is sound and matches the ticket's invariants: silent-by-default owned in one place, exactly four levels, no bundled logging library, no Node import on browser paths, the `error`→`debug` demotion correctly scoped (handled failures still reject; genuine worker faults stay at `error`), and the `LoggerService` never crosses the worker `postMessage` boundary. CI is green. Nice, well-structured change.

Requesting changes on three small items. (1) and (4) are inline; (2) is here because its line isn't in the diff.

### Blocking

**(2) Missed logger arg — `packages/sdk/src/credentials/keypair-vault.ts:108`**
`clear()` calls `swallow("delete transport key pair entry", () => this.#storage.delete(key))` without `this.#logger`, while the three sibling `swallow` calls in this same file (47/56/91) pass it. Since `swallow` no longer falls back to `console`, a storage-delete failure during credential clear is now silent **even when the consumer configured a logger**. Fix: add `, this.#logger` as the third arg.

### Non-blocking

**(3) `swallow()` contract change left two internal sites with no logger path.** `namespaces/permits.ts:156,175` ("clear decrypt cache") and `ethers/ethers-signer.ts:69` ("refresh wallet account") still call `swallow(label, fn)` with no logger, and neither class has a logger plumbed — so these best-effort failures are invisible even with a logger configured. Nothing awaited fails, so it's not a correctness bug, but it's incomplete vs the PR's goal of routing SDK diagnostics through the logger. Thread `LoggerService` into `Permits`/`EthersSigner`, or track as a fast follow. (The `swallow` sites in `token`/`wrapped-token`/`submit-transaction` wrap consumer callbacks — fine to stay silent.)

**Test gaps — the failure paths where (1) and (2) live are untested:**
- Provider: `provider.test.tsx` covers the warm *success* path but never forces `warmTransportKeyPair()` to reject, so nothing asserts the logged message — which is how the double prefix slipped through. Add a rejection case asserting `sink.warn` is called once with a single `[zama-sdk]` prefix.
- keypair-vault: the vault is never constructed with a logger sink, and the `clear()` test only checks regeneration. Add a failing `storage.delete` + assert `logger.warn` got the breadcrumb (catches (2)).
- Worker: the demotion test proves handled failures log at `debug` and still reject, but nothing asserts genuine worker faults (`handleWorkerError`/timeout) stay at `error` — the other half of the invariant. One `sink.error` assertion closes it.

**Nit:** `info` has zero call sites in the SDK (the docs describe it). Keeping the level for portability is correct; just noting the docs describe a level nothing currently emits.

### Ticket
This PR implements the stage-2 logger work, which we split out as **SDK-233**, but the description says "Closes SDK-232" (the parent). As written it would close the parent and orphan the child. Please retarget to **Closes SDK-233** — the SDK-232 parent resolves automatically once both children (SDK-230, merged; SDK-233, this) are done.

ankurdotb:
[AUTOMATED]

Re: the `GenericLogger` vs `LoggerService` point (the `export { LoggerService }` comment) — good question, but I think it conflates two different changes, and the ergonomic concern doesn't actually apply 🙂

The suggestion is **only** to stop *exporting* the `LoggerService` class and to type the public positions as `GenericLogger` — **the wrapper stays exactly as it is**. It is *not* "thread the raw optional logger everywhere" (that's the version that would force `?.` + manual prefixes). So you keep calling `logger.info("blabla")` — no `?.`, no `[zama-sdk]` at the call site:

- **No `?.`.** The `?.` is only needed when the value is `GenericLogger | undefined`. That optionality lives solely on the consumer *input* (`ZamaConfigBase.logger?`), which `buildZamaConfig` resolves once by always constructing a `LoggerService`. The resolved `ZamaConfig.logger` stays a **required** field, so since `GenericLogger`'s four methods are non-optional, `logger.info(...)` type-checks directly.
- **No manual prefix.** The `[zama-sdk]` prefix is added inside `LoggerService.info()` — a runtime behavior of the instance. The static type doesn't change which method runs; as long as the object is a `LoggerService` (always, from `buildZamaConfig`), the prefix is applied even when the variable is *typed* `GenericLogger`.

Mechanically it mirrors how `GenericStorage` already works — `MemoryStorage` / `IndexedDBStorage` are named only where constructed, but the flowing type is the interface:

```ts
// build.ts — unchanged: construct the wrapper here
const logger = new LoggerService(params.logger);   // owns no-op default + prefix

// every call site — unchanged: no ?., no manual prefix
this.#logger.info("blabla");                        // → "[zama-sdk] blabla" at runtime

// the ONLY change — widen the public/internal annotations:
interface ZamaConfig { logger: GenericLogger; /* required, not optional */ }
createWorker?: (chains: FheChain[], logger: GenericLogger) => unknown;
createRelayer:  (chain, worker, logger: GenericLogger) => RelayerSDK;

// …then index.ts can drop `export { LoggerService }`
```

Direction is safe: `LoggerService implements GenericLogger`, so it's assignable **to** `GenericLogger` (the way we use it); the reverse isn't, thanks to the `#private` fields — which is exactly the direction we want.

ghermet:
Addressed the review in `fix(sdk): address review on injectable logger; keep LoggerService internal [SDK-233]`.

**Blocking**
- **(1) Provider double prefix** — dropped the literal `[zama-sdk]` from `provider.tsx`; the prefix is owned solely by `LoggerService`. Added a warmup-rejection test asserting every logged message is the bare `"warm transport key pair failed"`.
- **(2) keypair-vault `clear()`** — now passes `this.#logger` to `swallow()`. Added a test that forces `storage.delete` to reject and asserts the `warn` breadcrumb.
- **(4) `export { LoggerService }`** — done as you described: the wrapper stays internal (still constructed once in `buildZamaConfig`), and every public/threaded position is now typed `GenericLogger` (mirroring `GenericStorage`/`GenericSigner`). Dropped the export. API reports regenerated — the diff is purely `LoggerService → GenericLogger` plus removal of the class.

**Non-blocking**
- **(3) `swallow()` sites** — threaded the logger into `Permits` (`clear decrypt cache`). **`EthersSigner` left as a deliberate fast-follow:** the signer is consumer-constructed *before* `createConfig`, so the SDK logger isn't reachable in its constructor without changing the public signer construction contract. Flagging rather than forcing it in here.
- **Test gaps** — added all three: provider single-prefix, keypair-vault `clear()` breadcrumb, and a worker base-client assertion that a genuine `handleWorkerError` fault stays at `error` (the other half of the SDK-230 invariant).
- **`info` nit** — added one real call site: `info("SDK initialised")` at the end of the `ZamaSDK` constructor, so the documented level isn't empty.
- **Docs** — reordered the configuration guide so the storage section precedes the logger section.

**Ticket** — retargeted the description to **Closes SDK-233**.

Verification: typecheck / lint / format clean, 1543 tests pass, API reports + LLM corpus regenerated.

ghermet:
Correction to my previous comment: the **`info` "SDK initialised" call site was *not* added** — an auto-fix hook stripped it before commit and I reported it in error. After review, leaving `info` with no call site is the right call anyway: it stays documented and retained for portability (a consumer's logger still exposes it), matching your nit. The commit has been amended accordingly; everything else in the prior comment stands.

ankurdotb (approved):

PR #458, opened 2026-06-23
refactor(sdk): replace @zama-fhe/relayer-sdk with @fhevm/sdk

ghermet:
## Summary

Replaces the legacy `@zama-fhe/relayer-sdk` low-level dependency with the new high-level `@fhevm/sdk` engine, then re-shapes the relayer layer around a multichain router with per-chain auth and a glossary-aligned decrypt surface. Net ~11k lines deleted.

- **Swap the FHE engine** — drive `@fhevm/sdk` (via its `viem` adapter) instead of `@zama-fhe/relayer-sdk`. `FhevmClient`/`BaseRelayer` are merged into a single `FhevmRelayer`, transport factories are unified onto `FhevmRuntimeConfig`, and the deep-delegation/worker-pool indirection is removed.
- **`ChainRouter` multichain layer** — the single-chain `RelayerSDK` is re-owned as a port behind a thin `ChainRouter` (`chains/router.ts`) that owns chain management (`chains` / `chain` / `switchChain`) and hands services the live backend for the active chain. Fhevm runtime config is applied once and per-chain auth params are threaded through, so credentials are scoped per chain rather than global.
- **Capability-scoped backend clients** — the FHE backend is split into capability-scoped clients (encrypt / decrypt / base), and `FhevmRelayer` exposes an explicit `init()` lifecycle. Lazy init is routed through `init()`, which also prefetches FHE keys for the active chain under that chain's auth and memoizes the result.
- **`FheChainAuth` public type** — chain relayer auth gets its own public `FheChainAuth` type keyed on a `__type` discriminator, decoupled from the low-level `@fhevm/sdk` `type` discriminator. `FhevmRelayer` translates `__type` → `type` via `toFhevmAuth` before every request, so the internal transport shape stays hidden. Exported from the root and `/chains` entry points.
- **Glossary-aligned decrypt verbs** — the relayer port and `DecryptionService` now speak `decryptValues` / `decryptPublicValues` / `delegatedDecryptValues`, matching the Zama glossary across the SDK and React hooks.
- **Centralize `@fhevm/sdk`-derived types** — `FhevmSdkClient`, `FhevmClientOptions`, `FhevmRuntimeConfig`, and the serialized boundary types live in `relayer-sdk.types.ts`, reusing `@fhevm/sdk`'s own named action types where publicly exported. The `RelayerSDK` interface is folded into the same module (deleting `relayer-sdk.ts`) so domain shapes and backend contract share one source of truth.
- **Cleartext E2E** — node and vite E2E suites run against the `@fhevm/sdk` cleartext host stack (deployed via the `fhevm` submodule + `deploy-cleartext-host.sh`), pinned to the v0.13 host stack.
- **Build/packaging** — ship CJS for the `./node` subpath, add a deprecated `./cleartext` compatibility entry point, and drop the now-dead iife worker-bundle plugin after worker removal.
- **Migration guide** — adds `MIGRATION-relayer-to-fhevm-sdk.md` documenting the relayer → `@fhevm/sdk` move and the deep-delegation deletion path.

Pinned to `@fhevm/sdk@1.1.0-alpha.8`.

## Test plan

- [x] `tsc --noEmit` across the monorepo (sdk, react-sdk, test apps)
- [x] `oxlint` + `oxfmt` (pre-commit)
- [x] Unit/integration suites green (~2018 tests)
- [x] Live integration suite across all non-mainnet chains + gated mainnet relayer
- [x] node E2E and vite E2E against the cleartext host stack
- [x] API reports regenerated (`pnpm api-report`)


packages/sdk/etc/sdk.api.md
```diff
@@ -12679,9 +12664,17 @@ export class RelayerRequestFailedError extends ZamaError {
...
+    createEIP712(publicKey: Hex, contractAddresses: Address[], startTimestamp: number, durationDays?: number): Promise<EIP712TypedData>;
+    delegatedUserDecrypt(params: DelegatedDecryptValuesParams): Promise<Readonly<Record<EncryptedValue, ClearValue>>>;
+    encrypt(params: EncryptParams): Promise<EncryptResult>;
+    fetchFheEncryptionKeyBytes(): Promise<FheEncryptionKey | null>;
+    generateTransportKeyPair(): Promise<TransportKeyPair>;
     getAclAddress(): Promise<Address>;
+    publicDecrypt(encryptedValues: EncryptedValue[]): Promise<DecryptPublicValuesResult>;
     terminate(): void;
+    userDecrypt(params: DecryptValuesParams): Promise<Readonly<Record<EncryptedValue, ClearValue>>>;
 }
```
ghermet: @ankurdotb Do you want these methods to match the `fhevm/sdk` client ? Would be breaking change on the public `sdk.relayer`

zama-cremaud:
## Might worth having a look:

### 1. Error classifier still keyed to the removed `@zama-fhe/relayer-sdk` — typed errors silently degrade with the new engine

**File:** `packages/sdk/src/utils/error.ts:306` (`isRelayerFetchError`) and `:434` (`isNotEntitledMessage`) — Confidence: 96/100

`packages/sdk/package.json` swaps `@zama-fhe/relayer-sdk ~0.4.4` for `@fhevm/sdk 1.1.0-alpha.6`, but `utils/error.ts` was not updated and still matches the legacy engine's error shapes. Verified against the actual `@fhevm/sdk@1.1.0-alpha.6` tarball:

- **`NotEntitledError` silently downgrades.** `isNotEntitledMessage` requires the literal substrings `"user address"` + `"is not authorized to user decrypt"`. `@fhevm/sdk` throws `AclUserDecryptionError` with `"User 0x… is not authorized to decrypt handle …"` (no "address", no "user decrypt"), so every ACL denial now falls through to a generic `DecryptionFailedError`, losing `encryptedValue`/`contractAddress`/`account`. This is exactly the SDK-239 regression class the code's own comments warn about.
- **Relayer 429 back-pressure mis-routed (breaks SDK-236).** `@fhevm/sdk` never sets `code: "RELAYER_FETCH_ERROR"` (zero hits in the package); it throws `RelayerResponseStatusError`/`RelayerResponseApiError` with a numeric `status`. So the `isRelayerFetchError` exclusion documented at `error.ts:329-338` can never fire, and a relayer 429 (`status: 429`) is caught by `hasStructuredRpcRateLimitSignal` and surfaced as `RpcRateLimitError` instead of `RelayerRequestFailedError` with `retryAfter`/`retryable`.
- **Drift guards are now orphaned.** `packages/sdk/src/utils/__tests__/error.test.ts:541-559` still does `require.resolve("@zama-fhe/relayer-sdk/node")` — the package is no longer a dependency of `packages/sdk` (it survives only transitively under `@fhevm/mock-utils` in `test/playwright`), so under strict pnpm these tests fail to resolve; and either way they pin the message/tag of an engine the runtime no longer calls.
- **Dead code:** `isRetryableRelayerError` (`error.ts:391`) has zero callers since `withRetry`/`relayer-utils.ts` were deleted; its docstring still claims it is "the single gate for `withRetry`". (Retry itself is not lost — `@fhevm/sdk`'s `RelayerAsyncRequest` does its own `_fetchWithRetry` — but the classifier is now unreachable orphan code.)

**Fix:** re-key `isNotEntitledMessage` / the relayer-error detection on `@fhevm/sdk`'s actual error classes (`AclUserDecryptionError`, `RelayerResponseStatusError`/`RelayerResponseApiError` names and `status`), restore the 429 → `RelayerRequestFailedError` routing, delete or rewire `isRetryableRelayerError`, and re-pin the drift-guard tests to `@fhevm/sdk`.

### 2. `CredentialService` freezes the initial chain's relayer — permits signed with the wrong EIP-712 domain after `switchChain`

**File:** `packages/sdk/src/credentials/credential-service.ts:66` (used in `#signPermit` at `:247`, `:260`, `:264`, `:267`) — Confidence: 90/100 (found independently by both bug-hunt passes)

`this.#relayer = config.router.relayer;` captures `chains[0]`'s backend once at construction. `ChainRouter.switchChain()` mutates the router in place and `get relayer()` is chain-dependent, but `CredentialService` is constructed once in `zama-sdk.ts:89` and never rebuilt (`LifecycleService` explicitly keeps credentials across chain-only changes). `signDecryptionPermit` is chain-specific (`toFhevmChain` projects per-chain `gatewayChainId` and `verifyingContractAddressDecryption` into the EIP-712 domain). After the wallet switches to chain B, `grantPermit` stores the permit under `scope.chainId = B` but signs it against chain A's domain — `DecryptionService` (which correctly reads `this.#router.relayer`) then replays an invalid permit, so every freshly granted permit on a non-initial chain fails to decrypt. Notably, the `TransportKeyPairVault` generator in the same constructor already reads `config.router.relayer` dynamically per call — only `#signPermit` was left pinned.

**Fix:** store the router, not the relayer — `this.#router = config.router;` and use `this.#router.relayer.…` in `#signPermit`, mirroring `DecryptionService`/`DelegationService`/`EncryptionService`.

### 3. `delegationStatusQueryOptions` reads the initial chain's ACL address after `switchChain`

**File:** `packages/sdk/src/query/delegation-status.ts:43` — Confidence: 85/100

`const acl = sdk.relayer.chain.aclContractAddress;` — but `sdk.relayer` is a plain `readonly` field set once in `zama-sdk.ts:61` from `config.router.relayer`, i.e. permanently `chains[0]`. Meanwhile `sdk.provider` (wagmi path) follows the currently connected chain. After a switch to chain B this queries chain A's ACL address on chain B — wrong contract / revert / bogus expiry wherever ACL addresses differ per chain. `DelegationService` itself was correctly migrated to the dynamic form (`this.#router.relayer.chain.aclContractAddress`, `delegation-service.ts:93/139/191`), and this file's own comment says it wants to stay consistent with `sdk.delegations.isActive()`.

**Fix:** resolve the ACL through the live router (same pattern as `delegation-service.ts:93`). Worth also auditing whether the frozen public `sdk.relayer` field should be a getter over the router — it's the root cause behind both 2 and 3.

zama-cremaud:
## + Non-blocking (to consider)

### 4. Persisted permits are silently wiped on upgrade — needs a consumer-facing note

**File:** `packages/sdk/src/credentials/permission-store.ts:42-46` — Confidence: 95/100

`PermissionSchema` changed incompatibly (dropped `signerAddress`/`chainId`/`signature`/…, added `contractAddresses`/`serializedPermit`) while the storage key derivation (`permits:${signer}:${chainId}:${delegator}`) is unchanged. On first `list()` after upgrade, `safeParse` fails, the whole scope is deleted, and `[]` is returned — no log, no migration, users silently re-authorize every permit. Acceptable for a major bump on the alpha line, but it must be called out in the consumer-facing changelog/release notes (the migration doc is an internal implementation doc and never mentions it). A debug-level log on the discard path would also help support.

### 5. `FhevmRelayer` hardcodes viem in core — ethers consumers are silently forced onto viem

**File:** `packages/sdk/src/relayer/fhevm-relayer.ts:21` (import), `:52-58` (`createPublicClient` from `chain.network`) — Confidence: 85/100

`docs/agents/conventions.md` ("framework-specific code lives in adapter sub-paths, not the core") and `docs/agents/gotchas.md` (chain reads live on `GenericProvider`). This new core file builds its own viem `PublicClient`, bypassing the `GenericProvider` configured via `createConfig`, and `web()`/`node()`/`cleartext()` all wire it unconditionally. `@fhevm/sdk` does force *a* framework client (no generic factory exists — verified in the tarball), so some adapter is unavoidable — but it ships both `/viem` and `/ethers` entry points, while this PR leaves no ethers relayer path despite `src/ethers/` adapters and `docs/agents/architecture.md` advertising ethers support. Consider moving the client construction into the adapter sub-paths (or a pluggable client factory) — or, if viem-only is the deliberate new baseline for the engine, documenting that decision explicitly.

zama-cremaud:
## Nit

In `packages/sdk/src/query/user-decrypt.ts:13` doc comment mentioned "keyed by encrypted handle" -> shouldn't it be "encrypted value"  instead of "encrypted handle" since its public surface?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -62,7 +63,7 @@ export class CredentialService {
       storage: config.permitStorage ?? config.storage,
       logger: config.logger,
     });
-    this.#relayer = config.relayer;
+    this.#relayer = config.router.relayer;
```
ghermet: [AUTOMATED] **Blocking bug — stale relayer after `switchChain`.**

`this.#relayer` snapshots `config.router.relayer` at construction. `ChainRouter.relayer` is a getter on the active `#chainId`, and `switchChain` (called from `LifecycleService` on wallet-account change) mutates it — but `CredentialService` is built once (`zama-sdk.ts:89`) and kept across chain switches (`handleWalletAccountChange` intentionally preserves credentials on chain-only changes).

After a switch, `#signPermit` (lines 247/260/264/267) signs against the **previous** chain's relayer → wrong KMS/EIP-712 domain, so the permit is rejected when `DecryptionService` (which reads `this.#router.relayer` lazily) verifies on the new chain.

Note line 54 in this same constructor already reads `config.router.relayer` lazily inside the vault generator — apply the same pattern here: store the router and read `.relayer` per-call in `#signPermit`.

Confidence: 90/100

packages/sdk/src/zama-sdk.ts
```diff
@@ -59,7 +58,7 @@ export class ZamaSDK {
   readonly #delegationService: DelegationService;
 
   constructor(config: ZamaConfig) {
-    this.relayer = config.relayer;
+    this.relayer = config.router.relayer;
```
ghermet: [AUTOMATED] Same stale-snapshot root cause as `credential-service.ts:66`. `this.relayer = config.router.relayer` captures the active-chain backend once; after `switchChain` it points at the wrong chain. This surfaces in `query/delegation-status.ts:43` (`sdk.relayer.chain.aclContractAddress`), which reads the wrong ACL address post-switch.

Consider exposing `relayer` as a getter that delegates to the router (`get relayer() { return this.#router.relayer; }`) so all readers stay aligned with the active chain.

Confidence: 82/100

packages/sdk/src/query/delegation-status.ts
```diff
@@ -40,7 +40,7 @@ export function delegationStatusQueryOptions(
       assertNonNullable(contractAddress, "delegationStatusQueryOptions: contractAddress");
       assertNonNullable(delegatorAddress, "delegationStatusQueryOptions: delegatorAddress");
       assertNonNullable(delegateAddress, "delegationStatusQueryOptions: delegateAddress");
-      const acl = await sdk.relayer.getAclAddress();
+      const acl = sdk.relayer.chain.aclContractAddress;
```
ghermet, line 43: [AUTOMATED] Concrete symptom of the `ZamaSDK.relayer` stale snapshot (`zama-sdk.ts:61`): after a chain switch this reads the ACL address from the previously-active chain and queries `getDelegationExpiry` on the wrong contract. Reading through a router-backed getter (or `sdk` router directly) fixes it.

Confidence: 82/100

packages/sdk/src/test-fixtures/services.ts
```diff
@@ -14,37 +14,39 @@ import { LoggerService } from "../services/logger-service";
...
+import type { RelayerSDK } from "..";
 
 export type CreateCredentialServiceFn = (
   config?: Partial<CredentialServiceConfig>,
 ) => CredentialService;
 
 export type CreateDelegationServiceFn = (overrides?: {
   provider?: GenericProvider;
-  relayer?: RelayerDispatcher;
+  relayer?: RelayerSDK;
```
ghermet: [AUTOMATED] Nit — legacy-shim shape. These three factory override types still expose `relayer?: RelayerSDK` (lines 29/37/42) and the bodies wrap it via `createMockRouter({ relayer: overrides.relayer ?? relayer })`, while the sibling `CreateLifecycleServiceFn` (line 49) already migrated to `router?: ChainRouter`. Per the repo's no-legacy-shim convention, migrate these to `router?: ChainRouter` too so fixtures speak one vocabulary and don't silently accept the pre-migration shape.

Confidence: 80/100

ghermet (commented):
[AUTOMATED]

## Code Review — PR #458 (`@fhevm/sdk` migration)

Reviewed with four parallel agents (2× bug-hunter, guideline-compliance, error-handling) plus manual validation against the working tree at the PR head. The migration is well-structured and the bulk deletion of the worker/deep-delegation layers is clean. One **blocking multichain correctness bug** stands out, independently flagged by three of the four agents and confirmed by tracing the code.

### 🔴 Blocking: stale relayer after `switchChain`
`ChainRouter.relayer` is a live getter keyed on `#chainId`, and `switchChain` (fired from `LifecycleService` on every wallet-account change, `lifecycle-service.ts:78`) mutates that id. Two long-lived objects capture `config.router.relayer` **once at construction** instead of reading it lazily:

- `CredentialService.#relayer` (`credential-service.ts:66`) — used in `#signPermit`. After a chain switch, permits are signed against the **previous** chain's KMS/EIP-712 domain while `DecryptionService` (correctly lazy) verifies against the new chain → permit rejected. Note the asymmetry: the vault generator in the same constructor (`credential-service.ts:54`) already reads `config.router.relayer` lazily. `DecryptionService`/`DelegationService`/`EncryptionService` all store the router and read `.relayer` per-call — `CredentialService` is the outlier.
- `ZamaSDK.relayer` (`zama-sdk.ts:61`) — surfaces concretely in `delegation-status.ts:43` (`sdk.relayer.chain.aclContractAddress`), reading the wrong chain's ACL address after a switch.

**Fix (both):** store the `ChainRouter` and read `.relayer` lazily, matching the pattern already used by the other services. See inline comments.

### Nits
- Legacy-shim in `test-fixtures/services.ts` (see inline).
- Stale JSDoc at `namespaces/decryption.ts:15` references `sdk.relayer.userDecrypt / delegatedUserDecrypt / publicDecrypt` — these relayer verbs were renamed in this PR to `decryptValues` / `delegatedDecryptValues` / `decryptPublicValues` and no longer exist. Update the comment.

### FYI (not blocking)
- `decryption-service.ts:333-346` zips `@fhevm/sdk`'s `decryptValues` result positionally and `continue`s on a `value === undefined` slot. If the backend ever returns fewer entries than requested (without throwing), the missing handle silently drops out of the result map and reads as `undefined`/zero downstream rather than an error. Worth an explicit length assertion if the `@fhevm/sdk` contract doesn't already guarantee 1:1 alignment.

Checked: worker-timeout error removal leaves no dangling references; `wrapDecryptError`/`errors/match.ts` exhaustiveness intact; relayer verbs consistently renamed; no constructable relayer leaked from public barrels.

zama-cremaud:
Relayer/WASM timeouts silently lose their "retryable" signal

**Files**: `packages/sdk/src/errors/decrypt.ts:130-134`, `packages/sdk/src/errors/relayer.ts:22-33`, `packages/sdk/src/utils/error.ts` (`isRelayerError`, ~L123-138)

The deleted `packages/sdk/src/errors/timeout.ts` (`WorkerTimeoutError`) was
explicitly documented as retryable ("the operation itself is retryable") and
was a distinct, `@public` exported class (confirmed via
`packages/sdk/etc/sdk.api.md` diff — both `WorkerTimeoutError` and
`WorkerRecycledError` are removed from the public API surface).

In the new code, an `@fhevm/sdk`-thrown timeout (a `Relayer*`-named class,
matched by `isRelayerError`'s name-prefix check) falls into:

```ts
// packages/sdk/src/errors/decrypt.ts:130
if (isRelayerError(error) || statusCode !== undefined) {
  return new RelayerRequestFailedError(message, statusCode, {
    cause: error,
    retryAfter: extractRetryAfter(error),
  });
}
```

```ts
// packages/sdk/src/errors/relayer.ts:22-33
constructor(message: string, statusCode?: number, options?: ...) {
  ...
  this.statusCode = statusCode;
  this.retryable = statusCode === 429;
  this.retryAfter = this.retryable ? options?.retryAfter : undefined;
}
```

A timeout carries no HTTP status code, so `statusCode` is `undefined`,
`retryable` evaluates to `false`, and `retryAfter` is nulled out. The result:
**a relayer/WASM timeout now surfaces identically to a genuinely terminal
error** (e.g. a rejected request or a 500) on the one public field
(`.retryable`) that exists specifically to let callers decide whether to
retry. `retryable` is documented `@public` API (`sdk.api.md`).

This isn't a hypothetical regression from a design-intent document — it's a
concrete, traceable behavior change: before this PR, a timeout was a distinct
type (`instanceof WorkerTimeoutError`) that `matchZamaError` could route to a
"retry" branch; after this PR, that type doesn't exist and the signal is
`retryable: false`.

**Mitigating context** (worth weighing, not dismissing the finding): `@fhevm/sdk`
retries internally (`fetchRetries: 2` in `FhevmRelayer`'s default options), so
some timeouts may already be absorbed before they ever reach SDK-level error
classification. That's a reasonable design if intentional, but it should be a
documented decision, not an incidental side effect of a type deletion —
currently there is no code comment, docstring, or migration-guide entry
explaining that timeout-retryability moved from "SDK-modeled and exposed" to
"opaque and internal to @fhevm/sdk."

**Suggested fix**: either special-case the relayer-timeout class(es) to set
`retryable: true` in `wrapDecryptError`/`wrapEncryptError` (mirroring the old
type's intent), or explicitly document in the PR/migration guide that
SDK-level retry signaling for timeouts is intentionally gone and why.

zama-cremaud:
Bounded per-operation timeout removed with no replacement

**Files**: `packages/sdk/src/node/config.ts` (diff vs merge-base), `packages/sdk/src/relayer/fhevm-relayer.ts:8-13,54`, `packages/sdk/src/relayer/types.ts` (`FhevmClientOptions`)

At merge-base, the `node()` preset exposed `poolSize`, `operationTimeout`
(default 30s — "rejects with `WorkerTimeoutError` and recycles the worker"),
`initTimeout` (default 60s), and `recycleWorkerOnTimeout` (default `true`).
All four are gone. Confirmed via the `sdk-node.api.md` diff, which shows all
four fields removed from the public `NodePoolOptions` type, and via
`fhevm-relayer.ts`, where the only per-call option carried is:

```ts
// packages/sdk/src/relayer/fhevm-relayer.ts:8-13
export type RelayerCommonOptions = {
  fetchRetries?: number;
  fetchRetryDelayInMilliseconds?: number;
};
// ...
this.#defaultOptions = { auth: this.#chain.auth, fetchRetries: 2 };
```

There is no `AbortController`, `setTimeout`, or any wall-clock ceiling
anywhere in this file, and `FhevmClientOptions` (`relayer/types.ts`) is
documented to carry only `batchRpcCalls`/`fheEncryptionKey` — no timeout
knob. `fetchRetries` bounds *retry count*, not wall-clock duration; a single
hung request (stuck relayer, stalled WASM init) has no configurable ceiling
at the SDK layer anymore, and no self-healing recycle behavior either (there
is nothing left to recycle — workers are gone).

Whether this actually causes indefinite hangs in practice depends on whether
`@fhevm/sdk` itself imposes an internal fetch timeout — this could not be
verified from the diff alone (`@fhevm/sdk` isn't installed in the review
worktree). That uncertainty is itself the point: this used to be an explicit,
documented, user-configurable contract, and now it's an unverified assumption
about a third-party package's internals.

**Suggested fix**: confirm whether `@fhevm/sdk` enforces its own timeout by
default; if not, surface an equivalent timeout knob through
`FhevmClientOptions`/`FhevmRuntimeConfig` before release, or explicitly
document that timeout-bounding is now the integrator's responsibility (e.g.
via their own `AbortSignal`-capable provider).

ghermet:
Thanks for both timeout findings — they're accurate. Addressing them separately:

**1. Timeouts losing the `retryable` signal — fixed in bb95860f.**
Confirmed: an `@fhevm/sdk` `RelayerTimeoutError` carries no HTTP status, so `RelayerRequestFailedError` (whose `retryable` was derived solely from `statusCode === 429`) surfaced it as `retryable: false` — indistinguishable from a terminal error, and a regression from the old retryable `WorkerTimeoutError`. The fix adds an `isRelayerTimeoutError` helper and threads an explicit `retryable` override through `RelayerRequestFailedError`, forced on relayer timeouts in both `wrapDecryptError` and `wrapEncryptError` (other relayer errors keep the status-based default). Unit tests cover both paths; the only public-API delta is an additive optional `retryable?: boolean` constructor option.

**2. Bounded per-operation timeout — intentional delegation to `@fhevm/sdk`, with one caveat worth surfacing.**
I was able to verify what the review worktree couldn't (`@fhevm/sdk` installed locally): `@fhevm/sdk`'s `RelayerAsyncRequest` **does** enforce its own global request timeout and throws `RelayerTimeoutError` on expiry — so there is no indefinite-hang path for relayer requests; the per-op timeout modeling and worker-recycle logic were dropped deliberately as part of removing the worker pool, not lost. So this isn't unbounded.

The caveat: that internal default is **1 hour** (`DEFAULT_GLOBAL_REQUEST_TIMEOUT_MS = 60 * 60 * 1000`), versus the old preset's configurable 30s `operationTimeout`, and while `@fhevm/sdk` exposes a per-request `timeout`/`signal` option, `FhevmRelayer` doesn't currently forward it — so integrators can't tighten that ceiling through `node()`/`FhevmClientOptions` today. That's a reasonable stance for a first cut (the transport owns its own deadline), but if a shorter, configurable ceiling matters before release it's a small follow-up to plumb `timeout` through `FhevmClientOptions`. Happy to do that here if you'd prefer it in-scope rather than a follow-up — let me know.

---
_Generated by [Claude Code](https://claude.ai/code)_

zama-cremaud:
The new `timeout`/`signal` option does not bound client initialization — only the relayer request phase — and a stuck init permanently wedges the client

**Files**: `packages/sdk/src/relayer/fhevm-relayer.ts` (12 bare `await this.#fhevm.init()` call sites), `packages/sdk/src/relayer/types.ts:79-86` (`RelayerOptions.timeout` JSDoc), `packages/sdk/src/relayer/__tests__/fhevm-relayer.test.ts`

Verified directly against the real shipped source of the pinned
`@fhevm/sdk@1.2.0-alpha.1` (extracted via `npm pack`, which ships full `.ts`
sources, not just `.d.ts`):

- `FhevmRelayer` calls `await this.#fhevm.init()` before every operation
  (encrypt, decrypt, public-decrypt, permit signing, …), with **no
  arguments**. `@fhevm/sdk`'s `init()` (`core/runtime/CoreFhevm-p.ts:266-277`)
  takes zero parameters:
  ```js
  init: {
    value: (): Promise<void> => {
      this.#readyPromise ??= Promise.all(
        [...this.#initFns].map((fn) => fn(this)),
      ).then(() => {});
      return this.#readyPromise;
    },
  ```
- What `init()` actually does, with no timeout/signal reaching any of it:
  on-chain protocol-version resolution (`ensureResolvedProtocolVersion`), and
  — on first encrypt/decrypt — a fetch of the package's own documented
  "~50MB FHE public encryption key" (`fetchFheEncryptionKeyBytes(f, {})`,
  called with **empty options**, so no signal is even threaded internally)
  plus a WASM module load (`initTfheModule`/`initTkmsModule`). The key-fetch
  path is a plain `fetch()` with no default timeout.
- The `timeout` option this PR added only arms `RelayerAsyncRequest`'s
  internal deadline (`core/modules/relayer/module/RelayerAsyncRequest.ts:228`),
  which is only started inside `run()` (line 325) — i.e. strictly *after*
  `init()` has already resolved, and only for the four relayer-request
  operation kinds (input-proof, public-decrypt, user-decrypt,
  delegated-user-decrypt). It never touches the init phase.
- This directly contradicts the new JSDoc it ships with:
  ```ts
  // packages/sdk/src/relayer/types.ts:79-86
  /**
   * Maximum total duration, in milliseconds, to wait for a relayer request
   * before giving up. This bounds the whole async operation — proof
   * generation or decryption, including the retry/backoff polling loop —
   * not a single HTTP call. Defaults to the `@fhevm/sdk` ceiling of one
   * hour... A per-call `timeout` overrides this default.
   */
  ```
  "The whole async operation" is not accurate — the one-time (per client
  instance) init phase is entirely outside the bound.
- **Worse than a missing bound**: `@fhevm/sdk`'s ready-promise is memoized
  with `??=`. A rejected promise is not `undefined`, so a failed `init()`
  is never retried — every subsequent call on that `FhevmRelayer`/`ZamaSDK`
  instance gets the same stuck or rejected promise forever. The only
  recovery is constructing a brand-new client. The deleted
  `packages/sdk/src/errors/timeout.ts` (`WorkerTimeoutError`) explicitly
  covered this exact scenario — its own documented text (still readable via
  `git show <merge-base>:packages/sdk/src/errors/timeout.ts`) says a worker
  operation "encrypt / decrypt / EIP-712 / **key fetch**... typically a
  stuck relayer or **WASM call**" would be caught, and "the Node pool
  **recycles the affected worker**... so it self-heals; the operation
  itself is retryable." The new on-calling-thread design has no worker to
  recycle, and its only timeout doesn't even reach this phase — so both the
  bound *and* the self-heal are gone for exactly the failure mode the old
  mechanism was built for.
- The new test file gives false confidence here: `fhevm-relayer.test.ts`
  mocks `init: vi.fn(async () => {})` — an instant no-op — and its
  `timeout`-related assertions only check that the `timeout` value is
  *merged into the options object* passed to a mocked downstream call. No
  fake timers, no real timeout ever fires, `signal` is never referenced. A
  test suite built this way cannot catch the gap described above.

This is realistically reachable, not a theoretical edge case: a slow/flaky
CDN mid-WASM-download, a stalled 50MB key fetch on a bad connection, or an
RPC endpoint that's down during protocol-version resolution all hang
indefinitely today, `timeout: 5000` or not.

**Suggested fix**: wrap `this.#fhevm.init()` calls in `FhevmRelayer` with the
configured `timeout`/`signal` (e.g. `Promise.race` against
`AbortSignal.timeout(...)`, or an upstream ask to `@fhevm/sdk` to accept
options on `init()`); ensure a failed/aborted init resets the memoized
ready-promise (or have `FhevmRelayer` catch-and-reset) so the client isn't
permanently wedged; correct the `RelayerOptions.timeout` JSDoc to state
plainly that it excludes the one-time init/key-load phase until fixed; add a
fake-timers test that actually exercises a real timeout firing during init,
not just object-shape merging.

ghermet:
Verified all of this against both the pinned `@fhevm/sdk@1.2.0-alpha.1` tarball and the live `js-sdk` submodule HEAD (`0c30b139d`) — the init/ready code is byte-identical, so you're right on every point: `init()` takes no args (`CoreFhevm-p.ts:266`), `??=` memoizes even a rejected ready-promise, the key fetch is called with empty options (`encrypt-p.ts:18`) so no signal threads in, our `timeout` only reaches the post-init relayer request, and the JSDoc overclaimed.

Rather than race `init()` and swap out `#fhevm` inside `FhevmRelayer` — a workaround that can't reset `@fhevm/sdk`'s private `#readyPromise` and would surprise callers — I've filed the real fix upstream (`init()` accepting `timeout`/`signal` + not memoizing a rejected init): zama-ai/fhevm#3076.

For now I've corrected the `RelayerOptions.timeout` JSDoc to state plainly that the bound excludes the one-time init/key-load phase and that a failed init doesn't self-recover, so the docs no longer overclaim while the upstream bound lands.

packages/sdk/src/services/lifecycle-service.ts
```diff
@@ -62,16 +66,16 @@ export class LifecycleService {
...
-    // RelayerDispatcher.switchChain is synchronous.
+    // ChainRouter.switchChain is synchronous.
     const nextChainId = next?.chainId;
     if (nextChainId !== undefined) {
       await swallow(
         "switch relayer chain",
-        () => this.#relayer.switchChain(nextChainId),
+        () => this.#router.switchChain(nextChainId),
         this.#logger,
       );
```
ghermet, line 82: [AUTOMATED] **Silent wrong-chain state after a failed `switchChain`.**

On a wallet account/chain change, `this.#router.switchChain(nextChainId)` runs inside `swallow(...)` — which is warn-only, never rethrows, and emits no event. If the wallet switches to a chain that isn't in the configured `chains` array, `ChainRouter.switchChain` throws `ConfigurationError` (`chains/router.ts:49-52`) and returns *before* mutating `#chainId`, so the router stays on the **previous** chain while the signer/provider are on the **new** chain.

Downstream `encryptValues` / `decryptValues` / permit-signing all read `this.#router.relayer` / `this.#router.chain` directly with no per-op re-validation against the signer's actual chain, and each relayer backend is bound to a fixed chain at construction — so no internal mismatch error is raised. The net effect is a silent cross-layer mismatch: the SDK keeps signing EIP-712 permits and issuing relayer requests against the **old** chain's domain/ACL while the wallet is on a different chain, surfaced only as a `warn` log the application can't observe.

Note the legacy `relayer-web` resolved the chain lazily per operation and would throw for an unconfigured chain, so this branch removes that per-op safety net — the failure mode is new to this PR.

Consider surfacing the failed switch rather than swallowing it — e.g. emit a typed `ZamaSDKEvent` / rethrow so the app can react, or guard the encrypt/decrypt/permit paths with a signer-chain vs router-chain check that throws a `ChainMismatchError`.

Confidence: 88/100

ghermet (commented):
[AUTOMATED]

## Code Review

Reviewed the migration with parallel bug-hunter (×2), guideline-compliance, and error-handling passes over the ~6.3k-line source diff, then independently validated every finding scoring ≥80.

**Result: the migration is clean.** No runtime bugs and no guideline violations surfaced. Two error-handling flags were dismissed on verification: the `init()` rejection-memoization is an acknowledged upstream `@fhevm/sdk` limitation already documented in this PR (with a stated discard-and-rebuild recovery path); and the `BigInt()` coercion in the signers runs inside `credentialService.signDecryptionPermit`'s try/catch, which maps any non-`ZamaError` to a matchable `SigningFailedError`, so it can't escape uncategorized.

One validated issue worth a look (non-blocking) — see inline comment on `lifecycle-service.ts`.

zama-cremaud (approved):

zama-cremaud:
## `RelayerOptions.timeout`/`signal` doesn't bound client init — but the real-world impact is narrower than it might first appear, and worth pinning down precisely

Following up on the init-phase timeout gap from the review round — I went back and verified this claim itself against the actual `@fhevm/sdk@1.1.0-alpha.8` source (via `npm pack`, not just `.d.ts`) and `viem`'s actual transport implementation, rather than taking my own earlier characterization at face value. The conclusion changed materially once I checked what's actually reachable, so writing it up precisely here.

### What's actually going on

`FhevmRelayer`'s three capability clients (`#base`/`#decrypt`/`#encrypt`) each lazily call `.init()` with no arguments — that's correct, because `@fhevm/sdk`'s `init()` is a genuinely zero-argument method (`CoreFhevm-p.ts`). The question is what runs *inside* that init chain, and whether any of it is a real, unbounded risk.

I traced all three network-shaped operations that could happen during init:

1. **The ~50MB FHE encryption key fetch** — already bounded. `#initEncrypt` explicitly pre-fetches it with `{ options: this.#defaultOptions }` (which carries the configured `timeout`) before calling the underlying `init()`. No issue here.

2. **WASM module loading (TFHE/TKMS)** — turns out this is *not* a network risk in the default configuration, and I was wrong to originally flag it as one. Per `init-p.ts`'s own contract comment: *"If the WASM asset has no URL: compile from embedded base64. `wasmAssetLoadMode` does not control WASM module loading."* Without a custom `runtime.locateFile` (which this SDK never sets — confirmed via `grep locateFile` across the whole `packages/sdk/src` tree, zero hits), asset resolution falls to:
   - **Node**: an auto-derived `file://` URL, read via `fs.readFile()` — local disk, not network.
   - **Browser**: the embedded base64 payload, decoded locally (`Buffer.from`/`fetch("data:...")`/`atob` — none of these are real network requests, even the data-URI `fetch()`).

   This only becomes a genuine unbounded-fetch risk if a consumer explicitly configures `runtime.locateFile` to return an `http(s)://` URL — an advanced, opt-in customization, not something reachable by default. I'd drop this from the concern list unless we want to caveat that specific advanced path.

3. **Protocol-version resolution** (`ensureResolvedProtocolVersion` → `getHostContractVersion` → `readContract`) — this is the one that's real, but the severity depends entirely on which transport backs the `publicClient`:

   ```ts
   // fhevm-relayer.ts constructor
   publicClient: createPublicClient({
     transport: typeof this.#chain.network === "string"
       ? http(this.#chain.network)
       : custom(this.#chain.network),
   }),
   ```

   - **`http()` transport** (every built-in chain preset — `mainnet`/`sepolia`/`hoodi`/`ingen`/`hardhat` all configure `network` as a plain URL string) — viem's own `http.ts` source: `timeout = timeout_ ?? config.timeout ?? 10_000`, plus a default `retryCount = 3` that *does* retry on `TimeoutError` (confirmed via `shouldRetry`'s fallthrough `return true`). So worst case here is roughly **~30–40s**, not indefinite — real, and not configurable via `RelayerOptions.timeout`, but bounded.
   - **`custom()` transport** (an injected EIP-1193 provider used in place of a dedicated RPC URL — not something any built-in preset does, but a supported, documented option) — `CustomTransportConfig` has **no `timeout` field at all**, and delegates straight to `provider.request.bind(provider)`. This is the one case where "hangs indefinitely" is literally accurate, since it depends entirely on the wallet/provider's own behavior.

### So, precisely, what's the actual bug

Not "init can hang forever" as a blanket statement — that's only true for the less-common `custom()`-transport configuration. What *is* true across the board: `RelayerOptions.timeout`'s own JSDoc says it *"bounds the whole async operation... not a single HTTP call"* — and that's not accurate. For the common `http()` case, a caller who sets e.g. `timeout: 500` expecting that to bound the *entire* operation is still exposed to up to ~30–40s of unconfigurable delay during a cold-start init before their own bound ever applies to the relayer request itself. For the `custom()` case, there's no bound at all, contradicting the JSDoc outright.

Root cause: `ensureResolvedProtocolVersion(fhevm: unknown)` doesn't accept an options parameter anywhere in its signature or the chain beneath it (`resolveProtocolContext` → `getHostContractVersion`) — unlike `fetchFheEncryptionKeyBytes`, which does. There's no equivalent to `moduleVersions`' pre-supply escape hatch for the protocol version either, so there's no way to skip this RPC call structurally, even with advanced config.

### Suggested fixes, roughly in order of preference/effort

**A. Upstream ask on `@fhevm/sdk`.** The cleanest fix: have `ensureResolvedProtocolVersion`/`getHostContractVersion` accept `options?: { timeout?; signal? }` threaded down into the underlying `readContract` call, symmetric with what already exists for `fetchFheEncryptionKeyBytes`. Out of this PR's direct control, but worth a tracked issue against `@fhevm/sdk` regardless of what we do locally.

**B. Configure a shorter viem timeout at the `publicClient` level, derived from `RelayerOptions.timeout`.** Concretely:
```ts
transport: typeof this.#chain.network === "string"
  ? http(this.#chain.network, { timeout: this.#defaultOptions.timeout, retryCount: /* … */ })
  : custom(this.#chain.network),
```
Doesn't make the JSDoc's "bounds the whole operation" claim fully true (init and the relayer request would still be two separate budgets, not one coherent one), but it closes almost all of the practical gap for the common `http()` path — the pre-relayer-request wait shrinks from viem's unconfigurable ~30–40s default down to whatever the caller actually configured. Cheapest fix that meaningfully helps, and doesn't require an upstream change. Doesn't touch the `custom()` case.

**C. `Promise.race` the SDK's own `init()` calls against `AbortSignal.timeout(...)`/the configured `signal`.** Covers *every* transport, including `custom()` — the only option on this list that actually addresses the genuinely-unbounded case. Caveat worth being explicit about: this can't cancel the underlying operation — `@fhevm/sdk` memoizes its own init promise internally, so a "timed out" init here means the SDK gives up waiting while the real request keeps running in the background and will eventually settle on its own. That's still strictly better than hanging the caller forever, but it's a client-side abandon, not a true cancellation, and worth documenting as such rather than presenting as a clean fix.

**D. Fix the JSDoc regardless of what else lands.** At minimum, `RelayerOptions.timeout`'s contract should say precisely what's true today: excludes the init phase; on a standard `http()`-backed chain that phase is bounded by viem's own internal default (~10s × up to 3 retries, not configurable via this option); on a `custom()`-backed chain (injected provider) it's unbounded. Zero-cost relative to A–C, and prevents the documented contract from actively misleading someone who reads it.

My inclination would be **B + D as the immediate, low-effort fix** (closes the common case, keeps the docs honest), with **A filed as a tracked upstream issue** since it's the only complete fix for the `custom()` transport case. Happy to be argued out of any of this if there's context I'm missing on why `custom()`-transport usage isn't a real-world concern here.

PR #477, opened 2026-06-23
chore(deps): batch dependency updates

ankurdotb:
## Summary

- `@zama-fhe/relayer-sdk` root dev dep `0.4.3` → `~0.4.4` (aligns with `packages/sdk` which already had `~0.4.4`; only used at root for playwright E2E fixtures)
- `viem` `2.52.2` → `2.53.1` (test apps + react-sdk dev; `packages/sdk` peer dep `>=2` unchanged)
- `wagmi` `3.6.16` → `3.6.17`
- `rolldown` `1.1.1` → `1.1.2`, `rolldown-plugin-dts` `0.25.2` → `0.26.0` (no api-report drift)
- `oxfmt` `0.54.0` → `0.55.0`
- `vitest` + `@vitest/*` `4.1.8` → `4.1.9`
- `oxlint` `1.69.0` → `1.70.0`
- `ethers` `6.16.0` → `6.17.0`
- `happy-dom` `20.10.3` → `20.10.6`

**Intentionally skipped:**
- `@types/node` `22.x` → `26.0.0` — major, staying on `^22`
- `react-router` `7.x` → `8.0.1` — major
- `nanoid` — pnpm trust-downgrade blocks `3.3.14` (manual publish, no OIDC); `3.3.15` (OIDC-signed) clears the 3-day `minimumReleaseAge` gate on 2026-06-24; not a supply chain issue

## Test plan
- [x] typecheck passes locally (verified)
- [x] lint passes (verified via pre-commit hook)
- [x] api-report: no drift from rolldown-plugin-dts 0.26.0 (verified locally)
- [x] CI green


packages/react-sdk/package.json
```diff
@@ -58,7 +58,7 @@
   },
   "devDependencies": {
     "@zama-fhe/sdk": "workspace:*",
-    "viem": "^2.52.2"
+    "viem": ">=2"
```
ghermet: I think this the other way around: `viem` version needs to be flexible as a prod dependency on `sdk` but can be a fixed version as a dev dependency on `react-sdk`

package.json
```diff
@@ -83,21 +83,21 @@
     "@vitest/coverage-v8": "^4.1.9",
     "@vitest/runner": "^4.1.9",
     "@vitest/ui": "^4.1.9",
-    "@zama-fhe/relayer-sdk": "0.4.3",
+    "@zama-fhe/relayer-sdk": "~0.4.4",
```
ghermet, line 87: I'm not sure if we need this dependency at the root? It is only used in the `sdk` package

packages/react-sdk/package.json
ankurdotb: Confirmed and fixed the prod side. `packages/sdk` dependencies `viem`/`zod` had quietly regressed from the long-standing broad `>=2`/`>=4` to `^2.52.2`/`^4.4.3` at a prerelease merge, and the batch update carried viem to `^2.53.1`. Restored both to `>=2`/`>=4` (995458e1 → rebased) so the published SDK stays flexible for consumers — peerDependencies.viem was already `>=2`.

On the react-sdk dev side: keeping it at `>=2` rather than pinning. The specific version only crept in via a Dependabot bump, and a broad range avoids that recurring churn; lockfile still resolves the latest 2.x in-workspace either way.

package.json
ankurdotb: Dug into it. It is not actually unused — it was anchoring a peer dependency. `@fhevm/mock-utils@0.4.2` (the e2e mock, latest published) hard-pins its `@zama-fhe/relayer-sdk` peer to a stale `0.4.1`. The root `~0.4.4` was silently force-upgrading that peer to `0.4.4` so the node e2e mock tests against the same relayer version the SDK ships. Removing it outright downgrades the mock to `0.4.1`; a `pnpm.overrides` entry is worse (produces dual `0.4.1` + `0.4.4` mock-utils copies).

So in 1b298adf I moved both `@zama-fhe/relayer-sdk` and `@fhevm/mock-utils` out of the root and into `test/playwright`, where the mock worker actually consumes them — keeping the `~0.4.4` pin. Root is now clean, lockfile resolves a single `mock-utils@0.4.2(relayer-sdk@0.4.4)`, typecheck green. Durable follow-up (handled separately): ask devex to bump mock-utils' peer to `0.4.4` so the pin can be dropped.

packages/sdk/package.json
```diff
@@ -160,7 +160,7 @@
   },
   "dependencies": {
     "@zama-fhe/relayer-sdk": "~0.4.4",
-    "viem": "^2.52.2",
+    "viem": "^2.53.1",
```
ghermet:
```suggestion
    "viem": ">=2",
```

packages/react-sdk/package.json
ghermet:
```suggestion
    "viem": "^2.53.1"
```

ankurdotb: Applied in 306c165b — agreed, devDeps aren't published so pinning is fine, and it matches every other devDep in the repo. Thanks, you were right; my >=2 was the odd one out.

packages/sdk/package.json
ankurdotb: Done in 995458e1 — restored to >=2 (and zod to >=4). These had regressed to ^ at a prerelease merge; now flexible again for the published package.

ghermet (dismissed):

ghermet (approved):

PR #482, opened 2026-06-24
fix(sdk): distinct decryption error causes — not-entitled vs RPC rate-limit [SDK-239]

zama-cremaud:
## Summary

Closes [SDK-239](https://linear.app/zama/issue/SDK-239). Decryption failures previously collapsed two unrelated causes into `DECRYPTION_FAILED`, forcing entitlement-/retry-aware consumers (server-side indexers, AI codegen) to pre-check on-chain out of band or string-match error messages. This adds two distinct, accurate, typed causes so consumers can branch deterministically.

Both causes are **classified at the worker source** (where the full error object still exists) and threaded through the worker `ErrorResponse`, then rebuilt into typed errors on the main thread by `wrapDecryptError` — structured-clone across the worker boundary otherwise strips `code`/`status`/`cause`, leaving only the message.

### `NotEntitledError` (`NOT_ENTITLED`, non-retryable)
- Carries `encryptedValue` / `contractAddress` / `account`.
- **Derived from the relayer SDK's own authoritative ACL check** (`validateAclPermissions`), which already runs `persistAllowed` before every decrypt and throws on denial — `classifyDecryptWorkerError` recognizes that failure and emits `NOT_ENTITLED`. **Zero extra on-chain reads, no cross-provider race, complete coverage.**
- The matcher is a deliberate bridge against the pinned `@zama-fhe/relayer-sdk`; guard tests read the installed dependency source and fail loudly if the not-entitled message **or** the `RELAYER_FETCH_ERROR` tag drifts, plus a check that the worker's pinned CDN `RELAYER_SDK_VERSION` matches the installed npm version. A `TODO` tracks keying off the relayer's typed `ACLUserDecryptionError` once its decrypt path surfaces it.
- The rarer "dapp contract not authorized" case is a dapp misconfig and stays `DecryptionFailedError` (documented).

### `RpcRateLimitError` (`RPC_RATE_LIMITED`, retryable, `retryAfter?`)
- The **consumer's RPC provider** throttling (HTTP 429 / JSON-RPC `-32005`) during the SDK's on-chain reads (worker decrypt path + delegation pre-validation + cleartext relayer).
- The **relayer's own 429 stays `RelayerRequestFailedError`** (excluded via the `RELAYER_FETCH_ERROR` tag; that's SDK-236). `hasStructuredRpcRateLimitSignal` promotes a raw `-32005`/`status:429` regardless of HTTP status, while worker-origin relayer 429s (top-level `statusCode`) stay relayer errors.
- Treated as a fatal-batch error so batch decrypt aborts once instead of amplifying the throttle with per-item retries (a shared abort flag also short-circuits the per-item fallback under concurrency).

Both new codes are registered in `matchZamaError`'s per-subclass type map (added after merging #488), so `NOT_ENTITLED` / `RPC_RATE_LIMITED` handlers receive `NotEntitledError` / `RpcRateLimitError` without a cast.

> **Design note:** an earlier revision used an eager main-thread `persistAllowed` pre-check; it was dropped after review because it duplicated the relayer's check (+50% on-chain reads, feeding problem #2 for the indexer persona) and could emit a false `NotEntitledError` under cross-provider chain-tip skew. The current design reuses the `classifyWorkerError`/`errorCode` machinery instead.

## Test plan
- [x] `pnpm --filter @zama-fhe/sdk test` green — units for `classifyDecryptWorkerError`, `wrapDecryptError` mapping, rate-limit taxonomy, delegation-validate rate-limit, batch abort, worker round-trip, cleartext `NotEntitledError`, the `matchZamaError` handler typing, and the real-dependency drift guards (not-entitled message + `RELAYER_FETCH_ERROR` tag + pinned CDN version).
- [x] `tsc` / `oxlint` / `oxfmt --check` / `size-limit` clean; api-extractor reports + `llms-full.txt` regenerated. CI green.

## Notes
- Base branch: `prerelease`. Public API: additive only (`NotEntitledError`, `RpcRateLimitError`, codes; both narrowed by `matchZamaError`).
- The header-derived `retryAfter` for consumer 429s (viem/Cloudflare deliver it only as a `Retry-After` header) is handled in the stacked follow-up #489, which consolidates the classifier on the main thread.


packages/sdk/etc/sdk-node.api.md
```diff
@@ -294,8 +294,15 @@ export type EncryptResult = {
...
 export interface ErrorResponse extends BaseResponse {
+    // (undocumented)
+    account?: string;
+    // (undocumented)
+    contractAddress?: string;
     // (undocumented)
     error: string;
+    errorCode?: string;
+    handle?: string;
+    retryAfter?: number;
```
ghermet: These fields should ideally be declared in their respective error classes not here

ghermet: or at least create a discriminated union with the error code

packages/sdk/src/errors/decrypt.ts
```diff
@@ -1,10 +1,44 @@
...
-import { extractHttpStatus } from "../utils/error";
+import {
+  extractHttpStatus,
+  extractRetryAfterMs,
+  hasStructuredRpcRateLimitSignal,
+  isRpcRateLimitError,
+} from "../utils/error";
+
+/** Read the `zamaErrorCode` the worker client attaches to cross-thread errors. */
+function readZamaErrorCode(error: unknown): string | undefined {
```
ghermet: this could be a function to narrow down unknown errors to typed ones:
```ts
function isZamaError<T extends ZamaError>(
  error: unknown,
  zamaError: T,
): error is T {
  try {
    assertObject(error, "error");
    return error instanceof Error && error.code === zamaError.code;
  } catch {
    return false;
  }
}
```

```ts
 // Provider RPC rate-limit, in priority order:
  // 1. classified at the worker source (`zamaErrorCode`) — authoritative;
  // 2. an unambiguous structured signal (-32005 / `status: 429`) on a raw
  //    main-thread error, promoted regardless of any HTTP status it carries.
  // Worker-origin relayer 429s carry a top-level `statusCode` (not `status`) and
  // are intentionally excluded here, staying RelayerRequestFailedError below.
  if (isZamaError(error, RpcRateLimitedError)) {
    return new RpcRateLimitError(error.message ?? fallbackMessage, {
      cause: error,
      retryAfter: error.retryAfter ?? extractRetryAfterMs(error),
    });
  }
```

packages/sdk/src/errors/entitlement.ts
```diff
@@ -0,0 +1,44 @@
...
+ * } catch (e) {
+ *   if (e instanceof NotEntitledError) {
+ *     // don't retry — wait for an ACL grant / backfill
+ *   }
+ * }
+ * ```
+ */
+export class NotEntitledError extends ZamaError {
+  /** The handle (encrypted value) the actor is not entitled to decrypt. */
+  readonly handle: string;
```
ghermet: [AUTOMATED] **Should fix (public-surface naming):** `NotEntitledError` is `@public` (exported from `src/index.ts`, present in `sdk.api.md`), so its `readonly handle` property is on the public glossary-governed surface. conventions.md:9 is explicit:

> **"Encrypted value", not "handle", on public surfaces.** Public APIs and docs use `encryptedValue(s)` / `EncryptedValue` … Keep `handle` only where it mirrors an on-chain term (contract ABI fragments, `isHandleDelegatedForUserDecryption`, KMS `ctHandles`).

An error property isn't one of the listed exceptions, and a pre-PR grep shows no other public surface uses `handle` (only ABI fragments, test fixtures, and worker internals — all allowed). Suggest renaming the public property + constructor arg to `encryptedValue` (and the docs example at `errors.md:317`). The doc comment already calls it "(encrypted value)", which confirms the intent. Note: the *internal* worker payload (`WorkerErrorClassification.handle`, `classifyDecryptWorkerError`, `readWorkerClassification`) may keep `handle` per the same rule's internal-layer carve-out — only the public class needs the glossary name.

Confidence: 88/100

packages/sdk/src/errors/fatal-batch.ts
```diff
@@ -15,6 +18,7 @@ export function isFatalBatchError(error: unknown): boolean {
   return (
     error instanceof SigningRejectedError ||
     error instanceof SigningFailedError ||
-    error instanceof ConfigurationError
+    error instanceof ConfigurationError ||
+    error instanceof RpcRateLimitError
```
ghermet, line 22: [AUTOMATED] **Consider (non-blocking):** Adding `RpcRateLimitError` here is correct, and the bulk-first batch attempt does abort once on a throttle (the common path). But the per-item fallback loop in `decryption-service.ts` runs through `pLimit`, and `pLimit` has no cancellation: when one worker rethrows the fatal error, `Promise.all` rejects, but the sibling workers keep draining the remaining items (and with `maxConcurrency >= items.length` they're all already in-flight via the `Promise.all` fast-path). So in the per-item fallback the "abort once instead of re-hitting the throttled endpoint" property doesn't fully hold — up to N extra calls can still fire.

Impact is bounded (one pass, no per-item retries) so this isn't a correctness bug, just a gap vs. the stated rationale, and there's no test exercising fatal short-circuit under concurrency. If you want the property to actually hold, a shared `aborted` flag checked at the top of the `pLimit` worker loop would do it.

Confidence: 78/100

packages/sdk/src/utils/error.ts
```diff
@@ -34,6 +36,225 @@ export function isContractCallError(error: unknown): boolean {
...
+}
+
+/**
+ * Best-effort extraction of a `Retry-After` delay (in milliseconds) from a
+ * provider error, when present. Returns `undefined` when no hint is available.
+ */
+export function extractRetryAfterMs(error: unknown): number | undefined {
+  const node = findInErrorChain(error, (n) => {
+    const v = n.retryAfter ?? n.retryAfterMs;
+    return typeof v === "number" && Number.isFinite(v);
```
ghermet: [AUTOMATED] **Nit (non-blocking):** the `retryAfter` predicate guards `Number.isFinite(v)` but not `v > 0`, so a provider sending `retryAfter: -1` or `0` passes through and surfaces as `RpcRateLimitError.retryAfter = -1000` / `0`. `rpc.ts` documents `undefined` as "unknown", so a negative/zero value is misleading — a consumer doing `setTimeout(retry, retryAfter)` would retry immediately, defeating the back-off. Consider tightening the predicate to `typeof v === "number" && Number.isFinite(v) && v > 0`.

Confidence: 76/100

ghermet (commented):
[AUTOMATED]

## Code Review — SDK-239 typed decryption error causes

Reviewed with 4 parallel agents (2× bug hunter, guideline compliance, error-handling auditor), then independently validated every finding ≥80 confidence against the code.

**The design is sound and the implementation is careful.** Classifying at the worker source and threading a small typed payload across the structured-clone boundary (because `code`/`status`/`cause` are stripped) is the right call, and the tricky bits hold up under scrutiny:
- The relayer-429 exclusion (`RELAYER_FETCH_ERROR` tag) is honored on both the worker path (`isRelayerFetchError` runs on the live error before the boundary) and the main-thread path (`hasStructuredRpcRateLimitSignal` keys on `status`, not the worker's top-level `statusCode`). Relayer 429s correctly stay `RelayerRequestFailedError`.
- The SDK-241 `isDelegated && statusCode === 500 → DelegationNotPropagatedError` mapping is preserved — the new rate-limit branches sit above it and a 500 isn't a rate-limit signal, so it falls through correctly.
- `retryAfter` is converted seconds→ms once at the worker source and threaded as ms; no double-conversion.

One should-fix (public-surface naming) and two non-blocking notes below.

**Considered and dropped** (low signal / speculative): `extractHttpStatus` is depth-1 while `findInErrorChain` is depth-6 — a real asymmetry, but the only impact path (deeply-nested status on a main-thread cleartext-relayer error) relies on unverified error shapes. The `error.test.ts` drift guard reads `@zama-fhe/relayer-sdk/node` while the web worker classifies errors from the CDN bundle build — same source tree today, but a bundle-only message drift wouldn't be caught; worth keeping in mind for the SDK-239 follow-up that keys off the typed `ACLUserDecryptionError`.

Nice work — tests are thorough and the design note in the PR body answered most of my "why not X" questions before I asked them.

ghermet:
[AUTOMATED]

## Follow-up suggestion (non-blocking): classify once, on the main thread

Not for this PR — it's correct and well-tested, ship it. Capturing this next to the existing `TODO(SDK-239 follow-up)` because it dovetails with the planned "key off the relayer's typed `ACLUserDecryptionError`" change.

### The smell

Classification currently lives in **two** places:
- **Worker-origin** errors → `classifyDecryptWorkerError` decides `NOT_ENTITLED` / `RPC_RATE_LIMITED` at the source.
- **Main-thread-origin** errors (cleartext relayer, delegation pre-check) → `wrapDecryptError` + `hasStructuredRpcRateLimitSignal` / `extractHttpStatus` decide the same things again, with different code.

Two paths that must agree but don't is what produced the `extractHttpStatus` (depth-1) vs `findInErrorChain` (depth-6) asymmetry — that's a *class* of drift bug, not a one-off.

Note the transport is already "serialize fields, rebuild on the other side": `handleResponse` rebuilds a plain `Error` and re-attaches `statusCode`/`retryAfter`/`handle`/`zamaErrorCode` (`worker.base-client.ts:177-198`). The only question is *where the semantic decision happens*.

### The shape

**Worker — mechanical, taxonomy-agnostic.** Replace source classification with a generic serializer that flattens the error *including its cause chain* (the one thing that genuinely must happen at the source — structured-clone destroys `cause`):

```ts
function serializeError(error: unknown, depth = 6): SerializedError {
  const e = toError(error);
  return {
    name: e.name,
    message: e.message,
    code: pick(e, "code"),         // -32005, RELAYER_FETCH_ERROR, CALL_EXCEPTION…
    status: pick(e, "status"),
    statusCode: pick(e, "statusCode"),
    retryAfter: pick(e, "retryAfter"),
    cause: depth > 0 ? serializeCauseChain(e, depth - 1) : undefined, // cause/error/info
  };
}
```

The worker stops importing `NotEntitledError` / `RpcRateLimitError` entirely.

**Main thread — the single existing classifier.** Teach `findInErrorChain` to walk a serialized `cause` array as well as live `cause`/`error`/`info`, then run `wrapDecryptError` on both origins:

```ts
const err = deserialize(response.error);
return wrapDecryptError(err, msg, { isDelegated, contractAddress, account });
```

`NOT_ENTITLED` needs nothing special from the worker: the relayer **message survives clone**, the handle is parseable from it, and `contractAddress` / `account` are **request context the calling main thread already holds** — inject them at wrap time instead of threading them across the boundary.

### Payoff

1. **One classification path** → the depth-asymmetry bug (and its whole family) disappears by construction.
2. **Worker decoupled from the taxonomy** → the planned switch to the relayer's typed `ACLUserDecryptionError` becomes a main-thread-only edit instead of both-sides.
3. **Reusable envelope** → `serializeError` serves every worker request type, not just decrypt.

Cost: define a `SerializedError` format + make `findInErrorChain` walk the serialized chain. Mostly mechanical, and you touch the worker boundary only once if it's done together with the relayer-typed-error change.

ghermet:
@zama-cremaud FYI this touches on the worker part which is removed in #458

packages/sdk/src/utils/__tests__/error.test.ts
```diff
@@ -140,3 +153,198 @@ describe("extractHttpStatus", () => {
...
+    });
+  });
+
+  // Drift guard wired to the REAL dependency (not a hand-copied constant): if a
+  // @zama-fhe/relayer-sdk bump reworffds its not-entitled message, the substrings
+  // isNotEntitledMessage keys on disappear here and this fails loudly — otherwise
+  // NOT_ENTITLED would silently downgrade to DECRYPTION_FAILED (the SDK-239 bug).
+  test("the installed @zama-fhe/relayer-sdk still emits the message our matcher keys on", () => {
+    const require = createRequire(import.meta.url);
+    const source = readFileSync(require.resolve("@zama-fhe/relayer-sdk/node"), "utf8");
```
ankurdotb, line 346: [AUTOMATED] **Consider** — this drift guard reads the **Node** build (`@zama-fhe/relayer-sdk/node`), but the browser worker (`relayer-sdk.worker.ts`) doesn't use the npm package at runtime — it `importScripts` the CDN UMD bundle (`relayer-web.ts`: `cdn.zama.org/relayer-sdk-js/${RELAYER_SDK_VERSION}/...`, with `RELAYER_SDK_VERSION` hardcoded to `"0.4.4"`, decoupled from the `~0.4.4` npm range). So browser-side `NOT_ENTITLED` accuracy hinges on the *CDN* message matching, which this guard never checks: a reworded CDN message — or a `RELAYER_SDK_VERSION` bump not mirrored in npm — silently downgrades `NOT_ENTITLED` → `DECRYPTION_FAILED` in the browser while this test stays green. Consider asserting the message against the web/UMD artifact too, or gating `RELAYER_SDK_VERSION` bumps on re-validating the matcher.

Confidence: 90/100

packages/sdk/src/utils/__tests__/error.test.ts
```diff
@@ -140,3 +153,198 @@ describe("extractHttpStatus", () => {
...
+    expect(source).toContain("is not authorized to user decrypt");
+    expect(source.toLowerCase()).toContain("user address");
```
ankurdotb: [AUTOMATED] **Nit** — these two assertions check the substrings **independently**, anywhere in the bundle, but `isNotEntitledMessage` requires both in the *same* message string. The relayer ships a sibling "dapp contract … is not authorized to user decrypt" message and uses "user address" in unrelated strings, so a future rewording of *only* the actor message would keep both substrings present (guard green) while breaking the matcher's same-string conjunction — the exact silent downgrade this guard exists to catch. Consider asserting one contiguous pattern, e.g. `/User address .* is not authorized to user decrypt handle/`.

Confidence: 85/100

packages/sdk/src/utils/error.ts
```diff
@@ -34,6 +36,228 @@ export function isContractCallError(error: unknown): boolean {
...
+  return false;
+}
+
+/**
+ * The relayer SDK tags its HTTP errors with `code: "RELAYER_FETCH_ERROR"`. These
+ * stay the relayer's domain ({@link RelayerRequestFailedError} / SDK-236), never
+ * {@link RpcRateLimitError}, even on a 429.
+ */
+function isRelayerFetchError(error: unknown): boolean {
+  return findInErrorChain(error, (node) => node.code === "RELAYER_FETCH_ERROR") !== undefined;
```
ankurdotb, line 102: [AUTOMATED] **Consider** — `RELAYER_FETCH_ERROR` is the *sole* discriminator keeping the relayer's own 429 out of the new retryable `RpcRateLimitError` (the relayer's 429 message literally contains "rate limit"). Unlike the not-entitled message, this tag has **no drift guard** — it's asserted only against hand-written fixtures, not the installed relayer source. If a relayer bump renames this code, the relayer's throttle flips from non-retryable `RelayerRequestFailedError` to retryable `RpcRateLimitError`, and consumer retry loops amplify the throttle — the exact harm the fatal-batch abort is meant to prevent. Consider a guard mirroring the not-entitled one, or keying off `cause.status` + relayer domain rather than the message.

Confidence: 85/100

packages/sdk/src/errors/base.ts
```diff
@@ -32,6 +32,10 @@ export const ZamaErrorCode = {
   NoCiphertext: "NO_CIPHERTEXT",
   /** Relayer HTTP request failed. */
   RelayerRequestFailed: "RELAYER_REQUEST_FAILED",
+  /** The configured signer/account is not entitled (ACL) to decrypt this handle. Don't retry — wait for a grant. */
```
ankurdotb: [AUTOMATED] **Nit** — `docs/agents/conventions.md`: *"'Encrypted value', not 'handle', on public surfaces. Public APIs and docs use `encryptedValue(s)` / `EncryptedValue`…"*. This JSDoc on the `@public` `ZamaErrorCode.NotEntitled` says "decrypt this **handle**." `NotEntitledError` itself already uses `encryptedValue` correctly — just this doc line is out of step. s/handle/encrypted value/.

Confidence: 85/100

ankurdotb (commented):
[AUTOMATED]

## Code review — distinct decryption error causes (SDK-239)

Strong, well-tested, additive PR. Two independent bug-hunt passes plus a manual trace found **no objective runtime bugs** — the classification logic, the worker round-trip field preservation, the fatal-batch ordering, and the not-entitled / rate-limit precedence are all sound. The retryable-vs-terminal classification (the core SDK-239 deliverable) works for real provider 429s.

**Verdict: Comment** — nothing here blocks. The notes below harden the *safety net* around the deliberately-temporary `@zama-fhe/relayer-sdk` string-matching bridge (which the code already flags as a bridge via its `TODO`), plus one doc nit.

**Tracked elsewhere (not re-raised here):** `RpcRateLimitError.retryAfter` is effectively `undefined` for real edge/Cloudflare 429s — the `Retry-After` header is never read. Filed as a change request on the stacked #489, which is the right home for it (the consolidated main-thread classifier).

Confidence on the inline items: 85–90/100.

packages/sdk/src/errors/decrypt.ts
```diff
@@ -30,12 +38,55 @@ export function wrapDecryptError(
...
   ) {
     return error;
   }
 
-  const statusCode = extractHttpStatus(error);
+  // Causes classified at the worker source and threaded across the boundary
+  // (structured clone otherwise leaves only the message). Discriminated on
+  // `errorCode`, so each branch gets exactly its own typed payload.
+  const classification = readWorkerClassification(error);
+  if (classification?.errorCode === ZamaErrorCode.NotEntitled) {
```
ankurdotb, line 52: [AUTOMATED] **Edge case (overlap):** this terminal `NotEntitled` branch is evaluated before the `isDelegated && statusCode === 500 → DelegationNotPropagated` branch below (L98), and `NotEntitled` is matched message-first in `classifyDecryptWorkerError` from the relayer's `persistAllowed` L1 read (no staleness tolerance). On the delegated path, a delegator whose grant just landed — or a lagging RPC — gives `persistAllowed=false` → terminal `NotEntitled` ("don't retry") for what's really a transient propagation/lag window, the opposite of the delegated-500 handling. Consider treating delegator-not-entitled as transient when `isDelegated`. Full rationale in the review summary.

packages/sdk/src/utils/error.ts
```diff
@@ -34,6 +36,224 @@ export function isContractCallError(error: unknown): boolean {
...
+      if (found) {
+        return found;
+      }
+    }
+  }
+  return undefined;
+}
+
+/** Structured (unambiguous) rate-limit signal: JSON-RPC -32005 or a numeric HTTP 429. */
+function nodeHasStructuredRateLimit(node: Record<string, unknown>): boolean {
```
ankurdotb, line 76: [AUTOMATED] **Detection gap (two real rate-limit shapes missed):** (1) ethers throws a 429 as `code: "SERVER_ERROR"` with `info.responseStatus: "429 …"` (a string) and no numeric status — `=== 429` won't match, so a retryable throttle on the worker's ethers L1 reads falls through to `DecryptionFailedError`. (2) viem's own `shouldRetry` treats a JSON-RPC `code === 429` as a rate-limit, but this only checks `-32005` / `status === 429`. Consider `|| node.code === 429` plus parsing the leading code out of `info.responseStatus`.

ankurdotb (commented):
[AUTOMATED]

## Review — error-cause overlap (beyond the retryAfter timing)

A closer look at 482 through the lens of *cause overlap*: where can two distinct typed errors arise from one real condition, and does the classifier's precedence pick the wrong one? Audited the decrypt taxonomy against the installed `@zama-fhe/relayer-sdk@0.4.4` + ethers source and the tests. One substantive overlap, one transport divergence, one detection gap — none blocking. **Since #489 consolidates the classifier, the fixes may sit better there than here — your call on 482 vs 489.**

**1. `NotEntitledError` (terminal) can shadow `DelegationNotPropagated` / `DelegationNotFound` (transient) on the delegated path.** `NotEntitled` is matched message-first (`classifyDecryptWorkerError`) from the relayer's `validateAclPermissions` → `persistAllowed(account)` — an **L1 `eth_call` with no staleness tolerance**, thrown *before* the relayer HTTP call. That read returns `false` **transiently** when the consumer's RPC serves a stale block (lagging chain head) or the grant/delegation just landed → terminal `NotEntitledError` ("don't retry — needs an on-chain grant") for what is really a 1–2 min propagation/lag window. Note the asymmetry: a delegated **gateway-sync** lag returns a 500 → `DelegationNotPropagated` ("wait & retry") ✅, but a delegated **L1-read** lag on the delegator's grant → terminal `NotEntitled` ("give up forever") — both are timing. (The genuine gateway-not-synced case where `persistAllowed` passes then the POST 500s classifies correctly; the gap is the L1-read-returns-false-transiently case.)
- *Consider:* when `isDelegated`, treat a delegator `persistAllowed=false` as transient (`DelegationNotPropagated`/retryable) rather than terminal — the call site already knows `isDelegated`. And/or document that `NotEntitled` can be a false-terminal under RPC staleness so indexers confirm against a fresh block before giving up.
- *Empirical check worth doing:* grant a delegation, immediately delegated-decrypt against a slightly-lagging RPC, confirm which error surfaces.

**2. Cleartext vs worker classification divergence (delegated path).** The relayer/worker path checks `persistAllowed(delegator)` → `NotEntitledError`. But the cleartext transport's `#assertDelegation` (`relayer-cleartext.ts`) only checks `isHandleDelegatedForUserDecryption` and throws generic `DecryptionFailedError` — it never checks `persistAllowed(delegator)`. So the same condition yields `NotEntitledError` in prod but `DecryptionFailedError` in local cleartext dev — exactly where devs wire up `instanceof NotEntitledError` handling. (The non-delegated cleartext path *does* throw `NotEntitled` at `relayer-cleartext.ts:434`, so this is delegated-only.) *Consider:* have `#assertDelegation` also check `persistAllowed(delegator)` → `NotEntitled`, mirroring the relayer's order.

**3. ethers HTTP 429 with a non-standard reason phrase → `DecryptionFailed` (retry signal lost).** The worker's L1 reads use ethers; on a 429 ethers throws `code: "SERVER_ERROR"` with `info.responseStatus: "429 …"` (a **string**) and no numeric top-level status. If the phrase isn't "too many requests"/"rate limit", both the structured check (string ≠ numeric `429`) and the message fallback miss → generic `DecryptionFailed`, losing the retryable `RpcRateLimitError`. *Consider:* parse the leading code from `info.responseStatus` (`/^(\d{3})/`) or treat `code === "SERVER_ERROR"` + 429; also add `node.code === 429` (viem's own `shouldRetry` treats JSON-RPC `code === 429` as a rate-limit).

**Low / known:** the `isDelegated && 500 → DelegationNotPropagated` heuristic also shadows a genuine relayer-500 fault (acknowledged in its doc); a paused ACL that *reverts* on the `persistAllowed` view would land in `DecryptionFailed` (no `AclPaused` signal on the decrypt path) — needs an empirical check.

**What's solid (verified against source):** relayer-429 vs consumer-429 cleanly split by the `RELAYER_FETCH_ERROR` tag; the two-substring AND-gate blocks a generic "not authorized" 429 from a false `NotEntitled`; the dapp-contract variant correctly stays `DecryptionFailed`; `DelegationNotFound`/`Expired` caught by the L1 pre-check before the relayer (non-overlapping); `-32005` + viem `status:429` both caught.

Net: `NotEntitledError` earns its place — it's the only typed cause for the common non-delegated ACL denial. The things to tune are the **delegated terminal-vs-transient boundary (#1)** and **cleartext/worker parity (#2)**.

packages/sdk/src/errors/decrypt.ts
ankurdotb: [AUTOMATED] Downstream impact, for visibility: #486 (`sdk.delegations.isPropagated`, SDK-241) depends on the delegated propagation window surfacing as `DelegationNotPropagatedError`. Because `delegatedUserDecrypt` runs in the worker — where this `NotEntitled` classification applies — the overlap flagged here would make that probe **rethrow** instead of returning `false` during the sync window, breaking its documented poll loop. So the "treat delegator-not-entitled as transient when `isDelegated`" fix is effectively a prerequisite for #486, not just a hardening of this path. Full trace: https://github.com/zama-ai/sdk/pull/486#issuecomment-4841352236

PR #483, opened 2026-06-24
docs(sdk): steer integrators & LLMs to sdk.decryption over sdk.relayer [SDK-238]

ghermet:
## Why

Closes [SDK-238](https://linear.app/zama/issue/SDK-238).

There are two reachable ways to decrypt and nothing signalled which to prefer:

- **High-level (intended):** `sdk.decryption.decryptValues` / `delegatedDecryptValues` / `decryptPublicValues` — assembles the credential bundle (ephemeral keypair, EIP-712 permit) internally, caches, and wraps relayer errors.
- **Low-level:** `sdk.relayer` (`RelayerDispatcher`) → `userDecrypt` / `delegatedUserDecrypt` / `publicDecrypt`, where the caller assembles credentials by hand.

`sdk.relayer` carried **no JSDoc**, and neither path named the canonical one. A new integrator — or an LLM generating code — picked one at random or reached for the familiar low-level call, then fought manual credential assembly. The gap mattered doubly for AI consumers, which bake the guess into generated code.

## What

A "prefer this" steer at **every point of use**, plus the prose layer that feeds `llms.txt`:

- **`sdk.relayer`** — JSDoc escape-hatch note pointing at the high-level namespaces, and marked `@internal` so it drops out of the public API report (still reachable at runtime).
- **`FheOperations.userDecrypt` / `publicDecrypt` / `delegatedUserDecrypt`** — each now links to its high-level `sdk.decryption.*` equivalent via `{@link}`.
- **`Decryption` namespace JSDoc** — declared the canonical decryption path; names `sdk.relayer.*` as the escape hatch.
- **`encrypt-decrypt.md` guide** — hint box steering to the high-level path. `llms.txt` is generated from docs + README (not JSDoc/api-reports), so this is the only layer that reaches LLM consumers; confirmed the steer now appears in `llms-full.txt`.

Regenerated `etc/sdk*.api.md` and `llms-full.txt`.

## Out of scope

The `userDecrypt` → `decryptValues` **rename** of the low-level verbs (the naming-mismatch half of the issue) is handled on `feature/sdk-replace-relayer-with-fhevm-sdk`. This PR only adds the steer and keeps the verb names as `prerelease` currently has them.

## Verification

- `pnpm --filter @zama-fhe/sdk typecheck` ✅ (and all-package typecheck via pre-commit hook)
- `pnpm api-report:sdk` ✅ (`{@link}` refs all resolve — no unresolved-link warnings)
- `pnpm lint` (oxlint + ast-grep) ✅
- `pnpm llm:build` ✅ — steer present in corpus

No behavior change — docs/JSDoc and generated artifacts only.


docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -251,6 +251,10 @@ function ConfidentialAction() {
 
 ### 3. Decryption of the encrypted data
 
+{% hint style="info" %}
+**Use the high-level decryption path.** `useDecryptValues` (and its core-SDK equivalent `sdk.decryption.decryptValues`) is the canonical way to decrypt: it assembles the decryption credentials — ephemeral keypair and EIP-712 permit — for you, caches results, and wraps relayer errors. The low-level `sdk.relayer.userDecrypt`, which makes you build the credential bundle by hand, is an escape hatch — reach for it only when you genuinely need that control.
```
ankurdotb: [AUTOMATED] `ephemeral keypair` → `transport key pair`.

This keypair is persisted (IndexedDB, 30-day TTL) and reused across sessions, so it isn't ephemeral — "ephemeral" is the term the docs use for the *opposite* thing (the in-memory session signature). The canonical glossary term is **transport key pair** (`migrate-v2-to-v3.md:641`). This hint feeds `llms.txt`, so the wrong term lands in the LLM corpus too.

Confidence: 95/100

packages/sdk/src/namespaces/decryption.ts
```diff
@@ -8,9 +8,15 @@ import type { GenericProvider, GenericSigner } from "../types";
 import { requireAlignedWalletAccount } from "../utils/alignment";
 
 /**
- * Public namespace for FHE decryption.
+ * Public namespace for FHE decryption — the canonical way to decrypt.
  *
- * Exposed as `sdk.decryption`. Owns the SDK-level guards (signer requirement on
+ * Exposed as `sdk.decryption`. Prefer these methods over the low-level
+ * `sdk.relayer.userDecrypt` / `delegatedUserDecrypt` / `publicDecrypt`: they
+ * assemble the credential bundle (ephemeral keypair, EIP-712 permit) for you,
```
ankurdotb: [AUTOMATED] Same fix: `ephemeral keypair` → `transport key pair`. The transport key pair is persisted/reused (not ephemeral), and "transport key pair" is the canonical glossary term (`migrate-v2-to-v3.md:641`).

Confidence: 95/100

packages/sdk/src/relayer/relayer-sdk.ts
```diff
@@ -37,10 +37,22 @@ export interface FheOperations {
   /** Encrypt plaintext values into FHE ciphertexts. */
   encrypt(params: EncryptParams): Promise<EncryptResult>;
 
-  /** Decrypt FHE encrypted values using the user's own credentials. */
+  /**
+   * Decrypt FHE encrypted values using the user's own credentials.
+   *
+   * Low-level: the caller assembles the credential bundle (transport keypair,
```
ankurdotb: [AUTOMATED] Nit (same naming pass): `transport keypair` → `transport key pair` (three words), matching the glossary and line 26 of this file ("Generate a transport key pair"). Aligning all three sites avoids introducing a new one-word variant.

Confidence: 80/100

ankurdotb (changes requested):
[AUTOMATED]

## Code Review — correctness & consistency

Well-scoped steer, and the substantive claims check out: the `{@link}`s all resolve, "caches results" genuinely caches decrypted plaintext (not just credentials), and the error-wrapping claim is accurate. One blocking naming issue, plus two non-blocking notes.

**Blocking — `ephemeral keypair` is inaccurate and off-glossary.** The keypair the high-level path assembles is the **transport key pair**, which is *persisted* (IndexedDB, 30-day TTL) and reused across sessions — so "ephemeral" is precisely backwards (the docs reserve "ephemeral" for the in-memory session signature). The canonical term is **transport key pair**, per the migration-guide glossary (`docs/gitbook/src/guides/migrate-v2-to-v3.md:641` — "Key glossary: transport key pair & FHE encryption key"). See inline comments: it affects the namespace JSDoc, the guide hint (which feeds `llms.txt`), and the new one-word "transport keypair" this PR adds in `relayer-sdk.ts`.

**FYI (non-blocking):**
- The description says marking `sdk.relayer` `@internal` makes it "drop out of the public API report." It doesn't — there's no `stripInternal` (it breaks the dts build), so the regenerated `sdk.api.md` in this PR still lists `relayer`, just annotated `// @internal`. The tag itself is fine; only the stated *effect* is off. Flagging in case the intent was to hide it from the public surface.
- `createEIP712` and `createDelegatedUserDecryptEIP712` are also reachable low-level credential builders that didn't get the steer — a minor gap against "every point of use." Optional.

ankurdotb (approved):

PR #486, opened 2026-06-25
refactor(sdk): absorb the delegation-propagation window internally [SDK-241]

ghermet:
## What & why

Closes [SDK-241](https://linear.app/zama/issue/SDK-241/no-queryable-signal-for-whether-a-delegation-has-propagated-to-the).

A delegation grant is immediate on the host chain, but delegated decryption only works once the gateway syncs it cross-chain — usually within **~10 blocks (a few seconds)**. During that window a delegated decrypt fails with a transient `DelegationNotPropagatedError`, and until now callers had to wait/poll and retry themselves.

This makes the SDK **absorb the window internally**: a delegated decrypt issued right after granting simply waits for the sync to land. No new API to learn, no wait-loop for callers to write.

## What this adds on top of prerelease

**Bounded internal retry on the delegated-decrypt path.** When a delegated decrypt fails with the transient `DelegationNotPropagatedError`, the SDK retries until the delegation propagates, then returns the value.

```ts
// Just works, even immediately after granting — rides out the sync window.
const values = await sdk.decryption.delegatedDecryptValues(encryptedInputs, delegatorAddress);

// Opt out to fail fast on the first not-propagated response.
await sdk.decryption.delegatedDecryptValues(encryptedInputs, delegatorAddress, {
  waitForPropagation: false,
});
```

`waitForPropagation` (default `true`) is available on both `delegatedDecryptValues` and `delegatedBatchDecryptValues`. The only public-surface change is this new optional option — everything else is behavior and docs.

## How it works

- Fixed-interval polling across a bounded budget: **30 s budget / 2 s interval → 16 attempts**. No open-ended loop; deterministic under fake timers.
- Only the transient `DelegationNotPropagatedError` retries — a missing/expired grant, RPC rate-limit, or any other error throws immediately.
- **Batch:** the initial batch attempt rides out the window for the whole set. The per-item fallback then runs with `waitForPropagation: false`, so a still-unsynced delegation records one per-item error instead of each item re-spending the full 30 s budget. Fatal errors (e.g. RPC rate-limit) short-circuit the remaining queued items.

The SDK-239 / #489 error classifier (delegated ACL denial → transient `DelegationNotPropagatedError`) is unchanged and already in prerelease — this PR builds on it.

## Docs

`DelegationNotPropagatedError` now reads "you rarely need to handle this directly," and the timing is corrected throughout: the window is **~10 blocks (seconds)**, not "1–2 minutes", and callers no longer need to wait or poll manually.

## Testing

- Retries on transient not-propagated until success; `waitForPropagation: false` fails fast on the first response; batch per-item fallback fails fast (no per-item budget re-spend); non-transient errors rethrow immediately.
- Full suites green; typecheck, `api-report:check`, and `docs:check-links` all pass.


packages/sdk/src/services/decryption-service.ts
```diff
@@ -124,6 +130,82 @@ export class DecryptionService {
...
+            contractAddress,
+            ...resolveDelegatedDecryptPermit(credentials, contractAddress),
+            delegateAddress: normalizedDelegate,
+          });
+        }),
+        5,
+      );
+      return true;
+    } catch (error) {
+      const wrapped = wrapDecryptError(error, "Failed to probe delegation propagation", true);
```
ghermet: [AUTOMATED] **Consider (non-blocking):** `wrapDecryptError(error, …, true)` maps *every* relayer HTTP 500 to `DelegationNotPropagatedError`, so this `catch` collapses any 500 — including a genuine gateway/relayer outage — into `return false`. The query then never enters its error state, and a caller polling with `refetchInterval` would loop indefinitely on a real outage rather than surfacing the failure.

I can see this is a deliberate heuristic (documented in `errors/decrypt.ts:18-20`, and a 500 is the only propagation signal the gateway gives), so this is not a bug — just flagging that the indefinite-poll-on-outage case has no escape valve. If that's acceptable for the product (docs already say "allow 1–2 minutes"), no change needed.

Confidence: 75/100

packages/sdk/src/query/is-delegation-propagated.ts
```diff
@@ -0,0 +1,50 @@
...
+> {
+  return {
+    ...filterQueryOptions(config.query ?? {}),
+    queryKey: zamaQueryKeys.delegationPropagation.scope(
+      config.encryptedInputs,
+      config.delegatorAddress,
+      config.delegateAddress,
+    ),
+    queryFn: async (context) => {
+      const [, { delegatorAddress, encryptedInputs: keyedInputs }] = context.queryKey;
```
ghermet: [AUTOMATED] **Nit (non-blocking):** `queryFn` pulls `encryptedInputs`/`delegatorAddress` back out of `context.queryKey` rather than reading from the `config` closure. It's functionally correct today (the keyed inputs are the same values, just sorted + checksummed), but every sibling factory (`delegationStatusQueryOptions`, `decryptValuesQueryOptions`, …) reads live data from the closure and uses the query key purely for cache identity. Reading from the closure here would match that convention and avoid coupling the call args to the key's serialization shape.

Confidence: 78/100

ghermet (commented):
[AUTOMATED]

## Code Review — SDK-241 isDelegationPropagated

Reviewed across 4 parallel agents (2× bug hunter, guideline compliance, error-handling auditor), then independently validated every finding ≥80 confidence against the code.

**No objective runtime bugs survived validation.** The feature is sound: three-layer architecture (service → `/query` factory → hook) is consistent, tests cover all four layers, and it is **fully guideline-compliant** (query-key `getAddress` checksumming, `@throws` condition-first + `{@link}` at end, no handle leakage on the public surface, `"use client"`, TanStack wrapper rule).

Two design-level observations below are **non-blocking** — both reflect deliberate, documented choices; flagging them only so the tradeoffs are on record.

Findings that were investigated and **refuted**:
- *Zero-value handles aren't short-circuited* → intentional and documented (`decryption-service.ts:136-137`): bypassing the zero-value short-circuit is the whole point of forcing a round-trip to exercise the gateway ACL.
- *`delegateAddress` cache divergence on wallet switch* → the query key includes `delegateAddress` (`query-keys.ts:178`), so a wallet switch produces a new key, not a stale hit.

Verdict: looks good to me.

packages/sdk/src/query/is-delegation-propagated.ts
ghermet: [AUTOMATED] Retracting this — it's a false positive. On verification, reading from `context.queryKey` is the established convention here, not the exception: `delegationStatusQueryOptions` itself does it (delegation-status.ts:39), and 12/12 query factories in this directory read their args from `context.queryKey` (confidential-balance, has-permit, total-supply, token-metadata, user-decrypt, …). `is-delegation-propagated.ts` is consistent with the codebase as written; switching to the closure would make it the odd one out. No change.

packages/sdk/src/services/decryption-service.ts
ghermet: [AUTOMATED] Confirmed working-as-designed, no change. A 500 is the only signal the gateway emits for an un-synced delegation, so collapsing it to `false` is exactly what makes this a readiness probe rather than a throwing decrypt — the behavior is intentional and documented in errors/decrypt.ts:18-20, and the doc comment tells callers to allow 1–2 min. Distinguishing a genuine outage from propagation lag isn't possible at this layer without protocol support; bounding the poll (max retries / timeout) is the caller's concern, not the probe's. Leaving as-is.

ankurdotb:
[AUTOMATED]

> ## ✅ Resolved in the latest push (HEAD `8eee1fb5`)
>
> #489 has since been merged into this branch, and the fix landed in the single classifier exactly as hoped: `wrapDecryptError` now evaluates the not-entitled branch **first** and, when `ctx.isDelegated`, returns the *transient* `DelegationNotPropagatedError` rather than terminal `NotEntitledError`. So during the propagation window `isDelegationPropagated` collapses the delegated ACL denial to `false` instead of rethrowing. (A direct, non-delegated signer denial still surfaces as terminal `NotEntitledError`.) The reasoning is sound because `#assertAllDelegationsActive` runs first — any *subsequent* relayer ACL denial on the delegated path is by construction a sync/lag condition.
>
> A dedicated **merge-guard test** locks in the branch ordering — `decryption-service.test.ts`, *"returns false when the delegated ACL check denies during the propagation window"* — driving the message-only not-entitled error and asserting `false`. All CI (`test` / `test-node` / `test-vite` / `test-nextjs` / `api-report`) is green on this head. No blocking issues remain. Thanks for the clean fix.
>
> <sub>Residual non-blocking items are unchanged and previously acknowledged (the multi-contract `pLimit` first-settled-rejection race; the empty-array `→ true` short-circuit; and the accepted infinite-poll-on-genuine-500 tradeoff you flagged in the thread above).</sub>

<details><summary>Original coordination note (for history)</summary>

## Coordination hazard with #482 / #489 (classify-once + `NotEntitledError`)

Flagging a sequencing risk rather than a code defect in isolation: #486's correctness is coupled to the decryption-error reclassification in #482/#489, and all three target `prerelease`.

**The dependency.** `isDelegationPropagated` returns `false` for exactly one wrapped type — `DelegationNotPropagatedError` — which `wrapDecryptError(…, isDelegated=true)` produces *only* for a numeric HTTP **500**; everything else rethrows. That's the same single-branch collapse @ghermet already flagged for the outage case in https://github.com/zama-ai/sdk/pull/486#discussion_r3473019425 — this is its other edge.

**The collision.** `delegatedUserDecrypt` runs **inside the worker** (`relayer-web.ts` → `#worker`; `relayer-sdk.worker.ts` calls the relayer-SDK instance there) — exactly where #482's `classifyDecryptWorkerError` runs. Once #482/#489 land, `wrapDecryptError` gains a `NotEntitled` branch evaluated **before** the `isDelegated && 500 → DelegationNotPropagated` branch. A delegation that has landed on L1 but not yet synced to the gateway — the exact window this probe exists for — yields `persistAllowed=false`, which now classifies as `NotEntitledError`, not `DelegationNotPropagatedError`. Since `NotEntitledError instanceof DelegationNotPropagatedError` is `false`, `isDelegationPropagated` would **rethrow during precisely the propagation window it is built to report as `false`**, breaking the documented `while (!ready)` poll loop.

This is the delegated-path overlap I raised on #482 in https://github.com/zama-ai/sdk/pull/482#discussion_r3491858290 ("treat delegator-not-entitled as transient when `isDelegated`") — surfacing here as a concrete downstream regression.

**Independent of the merge, the `false` branch is already narrow.** `extractHttpStatus` is numeric-only and `withRetry` retries 502/503/504 but not 500, so a *message-only* 500 or an *exhausted* 5xx also rethrows rather than returning `false` — the set of "still syncing" responses is wider than the single numeric 500 the probe collapses, even before #482. And the underlying premise (not-propagated ⇒ 500) is asserted only by mocks here; `DelegationNotPropagatedError`'s own doc calls it "a best-effort heuristic," and #482 encodes the contradictory model (ACL denial ⇒ not-entitled). Worth pinning down the real relayer response for an unsynced delegated handle (a testnet observation would settle it).

**Suggested resolution — one fix covers both.** Land the #482 change that classifies `isDelegated` + not-entitled as transient (→ `DelegationNotPropagated`, not terminal `NotEntitled`), keeping readiness semantics in the single classifier #489 is consolidating; then #486's `false` branch stays correct by construction. Whichever way it lands, **#486 shouldn't merge independently of the #482 classification decision**, and a regression test driving this probe with a `NotEntitled`-shaped relayer rejection (asserting `false`) would stop the two from merging green into a silent regression.

Confidence: 88/100


</details>

ankurdotb (approved):
[AUTOMATED]

Approving. Re-reviewed at HEAD 8eee1fb5 after #489 was merged in.

The one blocking concern from my earlier coordination note — the delegated propagation window reclassifying to terminal \`NotEntitledError\` and making \`isDelegationPropagated\` rethrow instead of returning \`false\` — is resolved in the right layer: \`wrapDecryptError\` maps a delegated ACL denial to the transient \`DelegationNotPropagatedError\` (not-entitled branch first, gated on \`ctx.isDelegated\`), and a dedicated merge-guard test locks in the branch ordering. Non-delegated signer denials still get terminal \`NotEntitledError\`. All CI green (test / test-node / test-vite / test-nextjs / api-report).

Residual items are non-blocking and previously acknowledged (multi-contract \`pLimit\` first-settled-rejection race; empty-array \`→ true\`; the accepted infinite-poll-on-genuine-500 tradeoff). Nice work.

ankurdotb (changes requested):
Hold fire taking a look on whether this makes sense

ankurdotb (changes requested):
[AUTOMATED]

Updating this review after re-examining the direction through the SDK's design principles ("clear-text in, clear-text out") rather than correctness alone.

**Direction: absorb the propagation wait inside the SDK — don't ship a readiness API.**

The corrected timing reframes the problem. Propagation is gated by the protocol's `delegation_block_delay` (~10 blocks; the E2E suite waits 15) — a few **seconds** on a fast host chain, not the "1–2 minutes" our docs claimed. (SDK-241's ~4s measurement wasn't an outlier; it's ~10 blocks at a fast block time.) The readiness-query request was an artifact of that docs bug: at minutes-scale you schedule around the wait, at seconds-scale you just want the decrypt to work. So the right question isn't "how do we expose propagation state?" — it's "why does the consumer ever meet this window at all?"

**Rework proposal**

- Drop `isPropagated` / `isDelegationPropagatedQueryOptions` / `useIsDelegationPropagated`.
- Instead: **bounded internal retry** on `DelegationNotPropagatedError` inside the delegated-decrypt path (cap ~30s, configurable/opt-out). A fresh grant then just works: `delegateDecryption()` → `decryptBalanceAs()`; the error only surfaces on a genuine outage, with corrected block-based copy.

**Why this is the right shape for this SDK**

- *Seals the leak instead of documenting it.* The docs-only alternative pastes a catch-retry loop into every consumer — and "document the delay" was already tried (SDK-48) and produced this ticket.
- *Safe by construction.* `#assertAllDelegationsActive` runs first, so missing/expired grants still throw immediately — the retry can only engage in the grant-exists-but-not-synced window. The #489 classifier fix (delegated ACL denial → transient) is exactly the invariant it leans on.
- *Precedented, no conflict with defer-to-transport.* The SDK already retries transient errors on its own relayer transport (`withRetry` on 502/503/504). The defer-to-viem/ethers direction covers consumer-owned chain RPC; the relayer is SDK-owned and propagation is protocol semantics.
- *Cheaper than polling.* The hook fires speculative relayer round-trips every tick; internal retry only retries the decrypt the consumer already asked for. Retries reuse the cached EIP-712 permit — no extra signatures.
- *React for free.* `useDecryptBalanceAs` shows `isPending` through the window; no new hook, no `refetchInterval` guidance.
- *Dissolves the open probe questions* (multi-contract rejection race, `false`-vs-throw semantics, poll caps, the query-only `delegateAddress` field, empty-array semantics) — they were all properties of the probe surface.

**Implementation caveat:** in cleartext mode "not propagated" is really "not allowed" (no gateway lag exists there), so skip or short-cap the retry on the cleartext path.

**Independent of the rework, worth landing either way:** the "allow 1–2 minutes" copy in `DelegationNotPropagatedError`, the `Delegations` JSDoc, the cleartext message, `guides/delegated-decryption.md`, and `reference/sdk/errors.md` should be corrected to the block-based framing (~10 blocks / seconds). I have that change in hand.

**Fallback** if bounded auto-retry is rejected on principle: docs-only (document the reactive catch pattern; point to `isActive` for "grant landed") — accepted cost being that every consumer owns the loop.

The earlier classifier work (delegated ACL denial → transient `DelegationNotPropagatedError`, merged via #489) is correct and stands regardless.

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -16,7 +16,7 @@ This guide uses `sdk.delegations` and `token.decryptBalanceAs`. Before starting,
 
 ## Example
 
-A complete delegation flow — grant, wait for propagation, then decrypt as delegate:
+A complete delegation flow — grant, then decrypt as delegate (the SDK rides out gateway propagation for you):
```
ankurdotb: This term is a bit hard to understand if you don't know what "gateway" is. Maybe just call it ACL propagation or stick to previous terms.

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -78,10 +78,12 @@ The expiration date must be **at least 1 hour in the future**. Passing a closer
 
 Each call grants delegation for a single `(contractAddress, delegateAddress)` pair and submits one on-chain transaction.
 
-### 2. Wait for gateway propagation
+### 2. Gateway propagation (handled for you)
 
-{% hint style="warning" %}
-After the delegation transaction is mined, wait **1–2 minutes** before calling `decryptBalanceAs`. The delegation is recorded on L1 immediately, but the gateway (on Arbitrum) must sync the ACL state via cross-chain event propagation. Attempting delegated decryption before propagation completes throws `DelegationNotPropagatedError`.
+After the delegation transaction is mined, the gateway (on Arbitrum) syncs the ACL state via cross-chain event propagation — usually within ~10 blocks (a few seconds). You don't need to wait or poll: the delegated-decrypt path rides out that window with a bounded internal retry (~30s), so a decrypt issued right after granting simply waits for sync.
```
ankurdotb: Rename to "Zama Gateway"

ankurdotb (changes requested):

packages/sdk/src/relayer/cleartext/relayer-cleartext.ts
```diff
@@ -475,8 +475,8 @@ export class RelayerCleartext implements RelayerSDK, Disposable {
       if (!delegatorAllowed[i]) {
         throw new DelegationNotPropagatedError(
           `Delegated decryption of encrypted value ${encryptedValues[i]!} was denied by the on-chain ` +
-            `ACL check for delegator ${delegatorAddress}. This is most commonly a propagation/lag window — ` +
-            `allow 1–2 minutes after granting before retrying.`,
+            `ACL check for delegator ${delegatorAddress}. This is most commonly a propagation/lag window ` +
+            `(~10 blocks, a few seconds) that the SDK normally retries across — retry shortly.`,
         );
```
ankurdotb: [AUTOMATED] **Cleartext should throw `NotEntitledError` here, not `DelegationNotPropagatedError`.**

Cleartext is host-chain-only — `#persistAllowed` reads the ACL directly on the local chain (`this.#config.aclContractAddress` via `this.#client`), with no gateway and no cross-chain step. So a failed check here can never be a propagation lag; it's authoritative and permanent. The retryable `DelegationNotPropagatedError` has no meaning in this mode — and now that `delegatedUserDecrypt` rides out that error with a bounded ~30s retry, this makes a genuinely-not-entitled delegator spin the full budget on a local misconfig before failing. `NotEntitledError` is the correct terminal label, and using it fixes the hang without having to teach the retry loop about transports.

```suggestion
        throw new NotEntitledError({
          encryptedValue: encryptedValues[i]!,
          contractAddress,
          account: delegatorAddress,
        });
```

Two follow-ons outside this hunk: (1) update the comment above (L462–467 still says "transient propagation/lag window … maps to `DelegationNotPropagatedError`"), and (2) drop the now-unused `DelegationNotPropagatedError` import (L58) or it trips `no-unused-vars`.

ankurdotb (commented):
[AUTOMATED]

Re-reviewed at HEAD `7d2e3cf5`. The rework cleanly implements the agreed direction: the readiness API (probe + `/query` factory + React hook) is fully removed (net-zero vs `prerelease`), and the delegated-decrypt path now absorbs the propagation window via a bounded retry (retries only on `DelegationNotPropagatedError`, host pre-check inside each attempt, ~30s cap, `waitForPropagation` opt-out, batch fallback fails fast so items don't each re-spend the budget). Nicely done.

One inline change requested (the cleartext item from my previous review), plus a minor non-blocking note below. Keeping this as `CHANGES_REQUESTED` for the cleartext fix.

Minor (separate, no suggestion): `DelegatedDecryptOptions` is used in the public `delegatedDecryptValues` / `delegatedBatchDecryptValues` signatures but isn't exported from the package (no definition block in `sdk.api.md`) — consumers can pass `{ waitForPropagation: false }` inline but can't name the type. Worth exporting it alongside the other public option types.

ankurdotb (approved):
[AUTOMATED]

Approving — re-reviewed at HEAD 6d09b880. Both items from my previous review are resolved, with matching test coverage and green CI:

- **Cleartext delegated ACL denial is now terminal.** It throws `NotEntitledError` instead of the retryable `DelegationNotPropagatedError` — correct, since cleartext is host-chain-only (`#persistAllowed` reads the local ACL, no gateway/propagation), so a denial there is authoritative and permanent. This also removes the ~30s bounded-retry spin on a local misconfig without teaching the retry loop about transports. Comment corrected, unused import dropped, and the test now asserts `NotEntitledError`.
- **`DelegatedDecryptOptions` is exported**, so consumers can name the type they pass to `delegatedDecryptValues` / `delegatedBatchDecryptValues` (api-report regenerated).

The broader rework is clean: the readiness API (probe + `/query` factory + React hook) is fully dropped (net-zero vs prerelease), and the delegated-decrypt path absorbs the propagation window via a bounded, opt-out retry that only engages on the transient error and keeps the host-chain pre-check inside each attempt. Nice work.

PR #488, opened 2026-06-25
refactor(sdk): type matchZamaError handlers by matched error subclass [SDK-231]

zama-cremaud:
## What

`matchZamaError(error, handlers)` routed by code but typed every handler to receive the **base** `ZamaError`, so a handler keyed on a code couldn't read that subclass's own fields without a cast (`TS2339`). This types the handler for each code to receive **that code's error class**, so subclass fields are reachable without a cast:

```ts
matchZamaError(error, {
  INSUFFICIENT_CONFIDENTIAL_BALANCE: (e) => `need ${e.requested}, have ${e.available}`, // ✅
  RELAYER_REQUEST_FAILED: (e) => e.statusCode,                                          // ✅
});
```

Closes [SDK-231](https://linear.app/zama/issue/SDK-231).

## How

- New `errors/match.ts` holds a hand-maintained `ErrorForCode` map (code → error class) and `matchZamaError`. The function moves out of `base.ts` so the low-level module stays free of subclass imports (no `base ⇄ subclass` cycle).
- A `Complete<T extends Record<ZamaErrorCode, ZamaError>>` wrapper around the map **fails the build** if a code is added without an entry (or mapped to a non-`ZamaError`) — drift protection that lives in `src/`, so `pnpm typecheck` enforces it.
- **Types-only:** the runtime body is unchanged (a single internal cast bridges the per-code param types at the `instanceof` dispatch).
- **Purely additive:** the `_` wildcard, base-field handlers, and handlers explicitly annotated `(e: ZamaError) => …` all still compile (params are contravariant).

## Tests / docs / artifacts

- `errors.test-d.ts`: asserts each rich code narrows to its subtype (`available`/`requested`, `statusCode`, `signerChainId`/`providerChainId`) + additivity assertions.
- Restores the subclass-field examples in `handle-errors.md` / `errors.md` that previously had to fall back to `instanceof` (the doc workaround SDK-231 was filed to remove).
- Regenerated `sdk.api.md` and the LLM corpus.

## Notes for reviewers

- The one thing the compiler can't guard is a *wrong-but-valid* class mapping (e.g. `CHAIN_MISMATCH → SigningRejectedError`) — many error classes are structurally identical, so this is only caught by the `test-d` assertions.
- Heads-up (out of scope here): the repo's `vitest --project typecheck` runner collects `*.test-d.ts` but doesn't actually type-check them (the project tsconfig excludes `__tests__`), so type-test assertions don't currently gate CI. Enabling real enforcement surfaces ~3 pre-existing failures in `credentials.test-d.ts` — worth a dedicated follow-up.

### DIFF
--- a/docs/gitbook/src/guides/handle-errors.md
+++ b/docs/gitbook/src/guides/handle-errors.md
@@ -89,8 +89,8 @@ matchZamaError(error, {
   SIGNING_REJECTED: () => toast("Please approve the transaction"),
   ENCRYPTION_FAILED: () => toast("Encryption failed -- please retry"),
   TRANSACTION_REVERTED: (e) => toast(`Transaction failed: ${e.message}`),
-  INSUFFICIENT_CONFIDENTIAL_BALANCE: () => toast("Insufficient confidential balance"),
-  INSUFFICIENT_ERC20_BALANCE: () => toast("Not enough tokens to shield"),
+  INSUFFICIENT_CONFIDENTIAL_BALANCE: (e) => toast(`Need ${e.requested}, have ${e.available}`),
+  INSUFFICIENT_ERC20_BALANCE: (e) => toast(`Need ${e.requested}, have ${e.available}`),
   BALANCE_CHECK_UNAVAILABLE: () => toast("Sign to verify your balance first"),
   ERC20_READ_FAILED: () => toast("Could not read token balance -- check your connection"),
   _: () => toast("Something went wrong"),
@@ -102,7 +102,7 @@ matchZamaError(error, {
 
 The `_` wildcard catches any `ZamaError` not explicitly handled. If the error is not a `ZamaError` at all (and no `_` is provided), `matchZamaError` returns `undefined`.
 
-Each handler receives the error typed as the base `ZamaError`, so `.code` and `.message` are available but subclass-specific fields are not. To read fields like `InsufficientConfidentialBalanceError.available` or `RelayerRequestFailedError.statusCode`, narrow with `instanceof` (step 2) inside the handler.
+Each handler receives the error class for its code, so subclass fields are available without a cast — `INSUFFICIENT_CONFIDENTIAL_BALANCE` hands you an `InsufficientConfidentialBalanceError` with `.available` / `.requested`, `RELAYER_REQUEST_FAILED` an error with `.statusCode`, and so on.
 
 ### 4. Handle specific errors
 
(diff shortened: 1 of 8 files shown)


ghermet:
[AUTOMATED]

## Can `Complete` be improved with `Extract<K>`?

Short answer: **the `Extract` pattern is genuinely nicer, but it is not a localized swap for `Complete` — it needs a broader refactor this PR deliberately avoids. Keep `Complete` as-is here.**

### What `Complete` enforces today
```ts
type Complete<T extends Record<ZamaErrorCode, ZamaError>> = T;
```
Completeness (every code present) + value type (every value is a `ZamaError`) on the **hand-maintained** map. As the PR notes, it cannot catch a *wrong-but-valid* mapping (`CHAIN_MISMATCH → SigningRejectedError`) because the subclasses are structurally identical.

### What `Extract<K>` would buy
Derive the map from a union of the error classes instead of hand-writing it:
```ts
type AllZamaErrors = ChainMismatchError | InsufficientConfidentialBalanceError | ...;
type ErrorForCode = { [K in ZamaErrorCode]: Extract<AllZamaErrors, { code: K }> };
```
This would **drop the hand-maintained map entirely** *and* close the wrong-mapping gap — the mapping comes from each class’s own `code`, so it cannot be mis-keyed.

### Why it is not a drop-in here
`Extract<T, { code: K }>` discriminates on a **literal** `code`. But every subclass inherits `readonly code: ZamaErrorCode` (the wide union) from the base (`base.ts:85`) and none narrows it. With a wide `code`, `Extract<AllZamaErrors, { code: "CHAIN_MISMATCH" }>` matches no member and collapses to `never` for **every** key (verified with a minimal `tsc --strict` probe).

To make `Extract` viable, each subclass’s `code` must be narrowed to its literal — cleanest via a generic base:
```ts
class ZamaError<C extends ZamaErrorCode = ZamaErrorCode> extends Error {
  readonly code: C;
  constructor(code: C, ...) { ... }
}
class ChainMismatchError extends ZamaError<typeof ZamaErrorCode.ChainMismatch> { ... }
```
That is a one-line change per subclass (~25 classes) **plus** a public-API-surface change (`sdk.api.md` shifts — base error becomes generic), well beyond SDK-231’s “types-only, purely additive” scope.

### Recommendation
Keep `Complete` as-is for this PR; the `test-d.ts` assertions already backstop the wrong-mapping gap. The `Extract`-derived map is a good **follow-up**, coupled with the `code`-narrowing work since both touch the public surface.

Optional nit: `Complete`’s doc comment could add one line noting it *cannot* catch a wrong-but-valid subclass mapping — that caveat currently lives only in the PR description, not the code.

Confidence: 95/100

ghermet (approved):

PR #489, opened 2026-06-25
feat(sdk): typed decryption error causes — not-entitled, RPC rate-limit, relayer back-pressure [SDK-239][SDK-236]

zama-cremaud:
## Summary

Closes [SDK-239](https://linear.app/zama/issue/SDK-239) and [SDK-236](https://linear.app/zama/issue/SDK-236).

This is the **single, consolidated SDK-239 PR**. It supersedes #482 (the initial worker-source classifier, folded in here) and #492 (SDK-236 relayer back-pressure, also folded in and closed) — they are subsumed by the final architecture below, so reviewing this one PR covers everything.

Decryption failures previously collapsed several unrelated causes into `DECRYPTION_FAILED`, forcing entitlement-/retry-aware consumers (server-side indexers, AI codegen) to pre-check on-chain out of band or string-match error messages. This adds distinct, accurate, typed causes so consumers can branch deterministically — and consolidates **all** error classification on the main thread.

## Architecture — classify once, on the main thread

The worker is now a **taxonomy-agnostic serializer**. The structured-clone worker boundary strips prototypes / `code` / `status` / `cause`, leaving only the message — so instead of classifying inside the worker, `serializeError` flattens an error and its `cause`/`error`/`info` chain into a structured-clone-safe envelope, and `deserializeError` rebuilds a real `Error` whose `.cause` chain mirrors the original. The single classifier `wrapDecryptError(error, fallback, ctx)` then runs once on the main thread, where the request context (`contractAddress` / `account` / `isDelegated`) is available. This removes the worker-side classifier entirely (no taxonomy imports in either worker) and kills the prior depth-1-vs-depth-6 chain-walk asymmetry.

### `NotEntitledError` (`NOT_ENTITLED`, terminal / non-retryable)
- Carries `encryptedValue` / `contractAddress` / `account`.
- Derived from the relayer SDK's **own** authoritative ACL check (`validateAclPermissions` → `persistAllowed`) — **zero extra on-chain reads, no cross-provider race**. Matched against the pinned `@zama-fhe/relayer-sdk`; drift guards read the installed dependency source (and the pinned CDN `RELAYER_SDK_VERSION`) and fail loudly if the message or the `RELAYER_FETCH_ERROR` tag drifts. `TODO` tracks keying off the relayer's typed `ACLUserDecryptionError` once its decrypt path surfaces it.
- Scope: the **direct signer** denial. The rarer "dapp contract not authorized" case is a dapp misconfig and stays `DecryptionFailedError` (documented).
- **Delegated path:** a "not entitled" verdict there comes from the *delegator's* `persistAllowed` L1 read, which is false *transiently* under propagation lag / a stale RPC block — so it is surfaced as the retryable `DelegationNotPropagatedError`, mirroring the delegated-500 handling, not the terminal `NotEntitledError`.

### `RpcRateLimitError` (`RPC_RATE_LIMITED`, retryable, `retryAfter?` in seconds)
- The **consumer's RPC provider** throttling (HTTP 429 / JSON-RPC `-32005`) during the SDK's on-chain reads. Detection covers viem (`status: 429`, `code: 429`) and ethers (`code: "SERVER_ERROR"` + `info.responseStatus: "429 …"`) so the retryable signal isn't lost to `DecryptionFailedError`.
- Treated as a fatal-batch error so batch decrypt aborts once instead of amplifying the throttle (a shared abort flag also short-circuits the per-item fallback under concurrency).

### `RelayerRequestFailedError` — relayer back-pressure (SDK-236)
- Gains `retryAfter` (seconds) + `retryable` (true on HTTP 429). The relayer's own 429 stays a `RelayerRequestFailedError` (excluded from the consumer signal via the `RELAYER_FETCH_ERROR` tag). The `Retry-After` header lives only on a non-cloneable `Headers`, so it is parsed to seconds at the worker source and threaded through the envelope.

## Retry: consumer RPC vs relayer

`withRetry` (the SDK's only custom retry, wrapping relayer pool calls) now scopes to **relayer transport** only: it retries relayer HTTP 502/503/504 (tagged) and relayer-boundary network failures, and **defers consumer-RPC faults to the integrator's viem/ethers client** (which already retries them) instead of double-retrying on top. Timeouts are owned by viem/ethers + the worker-level operation timeout; relayer 429 stays surfaced with `retryAfter`.

## Units

All public time values are **seconds** (matching the SDK's existing convention — `transportKeyPairTTL`, `registryTTL`, … — and the `Retry-After` header's own unit).

## matchZamaError

`NOT_ENTITLED` / `RPC_RATE_LIMITED` are registered in `matchZamaError`'s per-subclass type map (after #488), so handlers receive `NotEntitledError` / `RpcRateLimitError` without a cast.

## Test plan
- [x] `pnpm --filter @zama-fhe/sdk test` green — `wrapDecryptError` mapping (incl. delegated overlap), the rate-limit taxonomy (viem/ethers 429 shapes), serialize/deserialize worker round-trip, relayer back-pressure (`retryAfter`/`retryable`), `withRetry` relayer-vs-consumer scoping, cleartext parity, batch abort, the `matchZamaError` handler typing, and the real-dependency drift guards.
- [x] `tsc` / `oxlint` / `oxfmt --check` clean; api-extractor reports + `llms` regenerated. Merged latest `prerelease`.

## Notes
- Base `prerelease`. Public API: additive (`NotEntitledError`, `RpcRateLimitError`, codes; `RelayerRequestFailedError.retryAfter`/`retryable`; all narrowed by `matchZamaError`).
- Supersedes #482 (folded in) and #492 (closed).


packages/sdk/src/errors/rpc.ts
```diff
@@ -0,0 +1,34 @@
...
+ *   if (e instanceof RpcRateLimitError) {
+ *     // back off and retry; consider a higher-throughput RPC endpoint
+ *   }
+ * }
+ * ```
+ */
+export class RpcRateLimitError extends ZamaError {
+  /**
+   * Suggested delay before retrying, in milliseconds, when the provider
+   * supplied one (e.g. a `Retry-After` header). `undefined` when unknown.
```
ankurdotb: [AUTOMATED] Superseded — this comment was anchored to an old commit and is outdated. Current review (with the serialize/ethers-429 fix + `SerializedError` export) is here: https://github.com/zama-ai/sdk/pull/489#pullrequestreview-4607074800

ankurdotb (changes requested):
[AUTOMATED]

## Request changes — `RpcRateLimitError.retryAfter` is inert for the dominant real provider

Flagging here rather than on #482 because this PR is the right home: it consolidates the decryption-error classifier on the main thread, and the fix belongs in that single classifier.

**What works (not contested):** the retryable-vs-terminal *classification* — the actual SDK-239 deliverable — is sound. A real edge/Cloudflare 429 surfaces `status: 429` on viem's `HttpRequestError`, which `findInErrorChain` reaches, so `RpcRateLimitError` is produced and the fatal-batch abort fires.

**What doesn't:** the `retryAfter` hint. `extractRetryAfterMs` (`packages/sdk/src/utils/error.ts:272`) only reads numeric `retryAfter` / `retryAfterMs` **properties**. But viem — and Cloudflare and most edge-proxied public RPCs — deliver the backoff **only as the `Retry-After` HTTP header**: viem consumes it internally for its own retry loop (`utils/buildRequest`) and exposes it on `HttpRequestError.headers` (a `Headers` object), never as a numeric property. So for the dominant real-world rate-limiter, `RpcRateLimitError.retryAfter` is **always `undefined`**, while `errors/rpc.ts` publicly documents a `Retry-After` header as its source. The field is populated only by synthetic test fixtures that no provider emits.

**Why it must be handled here specifically:** this PR's `serializeError` envelope captures only the `retryAfter` / `retryAfterMs` scalars (`utils/error.ts:99-140`), so a `Headers` object cannot cross the worker boundary. The header therefore has to be parsed into a number (a) from the live error on the main-thread classifier path (cleartext / delegation pre-check), and (b) inside the worker's `serializeError` before structured clone for the worker decrypt path.

### Requested changes
1. In `extractRetryAfterMs`, also read `error.headers?.get('Retry-After')` and parse **both** RFC 7231 forms: delta-seconds (`"120"`) and an HTTP-date.
2. Capture the parsed value into the `retryAfter` scalar in the worker `serializeError` path so it survives structured clone.
3. Add a fixture test built from a real viem `HttpRequestError` (status 429 + a `Retry-After` header, nested under a `cause`) asserting **both** the classification and a populated `retryAfter`.

If header parsing is out of scope for this PR, the acceptable alternative is to **soften the `rpc.ts` JSDoc** (drop the "Retry-After header" claim) and track consumer-side `retryAfter` explicitly. Note: SDK-236 covers `retryAfter` only for the *relayer's* `RelayerRequestFailedError`, **not** this consumer-RPC `RpcRateLimitError` — so today it is untracked.

Confidence: 90/100.

ankurdotb:
[AUTOMATED] Minor (description hygiene): the PR **title and body still say `retryAfterMs`**, but the code at HEAD now uses **`retryAfter` in seconds** on both `RpcRateLimitError` and `RelayerRequestFailedError`, and `parseRetryAfterHeader` returns seconds. Worth syncing the title/description so reviewers, the changelog, and the api-report narrative match the implementation.

packages/sdk/src/utils/error.ts
```diff
@@ -34,30 +34,441 @@ export function isContractCallError(error: unknown): boolean {
...
+  status?: number;
+  /** relayer / node-fetch-style HTTP status. */
+  statusCode?: number;
+  /** Server-driven retry delay in **seconds** (numeric prop or parsed `Retry-After`). */
+  retryAfter?: number;
+  cause?: SerializedError;
+}
+
+/** Scalar signal fields carried verbatim across the boundary, in both directions. */
+const SERIALIZED_SCALAR_KEYS = ["code", "status", "statusCode", "retryAfter"] as const;
```
ankurdotb, line 114: [AUTOMATED] `serializeError` copies only these scalar keys, so ethers' `info.responseStatus` (the 429 status string) is dropped across the worker boundary. A worker-origin edge/Cloudflare 429 on the relayer-sdk's `persistAllowed` read is then rebuilt without the signal → `wrapDecryptError` returns terminal `DecryptionFailedError` instead of retryable `RpcRateLimitError`, and `isFatalBatchError` → false so the delegated batch doesn't abort (throttle amplification). Reproduced against HEAD. Fix: parse `info.responseStatus` to a numeric status at serialize time (like `Retry-After`), add a round-trip test, and walk all of `cause`/`error`/`info` (not just the first present key). Detail in the review summary. Confidence: 90/100

packages/sdk/src/worker/worker.types.ts
```diff
@@ -212,12 +213,17 @@ export interface SuccessResponse<T> extends BaseResponse {
...
+ * prototypes/codes/causes across the boundary); the main thread rebuilds it
+ * ({@link deserializeError}) and classifies it once in `wrapDecryptError`.
+ */
+export type ErrorResponse = BaseResponse & {
   success: false;
   error: string;
-  /** HTTP status code from the relayer, when available. */
-  statusCode?: number;
-}
+  serialized?: SerializedError;
```
ankurdotb, line 225: [AUTOMATED] `serialized?: SerializedError` is part of the `@public` `/node` `ErrorResponse`, but `SerializedError` isn't exported from `src/node/index.ts` (`ae-forgotten-export: none` silences the warning). Custom worker implementors get an opaque type. Suggest `export type { SerializedError }` from `src/node/index.ts`, or make `serialized` opaque. (The `statusCode` removal here is safe — nothing reads it off the response.) Confidence: 85/100

ankurdotb (changes requested):
[AUTOMATED]

## Re-review of #489 (consolidated SDK-239 + SDK-236)

> Note: this supersedes my earlier review — it was anchored to an old commit (`7a692f0d`) and went **outdated** across the rebases, so GitHub collapsed it out of the current Files view (that's why it wasn't visible). Re-posting on the current head.

**Prior review asks — all resolved ✅** (verified in code, not just described): `retryAfter` in seconds on both errors; `withRetry` scoped to relayer transport with consumer-RPC deferred to viem/ethers; ethers `SERVER_ERROR`/`info.responseStatus` + `code === 429` in the detector; delegated not-entitled → transient `DelegationNotPropagatedError`; cleartext parity (`#assertDelegation` checks `persistAllowed(delegator)`); relayer back-pressure (`retryAfter`/`retryable`); `matchZamaError` map. Great work consolidating the whole thing.

Two items to address, one rebase note:

**1. [Please fix] `serializeError` drops the ethers-429 signal across the worker boundary** (`utils/error.ts` — inline). The detector handles `info.responseStatus`, but `serializeError` copies only `code/status/statusCode/retryAfter` and recurses a single nested link, so `info.responseStatus` never survives the worker → main-thread round-trip. A real edge/Cloudflare 429 on the relayer-sdk's ethers `persistAllowed` read *inside the worker* is rebuilt without the signal → `wrapDecryptError` returns terminal `DecryptionFailedError` instead of retryable `RpcRateLimitError`, and `isFatalBatchError` → false so the delegated batch does **not** abort (the sibling workers keep hitting the throttled RPC — the amplification the fatal-batch flag exists to prevent). Reproduced against HEAD: `direct → RpcRateLimitError`, worker round-trip → `DecryptionFailedError`. The existing round-trip test only covers `-32005` (survives because it's on `code`). *Fix:* parse `info.responseStatus` into a numeric status at serialize time (same pattern as `Retry-After`); add a round-trip regression test with `info: { responseStatus: "429 …" }`; and walk all of `cause`/`error`/`info` rather than the first present key.

**2. [Leak, medium] `SerializedError` referenced in the public `/node` API but not exported** (`worker.types.ts` — inline). `ErrorResponse.serialized?: SerializedError` appears in `sdk-node.api.md`, but `SerializedError` isn't declared/exported there (`ae-forgotten-export: none` silences the warning, so the build passes) — custom worker implementors get an opaque type. Suggest `export type { SerializedError }` from `src/node/index.ts`, or make `serialized` opaque. (The `ErrorResponse.statusCode` removal itself is safe — nothing reads it off the response; status now lives in `serialized.statusCode`.)

**3. [#486 rebase note]** With #489 landing first (confirmed): the `wrapDecryptError` third-arg change (boolean `isDelegated` → object `ctx`) means #486's `isDelegationPropagated` probe callsite must change on rebase — `wrapDecryptError(err, msg, true)` → `wrapDecryptError(err, msg, { isDelegated: true })` — otherwise `ctx.isDelegated` is `undefined`, the delegated reclassification stops firing, and the probe throws instead of returning `false`. Also re-check #486's tests against the broadened semantics (a delegated not-entitled ACL denial now also → `DelegationNotPropagated`; the "dapp contract not authorized" variant stays `DecryptionFailedError`).

**Design note (not blocking):** the delegated path maps *every* not-entitled denial to retryable `DelegationNotPropagated`, so a genuinely-terminal delegated denial isn't self-terminating — a naive retry loop won't stop. Symmetric with the delegated-500 branch and the message hints at it, but worth a deliberate call.

Confidence: #1 90/100, #2 85/100.

ankurdotb (approved):
[AUTOMATED] Re-reviewed `b66ed015`. Both requested changes resolved and verified in code + tests:

- **ethers-429 across the worker boundary** — `serializeError` now lifts `info.responseStatus` (and walks all of `cause`/`error`/`info`), `deserializeError` rebuilds it, so a worker-origin ethers 429 classifies as `RpcRateLimitError` and the fatal-batch abort fires. Locked in by round-trip + `wrapDecryptError` tests, with a 503 negative test so it doesn't over-match.
- **`SerializedError`** — now exported from `/node`.

Approving. Two non-blocking items live outside this PR: on #486's rebase, switch its `wrapDecryptError` callsite to `{ isDelegated: true }`; and the delegated "every not-entitled → retryable `DelegationNotPropagated`" behavior is a deliberate, documented tradeoff (not self-terminating for a genuinely-terminal delegated denial).

PR #285, opened 2026-04-27
refactor(sdk): multichain relayer architecture with shared worker runtime

ghermet:
## Summary

This PR introduces `createConfig` — a unified config builder that replaces manual relayer/signer/provider wiring with a single function call per adapter (viem, ethers, wagmi).

### Config system
- **`createConfig`** entry points for viem (`@zama-fhe/sdk/viem`), ethers (`@zama-fhe/sdk/ethers`), and wagmi (`@zama-fhe/react-sdk/wagmi`)
- **Chain definitions** via `@zama-fhe/sdk/chains` — `sepolia`, `mainnet`, `hardhat`, `hoodi` with all FHE contract addresses baked in
- **Relayer factories** — `web()`, `node()`, `cleartext()` take only runtime options (e.g. `poolSize`); per-chain config (`relayerUrl`, `auth`, `network`, `executorAddress`) lives on `FheChain`
- **`ZamaConfig`** replaces manual `relayer + signer + provider + storage` wiring

### Chain-scoped configuration
- **`relayerUrl`** moved from `web()`/`node()` args to `FheChain` definition
- **`executorAddress`** moved from `cleartext()` args to `FheChain` definition
- **`auth`** field added to `FheChain` for relayer authentication (API key headers, etc.)
- **`network`** field on `FheChain` for node relayer RPC endpoint
- Custom chains use `{ ...sepolia, relayerUrl: "..." } as const satisfies FheChain`

### Multichain relayer architecture
- **`RelayerDispatcher`** owns chain management (`chains`, `chain`, `switchChain`) and worker lifecycle — simplified with non-null `chainId` and list constructor
- **`RelayerWeb`/`RelayerNode`** reverted to single-chain — injected worker/pool, lazy init with retry on failure
- **Relayer construction** moved into `RelayerDispatcher` constructor (was in `buildZamaConfig`)
- **Shared WASM runtime** — same `web()`/`node()` reference = same worker = same WASM module. Different calls = separate runtimes.

### Signer / Provider split
- **`walletClient` required** in viem config (was optional)
- **`ethereum` + `provider` variant** for ethers config
- Concrete signer types (`ViemSigner`, `EthersSigner`, `WagmiSigner`) with proper lifecycle callbacks

### Worker protocol
- Unified `INIT` message with `env:"web"|"node"` discriminant (replaces `NODE_INIT`)
- `chainId` on all operation payloads
- Lazy multi-instance `FhevmInstance` maps with concurrent deduplication

### Other changes
[...]

PR #490, opened 2026-06-25
fix(sdk): resolve node worker without import.meta.resolve for SSR bundlers [SDK-235]

ghermet:
## Problem

`@zama-fhe/sdk/node`'s worker factory resolved its entry with `import.meta.resolve("@zama-fhe/sdk/node")`. Vite's SSR transform — used by **Ponder** and other server-side bundlers — rewrites `import.meta.resolve` to `__vite_ssr_import_meta__.resolve`, which is **undefined at runtime**. `createWorker()` then throws and the Node relayer can never spawn its worker, breaking any server-side integration running inside a Vite-SSR runtime.

Surfaced via the Product Integrations take-home (a server-side ERC-7984 indexer on Ponder). Fixes [SDK-235](https://linear.app/zama/issue/SDK-235).

## Fix

`packages/sdk/src/worker/worker.node-client.ts` now uses `createRequire(import.meta.url).resolve(...)`:

```ts
const require = createRequire(import.meta.url);
const nodeEntry = pathToFileURL(require.resolve("@zama-fhe/sdk/node"));
return new Worker(new URL("relayer-sdk.node-worker.js", nodeEntry));
```

- **SSR-safe** — relies only on `import.meta.url` (preserved by the SSR transform) plus `node:module`/`node:url`. No `import.meta.resolve`.
- **Preserves the #285 chunk-location fix** — `require.resolve` resolves the package entry regardless of which rolldown chunk this code is tree-shaken into, so the worker path stays correct.

Requires the SDK to be externalized server-side, which is inherent to spawning a worker from a file (and the default for Ponder / Vite SSR).

## Tests

- New build-gated regression test in `worker.bundler.test.ts`: asserts **no shipped ESM chunk references `import.meta.resolve`** — the exact construct that breaks SSR consumers. Fails on the old code, passes after the fix.

## Verification

- RED → GREEN: test failed on `dist/esm/node/index.js`, passes after rebuild.
- Runtime check (real Node ESM, package self-reference): resolves to the existing `dist/esm/node/relayer-sdk.node-worker.js`.
- Worker suite: 90 passed / 5 skipped. `typecheck` ✅, `lint` ✅.

## Not a duplicate of SDK-46

SDK-46 (Done) fixed an earlier `new URL('relayer-sdk.node-worker.js', import.meta.url)` path-layout mismatch. The current code used `import.meta.resolve(...)`, so this Vite-SSR incompatibility is a distinct issue.

### DIFF
--- a/packages/sdk/src/worker/__tests__/worker.bundler.test.ts
+++ b/packages/sdk/src/worker/__tests__/worker.bundler.test.ts
@@ -8,11 +8,13 @@
  * Run `pnpm build` before running these tests.
  */
 import { describe, test, expect } from "vitest";
-import { readFileSync, existsSync } from "node:fs";
+import { readFileSync, existsSync, readdirSync } from "node:fs";
 import { resolve } from "node:path";
 
 const DIST = resolve(__dirname, "../../../dist");
 const hasBuild = existsSync(resolve(DIST, "relayer-sdk.worker.js"));
+const ESM_DIR = resolve(DIST, "esm");
+const hasEsmBuild = existsSync(ESM_DIR);
 
 describe.skipIf(!hasBuild)("worker build smoke tests", () => {
   const workerPath = resolve(DIST, "relayer-sdk.worker.js");
@@ -41,5 +43,22 @@ describe.skipIf(!hasBuild)("worker build smoke tests", () => {
   test("standalone worker file exports filename alongside code", () => {
     const content = readFileSync(indexPath, "utf-8");
     expect(content).toContain("relayer-sdk.worker.js");
+  });
+});
+
+describe.skipIf(!hasEsmBuild)("node worker resolution is SSR-bundler safe (SDK-235)", () => {
+  // `import.meta.resolve` is rewritten to `__vite_ssr_import_meta__.resolve` by
+  // Vite's SSR transform (used by Ponder and other server-side bundlers), where
+  // it is undefined at runtime — so the Node worker factory throws and can never
+  // spawn its worker. No shipped ESM chunk may reference it, regardless of which
+  // rolldown chunk the node worker client lands in.
+  test("no shipped ESM chunk uses import.meta.resolve", () => {
+    const files = readdirSync(ESM_DIR, { recursive: true }) as string[];
+    const offenders = files
+      .filter((file) => file.endsWith(".js"))
+      .filter((file) =>
+        readFileSync(resolve(ESM_DIR, file), "utf-8").includes("import.meta.resolve"),
+      );
+    expect(offenders).toEqual([]);
   });
 });
--- a/packages/sdk/src/worker/worker.node-client.ts
+++ b/packages/sdk/src/worker/worker.node-client.ts
@@ -1,5 +1,7 @@
 import { Worker } from "node:worker_threads";
 import { randomUUID } from "node:crypto";
+import { createRequire } from "node:module";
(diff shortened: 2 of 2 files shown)


packages/sdk/src/worker/__tests__/worker.bundler.test.ts
```diff
@@ -43,3 +45,20 @@ describe.skipIf(!hasBuild)("worker build smoke tests", () => {
...
   });
 });
+
+describe.skipIf(!hasEsmBuild)("node worker resolution is SSR-bundler safe (SDK-235)", () => {
+  // `import.meta.resolve` is rewritten to `__vite_ssr_import_meta__.resolve` by
+  // Vite's SSR transform (used by Ponder and other server-side bundlers), where
+  // it is undefined at runtime — so the Node worker factory throws and can never
+  // spawn its worker. No shipped ESM chunk may reference it, regardless of which
+  // rolldown chunk the node worker client lands in.
+  test("no shipped ESM chunk uses import.meta.resolve", () => {
```
ankurdotb, line 55: [AUTOMATED] Optional (non-blocking): nice cheap guard — asserting no shipped ESM chunk contains `import.meta.resolve` directly targets the regression that matters. It's a token-absence proxy rather than a behavioral check; if you ever want the positive asserted too, running the factory through Vite's `ssrTransform` and confirming it doesn't throw would do it. Not needed for this PR.

Confidence: 85/100.

ankurdotb (approved):
[AUTOMATED]

## Code Review — Approve

Verified this fixes SDK-235. Swapping `import.meta.resolve("@zama-fhe/sdk/node")` → `createRequire(import.meta.url).resolve(...)` removes the exact construct Vite's SSR transform rewrites to the undefined `__vite_ssr_import_meta__.resolve`. Cross-checked against Vitest [#6953](https://github.com/vitest-dev/vitest/issues/6953) / [#7039](https://github.com/vitest-dev/vitest/issues/7039): the transform preserves `import.meta.url` and `.env` but not the `.resolve()` **method**, so relying only on `import.meta.url` (via `createRequire`) is the canonical SSR-safe pattern.

Checked and confirmed:

- **Behavior-preserving resolution.** `pathToFileURL(require.resolve("@zama-fhe/sdk/node"))` yields the same `file://` URL the old code produced → same worker path `dist/esm/node/relayer-sdk.node-worker.js`, identical under pnpm's symlinked layout (both deref realpath). The #285 / SDK-46 chunk-location fix is preserved.
- **No `ERR_PACKAGE_PATH_NOT_EXPORTED`.** `exports["./node"]` has no `require` condition but a top-level `default`, so `require.resolve` matches `default`. `require.resolve` only computes a path — it never loads the ESM file as CJS — so the ESM-only `./node` subpath is not an obstacle here.
- **Generalizes** beyond Ponder to any Vite/esbuild SSR transform (Next.js, Remix, SvelteKit); the regression test scans every shipped ESM chunk, and CI builds before vitest so the test actually runs (not silently skipped).

One non-blocking note inline. FYI on the PR description: "Requires the SDK to be externalized server-side" is a touch stronger than necessary — `require.resolve` hits the real filesystem and the worker thread runs under real Node regardless of whether the bundler externalizes or inlines the SDK module; the only hard requirement is the worker file existing on disk in `node_modules` (it does). Doesn't affect correctness.

Confidence: 95/100.

PR #494, opened 2026-06-29
fix(sdk): configurable + diagnosable Node worker timeouts with self-healing [SDK-237]

zama-cremaud:
## Summary

Closes [SDK-237](https://linear.app/zama/issue/SDK-237).

The Node worker pool can hang a consumer's decryption with no way to bound it diagnosably. A per-request timeout *does* exist today, but it is:

1. **Not configurable** — hardcoded 30s/60s; `node({ poolSize, fheArtifactCacheTTL })` exposes no timeout.
2. **Not typed/diagnosable** — it rejects a plain `Error`, so it isn't matchable and, flowing through `wrapDecryptError`, collapses into `DecryptionFailedError` (the same conflation SDK-239 fights, for timeouts).
3. **No recovery + a latent cascade bug** — the stuck worker is never recycled; once the timed-out request settles, `#dispatch`'s `finally` decrements its active count, so least-connections sees the still-hung worker as *idle* and re-selects it → cascading timeouts.

## What this does

- **`WorkerTimeoutError`** (`OPERATION_TIMEOUT`, retryable) carrying `operation` / `timeout` / `elapsed` (both in **seconds**) / `worker`. Registered in the `matchZamaError` per-subclass type map and **passed through `wrapDecryptError`** — a timeout is no longer a `DecryptionFailedError`.
- **Configurable bounds (seconds)** — `node({ operationTimeout, initTimeout, recycleWorkerOnTimeout })`, validated at the factory boundary and threaded to the base worker client. Durations are seconds to match the rest of the public config (`fheArtifactCacheTTL`, `transportKeyPairTTL`).
- **Self-healing (Node pool only)** — on a timeout the Node client terminates the (presumed-stuck) worker so the next call lazily re-inits a fresh one. This kills the hung WASM thread and **fixes the cascade**. Gated by `recycleWorkerOnTimeout` (default `true`). The browser runs a single worker with no least-connections pool and `web()` exposes no timeout knobs, so the **web worker is left intact** — a browser timeout stays a typed `WorkerTimeoutError` reject without tearing the worker down.
- **Honest collateral** — recycling a worker aborts its other in-flight requests. Those siblings did **not** time out, so they reject with a plain retryable `Error` (mirroring the existing `terminate()` / `handleWorkerError` crash recovery) rather than a contradictory `WorkerTimeoutError`. Only the operation that actually exceeded its bound carries `WorkerTimeoutError`.
- **Diagnostics** — the pool stamps a `workerLabel` (`node-worker-N`) surfaced on the error.

```ts
relayers: { [sepolia.id]: node({ operationTimeout: 10 }) }

matchZamaError(error, {
  OPERATION_TIMEOUT: async (e) => {
    // worker was recycled — retry with your own backoff, or raise operationTimeout
  },
});
```

> A genuine timeout (no server response) carries no server-driven retry delay, so `WorkerTimeoutError` deliberately has no `retryAfterMs` — backoff is the consumer's policy. (Rate-limit/back-pressure delays — 429 `Retry-After` — are a different failure mode, handled on the SDK-236/239 line.)

## Test plan
- [x] `pnpm --filter @zama-fhe/sdk test` green — typed reject (operation/timeout/elapsed), `operationTimeout` override, recycle (terminate + re-init), `recycleWorkerOnTimeout: false`, **sibling requests reject as a plain recycle error (not a fake timeout)**, **web worker is not recycled on timeout**, node-config schema validation, `wrapDecryptError` passthrough, `matchZamaError` handler typing.
- [x] `tsc` (the `Complete<ZamaErrorCode>` guard forces the new code into the map) · type tests · `oxlint` · `oxfmt --check`; api-extractor reports + `llms` regenerated.

## Notes
- Base: `prerelease`. Public API: additive only (`WorkerTimeoutError`, `OPERATION_TIMEOUT`, `node()` timeout options).
- Recycle is intentionally Node-only and durations are in seconds; both differ from earlier revisions of this PR.

### DIFF
--- a/packages/sdk/src/worker/worker.base-client.ts
+++ b/packages/sdk/src/worker/worker.base-client.ts
@@ -27,7 +27,7 @@ import type {
   WorkerResponse,
 } from "./worker.types";
 import { deserializeError } from "../utils/error";
-import { WorkerTimeoutError } from "../errors/timeout";
+import { WorkerRecycledError, WorkerTimeoutError } from "../errors/timeout";
 
 /**
  * Timeout / recovery knobs shared by every worker client (node + web), so the
@@ -304,16 +304,17 @@ export abstract class BaseWorkerClient<TWorker, TConfig extends WorkerClientTime
     this.#worker = null;
     this.#initPromise = null;
     // The sibling requests still in flight did not time out — they are aborted
-    // collateral of recycling the worker. Reject them like a worker crash
-    // (see `terminate`/`handleWorkerError`) rather than mislabeling them as
-    // their own timeouts; only the operation that actually exceeded its bound
+    // collateral of recycling the worker. Reject them with a typed, retryable
+    // `WorkerRecycledError` (not their own `WorkerTimeoutError`, since they never
+    // exceeded a bound) so consumers can retry them instead of receiving a
+    // terminal `DecryptionFailedError` — the fate of a bare `Error` through
+    // `wrapDecryptError`. Only the operation that actually exceeded its bound
     // gets a `WorkerTimeoutError`.
-    const reason =
-      "Worker recycled after a timeout" +
-      (this.config.workerLabel ? ` (worker ${this.config.workerLabel})` : "");
     for (const [id, pending] of this.#pendingRequests) {
       clearTimeout(pending.timeoutId);
-      pending.reject(new Error(reason));
+      pending.reject(
+        new WorkerRecycledError({ operation: pending.type, worker: this.config.workerLabel }),
+      );
       this.#pendingRequests.delete(id);
     }
     this.logger.warn("[WorkerClient] recycled worker after a timeout", {
--- a/packages/sdk/src/worker/worker.base-client.ts
+++ b/packages/sdk/src/worker/worker.base-client.ts
@@ -27,6 +27,25 @@ import type {
   WorkerResponse,
 } from "./worker.types";
 import { deserializeError } from "../utils/error";
+import { WorkerTimeoutError } from "../errors/timeout";
+
+/**
+ * Timeout / recovery knobs shared by every worker client (node + web), so the
+ * base class can read them generically off `config`.
+ */
+export interface WorkerClientTimeoutConfig {
+  /** Per-operation timeout in **seconds**. Defaults to 30. */
+  operationTimeout?: number;
+  /** WASM-init timeout in **seconds**. Defaults to 60. */
+  initTimeout?: number;
+  /**
+   * Recycle (terminate + lazily re-init) the worker after an operation timeout,
+   * so a hung thread can't keep serving requests. Defaults to `true`.
+   */
+  recycleWorkerOnTimeout?: boolean;
+  /** Diagnostic label surfaced on {@link WorkerTimeoutError} (e.g. `node-worker-2`). */
+  workerLabel?: string;
+}
 
 /** Pending request tracker */
 interface PendingRequest<T> {
@@ -48,7 +67,7 @@ export const INIT_TIMEOUT_MS = 60_000;
  * Encapsulates all shared logic: pending request tracking, timeouts, init dedup, domain methods.
  * Subclasses implement the abstract hooks for platform-specific worker creation and messaging.
  */
-export abstract class BaseWorkerClient<TWorker, TConfig> {
+export abstract class BaseWorkerClient<TWorker, TConfig extends WorkerClientTimeoutConfig> {
   #worker: TWorker | null = null;
   #pendingRequests = new Map<string, PendingRequest<unknown>>();
   #initPromise: Promise<TWorker> | null = null;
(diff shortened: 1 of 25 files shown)


packages/sdk/src/worker/worker.base-client.ts
```diff
@@ -240,10 +277,49 @@ export abstract class BaseWorkerClient<TWorker, TConfig> {
...
+    // collateral of recycling the worker. Reject them like a worker crash
+    // (see `terminate`/`handleWorkerError`) rather than mislabeling them as
+    // their own timeouts; only the operation that actually exceeded its bound
+    // gets a `WorkerTimeoutError`.
+    const reason =
+      "Worker recycled after a timeout" +
+      (this.config.workerLabel ? ` (worker ${this.config.workerLabel})` : "");
+    for (const [id, pending] of this.#pendingRequests) {
+      clearTimeout(pending.timeoutId);
+      pending.reject(new Error(reason));
```
ankurdotb: [AUTOMATED] Siblings are rejected here with a bare `Error`. Through `wrapDecryptError` that classifies as **terminal `DecryptionFailedError`** — so a request aborted purely as *collateral* of another op's timeout (inherently retryable) reaches the consumer as `DECRYPTION_FAILED`, indistinguishable from a real decryption failure, and `matchZamaError` routes it to the terminal handler. The PR calls these "retryable," but nothing types them retryable. Consider a retryable *typed* error (a distinct "aborted-collateral, retry" cause, or a collateral-flagged `WorkerTimeoutError`) so consumers can branch/retry. Rationale in the review summary. Confidence: 80/100

ankurdotb (changes requested):
[AUTOMATED]

## Review — configurable + self-healing worker timeouts (SDK-237)

Solid PR: the recycle/self-healing machinery is correct (cascade fix, timer lifecycle / no double-settle, exactly-once sibling rejection, lazy re-init race, web-path-untouched, config validation — all traced and test-verified), durations are **seconds**, the API is additive, and the rebase onto merged-#489 is already reconciled (all three codes in the `Complete<>` map, `WorkerTimeoutError` passthrough on the `ctx` signature). Requesting changes on one semantic item + one test gap; two nits.

**1. [Change] Recycle-aborted siblings surface as *terminal* `DecryptionFailedError`, not retryable.** (see inline) `#recycleStuckWorker` rejects siblings with a bare `new Error("Worker recycled after a timeout …")`. That flows through `wrapDecryptError`, matches no typed passthrough/branch, and lands on `DecryptionFailedError` — terminal, non-retryable. So a request aborted purely as *collateral* of another op's timeout — inherently retryable — reaches the consumer as `DECRYPTION_FAILED`, indistinguishable from a genuine decryption failure; `matchZamaError` routes it to the terminal handler and it won't be retried. The PR documents these as "plain retryable Error," but there's no `retryable` field and it classifies terminal — the intent isn't realized in the type system. (It mirrors the existing crash-recovery path, so it's consistent rather than a regression — but the same gap applies there.) Please reject siblings with a **retryable typed** error so consumers can branch/retry: e.g. a dedicated "aborted-collateral, retry" cause, or a `WorkerTimeoutError` variant flagged as collateral — rather than a bare `Error` that degrades to terminal.

**2. [Change] No pool-level test for the cascade fix — the PR's headline claim.** The pool tests fully mock `NodeWorkerClient`, so the recycle path never runs at the pool boundary; the fix is only unit-tested on a single client. Please add a `NodeWorkerPool` test that times out a worker and confirms a subsequent dispatch re-inits that slot and succeeds (the hung worker isn't re-selected) — the exact scenario SDK-237 fixes.

**3. [Nit] Docs state "recycled (retryable)" unconditionally** (`errors.md`, `handle-errors.md`); recycling is default-`true` (`recycleWorkerOnTimeout`) and skipped for the web worker — "recycled by default" is accurate.

**4. [Nit] `WorkerTimeoutError.operation` surfaces internal `WorkerRequestType` codes** (`USER_DECRYPT`, …) on a public field — fine as a diagnostic, but worth a one-line doc note that it's an internal op code, not the public method name.

**FYI (not blocking):** `operationTimeout` / `initTimeout` have no upper bound; a value > ~24.8 days overflows `setTimeout`'s 32-bit limit → clamps to 1 ms → instant timeouts. Absurd config, noted for completeness.

Confidence: #1 80/100; #2 (test gap) high.

ankurdotb (approved):
[AUTOMATED] Re-reviewed at `a82041aa`. All requested changes addressed and verified in code + tests:

- **Sibling-abort → typed retryable `WorkerRecycledError`** (`WORKER_RECYCLED`) — replaces the bare `Error`; wired into `ZamaErrorCode`, the `Complete<>` map, the `wrapDecryptError` passthrough, and public exports, so a recycle-aborted sibling no longer collapses into terminal `DecryptionFailedError`.
- **Pool-level cascade test** — real `NodeWorkerClient`: timeout → terminate → next dispatch re-inits a fresh worker (the hung worker isn't re-selected).
- **Docs** now "recycled by default"; `operation` doc-noted as an internal op code (not the public method name).
- **Timeout cap** (`MAX_WORKER_TIMEOUT_SECONDS = 3600` + `z.lte`) closes the `setTimeout` 32-bit overflow.

The classify-once passthrough lists all typed causes (incl. both worker errors), units are seconds, and the API is additive. `WorkerRecycledError` is correctly kept distinct from `WorkerTimeoutError` without over-granular proliferation. Approving.

**Holistic note (not blocking):** after SDK-239/236/237 the decryption path has five retryable causes but only `RelayerRequestFailedError` exposes a programmatic `retryable` field — there's no `isRetryable()` helper / retryable-group / consolidated docs recipe, so consumers must hardcode a silently-growing code set. Filed as **SDK-248** for a uniform retryability signal (plus a "every typed cause passes through `wrapDecryptError`" completeness test). Explicitly out of scope for this PR.

PR #495, opened 2026-06-29
feat(examples): demonstrate confidentialTransferAndCall vault deposit [SDK-244]

zama-cremaud:
## What

Demonstrates the ERC-7984 `confidentialTransferAndCall` receiver-callback pattern (SDK-244, child of SDK-242) with a **real reacting contract** — a minimal confidential escrow vault — wired into the `react-wagmi` example. Until now the pattern was only exercised mechanically in the `test/` harness against OpenZeppelin's `ERC7984ReceiverMock` with a raw `data` field; no example app demonstrated the real use case.

## Changes

**Contract** (`contracts/`)
- `src/ConfidentialVault.sol` — minimal confidential escrow implementing `onConfidentialTransferReceived`: decodes the beneficiary from `data`, credits `shares[beneficiary]` via `FHE.add` + ACL, rejects a zero beneficiary by returning an encrypted `false` (atomic refund), guards `msg.sender == confidentialToken`. `withdraw()` returns the full balance and clears the slot to the uninitialized `bytes32(0)` sentinel — so the position reads as empty again and a repeat withdraw reverts (`NothingToWithdraw`). Plus `sharesOf()`.
- `script/DeployVault.s.sol` — deployment (env `CONFIDENTIAL_TOKEN`).

**Example** (`examples/react-wagmi/`)
- `VaultDepositCard` (`useConfidentialTransferAndCall`, `data = abi.encode(beneficiary)`), `VaultPositionCard` (vault-scoped permit + `useDecryptValues` + `withdraw`), gated on the bound confidential token. Config + `.env.example` vars.
- The position auto-refreshes after a deposit, and deposit failures surface the underlying revert reason (the SDK's wrapped `cause`) instead of a generic message.
- Bumps the example to `@zama-fhe/sdk` / `@zama-fhe/react-sdk` `3.3.0-alpha.3` (the `…AndCall` API is not in stable 3.2.0). No migration of existing app code was required.
- Docs: WALKTHROUGH section + README.
- Playwright e2e (`e2e/vault.spec.ts`, 5 tests).

## Verification

- `tsc --noEmit`, `next build`, `forge fmt`/`forge build` — green.
- Playwright suite **27/27** (5 new vault + 22 existing, no regressions).
- **End-to-end on real Sepolia**: deposit via `confidentialTransferAndCall` → decrypted vault position equals the deposited amount → withdraw → cUSDC round-trip conserved → post-withdraw `sharesOf` returns the empty sentinel and a repeat withdraw reverts. Vault deployed at `0xb13720bec167A576D715F5aA7C7d68b3dB0A4Ad7` (bound to cUSDC).

## Scope notes

- Scoped to `confidentialTransferAndCall` only. The `unwrapAndCall` leg of SDK-244 stays deferred (no on-chain `unwrapAndCall` yet; owned by Protocol Apps).
- The alpha pin should move back to stable once `3.3.0` ships.

### DIFF
--- a/contracts/src/ConfidentialVault.sol
+++ b/contracts/src/ConfidentialVault.sol
@@ -32,6 +32,9 @@ contract ConfidentialVault is IERC7984Receiver, ZamaEthereumConfig {
     {
         require(msg.sender == address(confidentialToken), UnauthorizedToken(msg.sender));
 
+        // Malformed `data` (< 32 bytes) reverts here, which reverts the whole transfer
+        // atomically — the deposit fails rather than refunding. The bound UI always encodes
+        // a full address, so this only guards against direct callers.
         address beneficiary = abi.decode(data, (address));
 
         // No beneficiary: return encrypted `false` so the token refunds the transfer atomically.
@@ -58,11 +61,12 @@ contract ConfidentialVault is IERC7984Receiver, ZamaEthereumConfig {
         euint64 amount = _shares[msg.sender];
         require(FHE.isInitialized(amount), NothingToWithdraw(msg.sender));
 
-        // Checks-effects-interactions: zero the balance before the outbound transfer.
-        euint64 zero = FHE.asEuint64(0);
-        FHE.allowThis(zero);
-        FHE.allow(zero, msg.sender);
-        _shares[msg.sender] = zero;
+        // Checks-effects-interactions: clear the balance before the outbound transfer.
+        // Reset the slot to the uninitialized (bytes32(0)) sentinel — not `FHE.asEuint64(0)`,
+        // which is an initialized handle. This makes `sharesOf` report empty and a repeat
+        // withdraw revert on the `isInitialized` guard above. (`euint64` is a user-defined
+        // value type, so `delete` is unavailable; wrap the zero handle directly.)
+        _shares[msg.sender] = euint64.wrap(bytes32(0));
 
         // The token runs the FHE ops inside `confidentialTransfer`, so it needs access to `amount`.
         FHE.allowTransient(amount, address(confidentialToken));
--- a/contracts/src/ConfidentialVault.sol
+++ b/contracts/src/ConfidentialVault.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.27;
+
+import {FHE, ebool, euint64} from "@fhevm/solidity/lib/FHE.sol";
+import {ZamaEthereumConfig} from "@fhevm/solidity/config/ZamaConfig.sol";
+import {IERC7984} from "@openzeppelin/confidential-contracts/interfaces/IERC7984.sol";
+import {IERC7984Receiver} from "@openzeppelin/confidential-contracts/interfaces/IERC7984Receiver.sol";
+
+/// @dev Minimal confidential escrow vault demonstrating the ERC-7984 `confidentialTransferAndCall`
+///      receiver-callback pattern. A depositor moves confidential tokens into the vault in a single
+///      transaction; `data` carries the beneficiary to credit. The beneficiary withdraws later.
+///      Illustrative example for the Zama SDK — not a production contract.
+contract ConfidentialVault is IERC7984Receiver, ZamaEthereumConfig {
+    IERC7984 public immutable confidentialToken;
+
+    mapping(address account => euint64 shares) private _shares;
+
+    event Deposit(address indexed from, address indexed beneficiary);
+    event Withdraw(address indexed beneficiary);
+
+    error UnauthorizedToken(address caller);
+    error NothingToWithdraw(address account);
+
+    constructor(IERC7984 token) {
+        confidentialToken = token;
+    }
+
+    /// @inheritdoc IERC7984Receiver
+    function onConfidentialTransferReceived(address, address from, euint64 amount, bytes calldata data)
+        external
+        returns (ebool)
+    {
+        require(msg.sender == address(confidentialToken), UnauthorizedToken(msg.sender));
+
+        address beneficiary = abi.decode(data, (address));
+
+        // No beneficiary: return encrypted `false` so the token refunds the transfer atomically.
+        if (beneficiary == address(0)) {
+            ebool rejected = FHE.asEbool(false);
+            FHE.allowTransient(rejected, msg.sender);
+            return rejected;
+        }
+
+        euint64 credited = FHE.add(_shares[beneficiary], amount);
+        FHE.allowThis(credited);
+        FHE.allow(credited, beneficiary);
+        _shares[beneficiary] = credited;
+
+        emit Deposit(from, beneficiary);
+
+        ebool accepted = FHE.asEbool(true);
+        FHE.allowTransient(accepted, msg.sender);
+        return accepted;
+    }
+
+    /// @dev Withdraw the caller's full vault balance back to their wallet.
+    function withdraw() external {
+        euint64 amount = _shares[msg.sender];
+        require(FHE.isInitialized(amount), NothingToWithdraw(msg.sender));
+
+        // Checks-effects-interactions: zero the balance before the outbound transfer.
+        euint64 zero = FHE.asEuint64(0);
+        FHE.allowThis(zero);
+        FHE.allow(zero, msg.sender);
+        _shares[msg.sender] = zero;
+
+        // The token runs the FHE ops inside `confidentialTransfer`, so it needs access to `amount`.
+        FHE.allowTransient(amount, address(confidentialToken));
+        confidentialToken.confidentialTransfer(msg.sender, amount);
+
+        emit Withdraw(msg.sender);
+    }
+
+    /// @dev Caller-decryptable confidential balance held for `account`.
+    function sharesOf(address account) external view returns (euint64) {
+        return _shares[account];
+    }
+}
--- a/examples/react-wagmi/src/components/VaultDepositCard.tsx
+++ b/examples/react-wagmi/src/components/VaultDepositCard.tsx
@@ -1,0 +1,113 @@
+"use client";
+
+import { useState } from "react";
+import { encodeAbiParameters, isAddress } from "viem";
+import { useConfidentialTransferAndCall } from "@zama-fhe/react-sdk";
+import type { Address } from "@zama-fhe/sdk";
+import { parseAmount } from "@/lib/parseAmount";
+import { SEPOLIA_EXPLORER_URL } from "@/lib/config";
+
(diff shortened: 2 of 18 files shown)


contracts/src/ConfidentialVault.sol
```diff
@@ -0,0 +1,78 @@
...
+        return accepted;
+    }
+
+    /// @dev Withdraw the caller's full vault balance back to their wallet.
+    function withdraw() external {
+        euint64 amount = _shares[msg.sender];
+        require(FHE.isInitialized(amount), NothingToWithdraw(msg.sender));
+
+        // Checks-effects-interactions: zero the balance before the outbound transfer.
+        euint64 zero = FHE.asEuint64(0);
```
ankurdotb: [AUTOMATED] **Blocking — root cause of the post-withdraw ghost position.**

Storing `FHE.asEuint64(0)` writes a freshly-allocated, *initialized* ciphertext handle into the slot. That handle is a computed (trivial-encryption) handle, so it is **never** `bytes32(0)`. The frontend's `isEncryptedValueZero()` only treats `0x00…00` / `"0x"` as empty (it never decrypts), so after a full withdraw `sharesOf` keeps returning a non-zero handle and the position card never returns to the "No vault position yet" state — it shows a `0.00` position with live Reveal/Withdraw buttons indefinitely (see VaultPositionCard.tsx:40).

Fix — clear the slot instead of storing an encrypted zero:

```solidity
// Checks-effects-interactions: clear the balance before the outbound transfer.
// `delete` resets the slot to the uninitialized (bytes32(0)) sentinel, so `sharesOf`
// reports empty and a repeat withdraw correctly reverts (see the guard above).
delete _shares[msg.sender];
```

This replaces the `zero` / `allowThis` / `allow` / assignment block (lines 62–65). The deposit path already tolerates an uninitialized slot via `FHE.add` (the first-ever deposit proves it), and the local `amount` handle keeps its ACL after the slot is cleared, so `allowTransient(amount, confidentialToken)` + the transfer below still work.

Confidence: 95/100

contracts/src/ConfidentialVault.sol
```diff
@@ -0,0 +1,78 @@
...
+
+        ebool accepted = FHE.asEbool(true);
+        FHE.allowTransient(accepted, msg.sender);
+        return accepted;
+    }
+
+    /// @dev Withdraw the caller's full vault balance back to their wallet.
+    function withdraw() external {
+        euint64 amount = _shares[msg.sender];
+        require(FHE.isInitialized(amount), NothingToWithdraw(msg.sender));
```
ankurdotb, line 62: [AUTOMATED] **Same root cause — this guard is currently ineffective after the first withdraw.** Once the slot holds `FHE.asEuint64(0)` (initialized), `FHE.isInitialized(amount)` is `true`, so a second `withdraw()` passes this `require` and proceeds to a 0-value `confidentialTransfer` (silent, no revert) plus a spurious `Withdraw` event. Switching the cleanup to `delete _shares[msg.sender]` (see comment below) makes this guard work as intended — a repeat withdraw then reverts with `NothingToWithdraw`.

Confidence: 92/100

examples/react-wagmi/src/components/VaultPositionCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+  // The vault holds an euint64 share per beneficiary; reading returns its handle.
+  const { data: shareHandle, refetch: refetchShares } = useReadContract({
+    address: vaultAddress,
+    abi: VAULT_ABI,
+    functionName: "sharesOf",
+    args: [connectedAddress],
+  });
+
+  const handle = shareHandle as EncryptedValue | undefined;
+  const hasPosition = handle !== undefined && !isEncryptedValueZero(handle);
```
ankurdotb, line 40: [AUTOMATED] **UI symptom of the contract issue.** Because `withdraw()` leaves an initialized encrypted-zero handle in the slot, `isEncryptedValueZero(handle)` returns `false` post-withdraw and `hasPosition` stays `true` — the card never falls back to the `vault-no-position` empty state. The clean fix is contract-side (`delete _shares[msg.sender]` in `ConfidentialVault.sol`); no change is needed here once that lands.

Note: bumping `vaultNonce` on withdraw (to force a remount) would *not* fix this — a remount re-reads the same non-zero handle.

Confidence: 95/100

contracts/src/ConfidentialVault.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// @inheritdoc IERC7984Receiver
+    function onConfidentialTransferReceived(address, address from, euint64 amount, bytes calldata data)
+        external
+        returns (ebool)
+    {
+        require(msg.sender == address(confidentialToken), UnauthorizedToken(msg.sender));
+
+        address beneficiary = abi.decode(data, (address));
```
ankurdotb, line 38: [AUTOMATED] FYI (non-blocking): `abi.decode(data, (address))` reverts if a direct (non-UI) caller passes `data` shorter than 32 bytes. The bound UI always encodes a full address, and a revert here reverts the whole transfer atomically (no fund loss), so this is safe — but a one-line comment noting that malformed `data` reverts the deposit (rather than refunding) would help readers who copy this receiver.

Confidence: 88/100

examples/react-wagmi/src/components/VaultPositionCard.tsx
```diff
@@ -0,0 +1,150 @@
...
+        </>
+      )}
+
+      {grantPermit.isError && (
+        <div className="alert alert-error card-status">
+          {grantPermit.error?.message ?? "Signing failed"}
+        </div>
+      )}
+      {decrypt.isError && (
+        <div className="alert alert-error card-status">{decrypt.error?.message}</div>
```
ankurdotb: [AUTOMATED] Nit (non-blocking): if `decrypt.error` is set but its `.message` is empty/undefined, this renders an empty red alert. The `grantPermit` alert above uses a `?? "Signing failed"` fallback — consider matching it here, e.g. `{decrypt.error?.message ?? "Decryption failed"}`. Mostly theoretical since SDK-thrown errors carry messages.

Confidence: 85/100

examples/react-wagmi/src/components/VaultDepositCard.tsx
```diff
@@ -0,0 +1,125 @@
...
+  onSuccess?: () => void;
+}
+
+// The SDK wraps transaction failures as `TransactionRevertedError` with the underlying
+// viem/RPC error in `cause`. Surface that reason so reverts aren't hidden behind a generic message.
+function depositErrorText(error: (Error & { cause?: unknown }) | null): string {
+  if (!error) return "";
+  const cause = error.cause as { shortMessage?: string; message?: string } | undefined;
+  const causeMsg = cause?.shortMessage ?? cause?.message;
+  return causeMsg ? `${error.message} — ${causeMsg}` : error.message;
```
ankurdotb: [AUTOMATED] Nit (non-blocking): same blank-alert edge as the position card — if `error.message` is `""` and there is no `cause`, `depositErrorText` returns `""` and the error alert renders empty. A final fallback string would make the failure path always show something. Theoretical for SDK-thrown errors.

Confidence: 70/100

ankurdotb (changes requested):
[AUTOMATED]

## Code Review — confidentialTransferAndCall vault example (SDK-244)

**Re-review of the fix:** the blocking ghost-position bug is **fixed correctly** and every inline comment is resolved — thank you.

- `withdraw()` now resets the slot to the `bytes32(0)` sentinel via `euint64.wrap(bytes32(0))` — the correct idiom (`delete` doesn't compile for a user-defined value type, good catch). Verified: `sharesOf` reports empty after a full withdraw, a repeat `withdraw()` reverts `NothingToWithdraw`, and re-deposit still works (`FHE.add` on the now-uninitialized slot = `amount`).
- The `abi.decode` FYI (clarifying comment) and both blank-alert nits (`"Deposit failed"` / `"Decryption failed"` / `"Withdrawal failed"` fallbacks) are addressed.

The contract is correct — ACL on every handle, the zero-beneficiary atomic refund, and CEI ordering all re-verified against the pinned OpenZeppelin ERC-7984 source.

## Remaining blocker — a regression test

Holding `CHANGES_REQUESTED` on one item: this withdraw→empty behavior should ship **with a test**. The ghost-position bug shipped precisely because `withdraw()` had no coverage, and nothing currently guards a regression.

The browser e2e can't cover it — `e2e/fixtures.ts` doesn't mock the `sharesOf` selector (it falls through to `"0x"`), so `hasPosition` is always false in the harness and the reveal/withdraw branch never renders. The right layer is a **Foundry (forge-fhevm) test** on `ConfidentialVault` asserting the behaviors that just broke:

- a deposit credits the beneficiary (`sharesOf` is initialized after `onConfidentialTransferReceived`);
- after `withdraw()`, `sharesOf(user) == bytes32(0)` — the assertion that pins this fix;
- a second `withdraw()` reverts `NothingToWithdraw`;
- a zero-beneficiary deposit refunds (the hook returns encrypted `false`).

The exact harness is your call — once a regression test lands I'm happy to approve.

ankurdotb (approved):
[AUTOMATED]

## Approved ✅ — regression test closes the last item (SDK-244)

The Foundry test (`contracts/test/ConfidentialVault.t.sol`) fully resolves the remaining blocker. It exercises the **real** deposit path (`ERC7984Mock.confidentialTransferAndCall` → `onConfidentialTransferReceived`) on the forge-fhevm cleartext harness and decrypts to assert plaintext — covering exactly the behaviors that broke, and more:

- deposit credits the beneficiary / debits the depositor; credits a different beneficiary; **accumulates** across deposits;
- **`sharesOf(holder) == bytes32(0)` after `withdraw()`** — the assertion that pins the `euint64.wrap(bytes32(0))` fix — plus funds returned to the holder;
- repeat `withdraw()` reverts `NothingToWithdraw` (and withdraw-without-position reverts);
- zero-beneficiary deposit refunds atomically (balance whole, nothing credited);
- bonus: the anti-spoof `UnauthorizedToken` guard and the Deposit/Withdraw events.

A revert to `FHE.asEuint64(0)` would fail both the clear-slot and repeat-withdraw tests, so the regression is genuinely guarded. The contract and components are unchanged since my last review, so the verified fix still holds. Nice work — LGTM.

**Nit (non-blocking, CI convention):** `.github/workflows/contracts-test.yml` runs `cd contracts && forge test -vvv` inline, while this PR also adds a `contracts:test` pnpm script. Consider having the CI step call `pnpm contracts:test` (or a `:ci` variant that carries `-vvv`) so the workflow invokes the pnpm namespace rather than duplicating the command. Doesn't block.

PR #497, opened 2026-06-29
feat(sdk): persist pending unshield state internally

ghermet:
## What

`WrappedToken` now owns pending-unshield persistence across the two-phase unshield flow:

- **Saves** the unwrap tx hash right after phase 1 submits (keyed on the wrapper address).
- **Clears** it once phase 2 finalizes (and on `resumeUnshield` success — both share `#waitAndFinalizeUnshield`).
- Both are best-effort (`swallow` → `logger.warn`); a storage failure never aborts the unshield.
- Adds **`WrappedToken.getPendingUnshield(): Promise<Hex | null>`** so apps can detect an interrupted unshield without hand-managing storage keys, then pass the hash to `resumeUnshield`.

## Why

The two-phase unshield is only recoverable after a reload/crash if the unwrap tx hash is persisted between phases. Previously each app wired this up by hand via the `UnshieldPhase1Submitted` event — which carries only `txHash`, not the wrapper address — forcing a fragile `activeUnshieldToken` correlation module. `WrappedToken` already has both `this.address` and `this.sdk.storage`, so it can do this itself. Loading/resuming stays caller-driven (auto-resume on load would fire a wallet tx unprompted).

The exported `savePendingUnshield` / `loadPendingUnshield` / `clearPendingUnshield` helpers are unchanged, so existing integrations keep working.

## Tests

TDD, 4 new cases in `wrapped-token.test.ts`: save-survives-interruption, clear-on-success, resume-clears, and `getPendingUnshield`. Full SDK suite green (1717 passed); typecheck, lint, and api-report regenerated (`getPendingUnshield` added to `etc/sdk.api.md` + `etc/sdk-query.api.md`).

## Follow-up (not in this PR)

Example apps (`example-ingen`, `example-hoodi`, `react-ethers`) can now drop the manual save/clear wiring and the `activeUnshieldToken` module — deferred.


ankurdotb:
Screenshot 2026-06-30 at 11 01 30
See this @ghermet and consider thoughts

docs/gitbook/src/guides/unshield-tokens.md
```diff
@@ -114,10 +108,14 @@ if (pending) {
...
-2. **`loadPendingUnshield`** -- on mount, check if there is an incomplete unshield.
-3. **`resumeUnshield`** -- pick up where the SDK left off by polling for the proof and submitting the finalize transaction.
-4. **`clearPendingUnshield`** -- clean up storage once finalization is confirmed.
+1. **`getPendingUnshield`** -- returns the unwrap transaction hash of an interrupted unshield, or `null` if none is pending. The SDK saved it automatically during phase 1.
+2. **`resumeUnshield`** -- picks up where the SDK left off by polling for the proof and submitting the finalize transaction. On success the SDK clears the persisted state for you.
+
+Resuming is intentionally caller-driven: surface a "resume" prompt rather than finalizing on load, so you never trigger a wallet transaction the user did not initiate.
+
+{% hint style="info" %}
+For custom flows, the lower-level `savePendingUnshield`, `loadPendingUnshield`, and `clearPendingUnshield` helpers remain exported from `@zama-fhe/sdk`. You only need them if you bypass `unshield()` / `resumeUnshield()` and orchestrate `unwrap` + `finalizeUnwrap` yourself.
```
ankurdotb: [AUTOMATED] This hint presents the raw helpers as a permanent "for custom flows" API, which works against SDK-197's goal of demoting them. Suggest reframing as lower-level helpers being phased out — e.g. "prefer `getPendingUnshield` + `resumeUnshield`; the raw helpers remain only until the examples migrate" — so the docs don't entrench a surface we intend to remove.

packages/sdk/src/query/invalidation.ts
```diff
@@ -50,6 +60,7 @@ export function invalidateAfterShield(queryClient: QueryClientLike, tokenAddress
 export function invalidateAfterUnshield(queryClient: QueryClientLike, tokenAddress: Address): void {
   invalidateBalanceQueries(queryClient, tokenAddress);
   invalidateUnderlyingAllowanceQueries(queryClient, tokenAddress);
+  invalidatePendingUnshieldQueries(queryClient, tokenAddress);
```
ankurdotb, line 63: [AUTOMATED] `pendingUnshield` is invalidated here, but `invalidateAfterUnshield` only runs on the mutations' `onSuccess`. The pending hash is saved mid-flow in `#waitAndFinalizeUnshield`, so on an *interrupted* finalize (the exact case this feature targets) the mutation rejects and no invalidation fires — an already-mounted `usePendingUnshield` keeps its cached `null` until a remount / window-refocus. The on-reload path works (`staleTime: 0` → refetch on mount), so this is non-blocking, but consider invalidating on `onSettled` in the unshield / unshieldAll hooks so in-session interruptions surface reactively. Confidence: 75.

packages/react-sdk/src/unshield/use-pending-unshield.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param tokenAddress - Address of the confidential wrapper contract.
+ * @returns Suspense query result with `data: Hex | null`.
+ *
+ * @example
+ * ```tsx
+ * const { data: pending } = usePendingUnshieldSuspense("0xWrapper");
+ * ```
+ */
+export function usePendingUnshieldSuspense(tokenAddress: Address) {
```
ankurdotb, line 53: [AUTOMATED] Nit: `usePendingUnshieldSuspense` has no test (the CI-flagged 60% / lines 54-56 coverage gap). A small test rendering it inside a `<Suspense>` boundary and asserting `data` resolves to the persisted hash / `null` would close it.

ankurdotb (commented):
[AUTOMATED]

Following up on the SDK-197 triage card @ankurdotb shared above — confirmed direction: **remove the four raw helpers from the public surface.**

**They all stay in the codebase — none become dead code:**
- `savePendingUnshield`, `loadPendingUnshield`, `clearPendingUnshield` are used by `WrappedToken` (`wrapped-token.ts:367, 499, 533`).
- `loadPendingUnshieldRequest` is used internally by `loadPendingUnshield` (`pending-unshield.ts:89`).

`WrappedToken` imports them via the relative `./pending-unshield` path, **not** the public barrel, and there is no `./token` subpath export — so the *only* public exposure is `packages/sdk/src/index.ts:88-92`. Removing those re-exports (plus the now-unused `PendingUnshieldRequest` type) fully de-publicizes them with zero impact on internal usage. Public callers keep `getPendingUnshield()` (read) + `unshield` / `resumeUnshield` (orchestrated write) — which is exactly SDK-197's "inspect, never write" intent.

**Prerequisite:** the only consumers of the public helpers are our 7 example apps (react-hoodi/ethers/ingen/bnb/viem/wagmi/turnkey), each via the fragile `activeUnshield` bridge. Migrate them to `getPendingUnshield` / `usePendingUnshield` first (this PR already plans that), then drop the public exports in the same PR or the immediately-following one. Pre-GA is the cheap window to remove public surface — once GA locks it in it is a deprecation cycle.

Otherwise this is clean — three-layer architecture, consistent address normalization, api-reports regenerated, meaningful TDD. Two smaller engineering notes inline; non-blocking overall, hence a comment rather than request-changes.

ghermet:
@ankurdotb Examples can't be updated in this PR because they rely on the npm package not the workspace package

zama-cremaud:
### 1. Stale pending-unshield state has a permanent dead end — no public API can clear it

**Severity: should fix before merge · Confidence: 88/100**
**Files:** `packages/sdk/src/token/wrapped-token.ts:527-533` (clear only on finalize success), `packages/sdk/src/index.ts:87` / `packages/sdk/src/token/index.ts:1-3` (public `clearPendingUnshield` removed)

`clearPendingUnshield` now runs in exactly one place: the success tail of `#waitAndFinalizeUnshield`, after `finalizeUnwrap` resolves. Realistic paths leave the persisted hash stale forever:

- The session is interrupted **after** the finalize tx is submitted but **before** the clear runs (tab closed/crash in that window), or the swallowed `storage.delete` fails.
- The unwrap gets finalized elsewhere (second tab, a retried resume racing itself). Note `resumeUnshield` **re-saves** the hash at the top of `#waitAndFinalizeUnshield` before validating it, so a second tab can even restore an entry the first tab just cleared.

Once stale: `getPendingUnshield()` / `usePendingUnshield()` return the hash indefinitely; every `resumeUnshield()` reverts with `TransactionRevertedError` ("already finalized") and — since clear only runs on success — never cleans up. The UI built exactly per this PR's docs (`ResumeUnshieldGuard`) shows a "Resume unshield" button in a permanent failing loop. For `unshieldAll` there is no remaining balance to run a fresh unshield that would overwrite-then-clear the entry.

The PR is self-aware here but the replacement guidance is not implementable: `docs/gitbook/src/guides/handle-errors.md` / `reference/sdk/errors.md` change the "unwrap already finalized" fix from *"call `clearPendingUnshield()`"* to *"the unshield is complete — stop prompting to resume"* — but an app cannot persist "stop prompting" across reloads, because the read API keeps returning the stale hash and the clear API was just un-exported. The only workaround is reaching into `sdk.storage` with the undocumented `zama:pending-unshield:<checksummedAddress>` key format.

**Suggested fix (either or both):**
- Add a public escape hatch: `WrappedToken.clearPendingUnshield(): Promise<void>` next to `getPendingUnshield()` (delegating to the internal helper).
- In `resumeUnshield`, when finalize reverts because the unwrap is already finalized, clear the persisted state before rethrowing (the revert is durable proof the entry can never succeed).

Also consider not re-saving in `#waitAndFinalizeUnshield` when entered via `resumeUnshield` (it persists whatever hash the caller passes, before any validation).

ankurdotb (approved):
[AUTOMATED]

Approving — re-reviewed at head `ef65f967`; all earlier comments are addressed and the three review threads resolved.

- **Surface reduction done.** The four raw helpers are `@internal` and removed from the public barrel; verified physically absent from the published `.d.ts` with no leaks or dangling imports. Public callers get `getPendingUnshield()` + orchestrated `unshield` / `resumeUnshield`.
- **`onSettled` invalidation added** to `useUnshield` / `useUnshieldAll` so an interrupted finalize surfaces reactively; `resumeUnshield` correctly needs none.
- **Suspense tests added; docs hints corrected.**
- **`prerelease` merge is clean** — #489's error-table rows and this PR's `getPendingUnshield` edits coexist; `pnpm llm:check` and `pnpm api-report:check` pass with no drift; CI green.

One follow-up, tracked in **SDK-251**: migrate the 7 example apps off the now-internal helpers before/with the release that ships this (they build against the published SDK today, so they don't gate this PR). That ticket also flags the `refactor:` vs `BREAKING CHANGE:` semver question for the release bump.

PR #517, opened 2026-07-02
refactor(sdk): single source of truth for WrappersRegistry config [SDK-127]

zama-cremaud:
## Summary

Closes [SDK-127](https://linear.app/zama/issue/SDK-127) (see "Deliberate deviation" below).

`WrappersRegistry` received its configuration through too many inconsistent paths. This PR unifies them **without any breaking change**:

- **`registryTTL` is now honored by all 8 query factories.** The 7 factories that hardcoded `DEFAULT_STALE_TIME_MS = 86400 * 1000` now read `staleTime` from `sdk.registry.ttlMs` (matching `listPairsQueryOptions`, which already read `registry.ttlMs`). The default TTL equals the old hardcoded value, so behavior is unchanged unless `registryTTL` is configured — which is exactly the bug being fixed. The hardcoded constant (internal, never exported) is removed.
- **`createWrappersRegistry()` now produces a registry consistent with `sdk.registry`.** It previously dropped the chain-derived registry addresses (`FheChain.registryAddress` from `config.chains`), so a custom chain known to `sdk.registry` was silently unknown to a created instance. It now inherits them, with explicit per-call overrides still taking precedence. The `ZamaSDK.md` reference already documented this as the intended behavior — the code was the bug.
- **`DefaultRegistryAddresses` is deprecated** (`@deprecated` JSDoc + docs hints) in favor of reading `registryAddress` from the chain presets (`sepolia.registryAddress`, …). Removal is deferred to the next major since it is a public export; the constructor's default merge is untouched.

## Deliberate deviation from the ticket

The ticket's "route all 8 query factories through `WrappersRegistry` methods" item is **intentionally not done**:

1. **The premise doesn't hold.** The class-level TTL cache only backs `listPairs` / `getConfidentialToken` / `getUnderlyingToken` — already routed. The 7 low-level registry methods are uncached pass-throughs to `readContract`, so routing the remaining factories through them would unify nothing.
2. **It would introduce a cache-poisoning race.** Registry methods resolve the address at call time via `provider.getChainId()` (global mutable state with wagmi). A mid-flight chain switch would store the new chain's data under the old chain's query key for up to 24h. The current query-key-pinned `readContract` pattern (SDK-92) fails loudly instead.

Two real issues discovered during this analysis are worth their own tickets (Linear write access unavailable from this session — see PR checklist):

- **(a) Next major:** remove `DefaultRegistryAddresses` + settle `createWrappersRegistry()` fate.
- **(b) Registry lookup freshness inconsistency:** the class cache holds negative lookups ("no wrapper") for 5 min (`NEGATIVE_CACHE_TTL_MS`) while TanStack factories cache `[false, 0x0]` for 24h — so `useConfidentialTokenAddress` can report "no wrapper" for a day after registration while imperative discovery sees it after 5 min. Plus the pre-existing `listPairs` address-resolution race above.

## Test plan

- [x] New: `staleTime` assertion across all 7 sdk-based factories with a non-default `registryTTL`
- [x] New: `createWrappersRegistry()` inherits chain-derived addresses; overrides take precedence
- [x] Full `sdk` (1218) and `react-sdk` (331) suites green; typecheck, lint, format green
- [x] API reports regenerated — only diff is the `@deprecated` tag on `DefaultRegistryAddresses`; no signature changes
- [x] `llms-full.txt` regenerated

### DIFF
--- a/docs/gitbook/src/reference/sdk/network-presets.md
+++ b/docs/gitbook/src/reference/sdk/network-presets.md
@@ -140,13 +140,19 @@ import { sepolia, mainnet } from "@zama-fhe/sdk/chains";
 
 ## DefaultRegistryAddresses
 
+{% hint style="warning" %}
+**Deprecated.** Read `registryAddress` from the chain presets instead (e.g. `sepolia.registryAddress`). This export will be removed in the next major version.
+{% endhint %}
+
 A convenience export of built-in registry addresses for known chains (Mainnet, Sepolia, Hoodi) as a `Record<number, Address>` map. Used internally by the [WrappersRegistry](./WrappersRegistry.md) class.
 
 ```ts
+// Deprecated
 import { DefaultRegistryAddresses } from "@zama-fhe/sdk";
 
-// { 1: "0xeb5015fF...", 11155111: "0x2f0750Bb...", 560048: "0x1807aE2f..." }
-console.log(DefaultRegistryAddresses);
+// Preferred
+import { sepolia } from "@zama-fhe/sdk/chains";
+console.log(sepolia.registryAddress);
 ```
 
 {% hint style="info" %}
--- a/docs/gitbook/src/reference/sdk/WrappersRegistry.md
+++ b/docs/gitbook/src/reference/sdk/WrappersRegistry.md
@@ -10,7 +10,7 @@ High-level read interface for the on-chain `ConfidentialTokenWrappersRegistry` c
 ## Import
 
 ```ts
-import { WrappersRegistry, DefaultRegistryAddresses } from "@zama-fhe/sdk";
+import { WrappersRegistry } from "@zama-fhe/sdk";
 ```
 
 ## Usage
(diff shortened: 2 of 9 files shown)


docs/gitbook/src/reference/sdk/network-presets.md
```diff
@@ -140,13 +140,19 @@ import { sepolia, mainnet } from "@zama-fhe/sdk/chains";
...
 
+{% hint style="warning" %}
+**Deprecated.** Read `registryAddress` from the chain presets instead (e.g. `sepolia.registryAddress`). This export will be removed in the next major version.
+{% endhint %}
+
 A convenience export of built-in registry addresses for known chains (Mainnet, Sepolia, Hoodi) as a `Record<number, Address>` map. Used internally by the [WrappersRegistry](./WrappersRegistry.md) class.
 
 ```ts
+// Deprecated
 import { DefaultRegistryAddresses } from "@zama-fhe/sdk";
```
ghermet, line 151: Is this relevant here? is this doc a migration doc?

ghermet (approved):

PR #520, opened 2026-07-05
feat(sdk): Option A — self-contained off-chain CleartextRelayer

poppyseedDev:
## Summary

**Draft — proposes Option A** for restoring cleartext support on top of the relayer→`@fhevm/sdk` migration.

Adds a **self-contained, off-chain `CleartextRelayer`** that implements the new `FhevmRelayerSDK` interface but is backed by the old off-chain cleartext logic — reading `CleartextFHEVMExecutor.plaintexts(handle)`, checking ACL in TypeScript, and signing with the hardcoded mock keys. `config/cleartext.ts` now builds `new CleartextRelayer(chain)` instead of routing cleartext through `@fhevm/sdk`'s `createFhevmCleartextClient`.

## Why

Keeps cleartext working for the V4 release using **only the mock contracts partners already run via forge-fhevm** — no new `Cleartext*` host set to deploy, no dependency on the ~1-month ACL/executor reconciliation. It is byte- and key-compatible with forge-fhevm as shipped today.

## What's here (~1112 LOC)

- `relayer/cleartext/cleartext-relayer.ts` — new `CleartextRelayer` implementing `FhevmRelayerSDK`
- `relayer/cleartext/{eip712,fhe-type,handle,constants}.ts` — supporting logic restored/reshaped from `10d02bf1`
- `config/cleartext.ts` + `chains/configs.ts` — wiring
- `relayer/cleartext/__tests__/*` — unit + live-anvil integration tests

## Trade-offs (vs Option B)

- ✅ Self-contained in `sdk/`; `@fhevm/sdk` untouched. Guaranteed forge-fhevm compatibility.
- ❌ Divergence is avoided, not resolved — `@fhevm/sdk`'s on-chain cleartext still ships and drifts; two implementations to maintain.

Option A is the lower-risk fallback; Option B (contained change inside `@fhevm/sdk`'s cleartext modules) remains the recommended holistic play. Opening this as a draft to make the concrete diff reviewable alongside that decision.


packages/sdk/src/relayer/cleartext/fhe-type.ts
```diff
@@ -0,0 +1,79 @@
...
+  return value in FheTypeIdToValueTypeName;
+}
+
+export function isValueTypeName(value: unknown): value is ValueTypeName {
+  return typeof value === "string" && value in ValueTypeNameToId;
+}
+
+export function fheTypeIdFromValueTypeName(name: string): FheTypeId {
+  if (!isValueTypeName(name)) {
+    throw new Error(`Unsupported FHE type '${name}'`);
```
ghermet: [AUTOMATED] **Nit:** these three guards throw a plain `Error` (also lines 69 and 76), but CONTRIBUTING.md §178 says *"all SDK errors extend `TokenError` … use specific subclasses (`EncryptionFailedError`, `SigningRejectedError`, etc.)"*. `cleartext-relayer.ts` in this same PR already uses `EncryptionFailedError` for equivalent value/type guards, so these are inconsistent — an `EncryptionFailedError` (or a dedicated type-mapping error) would let callers catch them uniformly.

Confidence: 88/100

packages/sdk/src/relayer/cleartext/cleartext-relayer.ts
```diff
@@ -0,0 +1,431 @@
...
+    }
+    const results = await Promise.all(
+      handles.flatMap((h) => [this.#persistAllowed(h, owner), this.#persistAllowed(h, contractAddress)]),
+    );
+    for (let i = 0; i < handles.length; i++) {
+      if (!results[i * 2]) {
+        throw new NotEntitledError({ encryptedValue: handles[i]!, contractAddress, account: owner });
+      }
+      if (!results[i * 2 + 1]) {
+        throw new DecryptionFailedError(
```
ghermet, line 399: [AUTOMATED] **Nit (optional):** the contract-side ACL denial throws `DecryptionFailedError`, but the symmetric owner-side denial a few lines up throws `NotEntitledError`. Both are entitlement failures — a consumer catching `NotEntitledError` to surface "no access" will miss this branch. Consider `NotEntitledError` here too for consistency (the message text is already clear either way, so this is minor).

Confidence: 80/100

packages/sdk/src/relayer/cleartext/handle.ts
```diff
@@ -0,0 +1,61 @@
...
+  index: number,
+  fheType: FheTypeId,
+  aclAddress: Address,
+  chainId: bigint,
+): Hex {
+  if (!Number.isInteger(index) || index < 0 || index > 255) {
+    throw new EncryptionFailedError("index must be an integer between 0 and 255");
+  }
+
+  const blobHash = keccak256(concat([toHex(RAW_CT_HASH_DOMAIN_SEPARATOR), mockCiphertext]));
```
ghermet, line 44: [AUTOMATED] **FYI:** the `ZK-w_rct` domain separator is applied twice in the TS flow — once inside `computeMockCiphertext` (`keccak256("ZK-w_rct" || inner)`) and again here (`keccak256("ZK-w_rct" || mockCiphertext)`), whereas on-chain `InputProofHelper` applies it once to the raw ciphertext. This is **harmless** — the on-chain verifier never re-derives the handle from the ciphertext (it only checks the submitted handle is in the signed list), and the relayer uses its own handle consistently everywhere. Flagging only because it reads like a bug to a future maintainer; a one-line comment noting the mock derivation intentionally diverges here would save someone a double-take.

Confidence: 80/100

ghermet (commented):
[AUTOMATED]

## Code Review — Option A CleartextRelayer

**Correctness: verified clean.** I traced the handle bit-packing, the `InputVerification` / `Decryption` EIP-712 domains, the input-proof and decryption-proof wire layouts, and the user / delegated / public-decrypt ACL paths against the actual `forge-fhevm` on-chain contracts (`InputProofHelper`, `InputVerifier`, `KMSVerifier`, `KMSDecryptionProofHelper`, `CleartextFHEVMExecutor`, ACL) and ran the unit suite (10/10 pass). No correctness or security issues found:

- Handle layout (`index<<80`, `chainId64<<16`, `fheType<<8`, version byte, `PREHANDLE_MASK`) matches `InputProofHelper.computeInputHandle` and the on-chain extraction masks.
- inputProof `[numHandles][numSigners][handles][sig][extraData=cleartexts]` and the `2 + 32n + 65s` offsets match `InputVerifier.verifyInput` / `CleartextFHEVMExecutor._tryReadCleartextFromProof`.
- EIP-712 domain uses `gatewayChainId` while the embedded handle / `contractChainId` use the host chain id — the split is correct and proven by the signature-recovery tests.
- Non-delegated user-decrypt correctly requires **both** owner and contract `persistAllowed` plus the `owner !== contract` guard; delegated path relies on `isHandleDelegatedForUserDecryption` (which subsumes the owner grant on-chain). Positional `Promise.all` alignment in the `flatMap` loop is correct — no off-by-one.
- All 15 `FhevmRelayerSDK` methods + `chain` are implemented; the `checkSignaturesArgs` field names match the consumer, so the boundary casts hide no shape mismatch.

Verdict: **non-blocking**. A few minor nits below.

**Considered and cleared** (not issues): the delegated-path `NotEntitledError` choice is correctly reasoned — local ACL is authoritative in cleartext mode with no gateway propagation lag, so a retryable `DelegationNotPropagatedError` would be wrong; `FORBIDDEN_CHAIN_IDS` covering only mainnet/sepolia is fine because the required `executorAddress` + mock-only `plaintexts` mapping already prevent real-ciphertext exfiltration on other chains; dropping `options` on encrypt/decrypt is safe since cleartext makes no authenticated relayer calls.

PR #521, opened 2026-07-06
feat(sdk): expose WrappedToken.wrap() for two-signature shield

ghermet:
## Summary

Exposes a public `WrappedToken.wrap()` method so callers can split the shield flow across two signatures (`approveUnderlying()` then `wrap()`) — an escape hatch on top of the automatic `shield()` orchestration.

- Adds `WrappedToken.wrap(amount, options?)` — validates ERC-20 balance and the allowance granted to the wrapper (public reads, no signing) and fails early with a clear error.
- Adds `WrapOptions { to?, onWrapSubmitted? }`.
- Adds `InsufficientAllowanceError` (code `INSUFFICIENT_ALLOWANCE`), wired into the `matchZamaError` dispatch map.
- Adds `wrap` transaction-operation metadata — a standalone wrap completes a shield, so it emits `ShieldSubmitted` with `shieldPath: "approveAndWrap"`.
- Exports `WrapOptions` and `InsufficientAllowanceError` from the package barrel.

The `shield()` `approveAndWrap` leg is unchanged behaviourally — it still emits `shield:approveAndWrap`. `wrap()` and the shield leg both call `submitTransaction({ operation, config: wrapContract(...), onSubmitted })` directly.

## Test plan

- New `packages/sdk/src/token/__tests__/wrap.test.ts` (5 tests): sufficient balance/allowance submits `wrap`; `options.to` routing; `onWrapSubmitted` fires; `InsufficientERC20BalanceError` on low balance; `InsufficientAllowanceError` on low allowance.
- `pnpm --filter @zama-fhe/sdk typecheck` — clean.
- Full suite: 217 files / 1921 tests passing (includes existing shield suite — no regression).


ghermet:
Superseded: rebuilding this branch cleanly on top of `prerelease` so the PR contains only the wrap work (was dragging in the entire unmerged @fhevm/sdk migration). New PR opened from feature/unbundle-shield-and-unshield-on-prerelease.

PR #522, opened 2026-07-06
feat(sdk): expose wrap() + useWrap for two-signature shield

ghermet:
## Summary

Adds a public `wrap()` escape hatch so callers can split the shield flow across two signatures (`approveUnderlying()` then `wrap()`), plus the matching React hook and query mutation options. `shield()` remains the recommended one-call surface; `wrap()` is for callers that genuinely need per-phase control.

- **`WrappedToken.wrap(amount, options?)`** — validates ERC-20 balance and the allowance granted to the wrapper (public reads, no signing), fails early with a clear error. Emits `ShieldSubmitted` with `shieldPath: "approveAndWrap"` (a standalone wrap completes a shield).
- **`WrapOptions { to?, onWrapSubmitted? }`**.
- **`InsufficientAllowanceError`** (code `INSUFFICIENT_ALLOWANCE`), wired into the `matchZamaError` dispatch map.
- **`wrapMutationOptions` + `invalidateAfterWrap`** query helpers.
- **`useWrap`** React hook (escape-hatch, mirrors the existing shield hooks; pairs with `useApproveUnderlying`).
- Exports the new public symbols from both package barrels.

The `shield()` `approveAndWrap` leg is behaviourally unchanged.

## Base branch

Targets `prerelease` (not `main`) — this work stacks on the in-flight `@fhevm/sdk` migration and uses its post-migration APIs.

## Docs

- **`guides/shield-tokens.md`** — new "Manual approve + wrap (escape hatch)" section covering the two-signature flow (Core SDK + React tabs) and warning that `wrap()` skips ERC-1363 `transferAndCall` routing, so it must not be used to recreate `shield()` by hand.
- **`reference/sdk/WrappedToken.md`** — documents `wrap()`.
- **`reference/sdk/errors.md`** + **`guides/handle-errors.md`** — document `InsufficientAllowanceError`.
- LLM corpus / `llms-full.txt` regenerated; API reports (`sdk.api.md`, `sdk-query.api.md`, `react-sdk.api.md`) regenerated.

## Test plan

- `packages/sdk/src/token/__tests__/wrap.test.ts` — balance/allowance sufficiency, `options.to` routing, `onWrapSubmitted`, and both insufficient-balance / insufficient-allowance error paths.
- `packages/react-sdk/src/shield/__tests__/use-wrap.test.tsx` — hook behaviour.
- `test/playwright/tests/wrap.spec.ts` — e2e coverage of the two-signature shield (approve + wrap) against USDT/cUSDT (the approve+wrap path, since USDT is not ERC-1363).
- Full unit suite: 218 files / 1925 tests passing. `@zama-fhe/sdk` and `@zama-fhe/react-sdk` typecheck clean.


ankurdotb:
Merge conflicts to be resolved

ghermet:
The merge conflicts have been resolved — `prerelease` was merged into this branch in `5ffc59c`, and all 12 CI checks pass on that head. `git merge-tree` against the current `prerelease` reports no conflicts (`prerelease` is fully contained in the branch). This looks ready for another review pass. 🤖

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/sdk/src/query/wrap.ts
```diff
@@ -0,0 +1,19 @@
+import type { WrappedToken } from "../token/wrapped-token";
+import type { TransactionResult } from "../types";
+import type { MutationFactoryOptions } from "./factory-types";
+import type { Address } from "viem";
+
+/** Variables for {@link wrapMutationOptions}. */
+export interface WrapParams {
```
ankurdotb: [AUTOMATED] `WrapParams` is standalone (`{ amount, to }`) instead of `extends WrapOptions`, and `wrapMutationOptions` (line 17) forwards only `{ amount, to }`. So `WrapOptions.onWrapSubmitted` — a documented capability on `WrappedToken.wrap()` that fires at *submit* time (which the hook's `onSuccess` can't replicate) — is unreachable via `useWrap`. Asymmetric with `ShieldParams extends ShieldOptions` (which forwards `...rest`). Suggest `interface WrapParams extends WrapOptions` and forward `...rest` in the mutationFn, matching `shieldMutationOptions`.

Confidence: 80/100

packages/sdk/src/events/sdk-events.ts
```diff
@@ -244,6 +244,13 @@ export const transactionOperationMetadata = {
       shieldPath: "approveAndWrap" as const,
     }),
   },
+  wrap: {
+    submittedEvent: (txHash: Hex) => ({
+      type: ZamaSDKEvents.ShieldSubmitted,
+      txHash,
+      shieldPath: "approveAndWrap" as const,
```
ankurdotb: [AUTOMATED] The standalone `wrap` operation emits `ShieldSubmitted` with `shieldPath: "approveAndWrap"` — the same shape as `shield()`'s internal wrap leg (line 244). A consumer listening for `ShieldSubmitted` (analytics, progress UI) can't tell a caller-initiated `wrap()` from a real `shield()`. This breaks the precedent the codebase set for exactly this problem: unshield's phases emit distinct `UnshieldPhase1/2Submitted` so they aren't conflated with standalone `unwrap`/`finalizeUnwrap`. No test asserts `wrap()`'s emitted event, so it doesn't look deliberate. Suggest a distinct `ZamaSDKEvents.WrapSubmitted`, keeping `ShieldSubmitted`/`approveAndWrap` exclusive to `shield()`.

Confidence: 65/100

packages/sdk/src/token/wrapped-token.ts
```diff
@@ -198,6 +200,73 @@ export class WrappedToken extends Token {
...
+    }
+    if (erc20Balance < amount) {
+      throw new InsufficientERC20BalanceError(
+        `Insufficient ERC-20 balance: requested ${amount}, available ${erc20Balance} (token: ${underlying})`,
+        { requested: amount, available: erc20Balance, token: underlying },
+      );
+    }
+
+    const allowance = await this.sdk.provider.readContract(
+      allowanceContract(underlying, userAddress, this.address),
```
ankurdotb: [AUTOMATED] Nit: the ERC-20 balance read earlier in this method is wrapped in try/catch → `ERC20ReadFailedError`, but this allowance read isn't — a transient RPC failure here escapes as a raw provider error instead of the typed SDK error the method's JSDoc advertises. It's consistent with the rest of the codebase (`#ensureAllowance`, `approveUnderlying` also read allowance unwrapped), so an internal inconsistency rather than a regression — trivial to align by wrapping it in the same try/catch.

Confidence: 70/100

packages/sdk/src/token/__tests__/wrap.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    wrappedToken: token,
+    signer,
+    provider,
+  }) => {
+    vi.mocked(provider.readContract)
+      .mockResolvedValueOnce(UNDERLYING)
+      .mockResolvedValueOnce(1000n) // balance ok
+      .mockResolvedValueOnce(50n); // allowance < 100
+
+    await expect(token.wrap(100n)).rejects.toMatchObject({
```
ankurdotb, line 103: [AUTOMATED] Nit (test hardening): this asserts only `code: InsufficientAllowance`, not the structured `requested`/`available`/`token` fields — a regression that swapped `requested`/`available` (or dropped `token`) would ship silently, and those fields drive consumer shortfall messaging. Suggest `.rejects.toMatchObject({ code, requested: 100n, available: 50n, token })` (same for the insufficient-balance test above). Separately, the happy-path test (~line 35) asserts `writeContract` was called with `wrap` but not `toHaveBeenCalledTimes(1)` — adding that pins the "wrap-only, no approve" invariant.

Confidence: 60/100

test/playwright/tests/wrap.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Approve only a small allowance...
+  await page.getByTestId("approve-amount-input").fill("100");
+  await page.getByTestId("approve-button").click();
+  await expect(page.getByTestId("approve-success")).toContainText("Tx: 0x");
+
+  // ...then attempt to wrap more than that. The SDK validates the allowance
+  // (public read, no signing) and fails early with InsufficientAllowanceError.
+  await page.getByTestId("wrap-amount-input").fill("1000");
+  await page.getByTestId("wrap-button").click();
+  await expect(page.getByTestId("wrap-error")).toBeVisible();
```
ankurdotb: [AUTOMATED] Nit: this only asserts `wrap-error` is visible. Since `wrap()` runs the ERC-20 balance check *before* the allowance check, if a prior test left the wallet balance below the amount the visible error would be `InsufficientERC20BalanceError` and this test would still pass — so it doesn't actually pin the allowance path it's named for. Assert the rendered error is the allowance one (text contains "allowance"/"approved", or key off the code).

Confidence: 55/100

ankurdotb (commented):
[AUTOMATED] ## Code review — `wrap()` + `useWrap` (two-signature shield)

Reviewed via a multi-agent sweep (correctness ×2, tests/e2e, guideline/architecture) plus a maintainer read. **Verdict: Comment (non-blocking) — merge-worthy once the two API items below are addressed.** No correctness or security bugs; CI is green; the change is additive.

### Design — a consistent, well-justified addition
`wrap`/`useWrap` completes the shield side of an already-established, symmetric escape-hatch pattern: shield now matches unshield (orchestrator + two per-phase hatches, at both the core and React layers) — `wrap`↔`unwrap`, `useWrap`↔`useUnwrap`, alongside the existing `approveUnderlying`. `react-sdk/AGENTS.md` pre-anticipated it, and both docstrings correctly steer to `shield()`. Justified on consistency alone, with a concrete consumer for the core `wrap()` (a TypeScript microservice wrapping the SDK). Minor non-blocking nuance: shield's two-phase is the non-ERC-1363 fallback (not protocol-inherent like unshield's), so `wrap()` is marginally less essential than `unwrap()` — but with `approveUnderlying` already shipped, adding its partner is the consistent move.

### Two items worth fixing before the public API/event contract ships
Inline on `query/wrap.ts` (dropped `onWrapSubmitted`) and `events/sdk-events.ts` (reused `ShieldSubmitted` event). Both small. The remaining inline notes are nits.

### Verified clean
Allowance semantics (`owner=signer, spender=wrapper`), `to` routes the mint recipient only, ERC-1363 bypass is intended escape-hatch behavior, error wiring complete/consistent, the example component has no cross-hook mis-wiring, api-report + barrels complete, and CI green.

packages/sdk/src/query/wrap.ts
ghermet: Fixed in 4bf484688. `WrapParams` now `extends WrapOptions` and the mutationFn forwards `...rest` (`token.wrap(amount, rest)`), matching `shieldMutationOptions`. `onWrapSubmitted` (and `to`) are reachable via `useWrap` now; added a test asserting the callback is forwarded.

packages/sdk/src/events/sdk-events.ts
ghermet: Fixed in 4bf484688. Added a distinct `ZamaSDKEvents.WrapSubmitted` + `WrapSubmittedEvent`; the standalone `wrap` op now emits it instead of `ShieldSubmitted`/`approveAndWrap`, so `ShieldSubmitted` stays exclusive to `shield()` — mirroring how `unwrap` emits its own event rather than reusing the unshield phases. Exported from the package index; api-report regenerated.

packages/sdk/src/token/wrapped-token.ts
ghermet: Fixed in 4bf484688. Wrapped the allowance read in the same `try/catch` → `ERC20ReadFailedError` as the balance read directly above it, and documented the `@throws`. (Left `#ensureAllowance`/`approveUnderlying` as-is — the inconsistency worth fixing was within `wrap()` itself, where the balance read was already wrapped and the allowance read was the odd one out.)

packages/sdk/src/token/__tests__/wrap.test.ts
ghermet: Fixed in 4bf484688. Both insufficiency tests now assert `requested`/`available`/`token` via `toMatchObject`, and the happy path pins the wrap-only invariant with `toHaveBeenCalledTimes(1)`. Also strengthened the mutation-forwarding test to assert `onWrapSubmitted` is forwarded.

test/playwright/tests/wrap.spec.ts
ghermet: Fixed in 4bf484688. The e2e now asserts `wrap-error` `toContainText("allowance")` instead of just visibility, so a low-balance wallet hitting the balance check (which runs first) can't pass the allowance-path test.

ankurdotb (approved):

PR #530, opened 2026-07-07
fix(sdk): unbundle unshield into request + finalize phases

ghermet:
## Summary

Wires the two-phase unwrap (request + finalize) internals while preserving the public unshield API. Callers keep using `WrappedToken.unshield` / `useUnshield`; the underlying `unwrap/*` escape hatches and their supporting types/tests are brought in line with the unbundled flow.

## Changes

- Change `unwrap` / `unwrapAll` to return the new `UnwrapResult` (a `TransactionResult` plus the decoded `unwrapRequestId`) and export the type (`types/token.ts`, `types/index.ts`, `index.ts`). `finalizeUnwrap`'s signature is unchanged — only its docstring was updated.
- Update the `useUnwrap`, `useUnwrapAll`, and `useFinalizeUnwrap` hooks plus their queries (`query/unwrap.ts`, `query/unwrap-all.ts`)
- Thread `UnwrapRequested` through `wrapped-token.ts` and test fixtures/mocks
- Update unwrap/finalize hook tests, events, integration, and wrapped-token tests
- Update the `unwrap-manual-form` test component

## Review follow-up

- Refactor `unshield` / `unshieldAll` to reuse the `unwrapRequestId` that `unwrap` / `unwrapAll` already decoded via a shared `#finalizeUnshield` tail, dropping the redundant receipt re-fetch and re-decode. `resumeUnshield` stays the only path that decodes from chain (it starts from a bare tx hash), so `UnwrapResult` now earns its place on the mainline flow.
- Fix the `UnwrapResult` docstring (whole result → `useFinalizeUnwrap` hook; `unwrapRequestId` → the scalar `WrappedToken.finalizeUnwrap`), a `mutateAsync` typo, add an `unwrapAll` `unwrapRequestId` assertion, and update the GitBook reference pages (`useUnwrap` / `useUnwrapAll` / `useFinalizeUnwrap` / `WrappedToken`) to the unbundled flow.

## Verification

- `pnpm typecheck` passes across all workspaces (run via pre-commit hooks)
- `sdk` + `react-sdk` unit suites green (1561 passed)
- oxlint / oxfmt / ast-grep clean


packages/sdk/src/token/wrapped-token.ts
```diff
@@ -508,15 +534,10 @@ export class WrappedToken extends Token {
       }
       throw new TransactionRevertedError("Failed to get unshield receipt", { cause: error });
     }
-    const event = findUnwrapRequested(receipt.logs);
-    if (!event) {
-      throw new TransactionRevertedError("No UnwrapRequested event found in unshield receipt");
-    }
+    const unwrapRequestId = this.#extractUnwrapRequestId(receipt.logs);
```
ankurdotb: [AUTOMATED] **Architectural: the orchestrated `unshield` path doesn't use the `unwrapRequestId` this PR adds — it re-derives it here.** _(confidence 85, non-blocking)_

`unshield()` (line 296) and `unshieldAll()` (line 320) already call `unwrap()`/`unwrapAll()`, which now extract and return `unwrapRequestId` (lines 409 / 445). But they keep only `unwrapResult.txHash` and hand that to `#waitAndFinalizeUnshield`, which re-fetches the receipt (line 530) and re-decodes the id on this line. So the new value is consumed **only** by the manual `unwrap` + `finalizeUnwrap` escape hatch; the mainline `unshield` flow gets a redundant decode on top of the (pre-existing) double `waitForTransactionReceipt`. It's also why the error text drifted from "unshield receipt" to "unwrap receipt" — this shared helper now serves both paths.

Not a bug, but the abstraction only half pays off: the audience that actually uses `unshield()` gets none of the "no manual log decode" benefit. The coherent shape is to resolve the id at each entry point and share a pure finalize tail:

- `unshield` / `unshieldAll`: `#finalizeUnshield(result.unwrapRequestId, result.txHash, ...)` — reuse what `unwrap()` already returned; no re-fetch, no re-decode.
- `resumeUnshield(txHash)`: the one path that legitimately starts from a bare hash → `waitForTransactionReceipt` + `#extractUnwrapRequestId` → same tail.

That makes `unshield = unwrap + finalizeUnwrap` composed from the same public primitives, lets `UnwrapResult` earn its place on the mainline, confines from-chain decoding to the resume case, and drops the message drift. Public signatures stay unchanged, so it's an internal refactor.

packages/sdk/src/types/token.ts
```diff
@@ -41,3 +43,16 @@ export interface UnshieldOptions extends UnshieldCallbacks {
   /** Skip confidential balance validation (e.g. for smart wallets). Default: `false`. */
   skipBalanceCheck?: boolean;
 }
+
+/**
+ * Result of {@link WrappedToken.unwrap} / {@link WrappedToken.unwrapAll}: a
+ * {@link TransactionResult} plus the `unwrapRequestId` decoded from the
+ * `UnwrapRequested` event.
+ *
+ * Pass the whole result (or just its `unwrapRequestId`) straight to
```
ankurdotb: [AUTOMATED] This doc comment says the whole result can go "straight to `WrappedToken.finalizeUnwrap`", but that's a compile error: the class method takes a scalar `EncryptedValue`, so `finalizeUnwrap(result)` fails with TS2345 (`UnwrapResult` is not assignable to `` `0x${string}` ``). Only `finalizeUnwrap(result.unwrapRequestId)` works — passing the whole object is valid solely for the `useFinalizeUnwrap` hook (its params are structurally a superset). Suggest: "Pass its `unwrapRequestId` to `WrappedToken.finalizeUnwrap`, or the whole result to the `useFinalizeUnwrap` hook." _(confidence 95)_

packages/react-sdk/src/unwrap/use-unwrap-all.ts
```diff
@@ -21,16 +25,21 @@ import { useWrappedToken } from "../token/use-wrapped-token";
  * @example
  * ```tsx
  * const unwrapAll = useUnwrapAll("0xWrapper");
- * unwrapAll.mutate();
+ * const finalize = useFinalizeUnwrap("0xWrapper");
+ *
+ * async function unshieldAll() {
+ *  const result = await unwrapAll.mutateAsync();
+ *  await finalize.mutateASync(result);
```
ankurdotb: [AUTOMATED] Typo: `mutateASync` → `mutateAsync` (capital `S`). It's inside a JSDoc `@example` so it won't fail the build, but it renders into the API docs verbatim and a copy-paste hits `finalize.mutateASync is not a function`. The sibling example in `use-unwrap.ts` has it correct. _(confidence 100)_

packages/sdk/src/token/wrapped-token.ts
```diff
@@ -430,23 +437,28 @@ export class WrappedToken extends Token {
       throw new DecryptionFailedError("Cannot unshield: balance is zero");
     }
 
-    return this.submitTransaction({
+    const result = await this.submitTransaction({
       operation: "unwrapAll",
       config: unwrapFromBalanceContract(this.address, userAddress, userAddress, encryptedValue),
     });
+
+    return { ...result, unwrapRequestId: this.#extractUnwrapRequestId(result.receipt.logs) };
```
ankurdotb, line 469: [AUTOMATED] Test-coverage gap: nothing asserts `unwrapAll()`'s own `unwrapRequestId`. Mutation-testing confirmed it — hardcoding a bogus id into *only* this return statement left all tests green, because the suite asserts `unwrap()`'s value (`wrapped-token.test.ts`) and the shared `#extractUnwrapRequestId`, but never `unwrapAll()`'s attach-the-field line. A future edit touching just this line would ship silently. Suggest one assertion on `result.unwrapRequestId` in the `unwrapAll` test. _(confidence 95)_

packages/sdk/src/token/wrapped-token.ts
```diff
@@ -468,6 +480,20 @@ export class WrappedToken extends Token {
...
+  /**
+   * Decode the `unwrapRequestId` from an unwrap/unshield receipt's logs.
+   * Falls back to the event's `encryptedAmount` on protocol versions that
+   * don't emit a dedicated `unwrapRequestId`. Throws when the receipt lacks
+   * the `UnwrapRequested` event entirely.
+   */
+  #extractUnwrapRequestId(logs: readonly RawLog[]): EncryptedValue {
+    const event = findUnwrapRequested(logs);
+    if (!event) {
+      throw new TransactionRevertedError("No UnwrapRequested event found in unwrap receipt");
```
ankurdotb, line 514: [AUTOMATED] FYI (non-blocking), two small things on this helper:

1. Raw `unwrap()`/`unwrapAll()` now throw here where they previously returned unconditionally. For a genuine revert this is a fine fail-fast (verified: the viem/ethers providers don't throw on revert, so a reverted tx reaches here with empty logs → this throw). For the pathological "mined but no `UnwrapRequested` event" case, `TransactionRevertedError` is a mild mislabel since the tx didn't revert — a distinct error would read better.
2. The `?? event.encryptedAmount` fallback on the next line looks unreachable: `decodeUnwrapRequested` requires ≥3 topics and always sets `unwrapRequestId` when it returns non-null, so the docstring's "falls back on protocol versions that don't emit a dedicated `unwrapRequestId`" can't currently trigger. Pre-existing (hoisted from the old unshield code), now shared by three call sites. _(confidence 80)_

ankurdotb (commented):
[AUTOMATED] ## Code review — unbundle unshield (`UnwrapResult`)

Reviewed on its own merits as a DX/refactor. **Verdict: Comment (non-blocking).** No correctness or security bugs; CI is green and the change is additive and backward-compatible (`UnwrapResult extends TransactionResult`). It's a worthwhile improvement — `unwrap()`/`unwrapAll()` handing back `unwrapRequestId` removes the manual `findUnwrapRequested` decode for direct callers. There's one architectural coherence point (inline on `wrapped-token.ts`) plus a few doc/test tidy-ups worth doing before merge.

_Method: two independent passes over the diff in a local worktree, plus a real build (`tsc` / `oxlint` / `api-extractor`, all clean) and mutation-testing of the touched tests. Every finding verified against the code._

### Items that can't be anchored inline (outside the diff)

**Docs — GitBook reference pages still teach the pre-PR pattern** _(confidence 90)_. `docs/gitbook/src/reference/react/useFinalizeUnwrap.md`, `useUnwrap.md`, `useUnwrapAll.md`, and `docs/gitbook/src/reference/sdk/WrappedToken.md` weren't updated: they still show the manual `findUnwrapRequested(receipt.logs)` decode this PR removes, and `WrappedToken.md` still documents `unwrap`/`unwrapAll` as `Promise<TransactionResult>` (now `UnwrapResult`, so `unwrapRequestId` is missing from the documented shape). This is the highest-impact doc gap — the public reference would keep steering users to the old, more verbose flow the PR is meant to retire. Recommend updating all four pages.

**PR description accuracy** _(confidence 95, FYI)_. The description says it will "Complete the `finalizeUnwrap` type signature" — that signature is unchanged (only its docstring changed); the actual signature change is `unwrap`/`unwrapAll` → `UnwrapResult`. Worth correcting so the history reads accurately.

See inline comments for the rest — the `wrapped-token.ts` note is the substantive one.

packages/sdk/src/token/wrapped-token.ts
ghermet: Done in 356d650f2. Extracted a pure \`#finalizeUnshield(unwrapRequestId, unwrapTxHash, operationId, callbacks)\` tail. \`unshield\`/\`unshieldAll\` now hand it the \`unwrapRequestId\` + \`txHash\` that \`unwrap\`/\`unwrapAll\` already returned — no second \`waitForTransactionReceipt\`, no second decode. \`resumeUnshield\` stays the only path that fetches the receipt and decodes from chain (it starts from a bare hash), then joins the same tail. Public signatures unchanged; the message drift is gone since the shared helper now owns the "unwrap receipt" wording. All unshield/persistence/callback/event tests still pass.

packages/sdk/src/types/token.ts
ghermet: Fixed. Reworded to: "Pass its \`unwrapRequestId\` to \`WrappedToken.finalizeUnwrap\`, or the whole result to the \`useFinalizeUnwrap\` hook" — so the scalar-vs-superset distinction is explicit.

packages/react-sdk/src/unwrap/use-unwrap-all.ts
ghermet: Fixed: \`mutateASync\` → \`mutateAsync\`.

packages/sdk/src/token/wrapped-token.ts
ghermet: Added. The \`unwrapAll\` test now mocks a known id and asserts \`expect(result.unwrapRequestId).toBe(unwrapRequestId)\`, so the attach-the-field line is covered independently of \`unwrap\`.

packages/sdk/src/token/wrapped-token.ts
ghermet: Left both as-is, with reasoning:

1. **`TransactionRevertedError` for the no-event case.** The error type is codified by three tests that assert `TransactionRevertedError` / `ZamaErrorCode.TransactionReverted` when the receipt carries no `UnwrapRequested` event (`wrapped-token.test.ts` unwrap + unshield, `callbacks.test.ts`). The realistic trigger is a genuine revert (empty logs), where the label is accurate; the "mined but no event" case is pathological and not currently reachable in a test. A distinct error is a public error-taxonomy change I'd rather not fold into this PR — happy to file a follow-up if you want the finer distinction.

2. **The `?? event.encryptedAmount` fallback.** It is load-bearing for the type, not runtime: `UnwrapRequestedEvent.unwrapRequestId` is declared optional (`readonly unwrapRequestId?: EncryptedValue`), so `event.unwrapRequestId` is `EncryptedValue | undefined` and returning it bare would not satisfy the `EncryptedValue` return type. So it's a type-narrowing fallback rather than dead code, even though the current decoder always populates the topic. I can tighten the docstring's "protocol versions" wording if you prefer, but kept the behaviour.

ghermet:
Addressed the review in 356d650f2. Summary of the non-inline items:

- **GitBook reference pages** — updated `useUnwrap`, `useUnwrapAll`, `useFinalizeUnwrap`, and `WrappedToken` to the unbundled flow: `unwrap`/`unwrapAll` now return `UnwrapResult` (`Promise<UnwrapResult>`), the examples pass the result straight to finalize, and the manual `findUnwrapRequested(receipt.logs)` decode is gone. (`event-decoders.md` still references `findUnwrapRequested` — that page documents the decoder itself, so it's intentionally unchanged.) `scripts/docs/check-links.mjs` passes.
- **PR description** — corrected: the signature change is `unwrap`/`unwrapAll` → `UnwrapResult`; `finalizeUnwrap`'s signature is unchanged (docstring only).

Inline threads answered individually. The `wrapped-token.ts` architectural note drove the main change — `unshield`/`unshieldAll` now reuse the id `unwrap` already decoded via a shared `#finalizeUnshield` tail, so `UnwrapResult` earns its place on the mainline and the double receipt fetch/decode is gone.

Verification: `sdk` + `react-sdk` unit suites green (1561 passed), typecheck clean across workspaces, oxlint/oxfmt/ast-grep clean.

ankurdotb (commented):
[AUTOMATED] ## Re-review — review fixes verified

Checked against the live head (`36ad6f4a`), not the commit messages. Every earlier finding is now addressed:

- **Architectural** — `unshield`/`unshieldAll` reuse the `unwrapRequestId` that `unwrap`/`unwrapAll` already return; `#waitAndFinalizeUnshield` is now the pure `#finalizeUnshield` tail; `resumeUnshield` owns the sole from-chain decode. The "unshield receipt" → "unwrap receipt" message drift is gone as a bonus. ✅ _(thread resolved)_
- **`UnwrapResult` doc comment** reworded — scalar `unwrapRequestId` for the class method, whole result for the hook. ✅ _(resolved)_
- **`mutateASync` typo** fixed. ✅ _(resolved)_
- **`unwrapAll` return-value assertion** added. ✅
- **GitBook reference pages + `llms-full.txt`** updated to the `UnwrapResult` pattern. ✅
- **Example all-path bug (found during this pass, already fixed in `36ad6f4a`)** — the `unwrap-manual` form gated `burn-handle` + finalize on `unwrap.data` only, so "Unwrap All → finalize" couldn't surface the request id or finalize, and `test-vite`/`test-nextjs` went red on `unwrap-manual.spec.ts:47` (both apps). Fixed correctly with `const unwrapResult = unwrap.data ?? unwrapAll.data;`, used for the display, the finalize gate, and `finalizeUnwrap.mutate(unwrapResult)`. This was the integration-level manifestation of the `unwrapAll` coverage note. E2E jobs are re-running to confirm green.

SDK-level CI is green (`test`, `test-node`, `integration`, `lint`, `api-report`, `build`, `check`, `docs`); `test-vite`/`test-nextjs` are re-running on the fix.

**Nits (non-blocking):**
- PR description still says it will "Complete the `finalizeUnwrap` type signature" — that signature never changed — and doesn't mention the reuse-in-`unshield` refactor now included.
- The E2E logs showed `CleartextFHEVMExecutor is above the contract size limit (26573 > 24576)` during the anvil deploy on both apps. Other confidential E2E flows still passed, so it wasn't the cause of the failure here — but worth a quick confirmation that it's expected and not a latent deploy problem.

Verdict: **looks good** — nothing blocking remains at the code level; happy to see it merge once the re-run `test-vite`/`test-nextjs` go green.

packages/sdk/src/token/wrapped-token.ts
ankurdotb: [AUTOMATED] The unit assertion added here covers `unwrapAll`'s return value. The *consumer* gap this pointed at then showed up in E2E — the `unwrap-manual` form read `unwrap.data` for the all-path — and was fixed in 36ad6f4a (`unwrapResult = unwrap.data ?? unwrapAll.data`). Leaving this open until the re-run test-vite/test-nextjs confirm green, then it can be resolved.

ankurdotb: [AUTOMATED] E2E is green on 36ad6f4a — `test-vite` and `test-nextjs` both pass. Consumer gap closed; resolving.

packages/sdk/src/token/wrapped-token.ts
ankurdotb: [AUTOMATED] Went a step past the docstring tweak, since the fallback is genuinely dead code: pushed `008d1e57`. `unwrapRequestId` is now **required** on `UnwrapRequestedEvent` / `UnwrapFinalizedEvent` (both decoders always populate it from the indexed topic, guarded by `topics.length >= 3`), so `#extractUnwrapRequestId` returns it directly — the `?? event.encryptedAmount` fallback and its misleading "protocol versions" docstring are gone. api-report regenerated the proper way (build → api-extractor); the only report change is the two event fields losing `?`. `tsc`, the affected unit tests (89), `oxfmt`, and `oxlint` are clean locally. Resolving.

ankurdotb (approved):

PR #532, opened 2026-07-07
chore(lint): enable oxlint type-checking and fix test type debt

ghermet:
## What

Turns on oxlint's `typeCheck` (full type-aware diagnostics) and makes the whole repo pass it. The flag surfaced **212 pre-existing type errors, all in test files** — production `src` was already clean.

## Why the errors existed

Test files were never type-gated:
- `pnpm typecheck` **excluded** tests, and the `tsconfig.vitest.json` that would have covered them was orphaned (wired to nothing).
- oxlint's `tsgolint` resolves the nearest `tsconfig.json` per file, and both packages excluded tests — so test files fell back to an inferred project with **no `paths`** (`@zama-fhe/sdk` resolved to built `dist`, causing `#private` mismatches) and **no vitest globals** (`vi`/`expect` undefined).

## Changes (test-side + config only — no production source touched)

- **`packages/{sdk,react-sdk}/tsconfig.json`** — include tests and add `"types": ["node", "vitest/globals"]`, so oxlint *and* `tsc` resolve tests via `src` paths with vitest globals. Tests are now genuinely type-checked.
- **`test-fixtures/relayer.ts`** — `createMockRelayer` returns the concrete `RelayerDispatcher` (was the looser `RelayerSDK` interface), cascading away ~16 errors across service/provider tests.
- **34 test files** — fix genuine drift against the frozen API: required `logger`/`isReady` mock fields, `cache`→`cachingService` option rename, `RelayerDispatcher`'s new `logger` constructor arg, branded `ChecksummedAddress` types, EIP-712 union shapes, and `noUncheckedIndexedAccess` on known-defined fixture arrays. All test-side (imports, non-null assertions, casts) — **no assertion behavior changed, no API signatures widened.**
- Apply `sortPackageJson: true` (oxfmt) to `package.json` files.

## ⚠️ CI impact

This **broadens what CI enforces** — test files are type-gated for the first time. New type errors in tests will now fail lint/typecheck.

## Verification

| Gate | Result |
|------|--------|
| `pnpm lint` (oxlint typeCheck + ast-grep) | exit 0 |
| `pnpm typecheck` | all packages Done |
| `pnpm format:check` | clean |
| `pnpm test:run` | 1762 passed, 5 skipped, no type errors |

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -21,7 +21,7 @@
     "restriction": "off",
     "nursery": "off"
   },
-  "options": { "denyWarnings": true, "typeAware": true },
+  "options": { "denyWarnings": true, "typeAware": true, "typeCheck": true },
   "env": { "builtin": true },
   "settings": {
     "react": { "version": "19.2.4" },
@@ -42,6 +42,7 @@
     "coverage/**",
     "examples/**",
     "hardhat/**",
+    "codemods/**",
     "packages/playwright/**",
     "packages/test-nextjs/**",
     "packages/test-nextjs/playwright/**",
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -6,7 +6,7 @@
   "printWidth": 100,
   "tabWidth": 2,
   "objectWrap": "collapse",
-  "sortPackageJson": false,
+  "sortPackageJson": true,
   "ignorePatterns": [
     "dist",
     "node_modules",
--- a/codemods/sdk-migration-v3/package.json
+++ b/codemods/sdk-migration-v3/package.json
@@ -7,8 +7,8 @@
     "test": "./test.sh"
   },
   "devDependencies": {
-    "codemod": "1.12.3",
     "@types/node": "^20.0.0",
+    "codemod": "1.12.3",
     "typescript": "^6.0.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -4,11 +4,15 @@
   "private": true,
   "description": "Monorepo for Zama SDK packages",
   "license": "BSD-3-Clause-Clear",
+  "author": "Zama",
   "repository": {
     "type": "git",
     "url": "https://github.com/zama-ai/sdk"
   },
-  "author": "Zama",
+  "workspaces": [
+    "packages/*",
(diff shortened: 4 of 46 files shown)


zama-cremaud (approved):

.oxlintrc.json
```diff
@@ -42,6 +42,7 @@
     "coverage/**",
     "examples/**",
     "hardhat/**",
+    "codemods/**",
```
ankurdotb, line 45: [AUTOMATED]

Optional: adding `codemods/**` here drops *all* oxlint coverage from the codemod scripts, not just the newly-enabled `typeCheck`. `codemods/` wasn't in `ignorePatterns` before, so root `oxlint` was linting `codemods/sdk-migration-v3/scripts/*.ts`, and that package has no lint/typecheck script of its own — so this is a net loss of static analysis for the migration tool (likely because the JSSG transform scripts don't pass the new type-aware rules).

Non-blocking and defensible for a peripheral one-shot tool, but worth either narrowing the exclusion to just what fails, giving codemods its own lint script, or at least noting the coverage change in the PR description.

Confidence: 90/100

ankurdotb (commented):
[AUTOMATED] — Clean, well-scoped config + test-only PR; the "no production `src` touched / no assertion behavior changed" claim holds. Build stays isolated via `tsconfig.build.json` (still excludes tests), so `dist/` and the published `.d.ts` are unaffected, and CI (incl. `api-report`'s `pnpm build` + the consumer `test-nextjs`/`test-vite`/`test-node` builds) confirms the `types` narrowing is safe. Test edits are type-only: non-null assertions guarded by preceding length checks, stale/dead call-args removed, no weakened assertions, `.skip`, `.only`, or `@ts-expect-error`. `sortPackageJson: true` reordered no resolution-significant `exports` conditions.

Two non-blocking notes:

1. See inline comment on `.oxlintrc.json` re: `codemods/**` losing lint coverage.

2. **Nit:** `tsconfig.vitest.json` (repo root) is now dead — it has zero references anywhere, and this PR folds test type-checking into the package `tsconfig.json`s. Good opportunity to delete it in this PR.

PR #535, opened 2026-07-07
feat(examples): SDK-149 Privacy Service — JSON-RPC wrapper, indexer, demo app

zama-cremaud:
## Summary

[SDK-149](https://linear.app/zama/issue/SDK-149) is a validation Epic for a two-part "Privacy Service" product idea. This PR bundles the three proof-of-concept pieces built to de-risk the technical design — not a shipped product, and not pre-empting the Epic's own build/no-build decision (which still gates on ICP interviews, a separate workstream).

- **`examples/zama-json-rpc`** (write-side) — a local JSON-RPC middleware that transparently rewrites plaintext-looking calls (`transfer(to, amount)`) into real encrypted ERC-7984 calls for any address Zama's on-chain wrappers registry confirms as a genuine confidential token. Zero private-key custody. See its `WALKTHROUGH.md`.
- **`examples/confidential-indexer`** (read-side) — a separate, stateful service that decrypts and caches confidential balances/transfer history for accounts that have delegated to it via the ACL, serving them over a small REST API. Ships as its own deployable by design (real key custody, fundamentally different operational model from the write-side). See its `WALKTHROUGH.md`.
- **`examples/rpc-demo-app`** — a small Next.js app built to record a demo video of both, styled like `examples/react-wagmi` but never importing the Zama SDK — it only speaks plain JSON-RPC/REST, like a real unmodified consumer would. Building it surfaced a genuine architectural finding: a real EIP-1193 wallet (MetaMask/Rabby) signs `eth_sendTransaction` client-side *before* any network call, so no wallet RPC reconfiguration can route through the wrapper's rewrite step — fixed with a persistent signer relay standing in for "the custodian's infrastructure behind the wrapper." Full writeup in its own `WALKTHROUGH.md`.

Each project also has real, live-Sepolia verification (not mocked) — see each `WALKTHROUGH.md`'s "Verified for real" section for tx hashes and evidence.

Draft: three independent branches merged for one combined review; the write-side and read-side still ship as separate deployables (no shared runtime) per an explicit product decision documented in each WALKTHROUGH.

## Test plan

- [ ] `zama-json-rpc`: 43 unit tests + 1 real e2e test against the live Sepolia relayer (`npm test`, `npm run test:e2e`)
- [ ] `confidential-indexer`: 19 unit tests + integration (real local Redis) + 1 real e2e test against historical Sepolia logs (`npm test`, `npm run test:integration`, `npm run test:e2e`)
- [ ] `rpc-demo-app`: typecheck + headless Playwright smoke tests (mocked wallet connection only, every downstream call real)


examples/zama-json-rpc/src/rpc/router.ts
```diff
@@ -0,0 +1,117 @@
...
+export async function handleSingleRequest(
+  request: unknown,
+  deps: RouterDeps,
+): Promise<JsonRpcResponse> {
+  if (!isValidJsonRpcRequest(request)) {
+    const id = (request as { id?: JsonRpcRequest["id"] } | null)?.id ?? null;
+    return failure(id, { code: RpcErrorCode.InvalidRequest, message: "Invalid Request" });
+  }
+
+  const zamaHandler = deps.zamaHandlers[request.method];
```
ghermet, line 49: [AUTOMATED] **Bug:** unguarded prototype-chain lookup on the handler map.

`deps.zamaHandlers` is a plain object literal (`buildZamaHandlers` in `zama/introspection.ts` returns `{ zama_... }`), so this lookup walks `Object.prototype`. A request with `"method": "toString"` or `"constructor"` resolves `zamaHandler` to the inherited function, invokes it, and returns a **bogus success result** (`"[object Undefined]"`, `[]`, …) instead of passing through to upstream. `valueOf` / `hasOwnProperty` etc. resolve to functions that throw and get mapped to a spurious SDK error. In all cases the `startsWith("zama_")` guard at line 59 is bypassed because this block returns first.

Practical blast radius is small (real Ethereum clients don't send these method names), but it's incorrect dispatch and a fingerprinting oddity. Trivial fix — guard to own keys:

```ts
const zamaHandler = Object.hasOwn(deps.zamaHandlers, request.method)
  ? deps.zamaHandlers[request.method]
  : undefined;
```

or build the map with `Object.create(null)`.

Confidence: 97/100

examples/rpc-demo-app/src/lib/indexerClient.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Plain `fetch()`, no SDK — the whole point of talking to confidential-indexer is
+ * that consuming it never requires FHE-aware code. 403/202/200 are all meaningful
+ * responses here (no delegation / delegated-but-not-decrypted-yet / real data),
+ * not error states, so this returns the status alongside the body rather than
+ * throwing on non-2xx.
+ */
+export async function fetchIndexer<T>(path: string): Promise<{ status: number; body: T }> {
+  const response = await fetch(`${INDEXER_URL}${path}`);
+  const body = (await response.json().catch(() => undefined)) as T;
```
ghermet, line 35: [AUTOMATED] Nit: `.catch(() => undefined) as T` is type-unsound — a malformed `200` body yields `undefined` cast to `T`, so a downstream `body.clearValue` access throws `Cannot read properties of undefined` rather than being handled as an error. In practice the consumers (`HistoryCard`) gate body access on `status === 200` and React Query surfaces the throw as `isError`, so it degrades into the existing error path — hence a nit, not blocking. Consider typing the return as `{ status: number; body: T | undefined }` so callers must handle the missing-body case.

Confidence: 85/100

examples/rpc-demo-app/src/lib/useRelayedSend.ts
```diff
@@ -0,0 +1,216 @@
...
+async function fetchRecentAudit(sinceMs: number): Promise<AuditEntry[]> {
+  try {
+    const auditUrl = new URL("/audit", ZAMA_RPC_URL).toString();
+    const response = await fetch(auditUrl);
+    if (!response.ok) return [];
+    const body = (await response.json()) as { entries: { timestamp: string; entry: AuditEntry }[] };
+    return body.entries
+      .filter((e) => new Date(e.timestamp).getTime() >= sinceMs)
+      .map((e) => e.entry);
+  } catch (error) {
```
ghermet, line 30: [AUTOMATED] Nit: a failing `/audit` fetch is swallowed to `[]` (console-only), so the "rewrote the calldata" annotation is silently dropped from the trace — the visualization then looks as if no rewrite happened. The send itself still works, so this is cosmetic/educational, but for a demo whose purpose is to *show* the rewrite, consider surfacing a small "audit unavailable" marker in the trace instead of quietly omitting the step.

Confidence: 88/100

ghermet (commented):
[AUTOMATED]

## Code Review — SDK-149 Privacy Service POC

Reviewed the three example apps (`zama-json-rpc`, `confidential-indexer`, `rpc-demo-app`) across four dimensions — bugs/security, guideline compliance, and error handling — with independent validation of every finding at confidence ≥ 80.

**Overall: solid POC.** A few things I specifically verified are done right:
- **SDK usage is compliant.** Both services build the high-level `ZamaSDK` via `createConfig` + transport factories (no legacy `@zama-fhe/relayer-sdk` / `createInstance` / `initSDK`), and the low-level calldata rewriting / raw-log reading is intrinsic to what a JSON-RPC middleware and an indexer must do — not a recompose of a `Token` flow.
- **ACL / authorization is sound.** Reads are gated on `DelegationStore.isKnownActive(account, contract)`, the on-chain log filter narrows to this service's own delegate (`topics[2]`), and `delegatedDecryptValues` enforces entitlement protocol-side — no unauthorized cleartext is cached or served.
- **bigint / decimal round-tripping and address-key normalization** check out.

**One real (low-impact) bug** and **two demo-scoped nits** inline below. Non-blocking for a draft POC.

<details><summary>FYI — background-refresh staleness (not flagged inline)</summary>

The balance-refresh and transfer-tracker poll loops catch-log-continue, which is the correct pattern for a background indexer. Missing balances surface as `202 pending` and every served balance carries `decryptedAtBlock`, so staleness is observable rather than silently-wrong; the transfer scan cursor only advances on full success, so partial batches self-heal on the next tick. If you later want to harden it, consider comparing `decryptedAtBlock` against head to expose a staleness threshold — but that's out of scope for a POC.
</details>

examples/confidential-indexer/src/server.ts
```diff
@@ -0,0 +1,51 @@
...
+ * query this REST API directly cross-origin. Not something to expose as-is beyond
+ * local development.
+ */
+function setCorsHeaders(res: import("node:http").ServerResponse): void {
+  res.setHeader("Access-Control-Allow-Origin", "*");
+  res.setHeader("Access-Control-Allow-Methods", "GET, OPTIONS");
+  res.setHeader("Access-Control-Allow-Headers", "authorization");
+}
+
+export function createHttpServer(params: { routerDeps: RouterDeps; logger: Logger }): Server {
```
ghermet, line 16: Any reason why this is a raw node server and not like a fastapi,express server?

ghermet:
### Trim the HTTP/CLI boilerplate — lean on a framework + DI

Both example services (`confidential-indexer`, `zama-json-rpc`) hand-roll the HTTP layer today: a `node:http` `createServer`, manual `setCorsHeaders`, a hand-written path-splitting router, and stores threaded through by hand. It works, but it's boilerplate competing for attention with the SDK call sites that are the actual point of these examples.

Two changes would make them read more like references:

**1. HTTP layer → Fastify (or Hono).** Out of the box: routing, CORS (`@fastify/cors`), error handling — and the real win, dependency injection via plugins + decorators. The stores, decrypt cache, and poller become plugins that decorate the instance (`app.delegationStore`, …) and own their lifecycle on `onReady`/`onClose`, so `app.close()` tears everything down. That deletes the manual wiring in `cli.ts` and makes the seams explicit. Hono is the lighter, Fetch-native alternative if we'd rather stay dependency-thin.

> Gotcha for whoever picks it up: don't `await` the first poll inside the `onReady` hook — the log scan + decrypt-propagation retries exceed Fastify's 10s readiness timeout and abort `listen()`. Kick the loop off; don't block readiness on its first cycle.

**2. CLI → incur (instead of commander).** Schema-first (Zod): one set of flag definitions drives parsing, `--help`, env fallback, and agent-discoverable surfaces (`--llms`, `--schema`, `--mcp`) for free — on-theme with the epic's "integration as a DevOps task" framing. For a daemon the root `run()` just boots the server and stays pending.

Follow-on (out of scope for this validation spike, but worth a line in the decision doc): with incur in place, the read API could be served as incur commands via `cli.fetch` — handing a Profile B analytics consumer an MCP endpoint + JSON Schema for free. Tradeoff: it replaces the deliberately ERC-7984-shaped REST contract (`403` not-entitled / `202` pending / `200` cached) with incur's `{ok,error}` envelope, and that contract is the clearest demonstration of the ACL model — so it's a proposal, not a build.

PR #536, opened 2026-07-08
feat(sdk): uniform retryability signal across decryption error causes [SDK-248]

zama-cremaud:
## Summary

After SDK-236 (relayer back-pressure), SDK-237 (worker timeouts), and SDK-239 (typed decryption causes), five typed causes are transient/retryable but only `RelayerRequestFailedError` exposed a programmatic `.retryable` field — the other four were "retryable by JSDoc" only, forcing consumers to hardcode the set of retryable codes.

- Added `readonly retryable: boolean` on `ZamaError`, defaulted from a new `RETRYABLE_BY_CODE: Complete<Record<ZamaErrorCode, boolean>>` map in `base.ts` — mirroring the existing `ErrorForCode` exhaustiveness guard in `match.ts`, so a new `ZamaErrorCode` added without a retryability entry fails the build. `RelayerRequestFailedError` is the one exception (retryability depends on HTTP status, not just the code) and overrides it per-instance.
- Added exported `isRetryable(error: unknown): boolean` and `retryAfterSeconds(error: unknown): number | undefined` accessors.
- Locked `wrapDecryptError`'s hand-maintained `instanceof` passthrough list behind a single exported array (`DECRYPT_PASSTHROUGH_ERROR_TYPES`) plus a test asserting every entry has a matching passthrough example and vice versa, so a dropped/added entry is caught instead of silently collapsing into terminal `DecryptionFailedError`.
- Added a consolidated "Retry transient failures" section (table + retry-loop recipe) to `handle-errors.md`, and filled in the `NotEntitledError` / `RpcRateLimitError` rows that were missing from the error-hierarchy table since SDK-239.

Closes SDK-248.

## Test plan
- [x] `pnpm --filter @zama-fhe/sdk typecheck` — clean
- [x] Full `@zama-fhe/sdk` vitest suite — 1238/1238 passing
- [x] Manually verified the compiler guard: removing an entry from `RETRYABLE_BY_CODE` fails `tsc` with `Property '...' is missing in type ... Record<ZamaErrorCode, boolean>`
- [x] `pnpm --filter @zama-fhe/sdk api-report` — regenerated, diff is purely additive (`isRetryable`, `retryAfterSeconds`, `retryable` moved to base)
- [x] `pnpm docs:check-links` — clean
- [x] `pnpm llm:build` — corpus regenerated


packages/sdk/src/errors/base.ts
```diff
@@ -83,6 +85,55 @@ export const ZamaErrorCode = {
...
+  [ZamaErrorCode.InvalidTransportKeyPair]: false,
+  [ZamaErrorCode.NoCiphertext]: false,
+  [ZamaErrorCode.RelayerRequestFailed]: false, // conditional — see doc above
+  [ZamaErrorCode.NotEntitled]: false,
+  [ZamaErrorCode.RpcRateLimited]: true,
+  [ZamaErrorCode.OperationTimeout]: true,
+  [ZamaErrorCode.WorkerRecycled]: true,
+  [ZamaErrorCode.Configuration]: false,
+  [ZamaErrorCode.DelegationSelfNotAllowed]: false,
+  [ZamaErrorCode.DelegationCooldown]: false,
```
ankurdotb: [AUTOMATED] **Please fix:** this should be `true`. `DelegationCooldownError`'s own thrown message is "Wait for the next block before retrying" — a per-block timing gate that a next-block retry resolves with no other change, structurally identical to `DELEGATION_NOT_PROPAGATED` above (correctly `true`). `isRetryable` is now a public contract, so `false` here mis-steers consumers against the error's own remediation. Flip to `true` + add an `isRetryable(cooldownError) === true` test. Confidence: 80/100

packages/sdk/src/errors/base.ts
```diff
@@ -83,6 +85,55 @@ export const ZamaErrorCode = {
...
+  [ZamaErrorCode.DelegationNotFound]: false,
+  [ZamaErrorCode.DelegationExpired]: false,
+  [ZamaErrorCode.InsufficientConfidentialBalance]: false,
+  [ZamaErrorCode.InsufficientERC20Balance]: false,
+  [ZamaErrorCode.BalanceCheckUnavailable]: false,
+  [ZamaErrorCode.ERC20ReadFailed]: false,
```
ankurdotb: [AUTOMATED] **Consider:** `ERC20_READ_FAILED` conflates a *network* fault (transient) with a *contract* error (terminal). A 429 that exhausted viem/ethers' internal retries lands here as `ERC20ReadFailedError` (this path does no rate-limit classification), so a throttled balance read is currently non-retryable. Consider splitting network-vs-contract at the throw site, or a one-line comment justifying the conservative `false`. Confidence: 70/100

packages/sdk/src/errors/base.ts
```diff
@@ -83,6 +85,55 @@ export const ZamaErrorCode = {
...
+  [ZamaErrorCode.DelegationExpiryUnchanged]: false,
+  [ZamaErrorCode.DelegationDelegateEqualsContract]: false,
+  [ZamaErrorCode.DelegationContractIsSelf]: false,
+  [ZamaErrorCode.AclPaused]: false,
+  [ZamaErrorCode.DelegationExpirationTooSoon]: false,
+  [ZamaErrorCode.DelegationNotPropagated]: true,
+  [ZamaErrorCode.ChainMismatch]: false,
+  [ZamaErrorCode.SignerNotConfigured]: false,
+  [ZamaErrorCode.WalletNotConnected]: false,
+  [ZamaErrorCode.WalletAccountNotReady]: false,
```
ankurdotb: [AUTOMATED] **Consider:** `WALLET_ACCOUNT_NOT_READY` means wallet-account discovery is "still resolving" — transient (a retry after it resolves succeeds or degrades to the terminal `WALLET_NOT_CONNECTED`). `false` is defensible ("await readiness rather than busy-loop") but contradicts the framing; flip or add a justifying comment. Confidence: 72/100

ankurdotb (dismissed):
[AUTOMATED]

## Review — uniform retryability signal (SDK-248)

High-quality, faithful implementation of the ticket: the `Complete<Record<ZamaErrorCode, boolean>>` guard mirrors `ErrorForCode` (a new code without a retryability entry fails `tsc`), `isRetryable`/`retryAfterSeconds` are clean, the `RelayerRequestFailedError` per-instance 429 override is correct (its own `retryable` field removed in favour of the base, still inherited — verified additive), the docs recipe matches the map, and the tests are substantive. Toolchain verified clean (api-report 0-diff, full vitest + type-tests, `tsc`, `llm:check`).

Requesting one change; two Considers and a nit.

**1. [Please fix] `DELEGATION_COOLDOWN` should be `retryable: true`** (inline). Its own thrown message is *"Wait for the next block before retrying"* — a per-block timing gate, retryable by definition, and structurally identical to its `DELEGATION_NOT_PROPAGATED` twin (correctly `true`). Since `isRetryable` is now a public contract, marking it `false` tells consumers *not* to retry in direct contradiction to the error's own remediation. One-line flip + a test asserting `isRetryable(cooldownError) === true`. It's one line, but it's a self-contradiction in the very contract this PR establishes, hence blocking.

**2. [Consider] `WALLET_ACCOUNT_NOT_READY = false`** (inline) — "still resolving" is transient (a retry once discovery resolves succeeds); defensible as "await readiness, don't busy-loop," but the value contradicts the framing. Flip, or add a one-line justification.

**3. [Consider] `ERC20_READ_FAILED = false`** (inline) — conflates a network fault (transient) with a contract error (terminal). Note: a **429 that exhausted viem/ethers' internal retries lands here as `ERC20ReadFailedError`, not `RpcRateLimitError`** (this path does no rate-limit classification), so a throttled balance read is currently non-retryable. Consider splitting network-vs-contract at the throw site, or documenting the conservative default.

**4. [Nit] `DECRYPT_PASSTHROUGH_ERROR_TYPES` completeness is self-referential** — the test catches a *dropped* entry (array ⇄ its own example list) but not a *never-added* new subclass, and unlike `RETRYABLE_BY_CODE`/`ErrorForCode` it has no `Complete<>` tie to the taxonomy. Not blocking (full exhaustiveness isn't the right target for a decrypt-only subset) — worth a one-line comment noting the scope difference.

**Verified clean:** API additive (no break; `retryable` inherited on `RelayerRequestFailedError`); accessors + invariants sound; docs table matches `RETRYABLE_BY_CODE` with no drift; seconds convention respected; no `SDK-\d` in committed `.md`; tests meaningful (terminal + retryable defaults, 429-vs-503, non-`ZamaError` inputs, malformed `retryAfter`).

Confidence: #1 80/100; #2 72/100; #3 70/100.

packages/sdk/src/errors/base.ts
```diff
@@ -92,10 +143,66 @@ export class ZamaError extends Error {
...
+ * ```ts
+ * try {
+ *   await sdk.decryption.decryptValues([{ encryptedValue, contractAddress }]);
+ * } catch (e) {
+ *   if (isRetryable(e)) {
+ *     // back off and retry (see retryAfterSeconds for a server-suggested delay)
+ *   } else {
+ *     throw e; // terminal — surface it
+ *   }
+ * }
```
ghermet, line 181: retry mechanism is baked into `fhevm/sdk` for relayer calls, this will be redundant post migration

packages/sdk/src/errors/base.ts
```diff
@@ -92,10 +143,66 @@ export class ZamaError extends Error {
...
+ * ```
+ */
+export function isRetryable(error: unknown): boolean {
```
ghermet:
```suggestion
export function isRetryable(error: unknown): error is ZamaError & { retryable: true } {
```

packages/sdk/src/errors/base.ts
```diff
@@ -92,10 +143,66 @@ export class ZamaError extends Error {
...
+ * {@link ZamaError} — unifying the `retryAfter` field that today lives on
+ * {@link RelayerRequestFailedError} and {@link RpcRateLimitError} only.
+ * `undefined` when the error isn't retryable, or is retryable but carries no
+ * server-driven delay (e.g. {@link WorkerTimeoutError}, {@link WorkerRecycledError},
+ * {@link DelegationNotPropagatedError} — retry those with your own backoff).
+ */
+export function retryAfterSeconds(error: unknown): number | undefined {
+  if (!(error instanceof ZamaError) || !("retryAfter" in error)) {
+    return undefined;
   }
```
ghermet: If you compose this method with `isRetryable` you can avoid duplicating this logic: 
```ts
if(isRetryable(e) && 'retryAfter' in e){
  return isValidRetryAfterSeconds(e.retryAfter) ? e.retryAfter : undefined;
}
return undefined

File: packages/sdk/src/errors/base.ts (hunk 4 above)
Comment (zama-cremaud) [reply packages/sdk/src/errors/base.ts:181]: Noted!

File: docs/gitbook/src/guides/handle-errors.md [hunk 7]
```diff
@@ -202,6 +207,46 @@ When `matchZamaError` returns `undefined` (because the error is not a `ZamaError
...
+
+async function decryptWithRetry(fn: () => Promise<bigint>, maxAttempts = 3): Promise<bigint> {
+  for (let attempt = 1; ; attempt++) {
+    try {
+      return await fn();
+    } catch (error) {
+      if (!isRetryable(error) || attempt >= maxAttempts) {
+        throw error; // terminal, or out of attempts -- surface it
+      }
+      const delaySeconds = retryAfterSeconds(error) ?? attempt * 2; // backoff when the server gives no hint
```
ghermet, line 236: `retryAfter` should directly be accessed from the error instead of using `retryAfterSeconds`

ghermet (approved):

PR #539, opened 2026-07-08
refactor(sdk): API surface audit — resolve all ae-forgotten-export leaks [SDK-209]

zama-cremaud:
## Summary

Resolves the `ae-forgotten-export` audit left open by #406, which silenced the rule (`logLevel: "none"`) rather than fixing the underlying leaks. Every symbol referenced by a public `.d.ts` but not exported from its entry point has been either **exported** (genuinely public) or **hidden** (internal type no longer named in a public signature), then the rule is flipped to `error` so the surface stays clean.

**Note: this diff was fully regenerated on 2026-07-13** after #458 (`replace @zama-fhe/relayer-sdk with @fhevm/sdk`, 324 files) merged into `prerelease` and deleted the entire "transport cluster" this PR previously touched (`relayer-sdk.ts`, `relayer-node.ts`, `relayer-web.ts`, `relayer-dispatcher.ts`, the `worker/*` layer). A rebase wasn't viable — those files no longer exist. The audit was re-run from scratch against current `prerelease`; the methodology and the validated fix patterns (structural retyping over named aliases, checking existing docs before export/hide calls) carried over, but the diff itself is new.

- `cb80eace` — re-export symbols already public at the main `@zama-fhe/sdk` entry point but missing from secondary entry points (`/web`, `/viem`, `/ethers`, `/node`, `/query`). Purely additive, zero design decisions.
- `0a536b3f` — export the remaining straightforward cases: `ErrorForCode`, `EthersCallProvider`/`EthersTransactionSigner`, `RelayerOptions`/`FhevmRelayerOptions`/`FhevmRuntimeConfig`/`FhevmClient`/`FhevmClientOptions` (the latter because `FhevmRelayerSDK extends Pick<FhevmClient, …15 methods…>` — `Pick<T,K>` keeps naming `FhevmClient` in the emitted `.d.ts` regardless, and hand-duplicating 15 method signatures to avoid that would just create drift risk against a fast-moving alpha dependency), `PermissionSchema`/`StoredTransportKeyPairSchema`/`transactionOperationMetadata` (already `export const` in their own module, just missing from the barrel — not a new design call). `RequiredBy` inlined at its single use site. `ChainRouter` exported as a type only, matching the same treatment its predecessor `RelayerDispatcher` already had. `FhevmRelayer` stays hidden — it leaked only via unnecessary narrowing on `NodeRelayerConfig.createRelayer`'s return type (nothing internal reads the narrowed type; same fix already applied last round to `RelayerNode`), so that narrowing is dropped in favor of the already-public base `RelayerConfig`/`FhevmRelayerSDK` contract.
- `bf6135ab` — the `Permits`/`Delegations`/`Decryption` namespace constructors name the concrete service classes (`CachingService`, `CredentialService | undefined`, `DelegationService`, `DecryptionService`), per review. This supersedes `66c18e4f`'s anonymous structural literals, which review rejected. Naming a class in a public `.d.ts` signature requires exporting it with the rule at `error`, so the four service classes (plus `CredentialServiceConfig`, referenced by `CredentialService`'s constructor) are **type-only exports tagged `@internal`** from the two entries that inline the namespaces (main and `/query`); `/query` additionally gains `DelegationNotFoundError`/`DelegationExpiredError` (already public at the main entry, referenced by `DelegationService.findInactiveDelegations`) — same additive pattern as `cb80eace`. `ae-internal-missing-underscore` is silenced in `api-extractor.base.json`: underscore-renaming the classes would defeat the point of naming them, and without the override the warnings get embedded into the `etc/*.api.md` report files. Alternatives verified and rejected: `stripInternal` (supported by `rolldown-plugin-dts` via `compilerOptions`) would elide *every* `@internal` declaration from the `.d.ts` — including `ZamaSDK.onWalletAccountChange`/`relayer`/`logger`/`emitEvent`, which react-sdk calls and the docs reference — breaking the react-sdk build; a `private` constructor + static factory re-leaks the same types through the factory's public signature.
- `9727bdec` — flip `"ae-forgotten-export"` from `"none"` to `"error"` in `api-extractor.base.json`, enforced by the existing `pnpm api-report:check` CI gate.

## Verification

- `pnpm api-report:check` (exact CI command, forced/uncached): 0 errors/0 warnings, exit 0, across all 6 sdk reports + 2 react-sdk reports — re-run in full after the review round (`bf6135ab`).
- `pnpm build`, `pnpm typecheck` (6/6, forced), `pnpm lint`, `pnpm vitest run` (sdk 1149 + react-sdk 236), `pnpm llm:check`: all green after the review round.
- Independently re-verified (not just trusting the implementation pass): rebuilt, re-ran the CI command forced/uncached, re-ran the full test suite, and manually spot-checked the service export pattern and `ChainRouter`'s export style for consistency before pushing.

## Test plan

- [x] `pnpm api-report:check` passes with the rule at `error`
- [x] `pnpm build && pnpm typecheck && pnpm lint && pnpm vitest run` all green
- [ ] CI green on this PR
- [ ] Reviewer sanity-check on the `FhevmClient`/`FhevmRelayerSDK` export call and the `FhevmRelayer` hide call


ghermet:
@zama-cremaud This is duplicating exports across entries, I'm not sure about what is the value of this ticket

ankurdotb:
> @zama-cremaud This is duplicating exports across entries, I'm not sure about what is the value of this ticket

@ghermet Can you make a judgement call on whether this is needed, and to what extent?

zama-cremaud:
@ghermet 

> This is duplicating exports across entries, I'm not sure about what is the value of this ticket

This is indeed intentional duplication.
Each subpath (/ethers, /viem, /node, /query, /web) compiles its own separate .d.ts file.
So a type used in that entry's public API has to be exported from that same entry.
'api-extractor' has no per-symbol exception, only a global on/off switch.

We've done the same kind of fix before here and there.
This PR just finishes it everywhere so we can turn 'ae-forgotten-export' into a permanent CI check, instead of patching "leaks" one at a time.

Seems fair to ask if that's worth it. I'll trust your jugement on this!

ghermet:
> @ghermet
> 
> > This is duplicating exports across entries, I'm not sure about what is the value of this ticket
> 
> This is indeed intentional duplication. Each subpath (/ethers, /viem, /node, /query, /web) compiles its own separate .d.ts file. So a type used in that entry's public API has to be exported from that same entry. 'api-extractor' has no per-symbol exception, only a global on/off switch.
> 
> We've done the same kind of fix before here and there. This PR just finishes it everywhere so we can turn 'ae-forgotten-export' into a permanent CI check, instead of patching "leaks" one at a time.
> 
> Seems fair to ask if that's worth it. I'll trust your jugement on this!

Okay i've looked into the [documentation](https://api-extractor.com/pages/messages/ae-forgotten-export/) this is useful to keep as an error despite the duplication because it prevents these issues:
- Third parties can't name the type to annotate their own code (import type { X } fails).
- The type is missing from generated API docs.
- The API can be literally uncallable (e.g. an unexported enum/union a consumer must pass a value from).

I'm not sure about commit `66c18e4f` approach to satisfy the api-report thought

packages/sdk/src/node/index.ts
```diff
@@ -10,15 +10,33 @@
...
+  ClearValue,
+  EIP712TypedData,
+  EncryptParams,
+  EncryptInput,
+  EncryptedValue,
+} from "../relayer/types";
 // Decrypt parameter/result types — aligned with the canonical Zama glossary (see main entry).
 export type { DecryptPublicValuesResult } from "../relayer/types";
 
+// SDK-209: re-export types already public at the main `@zama-fhe/sdk` entry point —
```
ghermet: You can drop the comments in the barrels

zama-cremaud:
@ghermet

> I'm not sure about commit `66c18e4f` approach to satisfy the api-report thought

Fair. Approch simplified by dropping redundancy on private field's type annotation. Let it infer from the constructor assignment instead.

packages/sdk/src/namespaces/permits.ts
```diff
@@ -27,16 +26,29 @@ import { requireAlignedWalletAccount, requireChainAlignment } from "../utils/ali
...
+            delegator?: Address,
+          ): Promise<SerializedTransportKeyPairWithPermissions>;
+          hasPermit(contracts: readonly Address[], delegator?: Address): Promise<boolean>;
+          warmTransportKeyPair(address: Address): Promise<void>;
+          revokePermits(contracts?: readonly Address[]): Promise<void>;
+          clearCredentials(): Promise<void>;
+          revokeTransportKeyPair(scopeId: string): Promise<void>;
+          warmTransportKeyPairScope(scopeId: string): Promise<void>;
+        }
+      | undefined;
```
ghermet: This needs to be typed as `CredentialService | undefined;`

packages/sdk/src/namespaces/permits.ts
```diff
@@ -27,16 +26,29 @@ import { requireAlignedWalletAccount, requireChainAlignment } from "../utils/ali
...
+  readonly #credentialService;
   readonly #logger: GenericLogger;
 
   /** @internal */
   constructor(opts: {
     signer: GenericSigner | undefined;
     provider: GenericProvider;
-    cachingService: CachingService;
-    credentialService: CredentialService | undefined;
+    cachingService: { clearForRequester(requester: Address): Promise<void> };
```
ghermet: This needs to be typed as `CachingService`

packages/sdk/src/node/config.ts
```diff
@@ -1,12 +1,15 @@
...
 import { FhevmRelayer } from "../relayer/fhevm-relayer";
 import type { RelayerOptions } from "../relayer/types";
 
-/** Node transport — drives the FHE backend directly on the calling thread. */
+/**
+ * Node transport config. `createRelayer` is an internal self-registration
+ * hook (see `RelayerConfig`) — `FhevmRelayer` is an implementation detail,
+ * not part of the public API, so this deliberately doesn't narrow its type
+ * beyond the base (mirrors `WebRelayerConfig`/`CleartextRelayerConfig`).
+ */
```
ghermet: Can we revert to previous comment?  these are llm thoughts

packages/sdk/src/namespaces/permits.ts
```diff
@@ -11,7 +10,7 @@ import { requireAlignedWalletAccount, requireChainAlignment } from "../utils/ali
  *
  * Exposed as `sdk.permits`. Owns the SDK-level guards (chain alignment, empty-array
  * short-circuit, decrypt-cache invalidation) and delegates the actual work to the
- * internal {@link CredentialService}.
+ * internal credential service.
```
ghermet: where's the link?

packages/sdk/src/namespaces/delegations.ts
```diff
@@ -19,13 +18,37 @@ import { requireConfigured } from "../errors";
...
   readonly #provider: GenericProvider;
-  readonly #delegationService: DelegationService;
+  readonly #delegationService;
 
   /** @internal */
   constructor(opts: {
     signer: GenericSigner | undefined;
     provider: GenericProvider;
-    delegationService: DelegationService;
+    delegationService: {
```
ghermet: This needs to be typed as `DelegationService`

packages/sdk/src/namespaces/delegations.ts
```diff
@@ -9,7 +8,7 @@ import { requireConfigured } from "../errors";
  *
  * Exposed as `sdk.delegations`. Owns the SDK-level guards (signer requirement, chain
  * alignment, delegator address resolution from the wallet account) and delegates the
- * actual work to the internal {@link DelegationService}.
+ * actual work to the internal delegation service.
```
ghermet: Where's is the link?

ghermet (approved):

PR #541, opened 2026-07-09
feat(sdk): chunk decryption requests to respect the relayer's per-request bit budget

zama-cremaud:
## Summary

Closes [SDK-252](https://linear.app/zama/issue/SDK-252/per-request-decryption-bit-budget-2048-bit-cap-isnt-surfaced-batching).

The relayer enforces a hard **2048-cleartext-bit cap per decryption request**, client-side, inside `@zama-fhe/relayer-sdk`. `sdk.decryption.decryptValues` / `delegatedDecryptValues` / `delegatedBatchDecryptValues` never surfaced this budget and never chunked requests to respect it — a batch whose handles summed past the cap was hard-rejected, and consumers that batch many handles (e.g. a server-side indexer decrypting amounts pulled from event logs) had to hardcode a chunk size against an unexported, undocumented constant. That guess silently breaks if the cap ever changes, or if a batch mixes handle types of different bit-widths (`ebool`=2 bits vs `euint256`=256 bits vs `eaddress`=160 bits, etc.).

- `DecryptionService#decrypt` now splits each contract's uncached handles into sub-requests that stay under the budget before dispatching to the relayer — fixes `decryptValues`, `delegatedDecryptValues`, and `delegatedBatchDecryptValues` at their single common choke point, with zero caller-facing change for well-formed input.
- `MAX_DECRYPTION_REQUEST_BITS` is exported (`@zama-fhe/sdk`) so the budget is discoverable.
- Reuses the already-vendored, zero-WASM-dependency FHE-type table (relocated from `relayer/cleartext/fhe-type.ts` to `utils/fhe-type.ts`, its true isomorphic scope) instead of re-deriving FHE-type bit-widths or importing the real `relayer-sdk`'s version (unsafe in isomorphic code — pulls in native/WASM `node-tfhe`/`node-tkms` at module load).

Deliberately out of scope (see PR discussion / ticket for rationale): packing handles from *different* contracts into a single relayer request (the wire protocol supports it, the current per-contract dispatch doesn't use it) — a larger, separate optimization tracked as a possible follow-up.

## Test plan

- [x] New unit tests for `encryptionBitsForHandle` / `chunkHandlesByBitBudget` (`packages/sdk/src/utils/__tests__/handles.test.ts`)
- [x] New integration-level test: 40 same-contract `euint64` handles (2560 bits) correctly split into a 32- and an 8-handle relayer call (`decryption-service.test.ts`)
- [x] Full `@zama-fhe/sdk` suite: 94 files / 1242 tests passing, no regressions in the 55 pre-existing fake-handle fixtures across 20 files
- [x] `typecheck` / `lint` / `format` clean on all changed files (sdk + react-sdk)


packages/sdk/src/index.ts
```diff
@@ -85,7 +85,11 @@ export type {
 export { WrappersRegistry, DefaultRegistryAddresses } from "./wrappers-registry";
 export type { WrappersRegistryConfig, ListPairsOptions } from "./wrappers-registry";
 export { Token, WrappedToken, type BatchBalancesResult, type BatchDecryptAsOptions } from "./token";
-export { ZERO_ENCRYPTED_VALUE, isEncryptedValueZero } from "./utils/handles";
+export {
+  ZERO_ENCRYPTED_VALUE,
+  isEncryptedValueZero,
+  MAX_DECRYPTION_REQUEST_BITS,
```
ghermet: [AUTOMATED] **Blocking (CI gate): API report not regenerated for the new public export.**

This adds `MAX_DECRYPTION_REQUEST_BITS` to the public barrel, but `packages/sdk/etc/sdk.api.md` still lists only its siblings (`ZERO_ENCRYPTED_VALUE`, `isEncryptedValueZero`) — no `.api.md` file is touched in this PR. `api-report:check` (in `.github/workflows/ci.yml`) regenerates the reports with api-extractor and diffs them, so it will fail on the stale report.

Fix: `pnpm api-report:sdk` and commit the updated `packages/sdk/etc/sdk.api.md`.

Confidence: 95/100

ghermet (commented):
[AUTOMATED]

## Code review — SDK-252 (request-chunking for decryption)

Reviewed for correctness, error handling, and conventions (two parallel agents plus a manual trace of the `DecryptionService` integration). Solid PR:

- Chunking is applied to the **uncached** handle set only; per-chunk results merge/cache back by handle; per-contract error classification and the `DecryptStart`/`DecryptEnd`/`DecryptError` events are all preserved. `pLimit(…, 5)` concurrency is unchanged — the task count just moves from per-contract to per-chunk.
- `encryptionBitsForHandle` reads the correct type byte (bits 8–15, matching `RelayerCleartext`'s decoder); the 256-bit fallback conservatively **over**-estimates for unknown/deprecated types, so a chunk can never exceed the real budget.
- Greedy chunking boundaries are correct (exact-fit stays packed; an oversized handle still gets a solo chunk so the loop always makes progress; empty input → `[]`).
- The `fhe-type.ts` relocation migrated all three importers with no re-export shim left behind, and the `MAX_CONTRACTS_PER_PERMIT` / `chunkContracts` duplication pattern the comments cite genuinely exists and is followed.
- Tests are meaningful (boundary, mixed widths, unknown-type fallback, and the 32+8 integration split).

One blocking-for-CI item — see the inline comment.

packages/sdk/etc/sdk.api.md
```diff
@@ -11245,6 +11245,9 @@ export function matchZamaError<R>(error: unknown, handlers: { [K in ZamaErrorCod
     _?: (error: unknown) => R;
 }): R | undefined;
 
+// @public
+export const MAX_DECRYPTION_REQUEST_BITS = 2048;
```
ghermet: Why would this constant be public?

packages/sdk/src/services/__tests__/decryption-service.test.ts
```diff
@@ -88,6 +88,50 @@ describe("DecryptionService", () => {
     );
   });
 
+  test("decryptValues splits one contract's handles across the 2048-bit request budget (SDK-252)", async ({
```
ghermet: No need to tag tickets in the codebase 👍

ghermet (approved):

PR #544, opened 2026-07-09
refactor(codemod): rewrite renamed credentials config keys, flag silent drops [SDK-249]

zama-cremaud:
## Summary

Closes [SDK-249](https://linear.app/zama/issue/SDK-249/codemod-rewrite-renamed-credentials-config-keys-flag-silently-dropped).

`codemods/sdk-migration-v3` had no deterministic rule for the credentials config-key renames from SDK-134/SDK-204 (`sessionStorage→permitStorage`, `keypairTTL→transportKeyPairTTL`, `sessionTTL→permitTTL`). `ZamaConfigBase` has no index signature, so TS's excess-property check only catches a stale key on a **direct** `createConfig({...})` literal — a config assembled via a variable or spread keeps the key past a v3 upgrade with no signal, and the runtime silently drops it.

- **ast-grep rewrite** (`rules/rename-credentials-config-keys.yml`) for the direct-literal case, scoped to `createConfig()`/`createZamaConfig()` call arguments via a relational `inside` guard. This is deliberate, not optional: `sessionStorage` collides with the browser's `window.sessionStorage` global, so an unscoped pattern would corrupt unrelated code.
- **Runtime guard** in `buildZamaConfig` (`packages/sdk/src/config/build.ts`) warns via the configured logger when it sees a stale top-level key — covers the variable/spread case the codemod structurally can't reach without reopening the collision risk above.
- **SDK-169 namespace-method renames** (`sdk.allow→sdk.permits.grantPermit`, etc. — 13 methods, `scripts/rename-sdk-namespace-methods.ts`): the ticket's third done-when item asked to verify the prompt-based coverage of the credentials class renames rather than assume it works. On inspection it was incomplete (no explicit per-method mapping), and this exact class of rename is syntactically safe the same way the existing hook renames are, so it's now deterministic — scoped by tracing local `new ZamaSDK(...)` bindings (same import-provenance technique as the existing `createZamaConfig→createConfig` script). `isDelegated→isActive` additionally gets a trailing comment flagging its expiry-check semantic shift (not just a rename — see the SDK-169 PR).
- **`ai-upgrade` prompt corrected**: dropped the now-redundant namespace-method bullet (handled deterministically now), added `StoredCredentials`/`CredentialsManagerConfig` — verified these don't have a 1:1 replacement type (they split into `StoredTransportKeyPair` + `Permission`), so they're flagged for the AI/manual tail rather than auto-rewritten.

Both new rules are scoped defensively (relational `inside` for the ast-grep rule, same-file binding-provenance tracing for the JSSG script) and documented as best-effort in the README's "Known limitations" — neither claims full coverage for configs/instances assembled indirectly.

## Out of scope (flagged, not fixed here)

While verifying the codemod's claimed "3.0.x floor," found that `ZamaProvider` took flat JSX props (not a `createConfig()` object) at `v3.0.1` — `createConfig`/`createZamaConfig` didn't exist yet at that tag. There's no rule for that structural migration, so a literal v3.0.1 consumer wouldn't be covered regardless of this PR. Different problem (relayer/chains reshape, not credentials) — noting here rather than expanding scope; happy to file a follow-up ticket if wanted.

## Test plan

- [x] `ast-grep test` — 18/18 (12 existing + 6 new), including `valid` cases proving `window.sessionStorage` and non-`createConfig` objects are untouched
- [x] `codemod jssg test` — all scripts green, including 2 new fixtures for the namespace-method rename (plain binding + `this.sdk` class-property + aliased import)
- [x] End-to-end chain (`test-e2e.mjs`) — extended `tests/_e2e` fixtures to exercise both new steps together with the existing chain; converges and is idempotent
- [x] `packages/sdk` — `tsc --noEmit` clean, 4 new unit tests for the `buildZamaConfig` guard, `oxfmt --check` clean
- [x] Full workspace `turbo run typecheck` clean


codemods/sdk-migration-v3/scripts/rename-sdk-namespace-methods.ts
```diff
@@ -0,0 +1,143 @@
...
+// `.allow(...)` or `.isAllowed(...)` on an arbitrary receiver could belong to
+// any unrelated class. So this only rewrites `X.method(...)` where X is traced,
+// same-file, to a local `new ZamaSDK(...)` binding imported from `@zama-fhe/sdk`
+// -- the same import-provenance technique as
+// core-rename-createzamaconfig-to-createconfig.ts.
+//
+// Known limitation (best-effort, not exhaustive): an SDK instance obtained any
+// other way -- a factory function, a React hook, destructuring, re-assignment
+// through an intermediate variable -- is NOT traced and is left for the
+// typecheck + manual fix / AI tail. See README "Known limitations".
```
ghermet: This comment section could use a bit of trimming, also no need to reference linear tickets

packages/sdk/src/config/build.ts
```diff
@@ -24,6 +47,7 @@ export function buildZamaConfig(
 ): ZamaConfig {
   const { storage, permitStorage } = resolveStorage(params.storage, params.permitStorage);
   const logger = new LoggerService(params.logger);
+  warnRenamedConfigKeys(params, logger);
```
ghermet: codemods are meant to avoid warnings like these

ghermet (approved):

PR #545, opened 2026-07-13
feat(sdk): unified sdk.events multi-listener event stream (SDK-100)

zama-cremaud:
## Summary

Progresses SDK-100. This implements the global-event-stream deliverable of the epic (multi-listener `sdk.events`), not the full epic — see "Scope notes" below for what's still open. Not tagged with a closing keyword since two of the epic's three sub-tasks aren't covered here.

Replaces the single-callback `ZamaSDK.emitEvent()` / `config.onEvent` proxy with `sdk.events`, a typed, multi-listener event service that lets independent consumers (logging, analytics, monitoring) subscribe to the SDK-wide event stream without stepping on each other or on per-operation callbacks.

- `sdk.events.on(type, listener)` — subscribe to one event type, returns an unsubscribe function
- `sdk.events.once(type, listener)` — auto-unsubscribes after the first matching event
- `sdk.events.subscribe(listener)` — catch-all subscription for every event
- `onEvent` in `createConfig`/the `ZamaSDK` constructor is now sugar for a single `subscribe()` call registered at construction time (fully backward compatible)
- A throwing listener is caught, logged via the configured logger, and never affects the operation that triggered the event or other listeners
- `sdk.dispose()` now also clears all registered listeners

## Why

The SDK already has a rich, well-typed `ZamaSDKEvent` union (~20 event types) but only supported a single global callback wired at construction time. That made it impossible for two independent concerns (e.g. an app's own logger and a vendored analytics SDK) to both observe SDK events without composing callbacks by hand. `EventService` centralizes fan-out, per-type filtering, and error isolation behind a small typed API, and is the first `*Service` class exposed on the public surface (all others remain internal, accessed only via `sdk.<namespace>`).

## Scope notes

SDK-100 has three sub-tasks; this PR covers only the second:

1. **Not covered.** Audit per-operation callbacks (`ShieldCallbacks`, `TransferCallbacks`, `UnshieldCallbacks`) for completeness — e.g. `ShieldCallbacks.onShieldSubmitted` is missing the `shieldPath` that `ShieldSubmittedEvent` carries, and `onApprovalSubmitted` is missing `ApproveUnderlyingSubmittedEvent`'s `step`.
2. **This PR.** Replace the single `onEvent` callback with a proper multi-listener `sdk.events` API.
3. **Not covered.** Reconcile the callback/event overlap so a consumer never needs to wire both to get a complete picture of one operation — currently still required for shield (see #1).

I looked at SDK-218 (workflow-engine refactor for shield/unshield resumability, `Backlog`, not a child of SDK-100) as a possible home for #1/#3 — its spec doesn't mention `onShieldSubmitted`/`shieldPath`/the callback-event overlap anywhere, so nothing currently owns those two sub-tasks. Filing follow-up tickets for #1 and #3 rather than asserting SDK-218 will absorb them.

Other notes:
- Investigated and closed PR #409 (abandoned prior attempt at this ticket, 5+ weeks stale, zero review, cancelled for Clear-Signing-scope reasons unrelated to the technical design) — this is a clean reimplementation, not a rebase, keeping the existing swallow-and-warn error semantics rather than #409's AbortSignal/re-throw approach (unprecedented elsewhere in the codebase).
- SDK-181 (a prior, more detailed design for this same feature, `Canceled` alongside the Clear-Signing initiative it was a prerequisite for) wanted `emit` to be async/awaited so listeners could gate SDK operations (e.g. clear-signing waiting for a UI prompt to render before opening the wallet). This PR's `emit` is synchronous, matching the `emitEvent` it replaces — sufficient for SDK-100's stated scope, but worth knowing that making `emit` async later, if Clear-Signing (SDK-180) resumes, will be a breaking change to an already-public, already-documented symbol.

## Commits

- `feat(sdk): add EventService with typed multi-listener subscribe`
- `refactor(sdk): inject EventService into ZamaSDK, retire emitEvent proxies`
- `feat(sdk): expose sdk.events public surface, update api-report`
- `docs(sdk): document sdk.events multi-listener API`
- `docs(sdk): clarify error events are passive-logging only`
- `style(sdk): fix oxfmt formatting on observe-sdk-events guide`
- `fix(sdk): guard once() against double-fire on reentrant emission`
- `docs(sdk): correct sdk.events guide claims flagged in review`

## Test plan

- [x] `pnpm --filter @zama-fhe/sdk test:run` — 1237/1237 passing, incl. 12 `EventService` tests (fan-out, type filtering, unsubscribe, `once` incl. pre-fire cancel and reentrant-emission double-fire regression, catch-all, throwing-listener isolation, `dispose`)
- [x] `pnpm --filter @zama-fhe/sdk typecheck` — clean
- [x] `pnpm api-report` — regenerated, diff reviewed (clean addition of `EventService`/`EventServiceConfig`/`TypedZamaSDKEventListener`, removal of internal `emitEvent`)
- [x] `pnpm docs:check-links` — 91/91 pages resolve
- [ ] File follow-up tickets for SDK-100 sub-tasks #1 and #3 (callback completeness audit, callback/event reconciliation)
- [ ] Coordinate with SDK-218 owner on the `shieldPath`/`step` callback gap


ghermet:
@zama-cremaud duplicates #409, rework the 409 PR if you need it for observability

PR #546, opened 2026-07-13
feat(credentials): opt-in shared ML-KEM keypair scope for shared-tenant operators [SDK-142]

zama-cremaud:
## Summary

Closes [SDK-142](https://linear.app/zama/issue/SDK-142/featcredentials-opt-in-shared-ml-kem-keypair-scope-for-shared-tenant).

The credentials refactor ([SDK-134](https://linear.app/zama/issue/SDK-134)) hard-codes one ML-KEM transport key pair per signer address. That's the right default for separately-tenanted deployments (browser/mobile dApps, treasury wallets), but it's pure overhead for a **shared-tenant Wallet-as-a-Service (WaaS) operator**: a store holding 1000 client wallets gets no defense-in-depth from 1000 separate key pairs — a breach compromises them all together regardless — while still paying 1000x the generation/storage/management cost.

This PR adds an opt-in `transportKeyPairScope` config option: signers configured with the same scope identifier (e.g. a tenant ID) share one transport key pair. Default behavior (no scope configured) is unchanged.

- **Default unchanged** — no scope configured behaves exactly like today, one key pair per signer.
- **Permits stay per-signer, always** — the EIP-712 signature is inherently tied to the signing wallet, scope or not.
- **Two revocation tiers that must never be conflated:**
  - Signer-level (`sdk.permits.clear()` / `revokePermits()`) only ever touches the calling signer's own permits — the shared key pair is untouched, because it was never stored under any individual signer's slot to begin with. This is enforced at the vault layer (`TransportKeyPairVault.clear()` always targets the per-signer key, never the scope key), not just by caller discipline.
  - Operator-level (`sdk.permits.revokeTransportKeyPair(scopeId)`, new) invalidates the shared key pair. Every permit in the scope embeds that key pair's public key, so they're all treated as stale on next access automatically — no permit is touched directly, and no wallet needs to be connected.

## Design notes / things I verified rather than assumed

- Checked PR #458 (`@fhevm/sdk` swap) — it touches zero files under `credentials/`, so this is orthogonal and safe to land independently of that merge.
- The `derivationSecret` ticket ([SDK-135](https://linear.app/zama/issue/SDK-135)) doesn't exist in the code yet (still Backlog) and its own draft spec keys its HKDF `salt` on `walletAddress`. Building the scope's identity abstraction (`scope ?? signerAddress`) now, before SDK-135 lands, avoids a breaking change to that salt input later — SDK-135 can just reuse the same identity concept when it's built. The "composes with derivationSecret" acceptance criterion is structurally ready but can't be verified end-to-end until SDK-135 ships.
- Documented (not "fixed", since `GenericStorage` has no compare-and-swap) a pre-existing race: two signers hitting a brand-new scope for the first time can race the underlying storage. This already existed per-signer (multi-tab/process); a scope just makes it more likely since more callers share one key.
- Node's default storage (`asyncLocalStorage`) isolates a fresh store per request, which would silently defeat scope-sharing entirely — flagged explicitly in the docs since it's the typical Node server recommendation.

## Test plan

- [x] Two signers in the same scope share one key pair; different/unscoped signers stay independent (`keypair-vault.test.ts`, `credential-service.test.ts`)
- [x] Permits stay per-signer regardless of scope (`credential-service.test.ts`)
- [x] `clear()`/`revokePermits()` never delete a scope's shared key pair
- [x] `revokeTransportKeyPair()` deletes the shared key pair and makes existing permits stale via the embedded public key (not via direct permit deletion)
- [x] `revokeTransportKeyPair()` / `warmTransportKeyPairScope()` guards: throw when no scope is configured, or when `scopeId` doesn't match
- [x] `pnpm typecheck && pnpm lint && pnpm test:run` (full `sdk` package: 1234 tests, 0 regressions) && `pnpm build`
- [x] `pnpm api-report` / `api-report:check` — clean diff (`revokeTransportKeyPair`, `warmTransportKeyPairScope`, `transportKeyPairScope`)
- [x] `pnpm docs:check-links` — 90 pages, all anchors resolve
- [x] `pnpm llm:build` — corpus regenerated


zama-cremaud:
Addressed a local review pass on this PR (2 blocking findings, 2 low-cost non-blocking improvements applied):

1. **`rotateScope()` could resolve successfully while the shared key pair was still live** — `clearScope()` was swallowing `storage.delete()` failures via the same best-effort helper `clear()` uses, but `rotateScope()` is the operator's suspected-compromise primitive; a false all-clear during an incident is the wrong failure mode. Fixed: it now propagates storage failures instead of logging-and-dropping them. Added tests at both the vault and `CredentialService` level asserting the rejection.
2. **A scope-sharing test passed independently of whether sharing actually worked** — two vaults in the "two signers share one key pair" test each had their own key generator, so the equality assertion held even under a broken implementation (both generators' first call is deterministic). Fixed to share one generator, matching the pattern already used two tests below it. Verified by re-injecting the exact regression and confirming the test now catches it.
3. (non-blocking) Renamed the internal `scope` config field to `keyPairScope` (private, not public API) to avoid reading ambiguously next to the pre-existing, unrelated `PermissionScope` concept in the same class.
4. (non-blocking) Strengthened the race-condition docs with concrete operator guidance (pre-warm the scope once before opening concurrent traffic) and a more precise severity note (cross-trust-boundary blast radius + correlated thundering-herd on rotation/TTL, not just "more likely").

Full typecheck/lint/test (1236 tests)/api-report/docs-check-links all green after the fix.

zama-cremaud:
Addressed a second review pass (v2). Both v1 blocking findings were independently re-verified as genuinely fixed (mutation-tested by the reviewer). One new advisory finding from v2, fixed rather than just documented:

**The "pre-warm before opening traffic" guidance I added in the previous commit recommended the wrong primitive.** `sdk.permits.warmTransportKeyPair()` silently no-ops without a connected wallet-account snapshot — a precondition inherited from its per-signer design that doesn't conceptually apply to a scope-wide key, and it's the most likely state an operator is in exactly when pre-warming ahead of traffic (no end-user connected yet). This was asymmetric with `rotateScope()`, which was already designed to need no wallet account at all.

Fix: added `sdk.permits.warmScope()` as the real snapshot-free counterpart to `rotateScope()`, rather than just caveating the docs. `TransportKeyPairVault.getOrCreate`/`.readStored` are refactored onto shared internal helpers so `warmScope()` can generate directly against the scope's storage key, with no signer address involved. `CredentialService.warmScope()` and `Permits.warmScope()` mirror `rotateScope()`'s throw-if-unscoped contract. Docs updated to recommend `warmScope()` instead, with the asymmetry called out explicitly so it doesn't get reintroduced.

Added tests at all three layers, including one that specifically pins down the property the review flagged: `warmScope()` succeeds with a signer configured but no connected wallet snapshot, in a case where `warmTransportKeyPair()` would silently do nothing.

Full typecheck/lint/test (1246 tests)/api-report/docs-check-links green.

ghermet:
[AUTOMATED]

## Code Review

No high-signal issues found. This is a well-scoped, thoroughly documented change.

**Reviewed for:** runtime bugs & security (2 independent passes), guideline/glossary compliance, and error handling.

Points specifically verified as **correct**:
- **Scope identity resolution** (`#identityKey` = scope key when configured, else per-signer key) is applied consistently across every read/create/dedup path in `keypair-vault.ts`.
- **Two-tier revocation isolation** — signer-level `clear()`/`clearCredentials()` deliberately never touch the shared scope key; only `clearScope()`/`rotateScope()` do. Behavior is correct and documented in the vault, the service, and `security-model.md`.
- **`rotateScope()` targeting & guards** — matches the configured scope, rejects mismatched/absent scope with `ConfigurationError`, and propagates storage-delete failures (non-best-effort, as documented).
- **Permit staleness after rotation** — permits embed the keypair public key and are pruned via `listUsableAndPrune`; no permit is mutated directly.
- **No stale in-memory keypair** — the vault holds no persistent keypair cache (`#pending` is an in-flight-generation dedup map cleared in `.finally()`), so rotation takes effect on the next access with nothing to evict.
- **Config/schema wiring** — `keyPairScope` threads through `build.ts`/`zama-sdk.ts`/fixtures; empty-string scope is rejected at config build.

Two items were investigated and dismissed after tracing the source: a suspected `clear()`-doesn't-wipe-scoped-keypair bug (intended two-tier design, documented) and a suspected rotate-doesn't-evict-in-memory-cache bug (no such cache exists).

docs/gitbook/src/concepts/security-model.md
```diff
@@ -147,6 +167,12 @@ Permits can be revoked programmatically via `sdk.permits.revokePermits()` or aut
 
 After revoking permits, the transport key pair remains in storage. Use `sdk.permits.clear()` to also wipe the key pair.
 
+**With a shared `transportKeyPairScope` configured, these signer-level operations never touch the shared key pair** — `sdk.permits.clear()` for one signer only ever removes that signer's own permits, exactly like `revokePermits()`. This is deliberate: one end-user disconnecting must never invalidate every other signer sharing the scope's operator-controlled key store.
+
+Invalidating the shared key pair itself is a distinct, operator-level operation: `sdk.permits.rotateScope(scopeId)`. It deletes the scope's key pair; every permit in the scope embeds that key pair's public key, so they're all treated as stale on next access — no permit needs to be touched directly, and no wallet needs to be connected. Use it for suspected compromise, periodic rotation, or scope decommissioning — never as a side effect of a single signer's revoke.
```
ghermet: `rotateScope` might be slightly misleading, this revokes a transport keypair with a specific scope.
I would use `sdk.permits.revokeTransportKeyPair()` overloaded with `scopeId` as an arg if you want to scope it.

packages/sdk/src/credentials/__tests__/keypair-vault.test.ts
```diff
@@ -124,3 +124,228 @@ describe("TransportKeyPairVault", () => {
...
+    // deterministic first call) regardless of whether the implementation under test
+    // actually shares anything.
+    const generator = makeGenerator();
+    const scoped = () =>
+      new TransportKeyPairVault({
+        generator,
+        storage,
+        ttl: TTL_SECONDS,
+        logger: makeLogger(),
+        keyPairScope: "tenant-1",
```
ghermet:
```suggestion
        scope: "tenant-1",
```

packages/sdk/src/credentials/keypair-vault.ts
```diff
@@ -66,16 +96,46 @@ export class TransportKeyPairVault {
   /**
    * Return the cached transport key pair, generating and persisting a fresh one if absent or expired.
    *
-   * @remarks Deduplicates concurrent calls — simultaneous requests share one generation promise.
+   * @remarks Deduplicates concurrent calls within this process — simultaneous requests
+   * share one generation promise. Across separate processes/instances sharing a scope,
```
ghermet: use the jsoc skill to trim the docs

packages/sdk/src/credentials/keypair-vault.ts
```diff
@@ -66,16 +94,35 @@ export class TransportKeyPairVault {
...
+  /**
+   * @param strict - When `true`, a `storage.set` failure during persistence rejects
+   *   instead of being logged and swallowed. Used by {@link warmScope}, which — like
+   *   {@link clearScope} — must not resolve successfully while its write silently failed.
+   */
+  async #getOrCreateByKey(
+    identityKey: string,
+    options?: { strict?: boolean },
+  ): Promise<StoredTransportKeyPair> {
+    const existing = this.#pending.get(identityKey);
```
ankurdotb: [AUTOMATED] **Blocking:** `strict` gets lost through the `#pending` dedup cache — `warmScope()`'s "not best-effort" guarantee isn't actually always honored.

`#getOrCreateByKey` dedupes purely on `identityKey` here, before ever looking at the caller's own `options.strict`. For a scoped vault, `getOrCreate()` (non-strict, called from `grantPermit`) and `warmScope()` (strict) resolve to the *same* identity key — so whichever call registers the pending promise first decides strictness for both:

- A live signer's `grantPermit` is in flight first → a concurrent `warmScope()` inherits non-strict semantics → a transient `storage.set` failure gets swallowed → `warmScope()` resolves successfully while nothing was persisted. That directly contradicts the class doc a few lines below ("a resolved promise here is meant to mean the key pair is actually warmed in shared storage").
- Reverse: `warmScope()` is in flight first → a concurrent `grantPermit()`'s `getOrCreate()` inherits *strict* semantics → a transient storage blip now makes an ordinary signer's permit grant throw, when it's supposed to be best-effort.

Realistic trigger: a WaaS server that pre-warms a scope while already serving live traffic on it. Confirmed independently by two separate review passes plus a direct trace of this method — not a single-source finding.

Possible direction: key `#pending` on `(identityKey, strict)` rather than `identityKey` alone, or have a strict caller that finds an existing non-strict in-flight promise fall through to its own persistence instead of adopting the shared one.

packages/sdk/src/namespaces/permits.ts
```diff
@@ -184,4 +194,48 @@ export class Permits {
...
+   * signature the relayer accepts independently of this SDK; one exfiltrated alongside
+   * the key remains usable until its own `permitTTL` expiry regardless of this call.
+   *
+   * @param scopeId - Must match the configured `transportKeyPairScope`, as a guard
+   *   against revoking the wrong scope by mistake.
+   * @throws if no signer is configured. {@link SignerNotConfiguredError}
+   * @throws if no scope is configured, or `scopeId` doesn't match it. {@link ConfigurationError}
+   * @throws if the underlying storage delete fails.
+   */
+  async revokeTransportKeyPair(scopeId: string): Promise<void> {
```
ankurdotb, line 223: [AUTOMATED] **Should-fix:** this method (and `warmScope()` below) is documented as "operator-level action, not an end-user one — no wallet account is required or touched," but it still goes through `#requireCredentialService`, and `CredentialService` only exists when a `signer` is configured on the `ZamaSDK` instance at all (see `zama-sdk.ts`). So a WaaS backend that wants to purely administer a shared tenant key — no end-user in the picture — still can't construct an SDK instance to call this without configuring *some* signer, which contradicts the "no wallet needed" framing.

This isn't just a docs nit — the coupling is real. Worth deciding whether `CredentialService`'s existence should stop depending on `signer` being configured when only scope-admin operations are needed, or whether the docs should instead say "no wallet account needs to be *connected*, but a signer must still be configured at construction.

packages/sdk/src/credentials/keypair-vault.ts
```diff
@@ -87,24 +134,94 @@ export class TransportKeyPairVault {
...
+   * instead of persisting it once the round trip completes — otherwise that write would
+   * land after this delete and silently resurrect the key this call just removed.
+   */
+  async clearScope(): Promise<void> {
+    if (this.#scope === undefined) {
+      return;
+    }
+    const key = transportKeyPairScopeStorageKey(this.#scope);
+    this.#epoch.set(key, (this.#epoch.get(key) ?? 0) + 1);
+    await this.#storage.delete(key);
```
ankurdotb, line 214: [AUTOMATED] **FYI, non-blocking:** the epoch bump happens before this `delete`, which correctly defeats the in-flight-generation-resurrection race for stores that preserve per-key write ordering (verified by the test further down in this file). Worth a one-line comment noting this relies on the backing `GenericStorage` implementation applying same-key writes in dispatch order — for an async DB/Redis adapter without that guarantee, a `set` dispatched by an in-flight generation just before this delete could in theory land after it. Not necessarily a bug (`MemoryStorage` and most conventional stores are fine here), just worth being explicit about the assumption given how central this method is to the operator's compromise-response story.

packages/sdk/src/config/__tests__/schema.test.ts
```diff
@@ -67,6 +67,25 @@ describe("createConfig validation", () => {
     ).toThrow("transportKeyPairTTL must be a positive integer number of seconds");
   });
 
+  test("rejects an empty transportKeyPairScope at createConfig", ({ relayer, provider }) => {
```
ankurdotb, line 70: [AUTOMATED] **Nit:** this covers the invalid (empty-string) case, but there's no test asserting the *positive* default case at this layer — that omitting `transportKeyPairScope` entirely resolves to `undefined` on the built `ZamaConfig` (the existing "validates and applies numeric defaults" test above never touches this field either). Checked manually and the code is already correct (`buildZamaConfig` in `config/build.ts` only parses when `!== undefined`), so this is a coverage gap, not a bug — but worth adding given `transportKeyPairScope`'s invariant #1 is "default behavior does not change."

docs/gitbook/src/concepts/permit-model.md
```diff
@@ -88,6 +88,12 @@ Permits can be removed in two ways:
...
+
+Invalidating the _shared_ key pair is a separate, operator-level operation: `sdk.permits.revokeTransportKeyPair(scopeId)`. It deletes the scope's key pair; every permit in the scope embeds that key pair's public key, so `hasPermit`/`grantPermit` treat them all as stale on next access — no permit needs to be touched directly, and no wallet needs to be connected. Unlike other credential-store writes, this call is not best-effort: a storage failure rejects rather than being logged and swallowed, since a resolved promise here is expected to mean the key pair is actually gone — the primitive an operator reaches for on suspected compromise. This only stops the SDK from reissuing or reusing the key going forward; it does not revoke any permit already issued under it, which stays independently valid until its own `permitTTL` expiry — see [Security Model](./security-model.md#shared-tenant-scope-b2b2c-waas-operators) for the full explanation and when a shared scope makes sense.
```
ankurdotb: [AUTOMATED] **Nit:** "treat them all as stale on next access" is directionally right but compresses two different mechanisms depending on timing: immediately after a revoke, `hasPermit()` returns `false` because `readStored` finds no key pair at all (short-circuits before ever reaching the public-key comparison); only once some later caller regenerates a key for the scope does the `keypairPublicKey`-mismatch check in `pruneUnusable` actually filter the old permit. Same end-user-visible answer either way, so not user-facing-wrong — just worth being precise about if this doc is ever used to explain *how* the staleness check works rather than just *that* it happens.

ankurdotb (changes requested):
[AUTOMATED]
## Review summary

Reviewed with a full multi-agent sweep (two independent bug hunts, a guideline-compliance check, an error-handling audit, and a protocol-compatibility boundary scan), plus a holistic architecture-fit pass (forward-compat check against SDK-135/136/137, a deletion-test read on the `scope` abstraction's seam depth, and a namespace-placement judgment) and a comparative look at how Turnkey/Privy/Dynamic/Fireblocks/dfns/Coinbase CDP handle analogous "one key shared across many managed accounts" scenarios. Also re-ran the actual test suite locally (80/80 passing across the `keypair-vault`, `credential-service`, `permits`, and `schema` test files) rather than relying on the PR description's self-reported numbers, and independently verified the unscoped default path resolves `transportKeyPairScope` to `undefined` as expected.

**Overall:** a well-built, heavily-tested feature that matches the SDK-142 ticket and its Slack-discussion history closely, and fits the existing `credentials/` facade — it correctly respects the ticket's "SDK-core, not a React hook" constraint (verified the existing `warmTransportKeyPair()` hook call in `provider.tsx` genuinely adapts transparently, with a test proving it) and doesn't leak any raw storage helpers onto the public surface. Two things below should be addressed before merge; the rest are non-blocking.

### Should-fix (see inline comments)
- **Blocking** — `keypair-vault.ts`: the `strict` flag on `warmScope()`/`getOrCreate()` can get silently lost through the `#pending` promise-dedup cache, undermining the exact "a resolved promise means verifiably warmed/revoked" guarantee these operator primitives exist to provide.
- **Should-fix** — `permits.ts`: `revokeTransportKeyPair`/`warmScope` are documented as needing no wallet, but still require a `signer` configured at SDK construction — a real contradiction with the "purely operator-level, no wallet needed" framing, not just a docs gap.

### Non-blocking (see inline comments)
- `keypair-vault.ts` `clearScope()`: the epoch-bump-before-delete ordering relies on an assumption about the backing store worth documenting explicitly.
- `schema.test.ts`: missing a positive-case test for `transportKeyPairScope` omitted → `undefined` (verified manually that the code itself is already correct).
- `permit-model.md`: minor doc-precision point on how `hasPermit()` actually decides "stale."

### Also worth a look, not tied to a single line
- The PR description's top-level Summary still says `sdk.permits.rotateScope(scopeId)` — the commit history shows this was renamed to `revokeTransportKeyPair()` after an earlier review round, but the description text at the top was never updated to match.
- On the opaque `transportKeyPairScope: string` design: checked this against Turnkey (sub-organization + platform-issued UUID), Privy/Dynamic (mandatory per-user isolation, no key sharing at all), and Fireblocks/dfns — the closest real analogs, since both genuinely share one key across many managed accounts, but scope it via a platform-issued *structured* ID, not a free-form string. The closer precedent for a developer-supplied opaque string is AWS KMS aliases / HashiCorp Vault namespaces, both well-established for scoping *encryption* capability specifically (as opposed to fund custody, which is what justifies the heavier structured-object approach elsewhere). Since what's shared here is a transport/encryption key pair, not custody, the string is a reasonable, industry-consistent choice — but it's currently identity *and* access grant with no second factor. Worth a doc note recommending a namespaced convention (e.g. `myapp:tenant-123`) to reduce accidental collision risk across unrelated integrators sharing a backend.
- SDK-142's stated urgency ("land before SDK-136/137 to avoid churning `allow()`") doesn't fully hold up on inspection: SDK-136 already shipped independently two months ago (PR #372) without ever needing this, and this PR never touches `grantPermit()`'s signature at all. The forward-compat claim that does hold: SDK-135's draft spec keys its HKDF salt on raw `walletAddress`, which would silently break under a shared scope — this PR correctly establishes that the salt needs to be scope-aware instead. Worth keeping in mind when SDK-135 is scoped, not an issue with this PR.
- Checked `scope`'s structural placement (config → `CredentialService` → `TransportKeyPairVault`) against a deletion-test lens: it's a real seam that owns actual behavior at each layer (epoch-based rotation safety, the strict/non-strict split, the operator error contract) — not shallow field-threading. No extraction into a shared identity-resolver utility needed now; that would be premature generality for a single consumer. Worth revisiting only if/when SDK-135 becomes a second one.

### Confirmed clean
No guideline violations (react-sdk boundary respected, doc conventions followed, storage-key namespacing is collision-free, `docs:check-links` passes), no regressions in the best-effort vs. not-best-effort error-handling tiering, and no protocol/relayer/KMS-boundary changes — this is a purely client-side SDK change.

packages/sdk/etc/sdk.api.md
```diff
@@ -11424,7 +11424,7 @@ export class Permits {
     hasDelegationPermit(delegator: Address, contracts: Address[]): Promise<boolean>;
     hasPermit(contracts: Address[]): Promise<boolean>;
     revokePermits(contracts?: Address[]): Promise<void>;
-    rotateScope(scopeId: string): Promise<void>;
+    revokeTransportKeyPair(scopeId: string): Promise<void>;
     warmScope(): Promise<void>;
```
ghermet: `warmScope` can just be `warmTransportKeyPair(scopeId)`

packages/sdk/src/credentials/keypair-vault.ts
zama-cremaud: Yes, good catch!

ankurdotb: [AUTOMATED] Confirmed fixed in 67661f8d — `#pending` is now keyed on `(identityKey, strict)` instead of `identityKey` alone, so a strict and non-strict call can no longer inherit each other's persistence guarantee. Independently re-verified this does trade the bug for a narrower residual race (a strict + non-strict call that genuinely overlap now run two independent generations racing storage) — but that's only reachable by violating the documented 'warm before opening traffic' sequencing; under correct usage it's moot, since the warm call's pending entry is already cleared and the key already persisted before real traffic would arrive. New test in keypair-vault.test.ts proves the original bug is fixed. Not blocking.

packages/sdk/src/namespaces/permits.ts
ankurdotb: [AUTOMATED] Addressed in 67661f8d via docs clarification — this now correctly states 'no wallet account needs to be connected, but a signer must still be configured on this SDK instance at construction time,' rather than implying no wallet is needed at all. Also folded into the `warmScope()` → `warmTransportKeyPairScope(scopeId)` rename for naming/safety symmetry with `revokeTransportKeyPair`. Independently re-verified the rename is complete and consistent everywhere (docs, api.md, all test call sites), and a full monorepo typecheck passes.

packages/sdk/src/credentials/keypair-vault.ts
ankurdotb: [AUTOMATED] Addressed in 67661f8d — `clearScope()`'s docstring now explicitly documents the same-key write-ordering assumption on the backing `GenericStorage`.

packages/sdk/src/config/__tests__/schema.test.ts
ankurdotb: [AUTOMATED] Addressed in 67661f8d — new test 'resolves transportKeyPairScope to undefined when omitted, preserving the default per-signer behavior' added, and passing.

docs/gitbook/src/concepts/permit-model.md
ankurdotb: [AUTOMATED] Addressed in 67661f8d — permit-model.md now correctly distinguishes the two mechanisms (immediate `false` from no stored key pair yet, vs. later `pruneUnusable` mismatch-based filtering once a caller regenerates).

ankurdotb (approved):
[AUTOMATED]
## Re-review: all prior items addressed

Re-verified commit 67661f8d against the previous review round (replies posted on each original thread with specifics). Summary:

- **Blocking dedup race — fixed.** `#pending` is now keyed on `(identityKey, strict)` instead of `identityKey` alone. Independently re-verified this can't reintroduce the original bug, and traced the one residual edge case it trades in for: a strict + non-strict call that genuinely overlap now run two independent generations racing storage — but that's only reachable by violating the documented "warm before opening traffic" sequencing, and is moot under correct usage. A new test proves the original failure mode is gone.
- **Signer-gating contradiction — addressed via docs clarification** (kept `CredentialService`'s construction gated on `config.signer`, rather than removing the requirement) **and** folded into a `warmScope()` → `warmTransportKeyPairScope(scopeId)` rename for symmetry with `revokeTransportKeyPair`. Verified the rename is complete everywhere (docs, `sdk.api.md`, every test call site) and a full monorepo typecheck passes clean.
- **Storage-ordering caveat, missing default-case test, and the `hasPermit` doc-precision nit** — all three addressed exactly as suggested.

Re-ran the full test suite directly rather than trusting reported numbers: **84/84 passing** (up from 80), plus `pnpm typecheck` across all 6 workspace packages.

No new issues found in the fixup itself. Approving.

PR #547, opened 2026-07-13
feat(credentials): optional derivationSecret wrapping for at-rest private key protection [SDK-135]

zama-cremaud:
## Why

Closes SDK-135.

The SDK persists the ML-KEM transport private key through the configured `GenericStorage`. Security at rest is delegated to that backend. This is correct when a real keystore exists (IndexedDB behind disk encryption, platform keychains, KMS-backed storage). Headless contexts have no such backend: CLI tools, agents on bare metal, local dev. There, an attacker who copies the storage reads the private key directly.

The old SDK derived an encryption key from the permit signature, then stored the signature next to the ciphertext. The lock and the key were in the same drawer. It also broke MPC wallets, which do not sign deterministically.

## What

An optional secret that wraps the private key before every storage write, and unwraps it on read. The secret comes from the deployment environment (secrets manager, mounted file, KMS). The SDK never stores it.

```ts
const config = createConfig({ chains, storage, ... });
const sdk = new ZamaSDK(config, {
  transportKeyPairDerivationSecret: process.env.ZAMA_TRANSPORT_KEY_PAIR_SECRET,
});
```

Note: the secret is a **constructor option**, not a config option. This differs from the ticket spec on purpose. The config stays plain data: you can spread it, copy it, and log it. The secret never touches it.

## How

- Key derivation: `HKDF-SHA256(secret, salt: "signer:<address>" | "scope:<scope>", info: "zama-sdk-keypair-wrapping-v1")`.
- Encryption: AES-256-GCM, fresh random 96-bit IV per wrap. The sibling plaintext fields (public key, timestamps, TKMS version) are bound as AAD, so storage-level tampering fails decryption.
- All keys are non-extractable WebCrypto `CryptoKey`s. The SDK zeroizes its copy of the raw secret after the first use.
- Each entry persists `wrappingVersion: 1`. An entry from an unknown scheme is preserved and reported, never deleted.

### Validation, at construction

- The secret must be 32+ bytes (`string | Uint8Array`).
- A present-but-`undefined` value throws: an unset env var must not silently downgrade to plaintext.
- Browser-like environments throw (best effort: `window`, `document`, or `importScripts` present). The option is for headless use only.

### Read policy

| Stored entry | Secret configured | Result |
| --- | --- | --- |
| plaintext | no | normal read (unchanged baseline) |
| plaintext | yes | regenerate wrapped |
| wrapped, decrypts | yes | normal read |
| wrapped, fails to decrypt | yes (wrong or rotated) | cache miss: delete, regenerate |
| wrapped | no | throw `KeyWrappingError`, keep the entry |
| unknown `wrappingVersion` | any | throw `KeyWrappingError`, keep the entry |
| scoped shape mismatch | any | throw `KeyWrappingError`, keep the entry |

Rule of thumb: self-heal when only this signer pays (a re-prompt). Fail loudly when silence would hide an operational error or destroy a shared scope's key.

### Errors

- New `KeyWrappingError` (`KEY_WRAPPING_FAILED`), exported and matchable.
- It is fatal in batch decrypts: the batch aborts instead of repeating the same systemic failure per handle.
- `hasPermit` / `hasDelegationPermit` never throw it. They return `false`.

### Not wrapped

Permits stay in plaintext. They contain an EIP-712 signature and contract addresses, which are public data.

## Docs

- Security model: threat model, disqualifying secret sources, rotation, algorithm table.
- Configuration guide: headless setup, environment rules.
- Error reference and changelog entries.

## Reviewer notes

- History rewritten since the last push (rebuilt on current `prerelease`, force-pushed). Earlier inline comments will show as outdated.
- The first commit is the squashed pre-rebuild history (implementation by @corentin, with fixes by @ghermet); the rest of the series reworks and extends it.
- The three previous review comments are addressed: 32-byte (256-bit) entropy floor with boundary tests, the two anti-pattern warnings in the security model (same-store sourcing, double-wrapping over a secure store), and a no-leak regression test covering event payloads and logs.
- Potential pre-existing issue, deferred: `warmScope` and `getOrCreate` can race on the same scope (single-flight is keyed per caller mode). Already present on `prerelease`.

## Out of scope

- Seeded keypair derivation (storage-free mode). Needs relayer WASM support.
- Permit wrapping.
- Storage-strategy guide (SDK-143).


packages/sdk/src/credentials/schemas.ts
```diff
@@ -11,6 +11,22 @@ import { MAX_CONTRACTS_PER_PERMIT, SECONDS_PER_DAY } from "./utils";
 const transportKeyPairTTLError = "transportKeyPairTTL must be a positive integer number of seconds";
 const permitTTLError = "permitTTL must be a positive integer number of days";
 const transportKeyPairScopeError = "transportKeyPairScope must be a non-empty string";
+/** 128 bits — the minimum to reject trivially-weak input. A length check can't verify actual
+ * entropy, so this is a floor, not a guarantee: callers must still source derivationSecret from
+ * a CSPRNG or secrets manager, never a human-memorable passphrase (HKDF does no key-stretching). */
+export const MIN_DERIVATION_SECRET_LENGTH_BYTES = 16;
```
ankurdotb: [AUTOMATED] **Should-fix:** raise `MIN_DERIVATION_SECRET_LENGTH_BYTES` from 16 to 32 (256 bits), matching AES-256's actual key size rather than the weaker "128-bit security level" argument for the IKM floor. HKDF's output is bounded by both the hash function's output size and the input entropy — using a 256-bit-entropy floor removes any argument that the wrapping key is weaker than the cipher it's feeding.

This ripples through:
- The `derivationSecretError` message on the line below (currently says "128 bits").
- `docs/gitbook/src/concepts/security-model.md`'s "the SDK rejects anything under 128 bits" line.
- The existing tests in `config/__tests__/schema.test.ts` (`"rejects a derivationSecret below the 128-bit entropy floor"` / `"accepts a derivationSecret at or above the 128-bit entropy floor"`) need updating to the new floor and values.
- While updating those tests, also add the exact boundary case (31 vs. 32 bytes) that's currently untested at either floor.

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,16 +59,44 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
 
+### Wrapped at rest (`derivationSecret`)
+
+The gap in the default model above is **headless contexts** — CLI tools, agentic operators on bare metal, local dev environments — where no platform keystore exists and no operator KMS sits in front of storage. There's nothing to delegate security to.
+
+`derivationSecret` (`string | Uint8Array`) closes that gap by wrapping the private key half before every write and unwrapping it after every read, transparently, inside the credential store. The SDK never manages or stores this value — it comes from your environment (an env var from a secrets manager, a mounted file, a KMS-unwrapped blob) and exists only for the lifetime of your process. It must carry real entropy: the SDK rejects anything under 128 bits, but a length check can't verify actual randomness, so source it from a CSPRNG or secrets manager, never a human-memorable passphrase — HKDF does no key-stretching, so the secret's raw entropy is the only thing standing between a storage-reading attacker and the private key.
```
ankurdotb: [AUTOMATED] **Should-fix:** this explains how to use `derivationSecret` correctly, but doesn't warn against the two anti-patterns that would defeat it entirely — both named in the April Slack thread and SDK-89/SDK-143 as the exact failure mode this feature exists to prevent:

1. Sourcing `derivationSecret` from something that ends up in the same (untrusted) store it's meant to protect — recreating "lock and key in the same drawer" one level up.
2. Configuring `derivationSecret` on top of storage that's *already* secure (platform keychain, KMS-backed storage) — double-wrapping adds complexity without adding security.

SDK-143 (a separate, not-yet-started ticket) will eventually cover this as part of a broader storage-strategy guide, but a reader of only this PR's docs isn't told not to do either of these right now, and this is the one place in the docs that specifically introduces the mechanism this warning applies to. Suggest a short callout here rather than waiting for SDK-143.

packages/sdk/src/config/types.ts
```diff
@@ -51,6 +51,22 @@ export interface ZamaConfigBase<TChains extends AtLeastOneChain = AtLeastOneChai
...
+   * manager, not a passphrase — this scheme does no key-stretching). Omit for the
+   * default: plaintext, security delegated to the storage backend. Composes with
+   * `transportKeyPairScope`: when both are set, the derivation binds to the scope, so
+   * every signer in the scope produces the same wrapping key — but every instance
+   * sharing that scope must then use the *same* secret, since a mismatch throws
+   * instead of silently regenerating (regenerating would clobber the scope's shared
+   * entry). Without a scope, changing this value is self-healing: the next
+   * `grantPermit()` regenerates the per-signer key pair transparently, not a crash.
+   */
+  derivationSecret?: string | Uint8Array;
```
ankurdotb: [AUTOMATED] **Nit:** there's no regression test asserting `derivationSecret` (or the full resolved config) can never be reached from an event payload, log line, or any other externally-observable surface. This holds today structurally — `ZamaSDKEvent` payloads (`events/sdk-events.ts`) are hand-built field-by-field, never a config spread, and `derivationSecret` is consumed only as a private field inside `TransportKeyPairVault` — but a future refactor that starts passing `config` around more broadly could reintroduce a leak silently. Worth a cheap explicit test locking this in rather than relying on it staying true by construction.

ankurdotb (changes requested):
[AUTOMATED]
## Review summary

Reviewed against SDK-135's own ticket text and the April 2026 Slack thread that originated both this feature and #546 (transportKeyPairScope) — they're the same conversation, forked into two tickets. The design principle behind `derivationSecret` (secret sourced entirely out-of-band from the SDK's own storage, never persisted or logged by the SDK) is a direct, deliberate fix for a named historical anti-pattern: the old scheme derived its wrapping key from the permit signature, which sat in the *same* storage as the ciphertext it protected — "lock and key in the same drawer."

Traced the full chain (config → `CredentialService` → `TransportKeyPairVault` → `keypair-wrapping.ts`) and confirmed by grep across every `storage.set()`/`logger.*()` call site, independently re-verified by an adversarial crypto review: **`derivationSecret` is never written to storage, never logged, never serialized** — held only as an in-memory field for the process's lifetime. The HKDF salt (`identity` = scope-or-signer-address) is public, which is correct — salts don't need confidentiality. This also correctly composes with #546's scope work exactly as predicted in that earlier review (identity is `scope ?? signerAddress`, deliberately not the storage-key string).

Also independently verified (not just from the PR's own account of its internal review): IV handling (fresh random IV every wrap, never reused), AAD determinism (wrap-side and unwrap-side metadata encoding match byte-for-byte), non-extractable key derivation, and the scoped-vs-unscoped error branching (the destructive clobbering bug the PR says it already caught and fixed in an earlier round) — walked every branch myself and found it correctly fixed. Guideline compliance and error-handling wiring are clean. Re-ran the test suite directly: **192/192 passing**.

Three items below should be addressed before merge.

### Should-fix (see inline comments)
- **`schemas.ts`**: the 128-bit entropy floor should be raised to 256 bits, matching AES-256's actual key size rather than a weaker security-level argument.
- **`security-model.md`**: add an explicit anti-pattern warning against sourcing `derivationSecret` from the same storage it protects, and against double-wrapping storage that's already secure. This is exactly the failure mode this feature exists to prevent — a reader of only this PR's docs currently isn't told not to do it. (SDK-143 will cover this more broadly later, but this PR is where the warning is first needed.)

### Nit (see inline comment)
- No regression test confirms `derivationSecret`/the full config object can never reach an event payload or log. True today structurally (`ZamaSDKEvent` payloads are hand-built, never a config spread), but worth locking in with a test rather than leaving it implicit.

### Also verified, no action needed
A sub-threshold edge case was checked and deliberately not raised as a finding: during a mixed-mode scope rollout (some instances configured with `derivationSecret`, some not, mid-transition), a plaintext-mode instance would silently discard a wrapped peer's shared-scope entry — the mirror image of the bug already fixed on the wrapped side. Not fixable without leaving one side permanently non-functional (a plaintext instance genuinely cannot use a wrapped entry), and the docs already require an atomic rollout across a scope. Worth being aware of, not a defect.

ghermet:
@ankurdotb Do we need this if  we assume most custodians have their own secure storage?

ankurdotb:
> @ankurdotb Do we need this if we assume most custodians have their own secure storage?

We should create docs and help target existing secure storage if an app, wallet, or custodian already has one. We shouldn’t “double wrap” / double encrypt, if they can already target a secure store. The examples we spoke of were like how browser wallets, mobile wallets, custodian backends etc already have secure stores.

But it’s true that until natively integrated, those secure stores may not be targettable (as this needs a deeper level of integration).

If you mean PR 547, it isn’t meant for custodians (this would be a double wrap). I think it’s useful in headless usage, or in cases where we can’t yet target secure stores.

docs/gitbook/src/changelog/alpha.md
```diff
@@ -25,3 +25,24 @@ const prepared = await sdk.offline.prepare({
 ```
 
 All eleven write operations are covered, including the shield legs and the two-phase unshield. See [Offline signing](../guides/offline.md) for the full guide: the handoff contract, request kinds, multi-transaction batches, and approval-delay behavior.
+
+## Transport key pair wrapped at rest
+
+A new `transportKeyPairDerivationSecret` constructor option on `ZamaSDK` encrypts the transport key pair's private half before it is written to storage, and decrypts it on read. It is for headless environments with no secure storage to delegate to: CLI tools, bare-metal agents, local dev. The constructor rejects it, on a best-effort basis, whenever `window` or `document` is defined or `importScripts` is a function (a browser main thread, Web Worker, or Service Worker), and rejects it when passed as `undefined` (an unset env var reaching the option), both with `ConfigurationError` instead of a silent plaintext downgrade. Omit it entirely and nothing changes: storage stays plaintext, as before.
```
ankurdotb: I think this is good, but it's a bit convoluted to understand when to use it, and what to do if it's NOT that kind of client app.

Like for example, a table might be a good way to represent this? With client type/environment as a column that might have "Headless (e.g., CLI tools, bare-metal agents...)" as an entry and what do in that scenario, and also advice on other scenarios like what do in web app, in a wallet, in a backend etc. (Fine if this is elsewhere in the doc, rather than in changelog.)

EDIT: Such a table is probably best off in the actual config guide, as that's the persistent place where they can read this.

docs/gitbook/src/changelog/alpha.md
```diff
@@ -25,3 +25,24 @@ const prepared = await sdk.offline.prepare({
...
+  walletClient,
+  relayers: { [sepolia.id]: node() },
+});
+
+const sdk = new ZamaSDK(config, {
+  transportKeyPairDerivationSecret: derivationSecret, // 32+ bytes of real entropy
+});
+```
+
+Wrapping failures surface as a new `KeyWrappingError` (`KEY_WRAPPING_FAILED`). `hasPermit` and `hasDelegationPermit` never throw it: they return `false`, so a permit check stays a safe read.
```
ankurdotb: I understand what it's saying, but since "wrapping" is also used for confidential tokens, can this be clarified?

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
+
+Disqualifying sources, all of which put the secret back inside the same compromise:
+
+- A bundled build-time env var. Bundlers inline `process.env` values at build time, so the secret ships in the artifact itself, not just at runtime, and reaches every consumer of that artifact including whoever steals the storage.
+- A secret written to the same store the wrapped key pair lives in. If the attacker's storage copy also holds the secret, wrapping adds a decode step, not a defense.
+- A human-memorable passphrase. HKDF performs no key-stretching, so the secret's own entropy is the only thing standing between a storage-reading attacker and the private key. A passphrase an operator can memorize does not carry that entropy.
+- The same secret shared identically across an entire fleet, with no per-instance or per-tenant separation. One captured instance then wraps every other instance's storage too.
```
ankurdotb: Maybe it's better to move this a bit further down after we explain the actual approach, and create a heading for what not to do, because it's weaker?

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
...and rejects it, on a best-effort basis, whenever `window` or `document` is defined (a main-thread browser) or `importScripts` is a function (a Web Worker or Service Worker), both `ConfigurationError`, and when the option key is present but its value is `undefined` (the unset-env-var footgun, also `ConfigurationError`). Keeping the secret out of a shipped bundle in the first place is the consumer's responsibility: it must come from the deployment environment, never a shipped bundle, and this guard only catches what it can detect. On first use the SDK imports it into a single non-extractable WebCrypto key and drops its own copy of the raw bytes for a `Uint8Array` input, which shortens the raw secret's in-memory lifetime. A `string` input can't be scrubbed this way: JS strings are immutable and interned, so a passed-in string may already have copies elsewhere in the heap that dropping this reference does not reach. `Uint8Array` is the strong form; prefer it when you control how the secret is decoded from its source. Either way, this does not stop code already running in your process from using the resulting key, and WebCrypto makes no promise about where the key material itself lives.
```
ankurdotb: - Is it relevant that it's not a `createConfig` option? 
- This paragraph is a bit verbose. I had difficulty following it

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
+
+The secret must carry real entropy. The SDK enforces a minimum length (32 bytes for a `Uint8Array`, 32 characters for a string), but length is not entropy: an encoded string (hex, base64) carries fewer bits than it has characters, so a 32-character hex secret is only 128 bits. Supply either 32 random bytes as a `Uint8Array`, or a string holding at least 256 bits of actual entropy (`openssl rand -hex 32`). Source it from a CSPRNG or secrets manager.
```
ankurdotb: - I think we should call this out as a warning/highlight on how to choose/provide this value (each custodian/wallet will have it's own method I suppose of how they do this).
- Is `openssl` the best example? Since it's a node package, better to suggest good crypto libraries / node-native ways?

I'm conscious of striking a balance between being too prescriptive here, vs providing more concrete guidance that is relevant for a node SDK.

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
+```
+
+The salt is the same identity used for storage keying (see [Shared-tenant scope](#shared-tenant-scope-b2b2c-waas-operators) below), the configured `transportKeyPairScope` if set, else the signer address, so a scope's signers derive one shared wrapping key, not one each. It is prefixed by kind (`scope:` / `signer:`) so a scope named after an address never collides with that signer's own key. Only the private key half is wrapped; the public key, timestamps, and TKMS version stay in plaintext alongside the ciphertext, but bound to it as AES-GCM additional authenticated data (`tkmsVersion` encoded as `null` when absent), so a storage-level attacker can't tamper with them (e.g. mismatching the public key, or extending `expiresAt`) without also failing decryption. Every permit stays in plaintext, unauthenticated by this scheme: permits are already public data (an EIP-712 signature and a list of contract addresses). Each wrapped entry also records the scheme version (`wrappingVersion: 1`), so an entry written by a future scheme is recognizable as such rather than as corruption.
```
ankurdotb: Also a bit long/convoluted

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
+{% endhint %}
+
+{% hint style="danger" %}
+**Two ways to defeat this, don't.** `transportKeyPairDerivationSecret` only helps if it's genuinely out-of-band from the storage it protects.
+
+- **Don't source `transportKeyPairDerivationSecret` from the same store it wraps.** Reading the secret from the same (untrusted) storage that holds the wrapped key pair puts the lock and key in the same drawer: an attacker who can read one can read both, and the wrapping buys you nothing. The secret has to come from somewhere that storage-level attacker can't reach: a secrets manager, a KMS-unwrapped blob, an injected env var.
+- **Don't wrap storage that's already secure.** If you're already on a platform keychain, a KMS-backed store, or any at-rest-encrypted backend, `transportKeyPairDerivationSecret` is redundant: you'd be double-wrapping, adding a key-management burden (rotation, distribution, loss) without adding security. It exists for the [headless-context gap](#wrapped-at-rest-transportkeypairderivationsecret) above, not as a default to layer on everything.
+
+{% endhint %}
```
ankurdotb, line 115: Repetitive with the "what not to do" list above? Maybe that list can be moved here?

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
+
+**Rotation is just changing the value, for a per-signer key pair.** `transportKeyPairDerivationSecret` is set once at `ZamaSDK` construction, like every other credentials option: there's no live "update config" call. Changing it (or losing it) doesn't corrupt anything: the next read fails to authenticate against the stored ciphertext, is treated exactly like a cache miss (deleted, not surfaced as an error), and the next `grantPermit()` regenerates a fresh key pair and re-persists it wrapped with the new secret. One extra wallet re-prompt, never a crash.
```
ankurdotb: It's not "just" another wallet re-prompt, they will have to re-sign everything they had permits signed for...right? Maybe we should be directing them to the transportKeypairTTL as what should be guiding this rotation?

docs/gitbook/src/guides/configuration.md
```diff
@@ -393,6 +393,42 @@ const config = createConfig({
...
+
+{% hint style="danger" %}
+**Headless only, enforced on a best-effort basis.** The `ZamaSDK` constructor rejects `transportKeyPairDerivationSecret` with `ConfigurationError` whenever `window` or `document` is defined (a main-thread browser context) or `importScripts` is a function (a Web Worker or Service Worker). Bundlers inline values at build time, so a secret shipped this way reaches every visitor's bundle and protects nothing; keeping the secret out of a shipped bundle in the first place is the consumer's responsibility, this guard only catches what it can detect. Environments with a platform keystore, or that ship a bundle, must delegate at-rest security of the transport key pair to the storage backend instead (the default, no option needed).
+{% endhint %}
+
+{% hint style="warning" %}
+Wrapping uses WebCrypto `crypto.subtle`, which is unavailable in non-secure contexts (plain `http://` on a LAN IP). Where it is missing, wrapping fails with [`KeyWrappingError`](../reference/sdk/errors.md#keywrappingerror).
```
ankurdotb: Is this "non-secure" context a scenario that someone might run into for the recommended setups where they might use this?

Was prompted of thinking of this because of scenarios like localhost development, where it might be served on `http://`. Or is that not relevant here?

docs/gitbook/src/reference/sdk/errors.md
```diff
@@ -87,6 +88,7 @@ The `_` wildcard catches any `ZamaError` not explicitly handled. Each handler re
 | `InvalidTransportKeyPairError`          | `INVALID_KEYPAIR`                     | Relayer rejected transport key pair (stale or malformed)                                                                          |
 | `TransportKeyPairExpiredError`          | `KEYPAIR_EXPIRED`                     | Transport key pair expired — user must re-sign                                                                                    |
 | `NoCiphertextError`                     | `NO_CIPHERTEXT`                       | No encrypted balance for this account                                                                                             |
+| `KeyWrappingError`                      | `KEY_WRAPPING_FAILED`                 | `transportKeyPairDerivationSecret` wrapping/unwrapping of the transport private key failed                                        |
```
ankurdotb: Is there an alternative to the term wrapping/unwrapping? Of course it's used in the keys/storage encryption context in technical contexts, but perhaps confusing because of how much we use wrap/unwrap elsewhere?

(I can just foresee this being something that trips up an engineer at a partner in a rush)

docs/gitbook/src/guides/configuration.md
ankurdotb, line 1: I like the clarity/directness on this MUCH better

docs/gitbook/src/reference/sdk/errors.md
```diff
@@ -275,6 +277,25 @@ try {
...
+
+```ts
+matchZamaError(error, {
+  KEY_WRAPPING_FAILED: () =>
+    showError("Credential storage misconfigured, check transportKeyPairDerivationSecret"),
+});
+```
+
+**How to handle:** Without a `transportKeyPairScope`, a mismatched secret never throws: it's treated as a cache miss and regenerates silently. A `KeyWrappingError` without a scope means a wrapped entry was read with no `transportKeyPairDerivationSecret` configured at all (restore the secret, or call `sdk.permits.clear()` to downgrade to plaintext deliberately), the entry was written by a wrapping scheme version this build doesn't recognize, or `crypto.subtle` is unavailable in the environment. With a `transportKeyPairScope` configured, it doesn't self-heal (silently regenerating would clobber the entry every other signer in the scope reads); verify every instance sharing the scope is configured with the same `transportKeyPairDerivationSecret`, including that none is missing it.
```
ankurdotb: A bit verbose/repetitive with other areas/documents?

ankurdotb (changes requested):
I can see a huge amount of work done here to clean this up, and deliver an MVP, so thank you for the work! I think you understood well what this feature was trying to do, and make the simplest version of it. Thank you for that.

In general, I think it's better to leave the primary guidance in the long-lasting/persistent sections, since the same info is repeated in multiple places.

- Configuration and error guide pages is probably the primary section the guidance should be
- Security model pages for any explainer and gotchas, config pages can link to this
- Changelog can be most brief and link off to existing pages in config/security model pages, rather than repeating everything again

docs/gitbook/src/concepts/security-model.md
ankurdotb: [AUTOMATED] Confirmed — traced this precisely. A regenerated keypair gets a brand-new public key, and permits are pruned by matching `keypairPublicKey` against whatever key is currently active, so every permit this signer (or scope) previously held goes stale the moment the secret changes, not just the keypair itself. Whether that's actually 'one' re-prompt or several depends on how the integrator calls `grantPermit()`: if they re-request the full set of previously-covered contracts in one call, it can widen into a single signature (same batching/chunking behavior as an ordinary multi-contract grant); if the app grants lazily per-contract as each is next accessed, the cost shows up as repeated re-prompts spread over time as the user touches each previously-permitted resource again. Worth stating both the mechanism (all permits go stale, not just the keypair) and that the actual prompt count depends on integrator-side granting pattern, not the SDK.

Confidence: 90/100.

docs/gitbook/src/guides/configuration.md
ankurdotb: [AUTOMATED] Your suspicion is correct — this caveat doesn't actually apply to this feature's real execution environment. 'Secure context' (HTTPS/localhost) gating `crypto.subtle` is a browser-only policy, enforced by the browser's Window/Worker implementation of the Web Crypto API spec; it has no equivalent in Node.js, Deno, Bun, or edge/Workers runtimes — there's no origin/protocol concept there for it to gate on. Since `assertDerivationSecretHeadless` already rejects any environment where `window`/`document`/`importScripts` are defined, the only environments that can ever reach this code path are exactly the ones where 'plain http:// on a LAN IP' isn't a meaningful constraint at all. The 'non-secure context' language reads like a generic WebCrypto caveat carried over without checking it against the headless-only restriction this same PR enforces. Localhost-over-http dev is a real scenario, but it's a browser scenario, and browser usage of this option is already blocked outright, for an unrelated reason.

Confidence: 75/100 — high confidence on the browser-vs-Node distinction itself; leaving some room since I haven't exhaustively checked every conceivable edge/serverless runtime's crypto.subtle implementation.

docs/gitbook/src/guides/configuration.md
```diff
@@ -393,6 +393,42 @@ const config = createConfig({
...
 
+### 10. (Optional) Wrap the transport key pair at rest (headless environments)
+
+By default, security for the persisted private key is delegated to your storage backend. In a headless context with no secure storage to delegate to (a CLI tool, an agent on bare metal, local dev), pass `transportKeyPairDerivationSecret` as a `ZamaSDK` constructor option instead. The value must arrive out of band: sourced from your process environment, a secrets manager, or a KMS-unwrapped blob, never persisted alongside the storage it protects.
+
+```ts
+import { createConfig } from "@zama-fhe/sdk/viem";
+import { ZamaSDK } from "@zama-fhe/sdk";
+import { node } from "@zama-fhe/sdk/node";
+import { sepolia } from "@zama-fhe/sdk/chains";
```
ankurdotb, line 416: [AUTOMATED] **Should-fix, confidence 85/100:** this example omits `storage`, and `getDefaultStorage()` resolves to `MemoryStorage` on Node — exactly the runtime this option is restricted to (`assertDerivationSecretHeadless` only allows it when `window` is undefined, which is also precisely when the default storage is in-memory and non-durable). Following this verbatim wraps a value that's never written to disk: no actual at-rest protection, and a silent re-prompt on every process restart that reads as a bug rather than the real signal (no persistent storage configured).

The `transportKeyPairScope` section one screen earlier already handles this correctly (`storage: myPersistentStorage, // must be shared across every signer in this scope`) — this example needs the equivalent line, e.g. `storage: myPersistentStorage, // required: the headless default is in-memory, so a wrapped entry would never reach rest`.

docs/gitbook/src/concepts/security-model.md
```diff
@@ -59,18 +59,63 @@ The wallet signs EIP-712 typed data to authorize FHE operations. The SDK trusts
...
 ### Transport key pair storage
 
-The transport private key is stored in plaintext in the configured storage backend (typically IndexedDB in browsers). There is no encryption-at-rest layer.
+By default, the transport private key is stored in plaintext in the configured storage backend (typically IndexedDB in browsers), with no encryption-at-rest layer. This is correct for most contexts: browser dApps have IndexedDB behind OS-level disk encryption, mobile apps have platform keychains, server operators typically have a KMS/Vault/HSM in front of their storage. See [Wrapped at rest](#wrapped-at-rest-transportkeypairderivationsecret) below for the opt-in alternative when none of that exists.
 
 ![Transport key pair storage](../images/credential-storage.svg)
 
 | Parameter  | Value                                                                    |
 | ---------- | ------------------------------------------------------------------------ |
 | Storage    | IndexedDB (browser), memory (tests), AsyncLocalStorage (Node.js)         |
```
ankurdotb: [AUTOMATED] **Should-fix, confidence 95/100:** this table still says Node's default storage is `AsyncLocalStorage` — confirmed against `config/resolve.ts`: `getDefaultStorage()` returns `MemoryStorage` on Node, full stop. `asyncLocalStorage` is a separate, opt-in export from `@zama-fhe/sdk/node`, not the fallback.

This is a distinct occurrence of the same mistake already caught and fixed on PR #613 (commit `9f2df9f`) — that fix corrected the shared-scope warning callout further down this same file (now reads "the storage recommended for Node.js servers"), but never touched this table row. Worth fixing here too, in the same pass, especially now that this section documents at-rest behavior in more detail.

docs/gitbook/src/concepts/security-model.md
ankurdotb: [AUTOMATED] **Should-fix, confidence 80/100:** the string-branch entropy floor (`schemas.ts`, `z.minLength(32)`) checks character count, not entropy — and this paragraph's own next sentence proves the gap: "a 32-character hex secret is only 128 bits." The doc's recommended command (`openssl rand -hex 32`) is fine because it happens to produce a 64-character string (32 random bytes, hex-encoded), clearing 256 bits — but that's a property of using the right byte count with that specific command, not something the schema actually verifies. A user who ran `openssl rand -hex 16`, or hand-typed a 32-character hex string, would pass validation with only 128 bits — half the claimed floor, silently. The `Uint8Array` branch doesn't have this issue (`byteLength >= 32` is genuinely 256 bits regardless of source).

Suggest either raising the string floor to cover common encodings (64 chars covers hex's worst case) or otherwise making this paragraph's warning enforced in the schema rather than only stated in prose.

ankurdotb (commented):
[AUTOMATED]
## Re-review: Simon's rework, follow-up on Ankur's review

A separate automated pass, non-duplicative with the docs-clarity review above (that one's about structure/terminology; this one's 3 technical/accuracy findings, detail in the inline comments below) and with the prior automated round (its 3 items are already resolved).

**Should-fix:**
1. The documented headless example never persists to disk — 85/100 (`configuration.md:404`)
2. A second, still-unfixed instance of the `AsyncLocalStorage`-is-the-Node-default claim, distinct from the one #613 already corrected — 95/100 (`security-model.md:68`)
3. The string-secret entropy floor checks character count, not entropy — 80/100 (`security-model.md:87`)

**Re-verified clean on this pass**: the strict/dedup fix, `#recover`'s scoped-vs-unscoped routing (all 10 combinations), AAD symmetry, secret zeroization, `evict()`/`tkmsVersion` (confirmed pre-existing on `prerelease`, not scope creep), and all four bug-fix claims from the commit messages (checked in code and tests, not just trusted). 302/302 tests pass, every CI gate green.

Also replied directly on the "not just one re-prompt" and "non-secure context" threads above — confirming both, with the underlying mechanism.

docs/gitbook/src/changelog/alpha.md
zmalatrax: totally, I'd rather keep the actual word (as it's literally "key wrapping"), will clarify to avoid misunderstanding with wrapped tokens

docs/gitbook/src/concepts/security-model.md
zmalatrax: indeed, we should hint at the Node crypto api first

docs/gitbook/src/guides/configuration.md
zmalatrax: removed the secure/non-secure context warning as it's not relevant in the headless context WebCrypto is actually used

docs/gitbook/src/reference/sdk/errors.md
zmalatrax: Yeah I agree.. kinda hard to coin a new term when this one is what it actually is, we could use "seal" or something similar but I'm not very fond of it

I think that context matters, as it's more about "key wrapping" than just "wrapping" and it's tied to the transport key pair. I'd say it's fairly contained and should be a an issue for partners' engineers to notice the difference, but I might be too optimistic 🤔

docs/gitbook/src/concepts/security-model.md
zmalatrax: I don't think that having a lower entropy (of 128 bits) is much of an issue, but it's free and always better to enforce the same entropy on both Uint8Array and string, so might as well do it

ankurdotb (dismissed):

ankurdotb (approved):

PR #549, opened 2026-07-13
feat(credentials): recover from KMS context rotation on decrypt [SDK-137]

zama-cremaud:
## Summary

When the KMS rotates its signing context, a permit signed under the previous context is rejected on the next `userDecrypt`/`delegatedDecryptValues` call. Today this surfaces as an opaque decrypt failure. This PR adds automatic recovery:

1. Detect the stale-context failure.
2. Invalidate just the one permit covering the affected contract (`CredentialService.invalidatePermit`, internal-only — not public API).
3. Re-plan via `grantPermit` (one fresh wallet prompt).
4. Retry once.
5. A second consecutive failure throws the new `StaleKmsContextError` (`STALE_KMS_CONTEXT`).

Applies identically to `decryptValues` and `delegatedDecryptValues` — both go through the same per-contract recovery path in `DecryptionService.#decrypt`. Other permits in the same scope are never touched; recovery is per-permission, not per-scope.

## Key finding during analysis

The ticket assumed the *relayer* surfaces a distinguishable "stale KMS context" error. That's no longer accurate after `@zama-fhe/relayer-sdk` → `@fhevm/sdk` (#458, merged into `prerelease`): `@fhevm/sdk` already detects this **client-side**, before any relayer round-trip — it reads the current on-chain KMS signers context and compares it against the permit's signed `extraData`. But it throws a **plain, unexported `Error`** with no dedicated class or code (not part of `@fhevm/sdk`'s public `exports` map). So detection here necessarily keys on the literal message text, pinned to `@fhevm/sdk@1.1.0-alpha.8`, with a drift-guard test (`error.test.ts`) that reads the installed dependency's source and fails loudly if the message is ever reworded upstream.

## Changes

- `errors/base.ts` / `errors/decrypt.ts` / `errors/match.ts` / `errors/index.ts` / `src/index.ts`: new `StaleKmsContextError` (`STALE_KMS_CONTEXT`), registered in `matchZamaError`.
- `utils/error.ts`: `isStaleKmsContextError` detection heuristic + drift-guard test.
- `credentials/credential-service.ts`: `invalidatePermit(contractAddress, delegator?)` — internal-only, deletes just the permit covering one contract in the (possibly delegated) scope.
- `services/decryption-service.ts`: `DecryptionStrategy.invalidatePermit`, wired for both direct and delegated paths; `#decryptContractWithStaleContextRecovery` — bounded (one retry) recovery wrapper inserted into the existing per-contract `pLimit` loop.
- Docs: `permit-model.md` (new "KMS context rotation" section), `reference/sdk/errors.md` (new error entry + details section + common-problems row).
- Tests: canary/drift-guard for the message heuristic, `CredentialService.invalidatePermit` (direct/delegated isolation, keypair untouched, requires wallet), `DecryptionService` (success path, bounded-retry-exhaustion, delegated path, non-stale-error passthrough, and an isolation test proving one contract's stale permit never affects another's already-granted permit in the same scope).

## Verification

- `pnpm typecheck` / `pnpm lint` — clean
- `pnpm exec vitest run --project sdk` — 1094/1094 passing
- `pnpm api-report` / `api-report:check` — clean diff (`StaleKmsContextError` + `ZamaErrorCode.StaleKmsContext` only), no TSDoc warnings
- `pnpm docs:check-links` — 89 pages, all links/anchors resolve
- `pnpm llm:build` — regenerated

## Not in scope

No public API changes beyond the new error class/code — consumers see at most one extra wallet prompt, then a successful result or a thrown `StaleKmsContextError`.


zmalatrax:
**Closing: superseded by protocol 0.14, rescoped as SDK-294.**

 This PR was built against `@fhevm/sdk@1.1.0-alpha.9`, where the decrypt path hard-asserted that a permit's `extraData` (contextId + epochId) matched the *current* on-chain KMS signers context. Routine rotation therefore broke every existing permit, and this PR added detection (message-text match on the internal assertion) plus recovery (evict permit, re-grant, retry once).

The protocol 0.14 line of `@fhevm/sdk`, including the `0.13.2` build `prerelease` pins today, removes that assertion entirely. At decrypt time, `fetchKmsSigncryptedShares` now resolves the signers context *from the permit's own extraData* via `readKmsSignersContextFromPermitExtraData`, reading the historical context on-chain from the KMSVerifier. Permits stay valid across rotations and age out at their natural expiry. The failure mode this PR recovers from can no longer occur.

Concretely, the branch no longer applies:
- The detection keys on the assertion message `"does not match KmsSignersContext extraData"`, which does not exist anywhere in `@fhevm/sdk@0.13.2`.
- The branch's own drift-guard test (which reads the installed `@fhevm/sdk` source) fails against 0.13.2, by design.
- A rebase onto `prerelease` conflicts in 10 files, all resolving toward code whose trigger is dead.

**What survives:** the recovery *flow* (invalidate the persisted permit, one fresh `grantPermit` prompt, retry once, typed error surface) is still the right shape for the residual failure mode: a **revoked** context, where the on-chain signers read reverts with `InvalidKmsContext`. That is rescoped in SDK-294.

PR #569, opened 2026-07-15
feat(sdk): consume chain-sourced /keyurl — ProtocolConfig fix + KMSGeneration digest verification

zama-cremaud:
## Summary

Re-scoped implementation of SDK-247. Of the ticket's 3 items, only one turned out to be real, unimplemented work — the other two either don't apply anymore or are already handled transparently by `@fhevm/sdk`. Full investigation trail in the linked ticket / conversation; short version:

- **`dataId` opaque-cache-key tolerance** — moot. The cache it was meant to integrate with (`fhe-artifact-cache.ts`, SDK-7) was deleted by #458 with no replacement; `@fhevm/sdk` already treats `dataId` as fully opaque.
- **`contextId`/`epochId` consumption** — the ticket's premise is wrong (they're not `/keyurl` fields); `@fhevm/sdk`'s `signDecryptionPermitV2` already resolves them transparently via a separate on-chain call. But investigating this surfaced a real, unrelated bug (fixed here, see below).
- **On-chain SHAKE256 digest verification** — the only genuinely new, security-relevant scope. Implemented here.

### 1. Fix: wire `ProtocolConfig` address through (latent bug)

`@fhevm/sdk`'s permit-signing path throws `'protocolConfigAddress is required on protocol v0.14.0+'` the instant a chain's `KMSVerifier` crosses v0.4.0 — and this SDK's `toFhevmChain()` hardcoded `protocolConfig: undefined` with no `FheChain` field to override it. Verified live: mainnet `KMSVerifier` = v0.2.0, Sepolia = v0.3.0 (both safe today), but nothing in this repo would prevent the throw once either upgrades. Adds `protocolConfigContractAddress?: Address` to `FheChain`, left `undefined` on every existing preset — zero behavior change, purely preventive.

### 2. Feat: verify FHE public key/CRS digest against on-chain `KMSGeneration`

Adds `kmsGenerationContractAddress?: Address` to `FheChain`. When set, `FhevmRelayer#initEncrypt` verifies the downloaded FHE public-key/CRS bytes against `KMSGeneration.getKeyMaterials`/`getCrsMaterials` before using them:

- Digest = `SHAKE256(domain_separator || bytes)`, truncated to 32 bytes (`"PDAT_KEY"` / `"PDAT_CRS"` 8-byte ASCII domain separators) — matches kms-core's hashing exactly, verified against the first-party Rust reference in `fhevm-engine/host-listener/src/kms_generation/digest.rs`. **Not** a plain `SHAKE256(bytes)`.
- Two distinct error types: `KeyDigestVerificationFailedError` (RPC/read failure — retryable) vs. `KeyDigestMismatchError` (confirmed mismatch — terminal), so a flaky RPC provider is never mistaken for a security incident.
- New `abi/kms-generation.abi.ts` + `contracts/kms-generation.ts` follow the existing `contracts/`+`abi/` factory convention (see `contracts/acl.ts`). The ABI is manually transcribed (not `pnpm abi:build`-generated) because `KMSGeneration` lives in the `fhevm` submodule, which isn't wired into this repo's Forge build — noted as a follow-up.

**Left `undefined` on every built-in preset.** Neither production (mainnet/Sepolia — confirmed via Zama's public contract-address docs) nor this repo's local `hardhat`/`anvil` mock stack (confirmed by reading `contracts/lib/forge-fhevm`'s `DeployFHEVMHost.s.sol`: only 5 proxies deployed, no `ProtocolConfig`/`KMSGeneration`) has these contracts deployed yet. So both fields are dead-but-ready: zero behavior change today, ready the moment a real deployment exists.

## Test plan

- [x] `pnpm typecheck && pnpm lint && pnpm test` — 1090 tests pass
- [x] New unit tests for `verify-key-digest.ts`: match / key mismatch / CRS mismatch / no-Public-digest-recorded / RPC failure / invalid dataId / correct on-chain call args
- [x] SHAKE256+DSEP construction cross-checked against an independent implementation (Python's `hashlib.shake_256`), not just structural parity with `@noble/hashes`
- [x] `to-fhevm-chain.ts` unit test: `protocolConfigContractAddress` present/absent
- [x] `pnpm api-report` + `pnpm llm:build` regenerated (new public `FheChain` fields + error classes)
- [ ] No live/e2e test is possible yet for the digest-verification path — neither prod nor local anvil has `KMSGeneration` deployed. Follow-up once a real deployment exists (production: needs protocol/relayer team confirmation; local: needs `forge-fhevm` submodule bumped to a version that deploys `ProtocolConfig`/`KMSGeneration`).

Part of SDK-247.


packages/sdk/src/abi/kms-generation.abi.ts
```diff
@@ -0,0 +1,41 @@
+// Manually transcribed — NOT generated by `pnpm abi:build`, unlike this file's
```
ghermet: I think it'd be better if this were generated from `contracts/lib/fhevm/host-contracts/contracts/interfaces/IKMSGeneration.sol`

packages/sdk/src/relayer/fhevm-relayer.ts
```diff
@@ -110,7 +113,17 @@ export class FhevmRelayer implements FhevmRelayerSDK {
...
+        const kmsGenerationContractAddress = this.#chain.kmsGenerationContractAddress;
+        if (kmsGenerationContractAddress !== undefined) {
+          await verifyFheEncryptionKeyDigest(
+            this.#publicClient,
+            kmsGenerationContractAddress,
+            keyBytes,
+          );
+        }
+        return this.#encrypt.init();
+      });
```
ghermet, line 128: Isn't `fhevm/sdk` supposed to do handle this?

zama-cremaud: Good question — dug into this before answering rather than guessing.

The source ticket (`fhevm-internal#1466`, relayer-side chain-sourced `/keyurl`) explicitly punts this: *"SDK-side verification (SHAKE256 of downloaded bytes against the on-chain digest) is the consumer of this work and is tracked separately"* — and in the Deviations section, more precisely: *"is tracked on the relayer-sdk, not here."* So the original intent was indeed for `@fhevm/sdk` to own this, not us.

In practice though, nobody picked it up there: no issue, no PR, no test stub anywhere in `zama-ai/fhevm` or `fhevm-internal` mentions `KMSGeneration`/SHAKE256/digest verification (checked via GitHub search across both repos). The RFC-005 keyurl test that did land (fhevm#2950, `fetchFheEncryptionKeySource.test.ts`) only covers `dataId`/`contextId`/`epochId` parsing — no digest check. `@fhevm/sdk` is at `1.1.0-alpha.9` (latest published) and still has zero references to any of this.

So: the ticket says it belongs upstream, but upstream hasn't built it, and SDK-247 is the only tracker that exists for it today. Keeping the implementation here for now so the security property actually ships — happy to drop/replace it if `@fhevm/sdk` picks this up later (the `kmsGenerationContractAddress` gate makes that a clean removal, not a breaking one). Worth a direct conversation with the fhevm-sdk team on intended ownership going forward, but didn't want that to block this PR.

packages/sdk/src/abi/kms-generation.abi.ts
zama-cremaud: Yes, good point, updated

ghermet (approved):

zmalatrax:
Sync on Aug 17 with devex to make the on-chain digest verification on fhevm/sdk instead

zmalatrax:
Superseded by https://github.com/zama-ai/fhevm/pull/3640

PR #570, opened 2026-07-15
ci(examples): typecheck, build & e2e example apps against their pinned SDK [SDK-250]

zama-cremaud:
## Summary

`examples/*` are standalone npm projects — deliberately **not** pnpm workspace members — that each pin a **published** `@zama-fhe/sdk` / `@zama-fhe/react-sdk` version in their own committed lockfile. Until now the only workflow touching them (`example-hoodi-e2e.yml`) covered a single app's Playwright run; nothing typechecked or built the rest. So a broken import or type error in any example could sit unnoticed until someone cloned it.

Adds `.github/workflows/examples.yml` with two jobs, each installing every example exactly as an end user would — `npm ci` against the committed lockfile, i.e. the **published** SDK release the example actually targets:

- **`build`** — matrix over all 9 apps in `examples/*`: `npm ci` → `npm run typecheck` → `npm run build --if-present` (the last is a no-op for the plain-Node apps that ship no build script).
- **`e2e`** — matrix over the 5 apps that ship a Playwright suite (`example-bnb`, `example-hoodi`, `react-ethers`, `react-viem`, `react-wagmi`). The tests mock the RPC at the network layer and boot each app's own `npm run dev`, so they need no chain, keys, or secrets — only the browser binary. GitHub sets `CI=true`, which flips `reuseExistingServer` off so Playwright always starts fresh.

The standalone `example-hoodi-e2e.yml` is deleted — its coverage is now the hoodi row of the `e2e` matrix.

Path-filtered (`examples/**`, `.github/workflows/examples.yml`) so it runs only when something relevant changes, mirroring the existing per-area workflow pattern rather than folding into the always-on `ci.yml`.

## What this does and doesn't catch

It validates the **example code** against the SDK version each app pins — a bad import, a type error, a broken UI flow. It intentionally does **not** build the in-repo SDK source and overlay it onto the examples: examples pin published releases, not `workspace:*`, so coupling them to unreleased source would just make the job fail on every in-flight API change. Catching SDK API drift before publish is a separate concern, owned by the package-level CI.

## Supporting example changes

- **Per-app dev ports** in each `playwright.config.ts` (3001–3004…) so the `e2e` matrix apps don't collide on `:3000` when they run in parallel.
- **Explicit `@types/node`** (and `types: ["node"]` in tsconfig) for `node-viem` / `node-ethers`, which were relying on an ambient install that isn't present under a clean `npm ci`.
- **Lockfile resyncs** for the node examples to match their `package.json`.

## Test plan

- [ ] Confirm the `Examples CI` check appears on this PR with both `build` and `e2e` jobs green
- [ ] Spot-check logs for one Next.js app and one plain-Node app in `build`
- [ ] Spot-check one `e2e` app's logs to confirm the dev server boots on its pinned port

## Follow-ups

- Registering `Examples CI` as a **required** status check in branch protection is a repo-settings change, left to apply separately.
- The matrix is a static list of the apps present today. New examples (e.g. those landing via other in-flight PRs) need adding to the relevant matrix when they merge.


.github/workflows/examples-ci.yml
```diff
@@ -0,0 +1,151 @@
...
+          cache-dependency-path: examples/${{ matrix.example }}/package-lock.json
+
+      - name: Download SDK tarballs
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0
+        with:
+          name: sdk-tarballs
+          path: /tmp/tarballs
+
+      - name: Install with workspace SDK overlay
+        # Point this example's own SDK dependency at the freshly-built workspace
```
ghermet: Is this what we want? example apps are using npm versions not workspace builds

ghermet (approved):

PR #574, opened 2026-07-17
chore(docs): docs:changelog automation + fold promotion into docs:retarget main

ankurdotb:
## What

Adds a deterministic **`pnpm docs:changelog`** scaffolder and folds the changelog **alpha→version promotion** into **`pnpm docs:retarget main`**, so the GitBook human-readable changelog stays in sync with `CHANGELOG.md` with minimal manual work. Paired with a new `sdk-changelog` skill (separate marketplace PR) that drafts the plain-language prose.

### `scripts/docs/changelog.mjs` (new)
- Parses `CHANGELOG.md` (mainline vs alpha, typed sections; tolerant of ⚠ BREAKING-CHANGES squash-noise and empty/full hash links).
- **Scaffold (default):** adds missing `## {version}` stubs on the right `{minor}.x` page (newest-first), creates new `v3-N.md` pages + wires the `SUMMARY.md` nav (2-space indent) and the `v3.md` index, refreshes the Alpha raw-material block. Leaves `<!-- TODO -->` markers for the skill to draft. **Idempotent + no-op-safe.**
- **`promoteChangelog()` (exported):** moves the Alpha tip into the shipped version page, resets Alpha, rewires nav, and repoints inbound + self-referential anchors. Version auto-derived from the newest mainline `CHANGELOG.md` entry (`--version` overrides). Own state-detection → zero-diff on re-run.

### `scripts/docs/retarget.mjs` (behavior change)
`pnpm docs:retarget main` now runs `promoteChangelog()` first, with strict ordering **promote → URL-rewrite → oxfmt(changed) → llm:build** — so the promoted page's `alpha/` URLs flip to stable and *every* written file is formatted before the corpus inlines it (the ordering whose absence gated the 3.3.0 release). `docs:retarget prerelease` never promotes. No-op-safe (exit 0, no writes; warns loudly if `CHANGELOG.md` looks stale at merge time).

### Also
- `docs/agents/changelog.md` — the GitBook changelog authoring convention (linked from `AGENTS.md`).
- `package.json` — the `docs:changelog` script.
- Tests: `changelog.test.mjs` (parser + promotion state-detection), `retarget.test.mjs` (round-trip byte-identity, twice-`retarget main`-is-zero-diff, a formatting-invariant test reproducing the #538 table-reflow class, failure isolation, no-op), and a new `check-target.test.mjs` (previously-uncovered `main()`).

## Verification

- `vitest run scripts/docs` → **86 passing** (4 files, no type errors).
- `pnpm docs:check-links` → **98 pages, all links/anchors resolve.**
- Corpus is a fixed-point of the committed docs (fresh `llm:build` → zero diff), so `llm:check` stays clean.
- Reviewed adversarially (promotion ordering, no-op-safety, double-promotion, failure isolation, parser hazards); findings fixed.

## Notes

- **Tooling only** — this PR does not refresh the Alpha page content; drafting the plain-language prose is the `sdk-changelog` skill's job (a follow-up content PR).
- `chore(docs)` type → no release entry (internal dev tooling, not shipped in the package).


scripts/docs/changelog.mjs
```diff
@@ -0,0 +1,954 @@
...
+  // Workhorse guard: already promoted. Normally that means retarget ran twice — a
+  // true no-op. But if the Alpha page STILL carries staged content while the newest
+  // mainline entry is already on its page, CHANGELOG.md is very likely STALE: the
+  // release commit (`chore(release): X.Y.Z [skip ci]`) that adds the NEW `## [X.Y.Z]`
+  // hasn't been pulled, so `target` derived to the PRIOR, already-promoted version.
+  // That would silently ship a release with no version page and no alpha reset — and
+  // a MISSING page is not a broken link, so nothing downstream (incl. check-links)
+  // catches it. Warn loudly, but stay a no-op (exit 0 preserved): retarget runs at
+  // merge time, so failing here would block the merge.
+  if (pageContent !== null && hasVersionSection(pageContent, target)) {
```
ghermet, line 754: [AUTOMATED] **FYI (non-blocking):** the "already promoted → no-op" guard keys off `hasVersionSection(pageContent, target)`, which only checks whether a `## {version}` heading/slug exists on the page. But `scaffoldChangelog` → `renderStubSection` also emits a `## {version}` section (raw bullets + `TODO_MARKER`) on the same `changelog/vX-Y.md` page. So if someone runs `pnpm docs:changelog` (scaffold) after the mainline release entry lands but before `pnpm docs:retarget main` (promote), promote sees the stub heading, decides "already promoted," and skips — leaving the stub's raw bullets + visible TODO marker on the shipped page and never resetting `alpha.md`.

This is **not reachable via CI or the documented release flow** (CI never runs scaffold/promote; the documented flow drafts prose on the Alpha page then runs `docs:retarget main`), and it isn't silent — the `alphaHasSubstance` branch below still fires the loud stderr warning, so a maintainer reviewing the retarget diff would see the TODO stub. So it's a low-severity, manual-only footgun rather than a bug in the intended path.

If you want to close the footgun cheaply, disambiguating the guard with something like `&& !pageContent.includes(TODO_MARKER)` near the version heading would let a stub still be promoted rather than mistaken for finished prose. Your call — safe to ship as-is.

Confidence: 45/100

ghermet (commented):
[AUTOMATED]

## Code Review

Reviewed with parallel bug-hunting, guideline-compliance, and error-handling agents, then independently validated the one finding that cleared the confidence bar.

**No blocking issues found.** The changelog parser, scaffold idempotency, promotion/anchor-rewiring, nav wiring (SUMMARY 2-space indent, newest-first, v3-N page creation), and the retarget `promote → URL-rewrite → oxfmt → llm:build` ordering all check out against the real docs tree. Error handling is solid: spawn failures propagate non-zero, missing `CHANGELOG.md` throws descriptively, no-op paths exit 0 with no writes, and `EditSet.flush()` runs last so a compute-phase throw leaves the tree untouched. Semver sorting is numeric (`3.10.0` before `3.9.0`), and string replaces use function replacers (no `$` pitfalls). 86 tests pass.

One non-blocking **FYI** inline about a latent footgun in the scaffold/promote interaction, plus two minor observations below. None are blocking.

### Minor observations (no action required)
- **Parser:** `VERSION_HEADING_RE` (changelog.mjs:68) matches only the bracketed `## [X.Y.Z](url) (date)` form, not the linkless `## X.Y.Z (date)` form semantic-release emits for the very first release. Impact is effectively zero — it only affects the oldest 1.x alpha, which every downstream consumer filters out by active-major. Flagging only because the heading comment implies it matches the semantic-release heading generally.
- **Classification:** `isAlphaVersion` (changelog.mjs:72) matches only `-alpha.N`; a hypothetical `-rc.N`/`-beta.N` prerelease would be classified as mainline and promoted. Latent only — the repo emits alpha prereleases exclusively today.

ghermet (approved):

PR #576, opened 2026-07-20
refactor(examples): improve example apps — viem/wagmi, React 19 form actions, a11y

ghermet:
## What

Refactor and modernization sweep across all seven example apps (`example-bnb`, `example-hoodi`, `example-ingen`, `react-ethers`, `react-viem`, `react-wagmi`, `react-turnkey-wallet`).

## Changes

- **viem/wagmi migration** — move ethers-based apps to viem/wagmi and drop the bespoke `lib/ethereum.ts`, `global.d.ts`, and `client-providers.tsx` shims.
- **React 19 form actions** — adopt `useActionState` with client-side validation. The submit button reads its pending flag from the action state (covering the client-side validation step), not the underlying mutation.
- **Accessibility** — replace `role="status"` alerts with `<output>` (SonarQube S6819) plus matching CSS; checksum + `pattern`-validate address inputs.
- **Tidy-ups** — extract a shared `PageChrome` layout, inline `parseAmount` at its single use sites, and refresh READMEs, WALKTHROUGHs, and e2e specs/fixtures.

134 files changed (+5,229 / −4,173).

## Verification

- `turbo run typecheck` — green across all packages.
- e2e: run in CI; draft pending a full local Playwright pass.

## Notes

Base is `prerelease` per repo convention (`main` is reserved for release/CI infra).


examples/example-ingen/src/components/DecryptAsCard.tsx
```diff
@@ -26,93 +24,103 @@ function formatExpiry(expiryTimestamp: bigint): string {
...
+        </button>
+      </form>
+      {errorMessage && (
+        <div className="alert alert-error card-status" role="alert">
+          {errorMessage}
         </div>
       )}
-      {decryptAs.isSuccess && decryptAs.data !== undefined && (
-        <div className="alert alert-success card-status">
+      {decryptAs.data && (
```
ankurdotb: [AUTOMATED] Falsy-zero regression: `decryptAs.data &&` is `false` when the decrypted balance is exactly `0n` — a legitimate result, but `0n` is falsy in JS. The success confirmation silently never renders for a zero balance, even though decryption succeeded.

`example-bnb` and `example-hoodi` have the correct guard:
```tsx
{decryptAs.isSuccess && decryptAs.data !== undefined && (
```
This file is the only one of the three that regressed to `decryptAs.data &&`. Suggest matching the other two.

Confidence: 92/100

examples/example-bnb/src/components/ShieldCard.tsx
```diff
@@ -1,78 +1,90 @@
...
+        <button type="submit" className="btn btn-primary btn-full" disabled={disabled || isPending}>
+          {isPending ? pendingLabel : "Shield"}
+        </button>
+      </form>
+      {errorMessage && (
+        <div className="alert alert-error card-status" role="alert">
+          {errorMessage}
+        </div>
       )}
       {shield.isSuccess && shield.data?.txHash && (
```
ankurdotb: [AUTOMATED] Stale success banner can render alongside a new validation error. Sequence: a successful shield sets `shield.isSuccess` (banner shows) → user resubmits with an amount that parses to `0n` (e.g. precision beyond the token's decimals slips past the native `min`/`step` constraints) → the action returns `"Enter a valid amount."` (line 34) *before* ever calling `shield.mutateAsync` again → `shield`'s mutation state is untouched from the prior call, so `shield.isSuccess` is still `true`. Result: the new error banner and the stale "Shielded!" success banner (this line) render simultaneously — a misleading double-signal on a financial action.

Pre-refactor the submit button was flatly `disabled` on invalid input, so this state was unreachable; adopting `useActionState` with post-submission validation is what opened the gap.

Same shape, same fix needed, in every app that adopted this pattern:
- `ShieldCard.tsx` / `TransferCard.tsx` / `UnshieldCard.tsx` in example-bnb, example-hoodi, example-ingen, react-ethers, react-viem, react-wagmi (lines ~86/106/92)
- `react-wagmi/src/components/VaultDepositCard.tsx:136` (same pattern, `deposit.isSuccess`)

Smallest fix: gate the success banner on `!errorMessage` too, or call `shield.reset()` when validation rejects the input.

Confidence: 85/100

examples/react-turnkey-wallet/src/components/react-turnkey-wallet/shield-card.tsx
```diff
@@ -16,11 +16,10 @@ export function ShieldCard({
...
 }) {
   const shield = useShield({ address: tokenAddress });
-  const [amount, setAmount] = useState("");
   const [phase, setPhase] = useState<"prepare" | "approve" | "wrap">("prepare");
 
-  function handleShield() {
-    const parsed = parseAmountSafe(amount, decimals);
+  function handleShield(formData: FormData) {
+    const parsed = parseAmountSafe(formData.get("amount") as string, decimals);
     if (!parsed) return;
```
ankurdotb: [AUTOMATED] Same root cause as the `useActionState` cards elsewhere, different shape: `if (!parsed) return;` silently no-ops on an invalid amount without ever calling `shield.mutate(...)` again. `<MutationStatus mutation={shield} />` (line 69) renders "✓ Success" whenever `mutation.isSuccess`, which is still `true` from a prior successful shield — react-query doesn't reset that state just because this render's submit was skipped. So a user who shields successfully, then submits a zero/invalid amount next, sees "✓ Success" persist with no indication the second submission did nothing.

Same shape in `transfer-card.tsx:19,62` and `unshield-card.tsx:23,70`.

Confidence: 85/100

examples/example-bnb/src/lib/queryKeys.ts
```diff
@@ -0,0 +1,7 @@
+import type { Address } from "@zama-fhe/sdk";
+
+// Query key for the public ERC-20 balance. Shared between BalancesCard (which reads it)
+// and the operation cards' onSuccess handlers (which invalidate it after shield/unshield/mint),
+// so a self-contained BalancesCard refreshes automatically without prop-drilling the value.
+export const erc20BalanceKey = (tokenAddress: Address, account: Address) =>
```
ankurdotb, line 7: [AUTOMATED] Nit: `erc20BalanceKey` doesn't normalize address casing, unlike the SDK's own key builder one package over:
```ts
// this file
["erc20-balance", tokenAddress, account] as const

// packages/sdk/src/query/query-keys.ts
["zama.confidentialBalance", { tokenAddress: getAddress(tokenAddress) }] as const
```
If `token.tokenAddress` (from the registry) and `account` (from `useConnection`) ever arrive in different casing for the same address, `BalancesCard`'s `useQuery` and the operation cards' `invalidateQueries` calls land on two different cache keys — the balance display would silently stop refreshing after a shield/unshield instead of erroring. Unconfirmed whether this actually manifests at runtime (depends on wallet/connector casing behavior), but routing both args through `getAddress()` before building the key would close it off, matching the SDK's own convention. Same shape across all 6 apps.

Confidence: 70/100 (real class of bug, unconfirmed manifestation)

examples/example-bnb/src/lib/hooks.ts
```diff
@@ -0,0 +1,53 @@
...
+// key with the operation cards' invalidations, so a shield/unshield/mint elsewhere
+// refreshes it without any prop plumbing.
+export function useUnderlyingBalance(
+  token: TokenWrapperPairWithMetadata,
+  account: Address,
+  options?: { enabled?: boolean },
+) {
+  const sdk = useZamaSDK();
+  return useQuery({
+    queryKey: erc20BalanceKey(token.tokenAddress, account),
```
ankurdotb, line 24: [AUTOMATED] Nit: `queryFn` closes over `token`/`account` from the hook's own parameters instead of extracting them from `context.queryKey`. Not exploitable today — single call site, and the key is built from the exact same values the closure captures, so they can't diverge — but it's the standard TanStack anti-pattern (per our tanstack-best-practices guidance) and the first thing that bites if this hook is ever reused with an independent identity source. Consider extracting params from `context.queryKey` for consistency with how the SDK's own query options factories are built.

Confidence: 60/100 (style/future-proofing, not a live bug)

examples/react-wagmi/src/lib/hooks.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { erc20BalanceKey } from "@/lib/queryKeys";
+import { useMutation, useQuery } from "@tanstack/react-query";
+import { useZamaSDK } from "@zama-fhe/react-sdk";
+import type { Address, TokenWrapperPairWithMetadata } from "@zama-fhe/sdk";
+import { parseAbi, parseUnits } from "viem";
+
+// Standard ERC-20 balanceOf ABI — parseAbi is required: viem does not parse
+// human-readable ABI strings automatically.
+const BALANCE_ABI = parseAbi(["function balanceOf(address) view returns (uint256)"]);
```
ankurdotb: [AUTOMATED] Consider: this hand-rolls the `balanceOf` ABI and a raw `readContract` call, while the same read in example-bnb/hoodi/ingen/react-viem's `hooks.ts` goes through `balanceOfContract` from `@zama-fhe/sdk`. Same data, just reimplementing an existing SDK helper — worth aligning for consistency across the fleet.

Confidence: 75/100

ankurdotb (commented):
[AUTOMATED]

## Code Review

Reviewed this as a modernization/tooling sweep (viem/wagmi migration, React 19 form actions, a11y) rather than a protocol-level change — it only touches `examples/` and generated docs, no `packages/sdk`, `packages/react-sdk`, or contracts.

**FHEVM/SDK-usage alignment**: solid. The examples already track the current SDK surface (`confidentialTransferAndCall`, `usePendingUnshield`/`useResumeUnshield`, internally-absorbed delegation-propagation retries) and this refactor doesn't regress any of it — it actually removes a stale manual-approval sequence from the WALKTHROUGH docs in favor of `useShield`.

Found 2 real behavior bugs and 3 minor nits (TanStack Query key/closure hygiene, one reinvented SDK helper) — all inline below.

One thing with no single line to anchor it: the examples still pin `@zama-fhe/sdk@3.3.0-alpha.9` while the repo's own packages are at `3.4.0-alpha.10`. Not urgent (that gap is just the `@fhevm/sdk` backend migration, documented as a no-op for `Token`/hooks consumers), but worth a conscious call on whether this sweep should also bump the pinned example version.

Everything else checked clean: provider/signer wiring, chain configs, address-checksum handling, `PageChrome` error propagation, and CLAUDE.md's "use the SDK method, don't recompose" rule.

ankurdotb (approved):

PR #580, opened 2026-07-22
fix(sdk): normalize ethers readContract integers to match viem

ghermet:
## Problem

`EthersProvider.readContract` cast ethers' decoded result straight to viem's `ContractFunctionReturnType` without reshaping it. ethers v6 decodes **every** Solidity integer as a `bigint`, whereas viem — the return-type contract both providers implement — narrows integer widths of ≤48 bits to `number`.

So `Token.decimals()`, typed `Promise<number>`, actually returned a `bigint` under ethers. Any caller treating it as a number broke — e.g. the `react-ethers` example's `minAmount`, whose `decimals - 1` threw *"Cannot mix BigInt and other types"* and blanked the page (surfaced as the `e2e (react-ethers)` failures on #576).

Only ethers-backed integrations were affected; `ViemProvider` already decodes `uint8` → `number`.

## Fix

Walk ethers' own parsed `ParamType` outputs (`fn.fragment.outputs`) and coerce small ints to `number`, recursing into arrays and tuples, so both providers honour the same return-type contract regardless of backend. No public API change (verified via api-report).

## Tests

Added coverage to `ethers.test.ts`: `uint8` → `number`, `uint256` stays `bigint`, small ints inside array outputs, and per-value normalization of multi-output calls. All 35 ethers tests pass; typecheck, lint, build, and api-report clean.

## Note

This does **not** by itself green the #576 examples CI — that job installs each example against its *pinned published* SDK, so it needs the separate example-side workaround (already on `refactor/examples-apps`). This PR fixes the root cause for the next SDK release.


packages/sdk/src/ethers/ethers-provider.ts
```diff
@@ -70,8 +111,28 @@ export class EthersProvider implements GenericProvider {
       this.#readProvider,
     );
     const fn = contract.getFunction(config.functionName);
-    return fn(...(config.args as readonly unknown[])) as Promise<
-      ContractFunctionReturnType<TAbi, "pure" | "view", TFunctionName, TArgs>
+    const result = await fn(...(config.args as readonly unknown[]));
```
zmalatrax: The signature of `fn` is `const fn: ethers.ContractMethod<any[], any, any>`

So currently `result`, `fallback` and `normalized` are all typed `any`, which always passes `satisfies` (#L131)

I'd rather annotate it to unknown `const result: unknown = ` to avoid the propagation of `any`

packages/sdk/src/ethers/ethers-provider.ts
```diff
@@ -70,8 +111,28 @@ export class EthersProvider implements GenericProvider {
...
+
+    const fallback =
+      outputs.length === 0
+        ? result
+        : outputs.map((output, index) => toViemShape((result as unknown[])[index], output));
+
+    const normalized =
+      outputs.length === 1 && firstOutput ? toViemShape(result, firstOutput) : fallback;
+
+    return normalized satisfies ContractFunctionReturnType<
```
zmalatrax: Following https://github.com/zama-ai/sdk/pull/580/changes#r3672345810, current code does this check `any satisfies ContractFunctionReturnType<T>`, which always passes.

Given the generic type, I'd rather avoid making an always-true check and simply cast it instead (if you annotate `result` to `unknown`, the `satisfies` check will fail)
```suggestion
    return normalized as ContractFunctionReturnType<
```

packages/sdk/src/ethers/ethers-provider.ts
```diff
@@ -70,8 +111,28 @@ export class EthersProvider implements GenericProvider {
...
+    // ethers decodes every integer as a `bigint`; reshape the result to match viem so both
+    // providers honour the same return-type contract (e.g. `decimals` is a `number`, not a
+    // `bigint`). ethers unwraps a single-output call to the bare value and returns an array-like
+    // `Result` for multi-output ones — mirror that split when walking the parsed output types.
+    const outputs = fn.fragment.outputs;
+    const [firstOutput] = outputs;
+
+    const fallback =
+      outputs.length === 0
+        ? result
```
zmalatrax: viem returns `undefined` for a function with no outputs, cf. [decodeFunctionResult](https://github.com/wevm/viem/blob/207c8bb62b1a3621ea04d865c3a73c873207a237/src/utils/abi/decodeFunctionResult.ts#L160-L165), while ethers return an empty array, cf. [ethers](https://github.com/ethers-io/ethers.js/blob/3ea4c226dd0b3a074abf90748de9d11192027f4f/src.ts/contract/contract.ts#L302-L306)

```suggestion
        ? undefined
```

packages/sdk/src/ethers/ethers-provider.ts
```diff
@@ -9,6 +10,46 @@ import type {
...
+  if (param.isArray() && arrayChildren && Array.isArray(value)) {
+    return value.map((element) => toViemShape(element, arrayChildren));
+  }
+
+  // Tuples/structs: ethers returns an array-like `Result`; normalize each component positionally.
+  if (param.isTuple() && components && Array.isArray(value)) {
+    return components.map((component, index) =>
+      toViemShape((value as unknown[])[index], component),
+    );
+  }
```
zmalatrax: This tuple branch is broken (regression), cf. [viem tuple decoding](https://github.com/wevm/viem/blob/207c8bb62b1a3621ea04d865c3a73c873207a237/src/utils/abi/decodeAbiParameters.ts#L305-L308)

It always produces an array, when viem returns objects for named tuples

It wasn't caught by tests, the ethers tests mock ethers.Contract and inject fragment.outputs, hence not the real ethers/viem. We should add diff tests between ethers provider and viem outputs, using the actual ethers package, with a stub provider and viem

```suggestion
  // Tuples/structs: ethers always returns an array-like `Result`, but viem returns a *keyed object*
  if (param.isTuple() && components && Array.isArray(value)) {
    const normalized = components.map(
      (component, index) => [component.name, toViemShape(value[index], component)] as const,
    );
    const hasUnnamedChild = components.length === 0 || components.some(({ name }) => !name);
    return hasUnnamedChild
      ? normalized.map(([, componentValue]) => componentValue)
      : Object.fromEntries(normalized);
  }
```

zmalatrax (changes requested):
Non-blocking: I'm wondering if we could directly leverage viem instead of re-implementing the viem decoder: using ethers at the transport layer only (provider.call) and viem to encode/decode. There might be other pros/cons I'm not aware of, though I know that we'd lose ethers' revert data decoding

packages/sdk/src/ethers/ethers-provider.ts
ghermet: Resolved by removing the hand-rolled decode entirely. `readContract` now delegates to the existing `ethersRead` helper (ethers transport + viem decode), so there's no `fn: ContractMethod<any,…>` and no `any` propagation — `ethersRead<T>` casts the viem-decoded result to `T` explicitly.

packages/sdk/src/ethers/ethers-provider.ts
ghermet: The `satisfies` is gone — `readContract` delegates to `ethersRead`, which does an explicit `return decodeFunctionResult(...) as T`. No more always-true type assertion.

packages/sdk/src/ethers/ethers-provider.ts
ghermet: Handled by construction now: viem's `decodeFunctionResult` returns `undefined` for zero-output functions, and `ethersRead` uses it directly. Added a test asserting `readContract` returns `undefined` for a no-output function.

packages/sdk/src/ethers/ethers-provider.ts
ghermet: You're right, and it's the reason I dropped the hand-rolled decoder rather than patching the tuple branch. `readContract` now decodes with viem (`decodeFunctionResult`) over ethers' `provider.call`, matching the `ethersRead` helpers — named tuples come back as keyed objects natively, so registry reads like `pair.tokenAddress` work. On the tests: they now feed real ABI-encoded return data (encoded via viem, i.e. what a node returns) through a stubbed transport and assert viem-identical shapes, including a named-tuple case mirroring `TokenWrapperPair` and an empty-output case. The old mock-`Contract`/inject-`fragment.outputs` approach is gone.

ghermet:
Went with your non-blocking suggestion — it's the right call. `EthersProvider.readContract` now uses ethers as the transport only (`provider.call`) and viem for encode/decode, converging onto the `ethersRead` helper that already backs the pre-built `readXxxContract` reads. That deletes `toViemShape` and its int-width heuristic, and makes all four inline comments moot by construction: no `any`-typed `satisfies`, `undefined` for empty outputs, and keyed objects for named tuples (fixing the registry regression).

On the con you flagged — losing ethers' revert-data decoding: I checked, and no SDK read path depends on it. `NoCiphertextError` comes from the decryption path (`errors/decrypt.ts`), not from decoding a read revert, and the one `readContract().catch(...)` site (`WrappedToken.#getUnderlying`) only resets state and rethrows. Reverts still surface as errors; they're just not ABI-decoded, which matches nothing we currently rely on.

Verification: typecheck, lint, 1157 sdk tests, build, and api-report all clean (no public API change). Pushed as c936c40d5.

packages/sdk/src/ethers/__tests__/ethers.test.ts
```diff
@@ -346,24 +346,125 @@ describe("EthersProvider", () => {
   });
 
   describe("readContract", () => {
-    test("creates an ethers Contract and calls the function with args", async ({
+    // EthersProvider uses ethers as transport only: it performs an `eth_call` and decodes the raw
+    // return data with viem. These tests stub `call` with real ABI-encoded return data (encoded via
+    // viem, i.e. exactly what a node returns) and assert the provider hands back viem-identical
+    // shapes — the return-type contract both providers implement. The earlier tests mocked
+    // `ethers.Contract` and injected `fragment.outputs`, so they exercised neither real decoding nor
+    // the ethers/viem divergence, which is how the named-tuple shape regressed unnoticed.
```
zmalatrax: imo this comment is useless, I'd avoid narrating past implementation issues/rationale in code comments

zmalatrax (dismissed):

packages/sdk/src/ethers/__tests__/ethers.test.ts
ghermet: Yeah that doesn't add any value here

zmalatrax (approved):

PR #581, opened 2026-07-27
docs: onboarding refresh — READMEs, guides, diagrams, reference pages

ghermet:
## Why

Onboarding new engineers surfaced documentation that has drifted from the shipped codebase. This is a docs pass to make the READMEs, guides, and diagrams accurate. (An `ONBOARDING.md` was initially added and then removed — it duplicated `AGENTS.md` + `CONTRIBUTING.md` + `README.md` with no unique niche; those three remain the onboarding path.)

## What changed

**Corrected docs that contradicted the code**
- `docs/agents/fhevm-sdk-1x-migration.md`: flip Finding 1 from **PARKED → SHIPPED in #458**. The `@zama-fhe/relayer-sdk` → `@fhevm/sdk` swap has landed (`@fhevm/sdk@1.1.0-alpha.9`, `src/worker/` removed); the doc still claimed it was parked, that we consume `relayer-sdk@0.4.3`, and referenced the deleted worker dir. Reframed as a historical record.
- `concepts/security-model.md`: fixed the crypto summary table (dropped the misleading "CDN integrity" default row; `@zama-fhe/sdk (WASM)` → `@fhevm/sdk`).

**Diagrams** (regenerated via `docs:diagrams`; orphaned SVGs removed; the corrected ones are now embedded)
- `security-key-derivation` → **`credential-storage`**: the old diagram depicted an AES-256-GCM / PBKDF2 encryption-at-rest scheme that does not exist — transport keys are stored **plaintext**.
- `security-wasm-integrity`: CDN-fetch-as-default → **bundled by default**, with the optional URL-hosted + SHA-verify path shown as secondary.
- `two-phase-polling` / `two-phase-invalidation` → **`balance-read`** / **`balance-refresh`**: the balance flow is a single `balanceOf()` pass with a clear-value cache, not two query phases.
- `sdk-modules`: removed the non-existent `BalanceCache` module.
- Embedded the two security diagrams in `security-model.md` and the two balance diagrams in `check-balances.md` (all four were previously orphaned).

**API reference — 6 public hooks had no page**
- Added `useWrap`, `useApproveUnderlying`, `useConfidentialTransferFromAndCall`, `useDecryptPublicValues`, `useDelegatedDecryptValues`, `useTotalSupply` and wired them into `SUMMARY.md`. The escape-hatch pages (`useWrap`/`useApproveUnderlying`) steer readers to `useShield`, per the "use the SDK method, don't recompose" rule.

**READMEs & CONTRIBUTING**
- `pnpm >=10` → `>=11` (root README + CONTRIBUTING; `engines` requires `>=11`).
- `node-ethers` / `node-viem`: `isDelegated()` → `isActive()`.
- `codemods/README.md`: heading typo + flat `codemods/<slug>/` layout.
- `codemods/sdk-migration-v3`: dropped the false "capped at unreleased 3.1" framing (packages are at 3.4.0-alpha.11).
- `react-turnkey-wallet`: removed the invalid `next build --webpack` flag.
- `CONTRIBUTING.md`: refreshed the monorepo-structure block to the real `pnpm-workspace.yaml` layout (`test/*`, `tools/*`, `codemods/*`, examples).

## Verification

- `pnpm docs:check-links` → 104 pages, all internal links/anchors resolve.
- Diagrams recompiled cleanly via `pnpm docs:diagrams`.
- Pre-commit hooks passed (typecheck + LLM corpus regenerated: `llms.txt`, `llms-full.txt`, `corpus-manifest.json`).

Docs-only; no source or public API changes.


codemods/README.md
```diff
@@ -54,7 +54,7 @@ major-line packages in order, or we add a thin orchestrator later.)
 
 - Scaffold new codemods with `npx codemod init`.
 - Use Codemod MCP when creating or refining codemods, especially when symbol definitions or cross-file references matter.
-- Validate package workflows with `npx codemod workflow validate codemods/<family>/<release>/workflow.yaml`.
+- Validate package workflows with `npx codemod workflow validate codemods/<slug>/workflow.yaml`.
```
ankurdotb, line 57: Mmmm I'd say family is correct here but dropping the `<release>` is correct

---

Also flagging two more things while I'm looking at this PR's diagrams/docs:

**1. `credential-storage.d2` — the ML-KEM/transport-keypair split is misleading.**
The diagram draws two sequential boxes — *"SDK generates ML-KEM key pair (client-side)"* → *"Transport key pair (public + private)"* — as if they're different objects in a pipeline. They're the same object: `TransportKeyPairVault`'s docstring calls it a vault for "ML-KEM transport key pairs" (one term), and the generator (`() => Promise<SerializeTransportKeyPairReturnType>`, `packages/sdk/src/credentials/keypair-vault.ts`) returns the transport keypair directly — there's no separate ML-KEM stage. Should be one box, not two.

Also, "(client-side)" is too narrow — ML-KEM keypairs aren't only generated client-side, they can be generated anywhere the SDK runs. The same diagram's own storage-backend row shows `AsyncLocalStorage (Node.js)` right alongside `IndexedDB (browser)`, which already contradicts a client-side-only framing.

**2. `zamaQueryKeys.confidentialBalance` vs `.confidentialBalances` — the split isn't explained, and the docs' invalidation example silently misses one.**
`query-keys.ts` has two disjoint root namespaces — `confidentialBalance` (singular, single-token) and `confidentialBalances` (plural, batched/multi-token) — with no shared prefix, so invalidating one never touches the other. Neither `query-keys.md` nor `check-balances.md` explains why there are two, and both show:
```ts
queryClient.invalidateQueries({ queryKey: zamaQueryKeys.confidentialBalance.all });
```
under `// Invalidate all balance queries` — which only invalidates the singular namespace. The SDK's own internal `invalidateBalanceQueries()` (`packages/sdk/src/query/invalidation.ts:60-68`) knows about the split and invalidates both; the public docs don't. Since `invalidateBalanceQueries` is already exported from `@zama-fhe/sdk/query`, the doc should probably just point readers at that helper instead of hand-composing keys that silently drop half the cache.

ankurdotb (changes requested):

codemods/README.md
ghermet: Both good catches — fixed in cd8b722d5.

**1. `credential-storage.d2` — merged the two boxes into one.** You're right that the ML-KEM stage and "Transport key pair" box are the same object (`TransportKeyPairVault` docstring calls it one term, and the generator returns the keypair directly). It's now a single box: *"SDK generates transport key pair (ML-KEM public + private)"*. Also dropped "(client-side)" — the storage-backend row already showed `AsyncLocalStorage (Node.js)`, so the client-side-only framing was self-contradicting.

**2. `confidentialBalance` vs `confidentialBalances` split.** Added a warning hint to `query-keys.md` spelling out that the two are disjoint root namespaces with no shared prefix (so invalidating one never touches the other), plus a dedicated `invalidateBalanceQueries` section. Both `query-keys.md` and `check-balances.md` now steer readers to `invalidateBalanceQueries` (which invalidates both) instead of hand-composing `confidentialBalance.*` keys, and the disconnect example now uses `invalidateWalletLifecycleQueries` for the same reason.

(The `codemods/README.md` note — agreed, `family` + dropping `<release>` is the intended state, so no change there.)

ankurdotb: Confirmed independently against `cd8b722d5` (not just eyeballing the diff):

- **Transport keypair** ✅ — the merged box ("SDK generates transport key pair (ML-KEM public + private)") matches `TransportKeyPairVault`'s docstring and its single-call generator exactly. No separate ML-KEM stage in the code, so one box is right.
- **`confidentialBalance`/`confidentialBalances`** ✅ — `invalidateBalanceQueries` is genuinely exported off `@zama-fhe/sdk/query` (checked the `package.json` export map, not just the source), and `invalidateWalletLifecycleQueries`'s `predicate: isZamaQuery` really does cover every namespace. New doc prose matches actual behavior.

Leaving the `family`/`<slug>` naming point open on my end — noted your call, not re-litigating it here.

examples/react-turnkey-wallet/README.md
```diff
@@ -51,7 +51,7 @@ Run the example checks:
 ```bash
 npm run lint
 npm run typecheck
-npm run build -- --webpack
+npm run build
```
ankurdotb: This removes `--webpack`, but that's a real, documented Next.js 16 flag (`next build --webpack`) — the explicit opt-out from Turbopack, which is the new default in Next 16 (`package.json` here pins `"next": "^16.2.9"`). It's not an invalid flag; removing it silently switches which bundler this example's documented build check exercises (Turbopack instead of webpack) rather than fixing anything.

CONTRIBUTING.md
```diff
@@ -29,11 +29,25 @@ pnpm test:run
...
-  test-nextjs/           # E2E test app (Playwright + Hardhat)
+  sdk/            # Core @zama-fhe/sdk (framework-agnostic + viem/ethers adapters)
+  react-sdk/      # @zama-fhe/react-sdk — React hooks over TanStack Query
+tools/
+  ast-grep/       # Custom AST lint rules (pnpm ast-grep:scan)
+test/
+  playwright/     # E2E runner (@zama-fhe/playwright) driving the test apps
+  test-nextjs/    # Next.js test app
+  test-vite/      # Vite test app
+  test-components/, test-tanstack-start/
```
ankurdotb: `test-tanstack-start/` doesn't exist anywhere in this repo — checked `find test/`, a repo-wide grep, and git history (`git log --all --diff-filter=A -- '*tanstack-start*'`), all zero hits. `test/` only has `playwright/`, `test-components/`, `test-nextjs/`, `test-vite/`. This line is also the only one in the new block missing a `#` comment / matching the alignment above it.

docs/diagrams/security-wasm-integrity.d2
```diff
@@ -1,67 +1,79 @@
...
   style.bold: true
 }
 
-# --- Flow ---
-cdn: "cdn.zama.org\n(TFHE WASM bundle)" {
-  shape: cloud
-  style.fill: "#E3F2FD"
-  style.stroke: "#1565C0"
+# --- Default: bundled ---
+default: "Default — bundled" {
```
ankurdotb: This diagram's binary model (bundled by default vs. any URL-hosted config always verify-or-fails) doesn't match the SDK's actual `wasmAssetLoadMode` surface — this repo's own `docs/gitbook/src/changelog/alpha.md` documents 5 real modes, including `trusted-direct-url`, which explicitly skips verification for speed. The `verify` branch further down implies that's universal for any URL-hosted path, which it isn't.

docs/gitbook/src/reference/react/useApproveUnderlying.md
```diff
@@ -0,0 +1,131 @@
+---
+title: useApproveUnderlying
+description: Low-level escape hatch that approves the wrapper contract to spend the underlying ERC-20.
+---
+
+# useApproveUnderlying
+
+Approve the confidential wrapper contract to spend the underlying ERC-20. Defaults to a max `uint256` approval, and resets an existing non-zero allowance to zero first when required (as tokens like USDT demand).
```
ankurdotb: Nit: "resets an existing non-zero allowance to zero first when required" stacks two modifiers back-to-back. Maybe: "Defaults to a max `uint256` approval. When required — as tokens like USDT demand — it first resets an existing non-zero allowance to zero."

docs/gitbook/src/reference/react/useApproveUnderlying.md
```diff
@@ -0,0 +1,131 @@
...
+
+Auto-invalidates the `underlyingAllowance` cache on success.
+
+{% include ".gitbook/includes/mutation-result.md" %}
+
+## Related
+
+- [useShield](./useShield.md) -- recommended path; handles approval and wrapping in one call
+- [useWrap](./useWrap.md) -- wrap already-approved underlying into confidential tokens
+- [useUnderlyingAllowance](./useUnderlyingAllowance.md) -- read the current allowance (auto-invalidated on success)
```
ankurdotb, line 131: Polish: this page correctly links to `useUnderlyingAllowance`, but `useUnderlyingAllowance.md` (the read counterpart, not touched by this PR) doesn't link forward to this new page. One-directional; worth a backlink there for discoverability.

docs/agents/fhevm-sdk-1x-migration.md
```diff
@@ -1,43 +1,55 @@
...
+> **This doc is now a historical record** of two migrations that have both landed. Keep it for the
+> rationale and the rename mapping (still an accurate reference for the current API); it no longer
+> describes pending work. For the user-facing description of the current backend, see
+> [`changelog/alpha.md`](../gitbook/src/changelog/alpha.md) and
+> [`concepts/architecture.md`](../gitbook/src/concepts/architecture.md).
+
+Internal feedback that originally set this split (track 1 was deferred first, then completed):
+_"going from relayer-sdk to fhevm/sdk is not in scope for now until future network upgrades. However,
+aligning on a single public glossary is something we can do already to be consistent across all
+products — just updating our public APIs."_ The dependency swap was subsequently prioritised and
```
ankurdotb: Nit: "prioritised" (BrE) — the rest of `docs/agents/*` consistently uses American `-ize` spelling (normalize, finalize, modernize). "prioritized" would match.

docs/gitbook/src/reference/react/useConfidentialTransferFromAndCall.md
```diff
@@ -0,0 +1,160 @@
...
+    useConfidentialTransferFromAndCall(tokenAddress);
+
+  async function handleTransfer() {
+    const { txHash, receipt } = await transferFromAndCall({
+      from: "0xOwner",
+      to: "0xReceiverContract",
+      amount: 500n,
+      data: "0xabcd",
+    });
+    console.log("Confirmed in block", receipt.blockNumber);
```
ankurdotb: Low-severity, inherited rather than introduced here: `receipt.blockNumber` doesn't exist on this SDK's `TransactionReceipt` type (`packages/sdk/src/types/transaction.ts` only declares `logs`). `useConfidentialTransfer.md` and `useConfidentialTransferFrom.md` have the identical snippet already, so this page faithfully copied a pre-existing template defect rather than adding a new one — flagging since it's now in 3 places.

docs/diagrams/balance-read.d2
```diff
@@ -0,0 +1,90 @@
...
+
+# --- Step 1: read the encrypted value (cheap) ---
+read: "Read encrypted value\nconfidentialBalanceOf (eth_call)\n→ bytes32 encryptedValue" {
+  style.fill: "#E3F2FD"
+  style.stroke: "#1565C0"
+  style.font-size: 14
+}
+
+# --- Step 2: clear-value cache lookup ---
+lookup: "Look up clear-value cache\nkey: token + owner + encryptedValue" {
```
ankurdotb: Minor imprecision: labels the cache key `token + owner + encryptedValue`, but the actual key (`packages/sdk/src/services/caching-service.ts`) is `(requester, contractAddress, encryptedValue)`. Equivalent on the normal balance path, just not quite the real field names.

ankurdotb (commented):
Rest of the sweep from earlier discussion — flagging the parts that can't be inline-anchored because the files aren't part of this PR's diff, plus a couple of nits.

### Not touched by this PR, but worth a follow-up

- **`examples/node-ethers/WALKTHROUGH.md`** (lines 24, 210) and **`examples/node-viem/WALKTHROUGH.md`** (lines 24, 220) still show `tokenA.isDelegated({...})` — this PR correctly renamed `isDelegated()` → `isActive()` in the sibling `README.md` files, but missed the `WALKTHROUGH.md` in the same two example folders. Worse than just a stale name: the actual example source calls `sdkA.delegations.isActive(...)`, not `tokenA.isDelegated(...)` — the walkthrough snippet is wrong on both the method name *and* the receiver, so copy-pasting it gives broken code.
- **Terminology split this PR creates adjacency for**: the 4 new reference pages here correctly say "base units" (matching the actual JSDoc in `packages/sdk/src/query/{transfer,shield,transfer-from}.ts`), but the pre-existing sibling pages right next to them in `SUMMARY.md`'s nav — `useShield.md`, `useConfidentialTransfer.md`, `useConfidentialTransferFrom.md`, `useUnderlyingAllowance.md` — still say "smallest unit" for the identical field. Since the new pages are the accurate ones, the fix is updating the *old* pages to match, not the new ones.
- **`docs/gitbook/src/images/session-allow-revoke.svg`** + its `docs/diagrams/session-allow-revoke.d2` source are a genuine orphan (zero references anywhere in the repo), pre-dating this PR (from #382). Not introduced here, but the PR description's "orphaned SVGs removed" claim isn't fully accurate as a blanket statement since this one survived.

### One more nit (outside this file's diff hunk, so no inline anchor)

- `docs/agents/fhevm-sdk-1x-migration.md:140` — "We are at `@zama-fhe/sdk@3.1.0-alpha.1`" is a stale pinned version (repo is well past that now). Pre-existing and untouched by this PR, but the new "historical record" framing added at the top of the file makes the staleness more conspicuous by contrast — worth refreshing while the file's already open.

ghermet:
Thanks — worked through the whole follow-up sweep in 86371ccc4 (rebased, now `3802bd17d`):

- **`node-ethers` / `node-viem` WALKTHROUGH.md** — good catch, these were broken on both count. Fixed the delegation snippets to `sdkA.delegations.delegateDecryption(...)` and `sdkA.delegations.isActive(...)` (with the required `contractAddress`), matching `src/index.ts`, and updated the section-4 nav table `isDelegated` → `isActive`. The `tokenB.decryptBalanceAs(...)` line was already correct (that one *is* on `Token`), so left it.
- **"smallest unit" → "base units"** — aligned the 4 pre-existing pages (`useShield`, `useConfidentialTransfer`, `useConfidentialTransferFrom`, `useUnderlyingAllowance`) to "base units", matching the JSDoc and the new sibling pages. Fixed the old pages, not the new ones, as you noted.
- **`session-allow-revoke` orphan** — confirmed zero references repo-wide (predates this branch, from #382) and removed both the `.d2` source and the `.svg`, so the PR's "orphaned SVGs removed" now holds true.
- **`fhevm-sdk-1x-migration.md:140`** — dropped the stale pinned `3.1.0-alpha.1` and reframed to past tense, consistent with the file's "historical record" framing at the top (which already uses "we *were* still in the 3.x prerelease line").

`docs:check-links` → 104 pages clean; typecheck + LLM corpus regen passed in pre-commit.

examples/react-turnkey-wallet/README.md
ghermet: Good catch — restored `next build -- --webpack` in ee67cb78c. Agreed it's the deliberate Turbopack opt-out (Next 16's new default), not an invalid flag, so removing it silently changed which bundler the documented check exercises.

CONTRIBUTING.md
ghermet: Fixed in ee67cb78c. Confirmed `test/test-tanstack-start` has **0** git-tracked files and isn't in `pnpm-workspace.yaml` — just leftover `dist/`+`node_modules/` on disk. Dropped it, and gave `test-components/` a description matching the rest of the block.

docs/diagrams/security-wasm-integrity.d2
ghermet: Fixed in ee67cb78c. Reworked the diagram to the real `wasmAssetLoadMode` surface — `embedded-base64`, `verified-blob`, `precheck-direct-url`, `trusted-direct-url`, with `auto` picking the best — and corrected the matching prose in `security-model.md`. The 'URL-hosted always SHA-verifies or fails' framing was wrong precisely because `trusted-direct-url` skips verification.

docs/gitbook/src/reference/react/useApproveUnderlying.md
ghermet: Reworded in ee67cb78c to your suggestion: "Defaults to a max `uint256` approval. When required — as tokens like USDT demand — it first resets an existing non-zero allowance to zero."

docs/gitbook/src/reference/react/useApproveUnderlying.md
ghermet: Added the backlink in ee67cb78c — `useUnderlyingAllowance.md` now links forward to `useApproveUnderlying` under Related.

docs/agents/fhevm-sdk-1x-migration.md
ghermet: Fixed → "prioritized" in ee67cb78c.

docs/gitbook/src/reference/react/useConfidentialTransferFromAndCall.md
ghermet: Fixed all three in ee67cb78c. `TransactionReceipt` only exposes `logs`, so the snippets now log the already-destructured `txHash` (dropping the unused `receipt` binding) instead of the nonexistent `receipt.blockNumber`.

docs/diagrams/balance-read.d2
ghermet: Fixed in ee67cb78c — the label now reads `requester + contractAddress + encryptedValue`, matching `caching-service.ts`.

ankurdotb (approved):

PR #593, opened 2026-07-28
docs(guides): present Core SDK and React consistently across guides [SDK-278]

ghermet:
## What

Makes the task guides document the framework-agnostic **core SDK** path alongside the **React hooks**, using the repo's existing paired `Core SDK` / `React` tab convention (as already used in `shield-tokens` and `transfer-privately`).

Motivation: a user couldn't find how to encrypt/decrypt inputs with the SDK alone (no React) — `encrypt-decrypt` was React-only. Auditing the rest surfaced an inconsistent presentation: some guides used a lone `SDK` tab, one used no tabs, and delegated-decryption showed no React at all.

## Changes

| Guide | Change |
|---|---|
| `encrypt-decrypt.md` | Add the missing core-SDK path (`sdk.encrypt`, `sdk.decryption.decryptValues` / `decryptPublicValues`) as paired tabs. |
| `check-balances.md` | `SDK` → `Core SDK`; fold standalone React steps into paired tabs (Core-only kept where no clean hook exists). |
| `unshield-tokens.md` | Fold the standalone React step into per-step tabs. Preserves the `#4-handle-interrupted-unshields` heading (linked from `changelog/v3-3.md`). |
| `operator-approvals.md` | Plain code blocks → paired tabs; full React component kept as its own step. |
| `delegated-decryption.md` | Add React tabs (`useDelegateDecryption`, `useDecryptBalanceAs`, `useRevokeDelegation`, `useBatchDecryptBalancesAs`) and a **new "Check delegation status" step** wiring `sdk.delegations.getStatus` / `useDelegationStatus` (referenced in Next steps but never shown). |
| `handle-errors.md` | Pair only where the surface genuinely differs (hook `.error` narrowing, `NoCiphertextError`, React Query `retry`/`retryDelay` via `isRetryable`/`retryAfterSeconds`); unwrap framework-neutral `matchZamaError` blocks to plain code. |

Guides intentionally left alone (don't fit the Core/React split): `nextjs-ssr`, `migrate-v2-to-v3`, `configuration`, `node-js-backend`, `authentication`, `relayer-api-keys`, `local-development`, `web-extensions`.

`llms-full.txt` is the regenerated LLM corpus (auto-updated by the pre-commit hook to stay in sync).

## Verification

- `pnpm docs:check-links` → ✓ 104 pages, all links/anchors resolve.
- No single-`SDK` tabs remain in any guide.
- Every paired hook/method verified against source (`Token`/`WrappedToken` methods, `setOperator(operator, until?)`, delegation hook params, `getStatus`, batch signatures).


ankurdotb (commented):
[AUTOMATED] Ran an independent verification pass against the actual SDK source (not just internal diff consistency) for every new "Core SDK" example and touched error/delegation flow, for Linear SDK-278. Method signatures, param shapes, and the delegation/error API surface check out almost everywhere — the eight inline comments are the real gaps. A few lower-priority items below that don't have a clean single-line anchor:

- **`encrypt-decrypt.md` still doesn't show an actual Node.js path.** The new Core SDK tab is viem-browser-flavored (`walletClient.getAddresses()`) and only links out to `node-js-backend.md`. SDK-278's premise is a backend developer who couldn't find this; after this PR they still have to stitch two pages together, and neither points at the runnable ground truth in `examples/node-viem/` or `examples/node-ethers/`.
- **PR description: "98 pages, all links/anchors resolve".** Re-ran `pnpm docs:check-links` myself: it's actually 104 pages now (likely stale from before the #581 merge added reference pages), but the substantive claim holds — 0 broken links/anchors.
- **`check-balances.md` steps 8 and 9 stayed standalone** even though the PR folds standalone React steps into paired tabs everywhere else — step 8's snippets are now near-duplicates of the new step 1 and step 6 tabs (step 8 "Single token" = step 1 + `refetchInterval`; step 8 "Multiple tokens" is character-identical to step 6 minus one comment). Not incorrect, just redundant now that this PR added the paired tabs those steps used to justify.
- **`check-balances.md:68`** — "no subsequent `balanceOf()` call prompts the wallet" reads broader than reality. `grantPermit`'s own JSDoc says it chunks into groups of ≤10 contracts and prompts once per chunk (`packages/sdk/src/namespaces/permits.ts:53-64`) — true for the 2-token example shown, but stated unconditionally.

Nothing here blocks SDK-278's stated acceptance criteria (checked those separately — all three are literally met), but the `useToken` hook-rules violation, the non-runnable `encrypt-decrypt.md` snippet, and the batch-decrypt error-swallowing would each produce broken behavior for someone who copy-pastes them, which cuts against the ticket's own goal of a working non-React path.

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -147,7 +212,26 @@ const balances = await Token.batchDecryptBalancesAs(tokens, {
...
 ```
 
-### 5. Revoke delegation (optional)
+{% endtab %}
+{% tab title="React" %}
+
+```tsx
+import { useBatchDecryptBalancesAs, useToken } from "@zama-fhe/react-sdk";
+
+const tokens = addresses.map((a) => useToken(a));
```
ankurdotb: [AUTOMATED] `useToken` is a hook (`useZamaSDK()` + `useMemo` internally — `packages/react-sdk/src/token/use-token.ts:23`) — calling it inside `.map()` is a Rules-of-Hooks violation: the hook-call count depends on `addresses.length`, so React throws "Rendered more hooks than during the previous render" the moment that length changes between renders. The Core SDK tab one section up (line 196) and this hook's own reference page (`useBatchDecryptBalancesAs.md`) both use the correct pattern instead — `sdk.createToken` wrapped in `useMemo`, with the reference page explicitly noting "not hooks — hooks cannot be called in a loop." Worth mirroring that here.

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -125,19 +197,12 @@ const tokens = addresses.map((a) => sdk.createToken(a));
 
 const balances = await Token.batchDecryptBalancesAs(tokens, { delegatorAddress: "0xDelegator" });
```
ankurdotb: [AUTOMATED] This call has no `onError` and nothing catches it. Per the real implementation (`packages/sdk/src/token/token.ts:440-446`), without `onError` a single failing token throws `DecryptionFailedError` and discards the whole `results` map — not a partial result. The React tab's equivalent (`await batchDecryptAs(...)`, a few lines down) has the same gap, so it's an unhandled rejection as written. Contrast with `Token.batchBalancesOf`, documented correctly in `check-balances.md` ("partial failure never rejects the whole batch") — these two batch APIs have opposite failure semantics; only one is described accurately here. Consider leading with the `onError` variant, or wrapping this in try/catch.

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -5,14 +5,59 @@ description: How to encrypt values and decrypt FHE encrypted values for custom c
...
 Here is a complete flow that encrypts a value, sends it to a custom FHE contract, reads back the encrypted value, and decrypts it:
 
+{% tabs %}
+{% tab title="Core SDK" %}
+
+{% code title="confidentialRoundTrip.ts" %}
+
+```ts
+const contractAddress = "0xYourContract";
+const [userAddress] = await walletClient.getAddresses();
```
ankurdotb, line 60: [AUTOMATED] This "complete flow" snippet isn't runnable as presented: no imports anywhere in the block, and three undefined identifiers — `sdk` (never constructed), `walletClient` (here — a viem object the Core SDK never exposes; every other guide's Core SDK tab in this PR builds `sdk` via `createConfig`/`new ZamaSDK(...)` instead), and `yourContractABI`. Separately, the `writeContract` call below resolves on broadcast, not inclusion (`packages/sdk/src/viem/viem-signer.ts:104` — a direct passthrough with no receipt wait), so the read-back a few lines later can see a stale/zero handle on a real network; every other write-then-read path in the SDK (`wrapped-token.ts:444`, the repo's own `examples/node-viem/src/index.ts:129`) awaits `waitForTransactionReceipt` first. This is the flagship example for SDK-278 (the whole ticket is about giving non-React devs a working example here), so it's worth making it actually copy-pasteable.

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -124,7 +172,7 @@ export default defineConfig({
 {% endtab %}
 {% endtabs %}
 
-See the [security model](../concepts/security-model.md#coop-coep-headers) for details.
+This applies to browser environments only. In Node.js, encryption is always multi-threaded and needs no headers — see the [Node.js backend guide](./node-js-backend.md). See the [security model](../concepts/security-model.md#coop-coep-headers) for details.
```
ankurdotb: [AUTOMATED] "In Node.js, encryption is always multi-threaded" is stronger than what the rest of the docs support. `packages/sdk/src/node/config.ts:6` describes the Node transport as driving the FHE backend "directly on the calling thread," and `singleThread` is a process-wide `runtime` option (`packages/sdk/src/config/types.ts:65-69`), not browser-only. `concepts/architecture.md:17` hedges correctly ("spawns an internal worker pool for multi-threaded encryption where the environment allows") — this line drops that hedge.

docs/gitbook/src/guides/encrypt-decrypt.md
```diff
@@ -5,14 +5,59 @@ description: How to encrypt values and decrypt FHE encrypted values for custom c
...
+
+// 1. Encrypt
+const { encryptedValues, inputProof } = await sdk.encrypt({
+  values: [{ value: 42n, type: "euint64" }],
+  contractAddress,
+  userAddress,
+});
+
+// 2. Send to contract
+await sdk.signer!.writeContract({
```
ankurdotb: [AUTOMATED] `sdk.signer` is `GenericSigner | undefined` (`packages/sdk/src/zama-sdk.ts:39`). Every SDK-internal call goes through `requireConfigured(...)`, which throws a catchable `SignerNotConfiguredError` instead of asserting — this `!` produces a raw `TypeError: Cannot read properties of undefined` if no signer is configured, which the `matchZamaError`/`instanceof ZamaError` patterns taught in `handle-errors.md` won't catch. Same pattern repeats at lines 93, 324, and 355.

docs/gitbook/src/guides/check-balances.md
```diff
@@ -37,6 +37,21 @@ const balance = await token.balanceOf(address);
...
+```tsx
+import { useConfidentialBalance } from "@zama-fhe/react-sdk";
+import { useAccount } from "wagmi";
+
+const { address } = useAccount();
+const {
+  data: balance,
+  isLoading,
+  error,
+} = useConfidentialBalance({ address: "0xEncryptedERC20", account: address });
```
ankurdotb: [AUTOMATED] This is now the first `useConfidentialBalance` example in the file (moved here from step 8 in `prerelease`), but it's ungated — no `enabled`, no `useHasPermit`. The very next section (line 63) warns "don't trigger this signature on render… gate `useConfidentialBalance` behind `useHasPermit`" — so a reader now copies the anti-pattern before reaching the warning against it. `UseConfidentialBalanceOptions` does expose `enabled` (`packages/react-sdk/src/balance/use-confidential-balance.ts:18-25`), so either gate this snippet or move the warning ahead of it.

docs/gitbook/src/guides/handle-errors.md
```diff
@@ -73,17 +73,34 @@ try {
 }
 ```
 
+{% endtab %}
+{% tab title="React" %}
+
+Hooks surface the same error classes on `.error` — narrow them with the same `instanceof` checks, no `try/catch` needed:
```
ankurdotb, line 79: [AUTOMATED] This is accurate for the `.error` declarative pattern shown right below it, but most of the OTHER new React tabs added by this PR use `mutateAsync` with a bare, uncaught `await` instead — `unshield-tokens.md:48,85-90,121`, `delegated-decryption.md:52,56,93,140,222,249`, `operator-approvals.md:35,59,115`, `encrypt-decrypt.md:86,225,348,590` — none of those catch or read `.error`. Worth a one-line caveat here that the "no try/catch needed" claim is specific to the `.error`-narrowing pattern, not `mutateAsync` usage in general.

docs/gitbook/src/guides/unshield-tokens.md
```diff
@@ -74,87 +101,53 @@ Callbacks are safe to use -- if one throws, the unshield still completes. The ty
 
 ### 3. Unshield your entire balance
 
-If you want to convert all confidential tokens back to public, use `unshieldAll()`. It reads the current encrypted balance and unshields the full amount directly, without decrypting it first.
+If you want to convert all confidential tokens back to public, use `unshieldAll()`. It reads the current encrypted balance and unshields the full amount directly, without decrypting it first. It accepts the same callback options as `unshield()`.
```
ankurdotb: [AUTOMATED] "It accepts the same callback options as `unshield()`" is precise as written but easy to misread as "same options." `unshield(amount, options?: UnshieldOptions)` includes `skipBalanceCheck` (`packages/sdk/src/types/token.ts:50-52`), but `unshieldAll(callbacks?: UnshieldCallbacks)` does not — `UnshieldCallbacks` is the base interface without it. Worth calling out explicitly since step 1 just told the reader about `skipBalanceCheck`.

docs/gitbook/src/guides/delegated-decryption.md
ghermet: Fixed. The React tab now builds Token instances with `useZamaSDK` + `useMemo` + `sdk.createToken` (mirroring the reference page), instead of `useToken` in a `.map()`.

docs/gitbook/src/guides/delegated-decryption.md
ghermet: Fixed. The Core SDK example now leads with the `onError` variant and notes that without it a single failing token rejects the whole call and discards the map. The React tab's `mutateAsync` is now wrapped in try/catch with the same caveat.

docs/gitbook/src/guides/encrypt-decrypt.md
ghermet: Fixed. The Core SDK snippet is now self-contained — imports, `createConfig` + `new ZamaSDK`, and a concrete `yourContractABI` — and it awaits `sdk.provider.waitForTransactionReceipt(txHash)` before reading the handle back, so the read can't race the broadcast. Added a note pointing Node.js backends at `node()` + the runnable `examples/node-viem`/`examples/node-ethers`.

docs/gitbook/src/guides/encrypt-decrypt.md
ghermet: Fixed. Reworded to match `architecture.md`: no COOP/COEP headers in Node, the backend runs on the calling thread and spawns a worker pool where the environment allows, with `runtime.singleThread` as the process-wide opt-out — dropped the unconditional 'always multi-threaded'.

docs/gitbook/src/guides/encrypt-decrypt.md
ghermet: Fixed at all four sites. Replaced `sdk.signer!` with `const signer = sdk.signer; if (!signer) throw new SignerNotConfiguredError("writeContract")`, so the failure is the catchable `ZamaError` subclass the handle-errors patterns expect rather than a raw TypeError.

docs/gitbook/src/guides/check-balances.md
ghermet: Fixed. The first `useConfidentialBalance` example is now gated with `useHasPermit(...)` + `enabled`, so a reader copies the gated pattern rather than the render-time-signature anti-pattern that step 2 warns against.

docs/gitbook/src/guides/handle-errors.md
ghermet: Fixed. Added a hint that 'no try/catch needed' applies to the declarative `.error` field; when you `await mutateAsync(...)` directly, the promise rejects, so wrap it in try/catch or read `.error` instead.

docs/gitbook/src/guides/unshield-tokens.md
ghermet: Fixed. Reworded to 'the same lifecycle callbacks as `unshield()` — but not the `skipBalanceCheck` option from step 1, which is specific to `unshield()`.'

ghermet:
Pushed `28ea8981b` addressing all eight inline comments (replied in-thread) plus the lower-priority items from the review body:

- **Non-runnable `encrypt-decrypt.md` Node path** — the flagship Core SDK snippet is now self-contained and copy-pasteable, and a new hint tells backend devs to swap `web()` → `node()` and points at the runnable `examples/node-viem`/`examples/node-ethers`.
- **`check-balances.md` steps 8/9 redundancy** — folded the duplicated step 8 (which just re-showed step 1 + step 6) into a focused "poll with `refetchInterval`" note that cross-references those steps.
- **`check-balances.md` grantPermit claim** — softened to reflect that it batches ≤10 contracts per prompt, so "single signature" holds only for a ≤10 set.
- **Page count** — updated the PR description 98 → 104 (stale from before the #581 reference-page merge, as you noted); `docs:check-links` still reports 0 broken links/anchors across all 104 pages.

Verified each source-behavior claim against the cited files before editing — they all checked out. Thanks for the thorough pass.

ankurdotb (approved):

PR #598, opened 2026-07-29
chore: pin @fhevm/sdk to 0.13.2

ghermet:
## Summary

Migrates the SDK's FHE backend to `@fhevm/sdk` **0.13.2** and fixes the fallout from that bump.

- **Pin `@fhevm/sdk` to `0.13.2`** and align with its new API.
- **Auth wiring** — seed relayer auth into the global runtime config so the eager encryption-key fetch is authorized (fixes the client-creation `403`).
- **`signDecryptionPermit` → `signLegacyDecryptionPermit`** — rename to match 0.13.2's permit API; API reports regenerated.
- **`await` the now-async relayer serialize methods** — 0.13.2 made `serializeSignedDecryptionPermit` and `serializeTransportKeyPair` async (they now return a `Promise`). `CredentialService` was validating the *unresolved Promise* against `SerializedPermitSchema`, which threw `SigningFailedError` and broke every user-decrypt path (balances stayed masked). Fixed by awaiting the call, and the relayer mocks were made async so they mirror the real contract instead of masking the regression.

## Test plan

- [x] Typecheck passes
- [x] Full unit suite — 2219 passed, no type errors
- [x] test-vite e2e — 60/60 passed
- [x] test-nextjs e2e — 60/60 passed
- [x] Node transport e2e — 8/8 passed


packages/sdk/package.json
```diff
@@ -166,7 +166,7 @@
     "api-report:check": "api-extractor run -c api-extractor.json && api-extractor run -c api-extractor.web.json && api-extractor run -c api-extractor.viem.json && api-extractor run -c api-extractor.ethers.json && api-extractor run -c api-extractor.node.json && api-extractor run -c api-extractor.query.json"
   },
   "dependencies": {
-    "@fhevm/sdk": "1.1.0-alpha.9",
+    "@fhevm/sdk": "^0.13.2",
```
zmalatrax: Keep the verson pin exact
```suggestion
    "@fhevm/sdk": "0.13.2",
```

packages/sdk/src/relayer/fhevm-relayer.ts
zmalatrax: d802bce re-introduces process-global auth, which I don't think is necessary. It also partially reverts `64800509` ("fixing the multichain auth collision"). The global config is single-valued, so a multi-chain setup with different API keys silently keeps only the first one.

Your diagnosis is right, though. The 403 is real, and `_initEncrypt`'s eager key fetch genuinely does read `auth` only from the global config. The underlying cause is that `fetchFheEncryptionKeyBytes` awaits `initPublicAction` *before* applying caller options, so priming via `#encrypt` triggers the unauthenticated fetch first and our `auth` never gets a chance to apply.

Warming through the **base** client avoids that: `_initBase` only resolves the frozen context and never fetches the key, so the authenticated fetch lands first. The cache is first-write-wins keyed by `relayerUrl`, so `#encrypt.init()`'s own fetch becomes a no-op hit, and auth stays per-chain.

```diff
-    this.#encryptInitPromise ??= this.#encrypt
+    this.#encryptInitPromise ??= this.#base
```

Verified locally with the `build.ts` seeding removed: `pnpm test:integration` passes.

Also needed alongside the suggestion: drop the `build.ts` seeding, inline `toFhevmAuth` back into this file and delete `chains/to-fhevm-auth.ts` (its only reason to exist was sharing with `build.ts`), and retarget the 5 prefetch assertions in `fhevm-relayer.test.ts` from `encryptClient` to `baseClient`.

packages/sdk/src/relayer/types.ts
```diff
@@ -226,7 +226,7 @@ export interface RelayerSDK extends Pick<
   | "generateTransportKeyPair"
   | "serializeTransportKeyPair"
   | "serializeSignedDecryptionPermit"
-  | "signDecryptionPermit"
+  | "signLegacyDecryptionPermit"
```
zmalatrax: `signDecryptionPermit` is not gone in 0.13.2. Upstream added `signLegacyDecryptionPermit` next to it and tagged the original `@deprecated`, but it is still present and functional, so this is our rename rather than upstream's.

At 0.13.x the two are behaviorally identical. `signDecryptionPermit` dispatches on compile-time constants (`SDK_PROTOCOL_API_MAJOR_VERSION = 0`, `MINOR = 13`), and the `MINOR <= 13` branch always calls `signDecryptionPermitV1`, which is exactly what the legacy function calls. `SignLegacyDecryptionPermitReturnType` is an alias for `SignedDecryptionPermit`, so the types match too. It also keeps working once testnet moves to v14, since the protocol guarantees backward compatibility and v14 chains still accept V1 permits.

The cost is real, though: `RelayerSDK` is public, exported from `src/index.ts:38` and reachable via `ZamaSDK.relayer`. This drops `signDecryptionPermit` from the `Pick` and deletes the implementation at `fhevm-relayer.ts:342`, so external callers hit `TypeError: relayer.signDecryptionPermit is not a function`.

```suggestion
  | "signDecryptionPermit"
```

The only real difference is deferred. At an `@fhevm/sdk` built for protocol v0.14, `signDecryptionPermit` starts emitting V2 permits instead of V1. I would rather float: that can only change when someone deliberately bumps the dependency, which is a moment we are reviewing anyway. Pinning to the legacy name is a standing commitment to V1 that nobody revisits, and it buys nothing today.

So my suggestion is to revert the rename and keep calling `signDecryptionPermit` internally too. Adopting V2 needs real thought about stored permits, since `SerializedPermitSchema` accepts any positive `version`. That work belongs in the PR that bumps to a 0.14-speaking SDK.

Reverting touches `relayer/types.ts:229`, the method plus three JSDoc references in `fhevm-relayer.ts`, both call sites in `credential-service.ts`, `test-fixtures/relayer.ts`, three test files, and regenerated `sdk.api.md` and `sdk-query.api.md`.

packages/sdk/src/credentials/credential-service.ts
zmalatrax: 0.13.2 adds a `tkmsVersion` tag to transport key pairs. We drop it in two places and never pass it back here.

`generateTransportKeyPair` now returns it as required, `serializeTransportKeyPair` emits it as optional, and `parseTransportKeyPair` accepts it back. But it never survives storage. `StoredTransportKeyPairSchema` (`credentials/schemas.ts:39-44`) does not declare it, and `zod/mini`'s `z.object` strips unknown keys instead of rejecting them, so `#readByKey`'s `safeParse` removes it with no error. Separately, `keypair-vault.ts:138-143` builds `stored` by hand-copying four fields, so `fresh.tkmsVersion` is never read even before the schema sees it. This call then reads back a key pair that has no tag to pass along.

Nothing is broken today, since the ambient TKMS version still matches what the stored keys were generated under. The problem is the first KMS or TKMS rotation. Upstream's `transportKeyPairToTkmsPrivateKey` deserializes the stored private-key bytes using `fhevmContext.tkmsVersion`, the current session value, then `verifyTkmsPublicKey` re-derives the public key and throws `invalid TransportKeyPairKeyPair` on mismatch. That reaches the caller as `Credential signing failed: invalid TransportKeyPairKeyPair` via `wrapSigningError`, and nothing evicts the vault entry, so the only escape is TTL expiry: 30 days by default, up to 365.

Supporting that rotation is why 0.13.2 added the field, so I would rather not ship the bump while discarding it. Five changes, which need to land together to typecheck:

1. `credentials/schemas.ts:39-44`: add `tkmsVersion: z.optional(z.string())` to `StoredTransportKeyPairSchema`. Optional on purpose, since key pairs already in users' storage from the alpha line have no such field and requiring it would invalidate all of them.
2. `credentials/types.ts:25-30`: add `tkmsVersion?: string` to `StoredTransportKeyPair`.
3. `keypair-vault.ts:138-143`: persist `fresh.tkmsVersion` in the `stored` object.
4. This call site and `services/decryption-service.ts:303`: pass `tkmsVersion: keypair.tkmsVersion` into `parseTransportKeyPair`, which already accepts it.
5. Map `invalid TransportKeyPairKeyPair` onto the existing `InvalidTransportKeyPairError` (`errors/credential.ts:12`, currently exported but never constructed) in `wrapSigningError` and `wrapDecryptError`, and evict the vault entry on that path so a mismatch self-heals on the next call.

zmalatrax (changes requested):

packages/sdk/src/relayer/fhevm-relayer.ts
zmalatrax: JSDoc must be updated

zmalatrax: https://github.com/zama-ai/sdk/pull/598/changes#r3682490960

packages/sdk/src/relayer/__tests__/fhevm-relayer.test.ts
```diff
@@ -141,15 +140,19 @@ describe("FhevmRelayer capability initialization", () => {
...
-    expect(relayer.serializeSignedDecryptionPermit({} as never)).toBe("serialized-permit");
+    await expect(relayer.serializeTransportKeyPair({} as never)).resolves.toBe(
+      "serialized-keypair",
+    );
+    await expect(relayer.serializeSignedDecryptionPermit({} as never)).resolves.toBe(
+      "serialized-permit",
+    );
     expect(clients.baseClient.init).not.toHaveBeenCalled();
     expect(clients.decryptClient.init).not.toHaveBeenCalled();
     expect(clients.encryptClient.init).not.toHaveBeenCalled();
```
zmalatrax: baseClient is called  when calling `relayer.serializeSignedDecryptionPermit`
decryptClient is called when calling `relayer.serializeTransportKeyPair`
(both serializeXxx are calling `initPublicAction`), encryptClient is (and should) still be untouched

initPublicAction does not run in these mock tests, so I'd rather simply delete the assertions that does not match the sdk behaviour

packages/sdk/src/test-fixtures/relayer.ts
```diff
@@ -21,7 +21,7 @@ export function createMockRelayer(overrides: Partial<RelayerSDK> = {}): RelayerS
     parseTransportKeyPair: vi.fn().mockImplementation((kp: unknown) => kp),
     serializeTransportKeyPair: vi
       .fn()
-      .mockReturnValue({ publicKey: TEST_PUBLIC_KEY, privateKey: TEST_PRIVATE_KEY }),
+      .mockResolvedValue({ publicKey: TEST_PUBLIC_KEY, privateKey: TEST_PRIVATE_KEY }),
```
zmalatrax: missing `tkmsVersion` in the fixtures, it does not match the actual signatures of fhevm/sdk 0.13.2

Worth defining a new constant TEST_KMS_VERSION and adding to generate and serialize

packages/sdk/src/credentials/__tests__/credential-service.test.ts
zmalatrax, line 24: Related to https://github.com/zama-ai/sdk/pull/598/changes#r3682603248

I'd rather add an assertion that parseTransportKeyPair was called with the tkmsVersion

```typescript
expect(relayer.parseTransportKeyPair).toHaveBeenCalledWith(
  expect.objectContaining({ tkmsVersion: TEST_TKMS_VERSION }),
);
```

zmalatrax: Actually might be better to add a dedicated test 🤔

packages/sdk/src/services/__tests__/decryption-service.test.ts
zmalatrax, line 1: Similar to https://github.com/zama-ai/sdk/pull/598/changes#r3682606186
I'd add a dedicated test to check that tkmsVersion was properly forwarded in parseTransportKeyPair

packages/sdk/src/relayer/fhevm-relayer.ts
zmalatrax: JSDoc to update: Serialization is not local anymore, as serialize calls initPublicAction which awaits on f.ready resolving frozen context, which make on-chain reads (rpc calls)
https://github.com/zama-ai/sdk/pull/598/changes/4757ab53ede334f7937ddab2225f676a248b3bd4..582beace3639b1c232c71b6e6ebe8be9ae7454d5#r3682490960

zmalatrax:
```diff
-   * keys. Reverse of {@link parseTransportKeyPair}. Purely local — no round-trip.
+   * keys. Reverse of {@link parseTransportKeyPair}.
```
https://github.com/zama-ai/sdk/pull/598/changes/4757ab53ede334f7937ddab2225f676a248b3bd4..582beace3639b1c232c71b6e6ebe8be9ae7454d5#r3682793100

zmalatrax:
```diff
-   * {@link parseSignedDecryptionPermit}. Purely local — no round-trip.
+   * {@link parseSignedDecryptionPermit}.
```
https://github.com/zama-ai/sdk/pull/598/changes/4757ab53ede334f7937ddab2225f676a248b3bd4..582beace3639b1c232c71b6e6ebe8be9ae7454d5#r3682799865

packages/sdk/src/relayer/__tests__/fhevm-relayer.test.ts
```diff
@@ -141,15 +140,19 @@ describe("FhevmRelayer capability initialization", () => {
 
     // The compound prefetch-then-init is memoized on the relayer, so two
     // concurrent encrypt calls warm the encryption capability exactly once.
-    expect(clients.encryptClient.fetchFheEncryptionKeyBytes).toHaveBeenCalledOnce();
+    expect(clients.baseClient.fetchFheEncryptionKeyBytes).toHaveBeenCalledOnce();
     expect(clients.encryptClient.init).toHaveBeenCalledOnce();
   });
 
-  test("serialization does not initialize any capability", () => {
+  test("serialization does not initialize any capability", async () => {
```
zmalatrax: Should be renamed considering https://github.com/zama-ai/sdk/pull/598/changes#r3682559866

zmalatrax (approved):
lgtm
Worth flagging that we're not mentioning any breaking change regarding sdk.relayer signature update from sync to async but it's low-level public api that might not actually be used much 🤔

PR #601, opened 2026-07-31
fix(examples): use bare relayer host in proxy default

zmalatrax:
Closes SDK-283

### DIFF
--- a/examples/react-ethers/WALKTHROUGH.md
+++ b/examples/react-ethers/WALKTHROUGH.md
@@ -57,12 +57,14 @@ const config = createConfig({
 `web()` creates the browser relayer backed by a Web Worker loaded from CDN. This example keeps the proxy URL fixed to the local Next.js app:
 
 ```
-Browser Worker → http://localhost:3000/api/relayer/keyurl
+Browser Worker → http://localhost:3000/api/relayer/v2/keyurl
                    ↓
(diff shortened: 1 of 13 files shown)


examples/react-ethers/WALKTHROUGH.md
```diff
@@ -57,13 +57,15 @@ const config = createConfig({
 `web()` creates the browser relayer backed by a Web Worker loaded from CDN. This example keeps the proxy URL fixed to the local Next.js app:
 
 ```
-Browser Worker → http://localhost:3000/api/relayer/keyurl
+Browser Worker → http://localhost:3000/api/relayer/v2/keyurl
                    ↓
-Next.js API route → RELAYER_URL/keyurl  (+ x-api-key header if set)
+Next.js API route → RELAYER_URL/v2/keyurl  (+ x-api-key header if set)
```
ghermet, line 62: Should we let user control the version ?

zmalatrax: I'd rather not, route versions should be abstracted away from the user, letting the fhevm/sdk handle it*
(e.g. `/v1` is not available anymore)
At some point in time, user-decrypt will move to `/v3/user-decrypt` while other routes will still be on `/v2` (/v2/user-decrypt and /v2/delegated-user-decrypt will still work until deprecated*)

ghermet (approved):

PR #612, opened 2026-08-06
docs(gitbook): document offline signing workflows

zmalatrax:
Closes SDK-289


docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+| `SetOperator`                                       | operator approval         | explicit `until` timestamp required                                  |
+| `TransferAndCall`                                   | single-transaction shield | ERC-1363 underlyings only                                            |
+| `ApproveUnderlying` + `Wrap`                        | two-transaction shield    | see the batch warning below                                          |
+| `Unwrap` / `UnwrapAll`                              | unshield phase 1          | `Unwrap` encrypts the amount; `UnwrapAll` reads the on-chain balance |
+| `FinalizeUnwrap`                                    | unshield phase 2          | public decryption happens during `prepare`                           |
+| `DelegateDecryption` / `RevokeDelegation`           | ACL delegation            | explicit expiry, or omit for permanent                               |
+
+## Multi-transaction flows
+
+`WrappedToken.shield()` and `WrappedToken.unshield()` need a live signer, so offline flows compose the same steps from `prepare` primitives. Offline flows are one of the few places where composing below the `Token` API is correct.
```
ankurdotb: I think this is slightly confusingly worded. Are we trying to say here that they should do `approveUnderlying` first and then call the `wrap` function instead?

docs/gitbook/src/changelog/alpha.md
```diff
@@ -11,7 +11,7 @@ description: Unreleased changes on the prerelease (alpha) line — not yet in a
 
 ## Offline signing (institutional custody)
```
ankurdotb: Note to self:

- Drop "institutional custody" from the title (already covered in the description)

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+});
+// sign, broadcast, fetch the receipt, then:
+const event = findUnwrapRequested(receipt.logs);
+if (!event) throw new Error("No UnwrapRequested event in receipt");
+
+const finalize = await sdk.offline.prepare({
+  kind: "FinalizeUnwrap",
+  from,
+  wrapper,
+  unwrapRequestIdOrAmount: event.unwrapRequestId,
```
ghermet, line 140:
```suggestion
  unwrapRequestId: event.unwrapRequestId,
```

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+```ts
+const signedTx = await custody.sign(prepared.unsignedTx);
+const txHash = await publicClient.sendRawTransaction({ serializedTransaction: signedTx });
+```
+
+{% hint style="warning" %}
+**Raw-signature APIs need an extra assembly step.** A raw HSM typically signs the EIP-1559 transaction digest and returns signature components, not a serialized transaction. Use your Ethereum library to compute the signing digest and insert the signature into the transaction envelope before broadcasting; do not treat the raw signature as signed transaction bytes.
+{% endhint %}
+
+Either way, watch the chain yourself: fetch the receipt for the transaction hash through your own provider.
```
ankurdotb: Too brief or sufficient?

docs/gitbook/src/guides/offline.md
zmalatrax: The docs mirrors the public API, the interface of FinalizeUnwrapRequest has the field `unwrapRequestIdOrAmount`:
https://github.com/zama-ai/sdk/blob/d92cf50514df1aed78027f0212299bbcbb221fc0/packages/sdk/src/types/offline-signing.ts#L112-L113
https://github.com/zama-ai/sdk/blob/d92cf50514df1aed78027f0212299bbcbb221fc0/packages/sdk/src/services/offline-service.ts#L268

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+| `DelegateDecryption` / `RevokeDelegation`           | ACL delegation            | explicit expiry, or omit for permanent                               |
+
+## Multi-transaction flows
+
+`WrappedToken.shield()` and `WrappedToken.unshield()` need a live signer, so offline flows compose the same steps from `prepare` primitives. Offline flows are one of the few places where composing below the `Token` API is correct.
+
+**Shield** mirrors the [shielding paths](./shield-tokens.md#shielding-paths): one `TransferAndCall` for ERC-1363 underlyings, otherwise `ApproveUnderlying` then `Wrap`.
+
+{% hint style="warning" %}
+**Dependent transactions must pin `nonce` and `gasLimit`.** Prepared before the first transaction mines, the second one hits two defaults that break: gas estimation reverts (the allowance does not exist yet) and the nonce read returns the same value twice. Pin both, or confirm each transaction before preparing the next. Against USDT-style underlyings with a non-zero allowance, reset it with a first `ApproveUnderlying` of `0n` (a third pinned-nonce transaction).
```
ankurdotb: "Against USDT-style underlyings..."

Are we trying to say this is for tokens with ERC-1363 or similar? Basically, what sort of tokens does this apply to? Otherwise, the directive is a bit vague.

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+
+const finalize = await sdk.offline.prepare({
+  kind: "FinalizeUnwrap",
+  from,
+  wrapper,
+  unwrapRequestIdOrAmount: event.unwrapRequestId,
+});
+```
+
+Neither phase needs an EIP-712 signature on top of the transaction: `Unwrap` encrypts the amount locally and `FinalizeUnwrap` runs the public decryption through the relayer, both during `prepare`.
```
ankurdotb: I think "encrypts the amount locally" might be slightly inaccurate, still contacts a relayer API no?

docs/gitbook/src/reference/sdk/Offline.md
ankurdotb, line 1: Change filename to lowercase or don't bother? Don't know if we're sticking to a convention.

docs/gitbook/src/guides/offline.md
ghermet: right my bad, that probably should be `unwrapRequestId` now there's no requestAmount anymore

ankurdotb (changes requested):

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: I chose based on the existing filenames, 15/20 have a capital letter in `/reference/sdk` (but only 1 in `/reference/react`) ; I'm not opinionated here, we can enforce lowercase, updating other files in a following PR

docs/gitbook/src/guides/offline.md
ankurdotb: Yes — this is meant to say exactly that (`ApproveUnderlying` then `Wrap`, or `TransferAndCall`), but it doesn't say it until two lines down. Suggest making it concrete right here:

```suggestion
`WrappedToken.shield()` and `WrappedToken.unshield()` need a live signer, so offline flows compose the same steps from `prepare` primitives instead — one call for `TransferAndCall`, or two for `ApproveUnderlying` then `Wrap` — mirroring the routing below. Offline flows are one of the few places where composing below the `Token` API is correct.
```

docs/gitbook/src/guides/offline.md
ankurdotb: Too brief, confirmed — it's missing confirmation/reorg guidance, which matters more here than on the atomic path since phase 2 of unshield (`FinalizeUnwrap`) only has a receipt log to go on:

```suggestion
Either way, watch the chain yourself: fetch the receipt for the transaction hash through your own provider, and wait for a few confirmations before acting on it — `waitForTransactionReceipt` returns on first inclusion, and a reorg after that point can invalidate a receipt you've already used (for example, to prepare `FinalizeUnwrap`).
```

docs/gitbook/src/guides/offline.md
ankurdotb: Neither — it's the opposite of ERC-1363. `TransferAndCall` (ERC-1363) never calls `approve`, so the reset only applies to the two-transaction `ApproveUnderlying`+`Wrap` path (non-ERC-1363 underlyings), and only when that underlying also has USDT's quirk of reverting on a non-zero-to-non-zero approve. "USDT-style" is doing double duty (path selector + specific quirk) — worth splitting out:

```suggestion
**Dependent transactions must pin `nonce` and `gasLimit`.** Prepared before the first transaction mines, the second one hits two defaults that break: gas estimation reverts (the allowance does not exist yet) and the nonce read returns the same value twice. Pin both, or confirm each transaction before preparing the next. On the two-transaction path (non-ERC-1363 underlyings), if the underlying also reverts on a non-zero-to-non-zero `approve` (USDT is the classic example), reset the allowance first with an `ApproveUnderlying` of `0n` (a third pinned-nonce transaction).
```

docs/gitbook/src/guides/offline.md
ankurdotb: Agreed, confirmed against source — `Unwrap` calls `encryptValues`, which is a relayer round-trip for the input-proof attestation, not a local-only operation. Same phrase also appears at line 49 (step 2) — flagged there too, worth fixing both in one pass:

```suggestion
Neither phase needs an EIP-712 signature on top of the transaction: `Unwrap` encrypts the amount (a relayer round-trip for the input-proof attestation) and `FinalizeUnwrap` runs the public decryption through the relayer, both during `prepare`.
```

docs/gitbook/src/reference/sdk/Offline.md
ankurdotb: Keep it capitalized — this directory's convention is PascalCase matching the exported name for pages about a single class/namespace (`ZamaSDK.md`, `WrappedToken.md`, `GenericProvider.md`, `Token.md`), vs. kebab-case for topic/utility pages (`errors.md`, `event-decoders.md`, `contract-builders.md`). `Offline` is the `sdk.offline` namespace, so it's already in the right family. No change needed.

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Approval delays
+
+Policy approval can take hours or days. The cryptographic proofs tolerate that, but the transaction still depends on chain state:
+
+- The input proof (`Unwrap`) and the public decryption proof (`FinalizeUnwrap`) embedded in the calldata have no on-chain expiry, and retries are safe: a dropped broadcast can be resubmitted, and a duplicate `FinalizeUnwrap` reverts instead of paying twice.
```
ankurdotb: "Retries are safe" is overstated. Only rebroadcasting the *same signed bytes* is safe to retry blindly; re-preparing `Unwrap` (or any non-`FinalizeUnwrap` kind) with a fresh nonce after what looks like a dropped transaction is a genuinely new transaction — if the original actually landed, that's a real double-debit, not a safe retry.

```suggestion
- The input proof (`Unwrap`) and the public decryption proof (`FinalizeUnwrap`) embedded in the calldata have no on-chain expiry. Only rebroadcasting the *same signed bytes* is safe to retry blindly — a duplicate `FinalizeUnwrap` reverts instead of paying twice. Re-preparing any other kind (including `Unwrap`) with a fresh nonce after what looks like a dropped transaction is a genuinely new transaction, not a safe retry — confirm the original never landed before re-preparing.
```

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+  kind: "ConfidentialTransfer",
+  from: "0xCustodyWallet",
+  token: "0xConfidentialToken",
+  to: "0xRecipient",
+  amount: 1000n,
+});
+// { kind: "ConfidentialTransfer", from: "0xCustodyWallet", unsignedTx: "0x02..." }
+```
+
+For a transfer, the amount is encrypted locally and the attested input proof is already inside the calldata. The result is JSON-safe and crosses a process boundary as-is. `unsignedTx` carries the whole EIP-1559 transaction (chain id, nonce, calldata, gas and fee caps); `from` travels alongside because an unsigned transaction has no sender field and the custodian needs it to pick the signing key.
```
ankurdotb: The encrypted input's proof binds to `from` (the FHE verifier checks it against the transaction sender), so a wallet-ID-to-address mapping error on the custodian's side reverts on submission — after signing and gas spend, not during `prepare`. Worth calling out explicitly since this is exactly the DFNS/Utila integration surface. Also folds in the "encrypted locally" wording flagged elsewhere on this page:

```suggestion
For a transfer, the amount is encrypted (a relayer round-trip for the input-proof attestation) and the attested input proof is already inside the calldata — bound to `from`, since the FHE verifier checks the proof against the transaction sender. If the signing key doesn't actually correspond to `from`, the transaction reverts on submission — after signing and gas spend, not during `prepare` — so get the wallet-ID-to-address mapping right before this step. The result is JSON-safe and crosses a process boundary as-is. `unsignedTx` carries the whole EIP-1559 transaction (chain id, nonce, calldata, gas and fee caps); `from` travels alongside because an unsigned transaction has no sender field and the custodian needs it to pick the signing key.
```

ankurdotb (commented):
One structural note that applies to the whole page, from the `sdk-documentation` skill's conventions (checked against this repo's own precedent, not just the skill's generic rule): `WrappedToken.md` and `ZamaSDK.md` — the two mature reference pages in this same directory — both use one `###` heading per method/param rather than tables. `Offline.md` uses tables for `PrepareOptions`, the request-kind list, and the Throws bullets. Not blocking, but worth aligning with house style in a follow-up so `Offline.md` doesn't read as a one-off.

The rest below is inline. Nothing here changes the verdict from the last pass — the API surface (types, signatures, error classes, kind list) is accurate against source — these are the operational/safety gaps a custody integrator would actually hit.

docs/gitbook/src/reference/sdk/Offline.md
```diff
@@ -0,0 +1,113 @@
...
+`expirationDate` must be at least 1 hour in the future; omit for permanent.
+
+## PrepareOptions
+
+Omitted fields fall back to live chain state:
+
+| Field      | Type                                                     | Default                   |
+| ---------- | -------------------------------------------------------- | ------------------------- |
+| `nonce`    | `number`                                                 | pending transaction count |
+| `gasLimit` | `bigint`                                                 | estimated                 |
```
ankurdotb, line 104: Pinning `gasLimit` doesn't just skip estimation — it's the only pre-flight check the offline path has at all (unlike atomic `shield`/`wrap`, which validate balance and allowance before submitting). The multi-transaction guide explicitly tells readers to pin `gasLimit` for dependent transactions, so it's worth a line in this table's surrounding text recommending an `eth_call` pre-flight when `gasLimit` is pinned, since a bad prepared transaction otherwise only surfaces after custody approval. Didn't attempt a suggestion — this reads better as a sentence after the table than a table-row edit.

docs/gitbook/src/reference/sdk/Offline.md
```diff
@@ -0,0 +1,113 @@
...
+| `fees`     | `{ maxFeePerGas: bigint; maxPriorityFeePerGas: bigint }` | estimated                 |
+
+`fees` carries both values in one object so they can only be pinned as a pair. Dependent transactions prepared before their predecessor mines must pin `nonce` and `gasLimit`; see the [batching warning](../../guides/offline.md#multi-transaction-flows) in the guide.
+
+## Related
+
+- [Offline signing guide](../../guides/offline.md) - the full workflow
+- [GenericProvider](./GenericProvider.md) - `prepareTransaction`, the provider hook behind `prepare`
```
ankurdotb, line 112: `GenericProvider.md` (not touched by this PR) doesn't actually document `prepareTransaction` — its `## Definition` block and custom-provider example both list only `getChainId`/`readContract`/`waitForTransactionReceipt`/`getBlockTimestamp`. It's a required interface member (`packages/sdk/src/types/provider.ts:71`), so anyone following that page to write a custom provider (a likely custody-platform path) ships one that can't satisfy `GenericProvider`. Pre-existing gap, but this PR is what makes the pointer load-bearing — worth a follow-up on `GenericProvider.md` itself rather than here.

docs/gitbook/src/reference/sdk/Offline.md
```diff
@@ -0,0 +1,113 @@
...
+
+Each request is a discriminated union member on `kind`. `from` is always the transaction sender.
+
+### Transfers and operators
+
+- `{ kind: "ConfidentialTransfer"; from; token; to; amount: bigint }`
+- `{ kind: "ConfidentialTransferFrom"; from; token; owner; to; amount: bigint }`
+- `{ kind: "SetOperator"; from; token; operator; until: number }`
+
+Amounts are cleartext; encryption happens during `prepare`. `ConfidentialTransferFrom` debits `owner` and requires `from` to be an approved operator. `SetOperator.until` is a required Unix timestamp: the offline payload is signed later, so no relative default is applied; pass a past value to revoke.
```
ankurdotb, line 72: The source's own docstrings (`offline-signing.ts:50,61`, `schemas/offline.ts:47-48`) say three times that it's a **positive** past value that revokes, not any past value — worth being precise here since `until: 0` passing schema validation but not being a revoke is exactly the kind of gap an integrator won't discover until it's on-chain:

```suggestion
Amounts are cleartext; encryption happens during `prepare`. `ConfidentialTransferFrom` debits `owner` and requires `from` to be an approved operator. `SetOperator.until` is a required Unix timestamp: the offline payload is signed later, so no relative default is applied; pass a **positive** past value to revoke (`0` is not a revoke).
```

docs/gitbook/src/reference/sdk/Offline.md
```diff
@@ -0,0 +1,113 @@
...
+| ------------ | ----------------- | --------------------------------------------------------------- |
+| `kind`       | `TransactionKind` | the request kind, echoed back                                   |
+| `from`       | `Address`         | transaction sender; the custodian picks the signing key with it |
+| `unsignedTx` | `Hex`             | RLP-encoded unsigned EIP-1559 transaction, ready to sign        |
+
+**Throws:**
+
+- `ConfigurationError` - invalid request, or the provider's chain differs from the configured chain
+- `EncryptionFailedError` - encryption produced no result (encrypting kinds)
+- `DecryptionFailedError` - public decryption failed (`FinalizeUnwrap`)
```
ankurdotb: This reads as the error for a failed `FinalizeUnwrap`, but it's actually the fallback among several classified outcomes (`wrapDecryptError` in `offline-service.ts` sorts into `NoCiphertextError`, `RelayerRequestFailedError`, `RpcRateLimitError`, `NotEntitledError`, `InvalidTransportKeyPairError`, or `DelegationNotPropagatedError` first):

```suggestion
- `DecryptionFailedError` - fallback for an unclassified public-decryption failure; a failed `FinalizeUnwrap` more often surfaces as `NoCiphertextError`, `RelayerRequestFailedError`, `RpcRateLimitError`, `NotEntitledError`, `InvalidTransportKeyPairError`, or `DelegationNotPropagatedError`
```

docs/gitbook/src/reference/sdk/Offline.md
```diff
@@ -0,0 +1,113 @@
...
+- `DelegationExpirationTooSoonError` / `DelegationSelfNotAllowedError` / `DelegationDelegateEqualsContractError` - delegation guards (`DelegateDecryption`)
```
ankurdotb, line 59: Missing from this list: `TransactionRevertedError`, thrown by both provider adapters when gas estimation reverts — exactly the scenario the guide itself describes for the two-transaction shield's `Wrap` leg prepared before `ApproveUnderlying` mines. It's the single most likely error on that path:

```suggestion
- `DelegationExpirationTooSoonError` / `DelegationSelfNotAllowedError` / `DelegationDelegateEqualsContractError` - delegation guards (`DelegateDecryption`)
- `TransactionRevertedError` - gas estimation reverted (e.g. the two-transaction shield's `Wrap` leg, prepared before `ApproveUnderlying` mines)
```

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+const unwrap = await sdk.offline.prepare({
+  kind: "Unwrap",
+  from,
+  token: wrapper,
+  to: from,
+  amount,
+});
+// sign, broadcast, fetch the receipt, then:
+const event = findUnwrapRequested(receipt.logs);
+if (!event) throw new Error("No UnwrapRequested event in receipt");
```
ankurdotb, line 133: `unshield-tokens.md` explicitly warns that nothing tracks a self-driven unwrap for you and that `unwrapRequestId` must be persisted by the caller. Worth repeating here — offline is the higher-risk population, given the approval window can run hours to days:

```suggestion
const event = findUnwrapRequested(receipt.logs);
if (!event) throw new Error("No UnwrapRequested event in receipt");
// Persist `event.unwrapRequestId` now: nothing tracks it for you, and approval on
// FinalizeUnwrap can take hours to days (see Approval delays below).
```

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+| `ApproveUnderlying` + `Wrap`                        | two-transaction shield    | see the batch warning below                                          |
+| `Unwrap` / `UnwrapAll`                              | unshield phase 1          | `Unwrap` encrypts the amount; `UnwrapAll` reads the on-chain balance |
+| `FinalizeUnwrap`                                    | unshield phase 2          | public decryption happens during `prepare`                           |
+| `DelegateDecryption` / `RevokeDelegation`           | ACL delegation            | explicit expiry, or omit for permanent                               |
+
+## Multi-transaction flows
+
+`WrappedToken.shield()` and `WrappedToken.unshield()` need a live signer, so offline flows compose the same steps from `prepare` primitives. Offline flows are one of the few places where composing below the `Token` API is correct.
+
+**Shield** mirrors the [shielding paths](./shield-tokens.md#shielding-paths): one `TransferAndCall` for ERC-1363 underlyings, otherwise `ApproveUnderlying` then `Wrap`.
```
ankurdotb: Two gaps: this never says how to determine which path applies (there's a signer-less, provider-only read for it: `wrappedToken.isPayable()`), and there's no `TransferAndCall` example at all in the guide — only in the reference page, where the raw-20-bytes-not-ABI-encoded `recipientData` detail is easy to miss (flagged separately on `Offline.md`). Suggestion covers the first gap; the missing example is probably a bigger addition than fits in a suggestion:

```suggestion
**Shield** mirrors the [shielding paths](./shield-tokens.md#shielding-paths): one `TransferAndCall` for ERC-1363 underlyings, otherwise `ApproveUnderlying` then `Wrap`. Check which one applies with `wrappedToken.isPayable()` — a provider-only read that works without a signer.
```

docs/gitbook/src/reference/sdk/Offline.md
```diff
@@ -0,0 +1,113 @@
...
+
+Amounts are cleartext; encryption happens during `prepare`. `ConfidentialTransferFrom` debits `owner` and requires `from` to be an approved operator. `SetOperator.until` is a required Unix timestamp: the offline payload is signed later, so no relative default is applied; pass a past value to revoke.
+
+### Shield
+
+- `{ kind: "TransferAndCall"; from; underlying; wrapper; amount: bigint; recipientData?: Hex }`
+- `{ kind: "ApproveUnderlying"; from; underlying; spender; amount: bigint }`
+- `{ kind: "Wrap"; from; wrapper; to; amount: bigint }`
+
+`TransferAndCall` is the single-transaction shield for ERC-1363 underlyings; `recipientData` is the recipient as 20 raw bytes, or omitted to self-shield. `ApproveUnderlying` + `Wrap` is the two-transaction path; USDT-style underlyings need an `amount: 0n` reset first when a non-zero allowance exists.
```
ankurdotb: Worth naming the failure mode: a 32-byte ABI-encoded address here won't decode to the intended recipient (the schema accepts raw 20 bytes or `"0x"`, not ABI encoding) — an easy mistake for anyone used to ABI-encoding addresses elsewhere in this API:

```suggestion
`TransferAndCall` is the single-transaction shield for ERC-1363 underlyings; `recipientData` is the recipient as 20 raw bytes (not ABI-encoded — a 32-byte ABI-encoded address will not decode to the intended recipient), or `"0x"`/omitted to self-shield. `ApproveUnderlying` + `Wrap` is the two-transaction path; USDT-style underlyings need an `amount: 0n` reset first when a non-zero allowance exists.
```

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+title: Offline signing
+description: How to build unsigned transactions that are signed and broadcast out-of-process by a custody platform, HSM, or policy engine.
+---
+
+# Offline signing
+
+`sdk.offline.prepare` builds a fully populated unsigned transaction and hands it to you. Signing and broadcasting happen out-of-process: an institutional custody platform, an HSM ceremony, a policy engine with human approval. The preparing process never holds the wallet private key.
+
+{% hint style="info" %}
+**Prefer the atomic API when you can.** If your signer can complete a signature inside one `Promise` (even a slow one that polls a custody API), implement a custom [`BaseSigner`](./node-js-backend.md#8-optional-use-a-custom-signer) and keep the one-call `Token` methods. Reach for `prepare` only when signing genuinely leaves the process.
```
ankurdotb: This links to `node-js-backend.md#8-optional-use-a-custom-signer`, which covers `GenericProvider`/`GenericSigner` but shows no `signTypedData` implementation. The actual `BaseSigner` walkthrough (with a working `signTypedData`) is on `GenericSigner.md#implementing-a-custom-signer`:

```suggestion
**Prefer the atomic API when you can.** If your signer can complete a signature inside one `Promise` (even a slow one that polls a custody API), implement a custom [`BaseSigner`](../reference/sdk/GenericSigner.md#implementing-a-custom-signer) and keep the one-call `Token` methods. Reach for `prepare` only when signing genuinely leaves the process.
```

docs/gitbook/src/guides/offline.md
```diff
@@ -0,0 +1,163 @@
...
+
+Neither phase needs an EIP-712 signature on top of the transaction: `Unwrap` encrypts the amount locally and `FinalizeUnwrap` runs the public decryption through the relayer, both during `prepare`.
+
+## Approval delays
+
+Policy approval can take hours or days. The cryptographic proofs tolerate that, but the transaction still depends on chain state:
+
+- The input proof (`Unwrap`) and the public decryption proof (`FinalizeUnwrap`) embedded in the calldata have no on-chain expiry, and retries are safe: a dropped broadcast can be resubmitted, and a duplicate `FinalizeUnwrap` reverts instead of paying twice.
+- The nonce can become stale if another transaction from the same wallet is mined first. Reserve or otherwise coordinate nonces across concurrent workflows, and re-prepare if the nonce is consumed.
+- Contract state can change while approval is pending, and explicit timestamps such as `SetOperator.until` or a delegation expiry keep advancing. Re-prepare when the transaction's assumptions no longer hold.
```
ankurdotb, line 152: Worth making this concrete for the USDC/USDT custodian case: a deny-listed address mid-approval-window turns an otherwise fully-approved, fully-prepared transaction into a guaranteed revert — a much more likely instance of "contract state can change" than the abstract phrasing suggests:

```suggestion
- Contract state can change while approval is pending, and explicit timestamps such as `SetOperator.until` or a delegation expiry keep advancing. This includes deny-list state on the underlying (e.g. `BlockedUser` on USDC/USDT) — an address blocked mid-approval turns an otherwise-valid prepared transaction into a guaranteed revert. Re-prepare when the transaction's assumptions no longer hold.
```

docs/gitbook/src/guides/offline.md
zmalatrax: i.e. the offline API should be refactored to `unwrapRequestId` instead of `unwrapRequestIdOrAmount` ?

I'd rather keep it as is for now (aligning docs) and then update it in a following PR, wdyt?

docs/gitbook/src/guides/offline.md
zmalatrax: Indeed it's poorly worded
What I was trying to say is that for offline shielding (resp. unshielding), there isn't a single function like `.shield()` (resp. `.unshield()`) that streamlines `approve()` and `wrap()` into a single call. So offline shielding requires the sdk user to call approve, and then wrap  (mentioned as primitives here)

Will reword

docs/gitbook/src/guides/offline.md
zmalatrax: we could add a code snippet to fetch the receipt indeed

docs/gitbook/src/guides/offline.md
zmalatrax: It's unrelated to ERC-1363, it's specific to USDT, which has an unusual constraint of reverting if you try to update your allowance while already having a non-zero allowance

It might too specific to include in the guide, and can definitely be reworded if we want to keep it

docs/gitbook/src/guides/offline.md
zmalatrax: mmh yes, I was thinking about the input proof generation client-side but the sdk communicates with the relayer before/after doing so anyway

will reword

docs/gitbook/src/guides/offline.md
zmalatrax: it feel a bit too specific (operational description), I'd rather keep the abstract "contract state can change", as it includes deny-list from tokens 🤔

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: until: 0 is accepted and revokes the operator on ERC-7984 (`block.timestamp <= until`)

docs/gitbook/src/guides/offline.md
ankurdotb: I think it's worth including, but call it out specifically as USDT-specific rather than "USDT-style". I was confused by what is covered under "USDT-style"

docs/gitbook/src/guides/offline.md
ankurdotb: Okay, acknowledge and agreed

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: for transferAndCall and approve it's valid, but for wrap, unless a proper allowance is already on-chain, eth_call preflight would fail anyway (as no allowance)

It's worth adding a sentence that pinning `gasLimit` skips estimation and callers should validate their assumptions

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/guides/offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/guides/offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/reference/sdk/Offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/guides/offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/guides/offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

docs/gitbook/src/guides/offline.md
zmalatrax: Resolved in 6414ff65a1a99234f5b68e3e32874bc956683690

ankurdotb (approved):

PR #572, opened 2026-07-16
feat(sdk): add DelegationService.getStatus

zama-cremaud:
## Summary

Addresses SDK-195 ([Linear](https://linear.app/zama/issue/SDK-195/make-query-modules-thin-adapters-over-token-methods-to-remove)), child of epic SDK-192.

**Ticket premise re-verified against current `prerelease`, mostly stale.** The ticket asked to make `query/*.ts` modules thin adapters over `Token`/`WrappedToken` methods instead of re-implementing orchestration. I read all ~30 `query/` modules, including all 8 the ticket named as highest-risk (`encrypt.ts`, `user-decrypt.ts`, `public-decrypt.ts`, `shield.ts`, `unshield.ts`/`resume-unshield.ts`/`finalize-unwrap.ts`, `transfer.ts`/`transfer-from.ts`) — **every one of them is already a one-line closure** delegating to a `Token`/`WrappedToken`/`ZamaSDK` method. This was already delivered by prior work (SDK-96 Token/WrappedToken split, SDK-197 pending-unshield fold, and historically SDK-27) before this ticket was ever written.

**One genuine surviving instance, not named by the ticket:** `query/delegation-status.ts` reimplemented the `isActive`-from-expiry derivation (`0n` → false, `MAX_UINT64` → true, else compare to chain time) instead of calling `DelegationService` — the exact "two places to check for the same bug" pattern the ticket warns about, just in `Delegations`, not `Token`.

## Changes

- `DelegationService.getStatus()` — resolves `{ isActive, expiryTimestamp }` from a single `getDelegationExpiry` read; `isDelegated()` now calls it internally (dedup).
- `Delegations.getStatus()` — public namespace method delegating to the service.
- New public `DelegationStatus` type, exported from `index.ts`.
- `query/delegation-status.ts` — `queryFn` now calls `sdk.delegations.getStatus(...)` instead of re-deriving `isActive` from a raw contract read. `DelegationStatusData` (the public `/query` subpath type `useDelegationStatus` imports) is kept as a re-exported alias — **no change needed in `react-sdk`**.
- `docs/gitbook/src/reference/sdk/delegation.md` — documents the new method.
- Tests added for `DelegationService.getStatus` and `Delegations.getStatus`; existing `query/delegation-status.ts` tests pass unchanged (same external behavior, same provider calls).

## Test plan

- [x] `pnpm typecheck && pnpm lint && pnpm format:check`
- [x] `pnpm test:run` — 1505 tests / 148 files, all passing (4 new)
[...]

PR #613, opened 2026-08-07
docs(gitbook): backfill 3.4.0 changelog with 4 missing features

ankurdotb:
## Summary

3.4.0's GitBook changelog page (`docs/gitbook/src/changelog/v3-4.md`) only covered the `@fhevm/sdk` backend migration and one bug fix. Six other things that shipped in `3.4.0` per `CHANGELOG.md` never got a plain-language write-up anywhere in `docs/gitbook/src/changelog/`:

- **credentials:** opt-in shared ML-KEM keypair scope for shared-tenant operators (SDK-142, #546)
- **sdk:** add `DelegationService.getStatus` (#572)
- **sdk:** chunk decryption requests to respect the relayer's per-request bit budget (#541)
- **sdk:** expose `wrap()` + `useWrap` for two-signature shield (#522)
- **sdk:** uniform retryability signal across decryption error causes, `isRetryable()`/`retryAfterSeconds()` (SDK-248, #536)
- **sdk:** two bug fixes — `ethers`/`viem` integer normalization (#580) and the unshield unbundle exposing `UnwrapResult` (#530)

This backfills all of them as new `###` sections / bullets on the existing `## 3.4.0` entry. Note `pnpm docs:changelog --check` does not flag this kind of gap — its scaffolder only detects a *missing* `## {version}` section, not a partially-documented one — so this was manually drafted following the `sdk-changelog` skill's conventions (bold-lead prose, name APIs exactly, state breaking posture, cross-check every claim against source) rather than via the automated scaffold path.

Each section was independently researched and cross-checked against current source (not just the one-line `CHANGELOG.md` description):

- **Shared transport key pair scope** — verified `transportKeyPairScope` on `createConfig`, `sdk.permits.warmTransportKeyPairScope`/`revokeTransportKeyPair`, and that omitting the option preserves the exact prior per-signer behavior (no breaking change).
- **`DelegationService.getStatus`** — verified the `{ isActive, expiryTimestamp }` shape, that it's a genuine addition (not a replacement) alongside pre-existing `isActive()`/`getExpiry()`, and the three `expiryTimestamp` sentinel states. Also caught that this same PR did a type-only rename (`DelegationStatusData` → `DelegationStatus`) on the `@zama-fhe/sdk/query` subpath with no compat alias and no `CHANGELOG.md` `BREAKING CHANGES` entry — mentioned inline since it's real and undisclosed, though low-blast-radius (type-only, undocumented type, zero prior doc references).
- **Chunked decryption** — verified the 2048-bit KMS gateway cap, the chunking mechanism in `chunkHandlesByBitBudget`, exactly which call paths are affected (`decryptValues`, `delegatedDecryptValues`, `delegatedBatchDecryptValues`, `Token.batchDecryptBalancesAs` and their hooks) vs. unaffected (`decryptPublicValues`), and confirmed there's genuinely nothing new to configure.
- **`wrap()`/`useWrap`** — verified the two-signature framing (pairs with existing `approveUnderlying()`), confirmed `shield()` itself is unchanged, and framed it explicitly as an escape hatch per this repo's "use the SDK method, don't recompose" convention rather than a `shield()` replacement.
- **Retryability signal** (added per [review feedback](https://github.com/zama-ai/sdk/pull/613#issuecomment-5215239113)) — verified `ZamaError.retryable`, `isRetryable()`, `retryAfterSeconds()` are all exported, build on the typed-causes work already documented in `v3-3.md`, and are already covered in the `handle-errors.md` guide (just missing from the changelog announcement).
- **Two bug fixes** (same review pass) — `#580` is a genuine user-visible fix (`Token.decimals()` returned `bigint` under `ethers`, `number` under `viem`); `#530` preserves the `unshield`/`useUnshield` public signature but exports a new `UnwrapResult` type on the `unwrap`/`unwrapAll` escape hatches.

**Also fixed two reference-doc gaps found while checking whether these features were documented anywhere besides the changelog:**

- `docs/gitbook/src/guides/decrypt-from-event-logs.md` claimed `decryptValues` "issues one decryption request per contract" — no longer accurate now that an oversized group chunks into multiple relayer requests. Corrected and cross-linked to the new changelog section.
- `docs/gitbook/src/reference/sdk/ZamaSDK.md`'s method reference was missing `permits.warmTransportKeyPair`, `permits.warmTransportKeyPairScope`, and `permits.revokeTransportKeyPair` entirely (the first predates 3.4.0; the latter two are the shared-scope feature this PR documents). Added all three alongside the rest of the `permits.*` methods.

All cross-reference anchors (`security-model.md`, `permit-model.md`, `shield-tokens.md`, `delegated-decryption.md`, `handle-errors.md`, the React reference pages) were checked against the actual heading text before use.

## Test plan

- [x] `pnpm exec oxfmt --check` on every changed file — clean
- [x] `node scripts/docs/check-links.mjs` — 107 pages checked, all internal links and anchors resolve
- [x] `node scripts/llm/build-llms.mjs` — corpus regenerated, `llms-full.txt` diff included in this PR
- [x] `node scripts/llm/check.mjs` — passes against the committed corpus in this PR (verified pre-commit that each regen matched exactly what was staged)
- [x] Every API name, signature, and behavioral claim cross-checked against current `packages/sdk`/`packages/react-sdk` source, not just `CHANGELOG.md`


zmalatrax:
Ran a quick review agent:
> The 3.4.0 changelog still appears to omit one shipped feature: #536 added the public retryability APIs `ZamaError.retryable`, `isRetryable()`, and `retryAfterSeconds()`.
>
> Since this PR aims to backfill the features missing from the GitBook release notes, should we add a section for it as well? The other remaining entries, #530 and #580, are narrower bug fixes and seem more like an editorial choice.

ankurdotb:
Good catch — #536 is a real gap, and it's a strong one: adds `readonly retryable` on `ZamaError` plus `isRetryable()`/`retryAfterSeconds()`, all exported, and already documented in the `handle-errors.md` guide's "Retry transient failures" step, just never announced on this page. Added as its own section, linked back to the 3.3.0 typed-causes work it builds on.

Went ahead and pulled in #530 and #580 too rather than leaving them editorial — checked both against source: #530 preserves the `unshield`/`useUnshield` public signature but does export a new `UnwrapResult` type on the `unwrap`/`unwrapAll` escape hatches (already on the reference pages per that PR, just missing here), and #580 is a genuine user-visible fix (`Token.decimals()` returned `bigint` under `ethers`, `number` under `viem`). Added both as bullets under Bug fixes — the 3.3.0 page lists its bug fixes exhaustively, so this brings 3.4.0 in line with that.

Pushed in 136d2b1.

docs/gitbook/src/changelog/v3-4.md
```diff
@@ -86,9 +86,108 @@ web({
...
+});
+
+const sdk = new ZamaSDK(config);
+
+// Create the scope's key pair once, before opening concurrent traffic to it.
+await sdk.permits.warmTransportKeyPairScope("tenant-123");
+```
+
+{% hint style="warning" %}
+**Sharing only works if every signer in the scope reads and writes the same storage instance.** The Node `asyncLocalStorage` default isolates a fresh store per request, which would silently regenerate and discard the "shared" key pair on every request.
```
zmalatrax: > the Node `asyncLocalStorage` default isolates a fresh store per request

Isn't the Node default per-instance `MemoryStorage`?

ankurdotb: Good catch — you're right, and it's a real error, not just imprecise wording. `resolveStorage()`'s actual code-level fallback on Node (`packages/sdk/src/config/resolve.ts`) is `new MemoryStorage()`, not `asyncLocalStorage` — that one's an opt-in export from `@zama-fhe/sdk/node` a caller has to explicitly wire in.

What's true, and what I was actually reaching for: `configuration.md`'s own storage table recommends `asyncLocalStorage` for "Node.js servers" specifically because it isolates the transport key pair per request — which is exactly the behavior that defeats a shared `transportKeyPairScope`. So the risk is real, but "the Node default" was the wrong label for it. Reworded to "the storage recommended for Node.js servers" in both this new hint and the pre-existing `security-model.md` hint I'd copied the phrasing from (it had the same ambiguity). Pushed in 9f2df9f.

zmalatrax (approved):
lgtm, ty!

PR #126, opened 2026-03-19
feat(example-hoodi): add on-chain ACL delegation flows (SDK-41)

zama-cremaud:
## Summary

  Extends the `example-hoodi` quickstart app with the three SDK delegation
  operations, demonstrating the full on-chain ACL delegation lifecycle using
  `@zama-fhe/react-sdk` hooks.

  ### New components

  - **`DelegateDecryptionCard`** — grants decryption access to a delegate
    address via `useDelegateDecryption`. Supports permanent delegation
    (default) or a datetime-bound expiry via a datetime picker. Form resets
    on success.
  - **`RevokeDelegationCard`** — revokes a previously granted delegation via
    `useRevokeDelegation`. Address validated before enabling the button.
  - **`DecryptAsCard`** — decrypts another wallet's confidential balance as a
    delegate via `useDecryptBalanceAs`, with a live `useDelegationStatus`
    indicator (fires as soon as a valid owner address is entered) and typed
    error handling for `DelegationNotFoundError` / `DelegationExpiredError`.

  ### Wiring (page.tsx)

  Cards are split across two visual sections with `<hr>` separators and JSX
  comments:
  - **Owner perspective** — Grant Decryption Access + Revoke Decryption Access
  - **Delegate perspective** — Decrypt Balance On Behalf Of

  All `key` props include `address` and `selectedToken` so cards remount
  (inputs + state reset) on wallet or token change, consistent with the
  existing core operation cards.

  ### Tests

  `e2e/main.spec.ts`: 18 → 19 tests
  - "renders all operation cards" extended to include the 3 delegation card titles
  - "action buttons are disabled before metadata loads" extended to include
    "Decrypt Balance" (disabled until `cTokenMetadata` resolves)
  - New: "delegation buttons are disabled when no address is entered"

  ### Documentation

  - `README.md`: operations table 4 → 7 rows, test count 18 → 19
  - `WALKTHROUGH.md`:
    - Supported operations table: 4 → 7 rows (with Source file column)
    - New **Delegation walkthrough** section (Steps 9–11): grant, decrypt-as,
      revoke — including a note on the revocation cache behaviour
    - New **Delegation hooks** code block in SDK integration details
    - Troubleshooting: entry documenting why a delegate can still decrypt
      after revocation (handle-keyed IndexedDB cache, by design)
    - Tech stack: delegation hooks added to `@zama-fhe/react-sdk` row

  ## Base branch

  Built on top of `feature/sdk-33-create-example-hoodi-example-app-using-ethers`
[...]

PR #623, opened 2026-08-12
ci: verify Aikido Safe Chain installer instead of piping curl to sh

ankurdotb:
Replaces the `curl … | sh` pipe-to-shell that installs Aikido Safe Chain in the release workflows with the security team's **pinned + download-then-verify** pattern.

## Alerts

Wiz IaC **#126 / #127 / #128** — *"workflow should not execute suspicious remote payloads"* — at `release.yml` (publish-from-tag + release jobs) and `release-preview.yml`.

## Before

```sh
curl -fsSL .../releases/latest/download/install-safe-chain.sh | sh -s -- --ci
```

No integrity check; pipes an unpinned remote script straight into a shell — in a repo that otherwise SHA-pins its Actions and enforces a 72h package-age floor.

## After (Aikido Safe-Chain user guide, GitHub CI section)

The pinned version lives in one workflow-level env var per file:

```yaml
env:
  # Pinned safe-chain release used by the verified installer below; bump as newer versions are vetted.
  SAFE_CHAIN_VERSION: "1.5.15"
```

Each step resolves the installer asset URL **and its GitHub-published SHA-256 digest** from the releases API for that pinned tag, downloads the script, **verifies the hash**, then executes:

```sh
SAFECHAIN_URL="https://api.github.com/repos/AikidoSec/safe-chain/releases/tags/${SAFE_CHAIN_VERSION}"
SC_INSTALL_URL=$(curl -fsSL "$SAFECHAIN_URL"  | jq -r '.assets[] | select(.name=="install-safe-chain.sh") | .browser_download_url')
SC_INSTALL_HASH=$(curl -fsSL "$SAFECHAIN_URL" | jq -r '.assets[] | select(.name=="install-safe-chain.sh") | .digest[7:]')
curl -fsSL "$SC_INSTALL_URL" -o install-safe-chain.sh
[ "$(sha256sum install-safe-chain.sh | awk '{print $1}')" = "$SC_INSTALL_HASH" ] \
  && echo "=> SHA256 install script matched" \
  || { echo "=> SHA256 mismatch — aborting"; rm -f install-safe-chain.sh; exit 1; }
bash ./install-safe-chain.sh --ci
```

- **Pinned** to `releases/tags/${SAFE_CHAIN_VERSION}` (currently `1.5.15`, the latest), matching the Notion doc's pinned-tag structure. Bumping the version is a one-line change per file.
- `curl` / `jq` / `sha256sum` are preinstalled on `ubuntu-latest`; steps already run under `bash -euo pipefail` (so the step `shell:` is left at the repo default rather than downgraded to plain `bash`), meaning any failed fetch/verify aborts the install.

## Verification

- Ran the exact block against real GitHub for `releases/tags/1.5.15`: resolves the asset, hash `de0565…` **matches**, downloaded script is valid bash (**not** executed), and a **tampered hash is correctly rejected**.
- Simulated the workflow env var (`SAFE_CHAIN_VERSION=1.5.15`) → URL builds and the API resolves `tag_name: 1.5.15`.
- YAML parses; `actionlint`/shellcheck clean on the new blocks; oxfmt + typecheck green.

> Security team owns #126–128 — this implements the Aikido user-guide pattern for their review. Wiz IaC only scans `main`, so clearance appears after this reaches `main` and the daily scan re-runs.

### DIFF
--- a/.github/workflows/release-preview.yml
+++ b/.github/workflows/release-preview.yml
@@ -19,6 +19,8 @@ env:
   # pnpm-workspace.yaml — mirror both the age (72 h) and the exclusion here.
   SAFE_CHAIN_MINIMUM_PACKAGE_AGE_HOURS: "72"
   SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS: "@fhevm/sdk,@zama-fhe/sdk,@zama-fhe/react-sdk"
+  # Pinned safe-chain release used by the verified installer below; bump as newer versions are vetted.
+  SAFE_CHAIN_VERSION: "1.5.15"
 
 jobs:
   validate-branch:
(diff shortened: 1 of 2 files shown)


zmalatrax (approved):
lgtm

PR #633, opened 2026-08-18
feat: revoked kms context recovery

zmalatrax:
Closes SDK-294

`userDecrypt` fails when the chain revokes the KMS context of a permit. The contract reverts with `InvalidKmsContext`. The SDK finds this revert by its selector in the raw revert data. The SDK then repairs itself. It removes the dead permits. It grants a new permit under the current context. It tries the decrypt one more time. If this also fails, it throws a typed `RevokedKmsContextError`.

**Design choice: the re-grant repairs the scope, not the call.** A revoked context kills all permits in the scope. Thus the recovery removes all permits in the scope. It signs one new permit that covers the old permits plus the requested contracts. This costs one wallet prompt. Parallel decrypt calls share one recovery per scope. They wait for the first recovery and find their contracts covered. Thus a revocation costs the user one signature, not one for each call. The delegated retry loop shares the same recovery budget. Thus retries cannot start a second recovery.

**Why we retry only one time.** `@fhevm/sdk` keeps the on-chain validity result in a cache for 15 minutes. This value is hardcoded. The SDK cannot force a refresh. Thus the retry can fail with the same error in this window. The recovery does not loop. It removes, re-grants, retries one time, and then throws. The error text tells the user about this window. We will change this when upstream gives cache control.


packages/sdk/src/credentials/permission-store.ts
```diff
@@ -118,16 +119,20 @@ export class PermissionStore {
...
-      await this.#untrackScope(scope);
+      await this.clearScope(scope);
     } else {
-      await this.#storage.set(key, next);
+      await this.#storage.set(permissionScopeKey(scope), next);
     }
   }
 
+  /** Delete every permit in the scope and untrack it from the per-signer index. */
+  async clearScope(scope: PermissionScope): Promise<void> {
```
eitjuh, line 131: Nit (also not specific to this PR):
clearScope() should probably be serialized with permit writes. Otherwise a concurrent append() can leave fresh scope data missing from the signer index, causing clearAllForSigner() to miss it. A signer-level lock would protect both the scope entry and shared index.

Now, you can basically have this race condition:

Clear scope:
- (clear scope 1): delete scope data
- (concurrent grant)
  - write fresh permit
  - read index, scope already listed
  - skip tracking
- (clear scope 2): remove scope from index

packages/sdk/src/services/decryption-service.ts
```diff
@@ -146,22 +145,27 @@ export class DecryptionService {
   ): Promise<Record<EncryptedValue, ClearValue>> {
     const normalizedDelegator = getAddress(delegatorAddress);
     const normalizedDelegate = getAddress(delegateAddress);
+    // One evict-and-regrant across the whole propagation-retry loop, not one
+    // per attempt: each retry would otherwise re-arm the recovery and spend a
+    // fresh wallet prompt per 2s attempt while the upstream validity cache
+    // keeps serving the revoked context.
+    const recovery: RecoveryBudget = { spent: false };
```
eitjuh, line 153: This state is recreated every time delegatedDecryptValues() is called. Since the batch fallback calls this method again per item, each item gets another recovery attempt and potential wallet prompt. Is the idea for the batch to share one recovery state across all these calls? Not sure if this is deliberate or not

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -448,4 +449,62 @@ export class CredentialService {
...
+    delegator?: Address,
+  ): Promise<SerializedTransportKeyPairWithPermissions> {
+    const signer = this.#requireSigner("recoverPermits");
+    const account = signer.requireWalletAccount("recoverPermits");
+    const scope = this.#permissionScope(checksum(account.address), delegator);
+    const key = permissionScopeKey(scope);
+    let recovery = this.#permitRecoveries.get(key);
+    if (!recovery) {
+      recovery = this.#evictAndRegrantScope(scope, contracts, delegator).finally(() => {
+        this.#permitRecoveries.delete(key);
```
eitjuh: probably an edge case, and maybe i'm missing something, but doesn't this delete mean we only de-dupe recoveries that overlap in time?

if decrypt A recovers and we wipe this map entry, then decrypt B (which started with the same dead permit, but whose relayer call is still in flight) comes back with InvalidKmsContext after this, we'll treat it as a brand new revocation. that would clear the permit A just signed and prompt the user again.

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -448,4 +449,62 @@ export class CredentialService {
...
+  async #evictAndRegrantScope(
+    scope: PermissionScope,
+    contracts: readonly Address[],
+    delegator?: Address,
+  ): Promise<void> {
+    // Re-sign the union of what the scope covered, not just this call's
+    // contracts: one signature restores every widened permit, so sibling
+    // decrypt calls joining the recovery find their contracts already covered.
+    let prior: ChecksummedAddress[] = [];
+    await swallow(
```
eitjuh: Swallow seems a bit destructive here? If #store.list throws, prior stays [] and we still clearScope() then re-grant only this call’s contracts. list can also clearScope itself on a parse failure, so a failed/empty read plus the wipe below looks like it can drop other contracts the scope already covered?

packages/sdk/src/credentials/permission-store.ts
zmalatrax: indeed, it's worth refactoring in a follow-up PR!
Besides clearScope/grant race, there is no serialization of PermissionStore mutations, for 2 concurrent grantPermit, user will have to sign twice, but only the last write is persisted

I agree on the design with a mutex per signer; In browsers we could also use web locks api to avoid a cross-tab race 🤔

packages/sdk/src/services/decryption-service.ts
zmalatrax: The per-item fallback of the batch decryption is guarded by `isFatalBatchError`: https://github.com/zama-ai/sdk/blob/1bfc41179d9cde363defcadc703567d35cc9cb24/packages/sdk/src/services/decryption-service.ts#L245-L248
https://github.com/zama-ai/sdk/blob/1bfc41179d9cde363defcadc703567d35cc9cb24/packages/sdk/src/errors/fatal-batch.ts#L30
So if a batch decryption request hits the `RevokedKmsContextError` error, it will try recovery (i.e. 1 wallet prompt), then retry the batch decryption request (the unresolved requests), if it fails again with a fatal batch error (either `RevokedKmsContextError` or another one), it will abort the batch and skip the per-item fallback. If it fails with a non-fatal error, the per-item fallback runs as usual

zmalatrax: Though, looking at other batch decryption functions, notably on `Token`,  we have `Token.batchDecryptBalancesAs` that relies on `sdk.decryption.delegatedBatchDecryptValues` so all good. `Token.batchBalancesOf` however loops over `Token.balanceOf` per token (concurrency of 5);  per-token errors are swallowed and `isFatalBatchError` is only checked after all requests have been processed --> might trigger multiple wallet prompts on revoked kms context
We can add the same short-circuit logic to it, as a quick fix in this PR

But there is an architectural difference between delegated and direct decryption of token balances. I'm starting to think that both should rely on functions from `DecryptionService`: refactoring `Token.batchBalancesOf` by introducing a function `batchDecryptValues` (similar to `delegatedBatchDecryptValues`)

packages/sdk/src/credentials/credential-service.ts
zmalatrax: Indeed it is temporal-only: if A re-signed a new permit (with a valid kms context), then decrypt B (still with the old permit) will trigger a new recovery (another wallet prompt to sign a new permit) because there are no other recovery running at the same time..

Note that there shouldn't be any coverage lost though, B recovery reads the prior coverage before clearing, so it signs the union of pre-existing A's permit and what B needs

But in any case, it potentially triggers one redundant wallet prompt, so it needs to be refactored. 
Re-using your example, if decrypt A fails and recover by signing a new permit and then decrypt B, that still had the now-cleared dead permit, fails as well, we could check whether the permit's signature used by B still exist or not to re-sign a permit:
- still there: genuine new revocation, evict and re-sign
- gone: it was already replaced (here by A), skip the recovery and retry with the current permit

packages/sdk/src/credentials/credential-service.ts
zmalatrax: Indeed, would be better to bundle read/clearScope in the same swallow, if read fails we return the pre-existing (dead) permits

Not sure to follow on the list part, if the permit schema cannot be parsed, then it returns an empty array []
The root cause being that the stored permits are unparseable, the coverage is void anyway, no?
https://github.com/zama-ai/sdk/blob/1bfc41179d9cde363defcadc703567d35cc9cb24/packages/sdk/src/credentials/permission-store.ts#L36-L47

packages/sdk/src/services/decryption-service.ts
zmalatrax: This architectural difference refactor is actually already tracked in https://linear.app/zama/issue/SDK-167/refactor-symmetric-batch-decryption-in-decryptionservice-unify-token

eitjuh (dismissed):

zmalatrax:
A few changes made to adapt revoked kms context recovery to offline permit that'll land with 634, in https://github.com/zama-ai/sdk/commit/52ddb7b9a6973eacf88893fd303d45407b982678:
- Surface the error `RevokedKmsContextError` when re-grant signing failed, wiith the signing failure as `cause` (would happen in the context of offline permit, recovery self-heal cannot sign on its own)

- Another issue, the recovery was first wiping all permits in the scope, leaving the scope empty if re-grant failed:
> **Recovery no longer destroys the scope on failure (new).** The self-heal cleared the whole scope before re-granting. In an offline-permit session the re-grant always fails, so one revoked permit would wipe every stored permit in the scope, including valid ones the session cannot re-sign, and the user would have to redo the offline ceremony for all of them. Now, when the re-grant throws (any failure, rejection included), the permits not involved in the failed decrypt are restored. Only the dead permit stays evicted.

I wanted to keep the refactor minimal, so I did restore-on-failure, rather than "make-before-break" (granting the new permit, then destroying the old scope) because the latter required more involved structural changes

eitjuh (dismissed):

eitjuh (approved):

PR #634, opened 2026-08-18
feat: offline permit flow

eitjuh:
# Summary

Adds the offline decryption-permit flow (SDK-295): `sdk.offline.preparePermit` builds the unsigned EIP-712 typed data for a permit without signing it, and `sdk.permits.registerPermit` verifies an out-of-process signer's signature and persists the permit. Closes the gap left by the existing offline-transaction `prepare` pipeline, which covers transactions but not decryption permits — needed for custody partners (HSM, policy engines, out-of-process signers) that cannot sign in-process.

`registerPermit` cross-checks every unsigned field on the prepared payload (`chainId`, `contracts`, `startTimestamp`, `durationDays`, `delegatorAddress`) against what's actually embedded in the signature-verified EIP-712 payload, rejecting any mismatch as `PreparedPermitMismatchError` — these fields travel alongside the signature but aren't themselves covered by it. `preparePermit` fails fast (`ConfigurationError`) on self-delegation and durations over the V1 protocol's 365-day cap, and uses strict transport-key-pair persistence so an unpersisted key can never silently strand an out-of-process signing ceremony.

Adds matching React hooks (`usePreparePermit`, `useRegisterPermit`), typed errors (`TransportKeyPairChangedError`, `PreparedPermitChainMismatchError`, `PreparedPermitExpiredError`, `PreparedPermitMismatchError`), and full documentation (guide section, SDK/React reference pages, error reference, changelog entry).

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [x] Docs

## Validation

- `pnpm typecheck` (all workspace packages)
- `pnpm lint` (`@zama-fhe/sdk`, `@zama-fhe/react-sdk`)
- `pnpm format:check`
- `pnpm test:run` — 1826 tests passing, including new unit coverage for `preparePermit`/`registerPermit` (shape, delegation, chain-switch races, tamper/mismatch cases, strict key-pair persistence and epoch races) and the two new React hooks
- `pnpm api-report` — regenerated for `@zama-fhe/sdk` and `@zama-fhe/react-sdk`, no `ae-forgotten-export` warnings
- `pnpm docs:check-links` — all internal links and anchors resolve
- `pnpm llm:build` — LLM corpus regenerated from the updated docs
- Added (and left properly `skipIf`-gated behind real credentials, consistent with the existing DFNS test) an integration test exercising `preparePermit` → DFNS `eth_signTypedData_v4` → `registerPermit` end to end

## Related

Closes SDK-295

## Demo (if possible)

No UI changes — this is an SDK/API surface addition.


soonum:
@cla-bot check

docs/gitbook/src/reference/react/usePreparePermit.md
```diff
@@ -0,0 +1,91 @@
...
+### request
+
+`PreparePermitRequest`
+
+| Field          | Type        | Default                                       | Meaning                                             |
+| -------------- | ----------- | --------------------------------------------- | --------------------------------------------------- |
+| `signer`       | `Address`   | required                                      | address expected to sign the returned `eip712`      |
+| `contracts`    | `Address[]` | required, max 10, no chunking                 | contract addresses to authorize                     |
+| `delegator`    | `Address`   | none — self permit; must differ from `signer` | delegator address, for a delegated permit           |
+| `durationDays` | `number`    | the SDK's configured `permitTTL`, max 365     | permit validity window in days, a V1 protocol limit |
```
ankurdotb, line 62: At the SDK level we were trying to standardise all input values for time asked from the user to second or milliseconds (check `transportKeypairTTL`). The exception we had was if the value/property was an exact onchain ABI, where we try and match the onchain ABI but I suppose that could be internal.

Does it make sense to stick to `permitTTL` as input here, or do we ask for `durationDays` elsewhere, e.g., in online signing flows?

docs/gitbook/src/reference/react/usePreparePermit.md
ankurdotb, line 1: Note: these methods should work not just in React SDK, but also in the core SDK as they are different contexts the two SDKs are used in. Offline prep/signing more likely to be used in a backend scenario where React is not used.

EDIT: Seems covered later

ankurdotb (commented):
Early review on some bits that sparked questions

docs/gitbook/src/reference/react/usePreparePermit.md
eitjuh: GrantPermit doesn’t take a duration, it always uses config permitTTL (days), although it's abstracted away in the SDK. I used durationDays here because preparePermit builds the EIP-712 payload the custody signer actually sees, and that message field is durationDays. I thought it would be confusing to take permitTTL on the request and then show durationDays when signing.

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -187,6 +209,237 @@ export class CredentialService {
...
+      throw new PreparedPermitExpiredError(
+        `registerPermit: the prepared permit's validity window (starting ${parsed.startTimestamp}, ` +
+          `${parsed.durationDays}d) has already elapsed — call preparePermit again.`,
+      );
+    }
+
+    const signerAddress = parsed.signerAddress;
+    const keypair = await this.#vault.getOrCreate(signerAddress);
+    if (keypair.publicKey !== parsed.transportPublicKey) {
+      throw new TransportKeyPairChangedError(
```
zmalatrax: Should use `#vault.readStored(signerAddress)` here, as the eip712 already bundles the transport public key used, it's useless to create a new transport keypair that will throw right after anyway
(readStored returns `null` if the key does not exist)

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -187,6 +209,237 @@ export class CredentialService {
...
+      }
+      throw wrapSigningError(error, "registerPermit: signature verification failed");
+    }
+
+    // `parsed.contracts`/`startTimestamp`/`durationDays`/`delegatorAddress` travel
+    // alongside `eip712`/`signature` but are not themselves covered by the
+    // signature — only what's embedded in `signedPermit.eip712.message` is
+    // cryptographically verified. Refuse to persist a permit whose storage
+    // metadata disagrees with what the signer actually signed.
+    this.#assertPreparedPermitMatchesSignedPayload(parsed, signedPermit);
```
zmalatrax: I'm not sure those "unsigned" metadata fields are required, they are initially copied from the request arg in preparePermit, and then they're directly available from the constructed `eip712` object

And in registerPermit, they're also available from the now-signed `eip712` field/object

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -187,6 +209,237 @@ export class CredentialService {
...
+      serializedPermit,
+      startTimestamp: parsed.startTimestamp,
+      durationDays: parsed.durationDays,
+    };
+    const scope: PermissionScope = {
+      signerAddress,
+      chainId: activeChainId,
+      delegatorAddress: parsed.delegatorAddress ?? signerAddress,
+    };
+    await this.#store.append(scope, [permission]);
```
zmalatrax: Services integrating the offline permit will usually trigger `registerPermit` on webhooks or similar callback, so it could be called multiple times (re-delivered, re-tried registration) --> registerPermit should be idempotent

I'd rather use `this.#store.replace(scope, serializedPermit.signature, permission)`, it fallbacks to `append` so identical on first registration, then re-registration overwrite their own entry instead of duplicating it

packages/sdk/src/credentials/types.ts
```diff
@@ -98,3 +98,50 @@ export interface SerializedTransportKeyPairWithPermissions {
...
+  contracts: ChecksummedAddress[];
+  /** Chain ID the permit is bound to. Must match the active chain at register time. */
+  chainId: number;
+  /** Unix timestamp (seconds) the permit becomes valid. */
+  startTimestamp: number;
+  /** Validity window length in days from {@link startTimestamp}. */
+  durationDays: number;
+  /** Transport key pair public key embedded in {@link eip712}, hex-encoded. */
+  transportPublicKey: Hex;
+}
```
zmalatrax, line 142: cf. https://github.com/zama-ai/sdk/pull/634/changes#r3828223025
I'd keep it minimal, { version, eip712, signerAddress } (`SerializedPermit` minus the `signature` field

This way, no need for `#assertPreparedPermitMatchesSignedPayload`, `PreparedPermitMismatchError` and related tests about "persist unsigned field, verify against signed original"

For "metadata" use, drilling into `eip712` is fine imo

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -187,6 +209,237 @@ export class CredentialService {
...
+    const durationDays =
+      request.durationDays !== undefined
+        ? parseSchema(PermitTTLSchema, request.durationDays)
+        : this.#permitTTL;
+    if (durationDays > MAX_V1_PERMIT_DURATION_DAYS) {
+      throw new ConfigurationError(
+        `preparePermit: durationDays (${durationDays}) exceeds the V1 permit maximum of ` +
+          `${MAX_V1_PERMIT_DURATION_DAYS} days.`,
+      );
+    }
```
zmalatrax: Not needed, the fhevm/sdk is already validating when creating the unsigned permit (`createUnsignedLegacyDecryptionPermitEip712` through `_validateDecryptionPermitEip712V1`), rejecting empty contracts, >10 addresses and durationDays>365

packages/sdk/src/credentials/__tests__/credential-service-offline-permit.test.ts
```diff
@@ -0,0 +1,422 @@
...
+    });
+    const credentialService = createCredentialService({ router });
+
+    // Switch chains between preparePermit's snapshot of the relayer/chainId and its
+    // relayer call, simulating a race with a concurrent switchChain().
+    vi.mocked(relayerA.createUnsignedLegacyDecryptionPermitEip712).mockImplementationOnce(
+      async () => {
+        router.switchChain(2);
+        return {
+          domain: { name: "Decryption", version: "1", chainId: 1n, verifyingContract: A },
```
zmalatrax: The permit could be created from the original relayerA, similar pattern as in L131-L136 (same file)

>This hand-writes a third copy of the EIP-712 shape with a hardcoded `chainId: 1n`, which makes `expect(prepared.chainId).toBe(1)` tautological and defeats the fixture's own design (its comment says the domain derives from the chain precisely for multi-chain-router tests). Lines 131-136 already show the right move: capture the default implementation, switch the chain, delegate. Then asserting on `prepared.eip712.domain.chainId` actually proves the snapshot held.
>
> Adapted to the chain-switch test at line 311, roughly:
>
> ```typescript
> // Switch chains between preparePermit's snapshot of the relayer/chainId and its
> // relayer call, simulating a race with a concurrent switchChain(). Delegates to
> // the default mock implementation — only the timing is under test here.
> const originalCreate = vi
>   .mocked(relayerA.createUnsignedLegacyDecryptionPermitEip712)
>   .getMockImplementation();
> assertNonNullable(originalCreate, "relayerA.createUnsignedLegacyDecryptionPermitEip712 mock implementation");
> vi.mocked(relayerA.createUnsignedLegacyDecryptionPermitEip712).mockImplementationOnce(
>   async (params) => {
>     router.switchChain(2);
>     return originalCreate(params);
>   },
> );
> ```

packages/sdk/src/credentials/types.ts
eitjuh: Fixed

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/__tests__/credential-service-offline-permit.test.ts
eitjuh: Fixed

packages/sdk/src/credentials/credential-service.ts
eitjuh: fixed it within the permitTTLSchema

docs/gitbook/src/changelog/alpha.md
```diff
@@ -26,6 +26,19 @@ const prepared = await sdk.offline.prepare({
 
 All eleven write operations are covered, including the shield legs and the two-phase unshield. See [Offline signing](../guides/offline.md) for the full guide: the handoff contract, request kinds, multi-transaction batches, and approval-delay behavior.
 
+## Offline decryption permits
+
+Decryption permits now have an offline flow, closing the gap the guide above used to call out: `sdk.offline.preparePermit` builds the unsigned EIP-712 typed data for a permit — no signer required — and `sdk.permits.registerPermit` verifies and persists the signature a custody partner returns for it.
```
zmalatrax: > Decryption permits now have an offline flow, closing the gap the guide above used to call out:

Not sure about mentioning a missing gap in the changelog, I'd rather state it now supports offline permit

thinking out loud*, offline transaction are only released on alpha, not stable, maybe worth to refactor the whole offline transaction/permit block into one 🤔

docs/gitbook/src/guides/offline.md
```diff
@@ -152,13 +152,42 @@ Policy approval can take hours or days. The cryptographic proofs tolerate that,
...
+const signature = await custody.signTypedData(prepared.eip712);
+```
+
+Then register the signature. This verifies it against `prepared.eip712` and persists the permit — no further wallet interaction:
+
+```ts
+await sdk.permits.registerPermit(prepared, signature);
+```
+
+One permit per call: unlike `grantPermit`, `preparePermit` never widens an existing permit or chunks a request over 10 contracts — `contracts` maps to exactly one signature. `prepared` is a V1 permit (`version: 1`): it stays valid after the chain upgrades to a newer protocol version, and the relayer accepts V1 and V2 permits side by side, so there's no forced migration when a V2 flow ships.
```
zmalatrax: Not sure if worth mentioning future V2 permits as they're not part of the sdk yet
> it stays valid after the chain upgrades to a newer protocol version, and the relayer accepts V1 and V2 permits side by side, so there's no forced migration when a V2 flow ships.

docs/gitbook/src/concepts/permit-model.md
```diff
@@ -106,6 +106,10 @@ The SDK automatically manages permits when the wallet state changes:
 
 See [ZamaSDK.onWalletAccountChange](../reference/sdk/ZamaSDK.md#onwalletaccountchange) for programmatic access to these transitions.
 
+## Offline permits
+
+The lifecycle above assumes a connected wallet signs `grantPermit`'s EIP-712 message in-process. Custody partners — HSMs, policy engines, out-of-process signers — cannot do that, so `sdk.offline.preparePermit` / `sdk.permits.registerPermit` split the same signature out of the SDK: prepare the unsigned typed data, sign it externally, then register the signature. See the [Offline signing guide](../guides/offline.md#offline-permits) for the workflow.
```
zmalatrax: I couldn't tell what exactly, but there is something off in the flow/wording when reading it,

zmalatrax (approved):
lgtm, ty!

PR #636, opened 2026-08-20
fix(examples): switch Sepolia/Hoodi explorer links to BlockScout [SDK-297]

ankurdotb:
# Summary

Point react-wagmi, react-ethers, react-viem (Sepolia), example-hoodi (Hoodi), and react-turnkey-wallet (Sepolia + mainnet) tx-explorer links at BlockScout instead of Etherscan. For the wagmi-registered apps (react-wagmi, example-hoodi), also replace the imported stock `wagmi/chains`/`viem/chains` chain object with a custom `defineChain(...)` chain (the same pattern `example-ingen` already uses for its own network) so the wallet's "add network" prompt points at BlockScout too, not just the tx-link constant. The `switchChain()` call in both apps uses the custom chain object's `.id` rather than the scalar constant, matching `example-ingen`'s own established convention for that call site.

- Sepolia apps (react-wagmi, react-ethers, react-viem) → `eth-sepolia.blockscout.com`
- example-hoodi → `eth-hoodi.blockscout.com`
- example-hoodi's README/WALKTHROUGH docs updated to match
- node-ethers/node-viem have no explorer-link *code* (CLI scripts), but their READMEs named Etherscan in prose — reworded to Blockscout
- react-turnkey-wallet wasn't in the original ticket scope but has its own Etherscan-pointing `explorerUrl` (Sepolia + mainnet branches) — extended the swap here too, to `eth-sepolia.blockscout.com` / `eth.blockscout.com`

Out of scope (unchanged): `example-polygon-amoy` (BlockScout doesn't support Polygon Amoy — see SDK-290), `example-ingen` (already a Blockscout deployment), `example-bnb`.

## Scope

- [ ] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [x] Examples (`examples/`)
- [ ] CI/workflows
- [ ] Docs

## Validation

- `npm run typecheck` and `npm run build` in every touched app (react-wagmi, react-ethers, react-viem, example-hoodi, react-turnkey-wallet) — all pass.
- Grepped `examples/` for leftover `etherscan` (URLs and prose) — none remain, aside from three explanatory code comments that name "Etherscan" by design (explaining why the custom chain exists).
- `pnpm llm:check` passes — the regenerated LLM corpus (`llms-full.txt`) matches the committed one, re-verified after a mid-PR rebase onto `prerelease`.
- No existing unit/e2e test asserts on the explorer domain, so none needed updating.

## Related

Closes SDK-297 (ticket updated to reflect the extended scope — node-ethers/node-viem docs, react-turnkey-wallet). See also PRO-502 (same Etherscan→BlockScout swap for docs.zama.ai) and SDK-290 (why `example-polygon-amoy` is excluded).

## Demo (if possible)

N/A — link/config swap, no visual UI change.


examples/example-hoodi/src/app/page.tsx
```diff
@@ -2,12 +2,11 @@
 
 import { ActionScreen, AppHeader, TokenSelector } from "@/components/PageChrome";
 import { NoTokenWorkspace, SelectedTokenPanel } from "@/components/SelectedTokenPanel";
-import { HOODI_CHAIN_ID } from "@/lib/config";
+import { hoodi, HOODI_CHAIN_ID } from "@/lib/config";
 import { useListPairs } from "@zama-fhe/react-sdk";
 import type { Address, TokenWrapperPairWithMetadata } from "@zama-fhe/sdk";
 import { useEffect, useMemo, useState } from "react";
 import { formatEther } from "viem";
-import { hoodi } from "viem/chains";
```
ankurdotb, line 10: I am not sure if this is the best pattern to keep or remove

examples/react-viem/src/app/page.tsx
```diff
@@ -3,7 +3,6 @@
 import { useEffect, useMemo, useState } from "react";
 import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { formatEther, createPublicClient, http } from "viem";
-import { sepolia } from "viem/chains";
```
ankurdotb, line 6: Same comment here as for Hoodi above

examples/react-viem/src/lib/config.ts
```diff
@@ -1,10 +1,21 @@
...
-export const SEPOLIA_EXPLORER_URL = "https://sepolia.etherscan.io";
+export const SEPOLIA_EXPLORER_URL = "https://eth-sepolia.blockscout.com";
 const SEPOLIA_RPC_DEFAULT = "https://ethereum-sepolia-rpc.publicnode.com";
 // Use || not ?? — Next.js replaces unset NEXT_PUBLIC_* with "" (empty string) at build time,
 // not undefined. "" is not nullish, so ?? would use the empty string as the URL.
 export const SEPOLIA_RPC_URL = process.env.NEXT_PUBLIC_SEPOLIA_RPC_URL || SEPOLIA_RPC_DEFAULT;
+
+// Custom chain definition (rather than importing viem's stock `sepolia`) so the
+// wallet's "add network" prompt points at Blockscout instead of the baked-in Etherscan entry.
+export const sepolia = defineChain({
```
zmalatrax, line 15: worth adding the flag `testnet: true`

examples/react-viem/src/lib/config.ts
```diff
@@ -1,10 +1,21 @@
...
+  id: SEPOLIA_CHAIN_ID,
+  name: "Sepolia",
+  nativeCurrency: { name: "Sepolia Ether", symbol: "ETH", decimals: 18 },
+  rpcUrls: { default: { http: [SEPOLIA_RPC_URL] } },
+  blockExplorers: { default: { name: "Blockscout", url: SEPOLIA_EXPLORER_URL } },
```
zmalatrax: Do we need all the fields from the chain definition? (besides testnet: true mentioned in the comment above)

For reference, the built-in sepolia definition, main diff is multicall3 and, ensUniversalResolver and apiUrl of the block explorer
https://github.com/wevm/viem/blob/main/src/chains/definitions/sepolia.ts

examples/react-viem/src/lib/config.ts
zmalatrax: Following https://github.com/zama-ai/sdk/pull/636#discussion_r3820610268

I think that you could spread the actual `sepolia` config and only re-define the blockExplorer object

```ts
import { defineChain } from "viem";
import { sepolia } from "viem/chains";

export const sepoliaChain = defineChain({
  ...sepolia,
  blockExplorers: {
    default: {
      name: "Blockscout",
      url: "https://eth-sepolia.blockscout.com",
      apiUrl: "https://eth-sepolia.blockscout.com/api",
    },
  },
});
```

And you can update in the react-* the `SEPOLIA_EXPLORER_URL` const
`export const SEPOLIA_EXPLORER_URL = sepoliaChain.blockExplorers.default.url;`

examples/example-hoodi/src/app/page.tsx
zmalatrax: If we want to use blockscout, then we should override the built-in chain definitions, though limiting the override to the block explorer only would reduce maintenance/potential drift (even though chain definitions are quite stable)

examples/example-hoodi/src/lib/config.ts
```diff
@@ -1,8 +1,28 @@
...
-export const HOODI_EXPLORER_URL = "https://hoodi.etherscan.io";
 const HOODI_RPC_DEFAULT = "https://rpc.hoodi.ethpandaops.io";
 export const HOODI_RPC_URL = process.env.NEXT_PUBLIC_HOODI_RPC_URL || HOODI_RPC_DEFAULT;
+
+// Extends viem's stock Hoodi definition instead of redefining it, so multicall3 and the
+// testnet flag stay inherited (and keep up with viem). Only two fields are overridden:
+// the block explorer (Blockscout renders confidential tokens and transfers better than
+// Etherscan) and the RPC, so both the app and the wallet's "add network" prompt use the
+// endpoint configured via NEXT_PUBLIC_HOODI_RPC_URL.
+export const hoodiChain = defineChain({
```
zmalatrax: nit: you could export as `hoodi` by renaming the viem (resp. wagmi) import of the chain:

```ts
import { hoodi as viemHoodi } from "viem/chains"

export const hoodi = defineChain({
  ...viemHoodi,
}
```

examples/example-hoodi/src/lib/config.ts
```diff
@@ -1,8 +1,28 @@
...
 export const HOODI_CHAIN_ID_HEX = `0x${HOODI_CHAIN_ID.toString(16)}`;
-export const HOODI_EXPLORER_URL = "https://hoodi.etherscan.io";
 const HOODI_RPC_DEFAULT = "https://rpc.hoodi.ethpandaops.io";
 export const HOODI_RPC_URL = process.env.NEXT_PUBLIC_HOODI_RPC_URL || HOODI_RPC_DEFAULT;
+
+// Extends viem's stock Hoodi definition instead of redefining it, so multicall3 and the
+// testnet flag stay inherited (and keep up with viem). Only two fields are overridden:
+// the block explorer (Blockscout renders confidential tokens and transfers better than
+// Etherscan) and the RPC, so both the app and the wallet's "add network" prompt use the
+// endpoint configured via NEXT_PUBLIC_HOODI_RPC_URL.
```
zmalatrax: nit: I'd only keep the "Extends viem's stock Hoodi definition instead of redefining it", the rest of the comment does not provide much

zmalatrax (dismissed):
lgtm, 2 nit comments, but non-blocker

zmalatrax (approved):

PR #638, opened 2026-08-24
feat: add polygon support

zmalatrax:
Closes SDK-285

### DIFF
--- a/docs/gitbook/src/guides/authentication.md
+++ b/docs/gitbook/src/guides/authentication.md
@@ -5,7 +5,7 @@ description: How to authenticate with the relayer using a backend proxy or a dir
 
 # Authentication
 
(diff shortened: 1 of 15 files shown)


ankurdotb (commented):
`polygon` is the right sort of term to call the mainnet preset.

How do we choose what RPC to set as default? Random? Best performing over certain time?

eitjuh (approved):

zmalatrax:
> How do we choose what RPC to set as default? Random? Best performing over certain time?
The other chains are relying on publicnode RPCs in the presets and examples, so I followed that pre-existing convention

I'm not opinionated on this, I know that publicnode are working free RPCs on mainnet-like networks, contrary to drpc for example

PR #645, opened 2026-08-25
fix: self-heal stale pending-unshield pointer

zmalatrax:
Closes SDK-296

### DIFF
--- a/packages/sdk/src/token/wrapped-token.ts
+++ b/packages/sdk/src/token/wrapped-token.ts
@@ -435,7 +435,7 @@ export class WrappedToken extends Token {
    * @param unwrapTxHash - The transaction hash of the previously submitted unwrap.
    * @param callbacks - Optional progress callbacks.
    * @returns The finalize transaction hash and mined receipt.
-   * @throws if the unwrap request was already finalized on-chain — the persisted
+   * @throws if the unwrap request was already finalized on-chain; the persisted
    *   pending-unshield state is cleared first. {@link UnshieldAlreadyFinalizedError}
    *
    * @example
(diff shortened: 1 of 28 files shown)


packages/sdk/src/token/wrapped-token.ts
```diff
@@ -587,16 +618,69 @@ export class WrappedToken extends Token {
...
-   * Throws when the receipt lacks the `UnwrapRequested` event entirely.
+   * Only logs emitted by this wrapper count — a caller-supplied tx hash from a
+   * different wrapper must fail here, not decode a foreign request id. Logs
+   * without an emitter address (older custom adapters) are kept.
+   * Throws when no matching `UnwrapRequested` event is found.
    */
   #extractUnwrapRequestId(logs: readonly RawLog[]): EncryptedValue {
-    const event = findUnwrapRequested(logs);
+    const ownLogs = logs.filter(
+      (log) => log.address === undefined || getAddress(log.address) === this.address,
```
eitjuh, line 628: With the help of my quant:

This new check tries to protect against someone passing in a transaction hash from a different wrapper contract. It does this by looking at which contract emitted the log (log.address) and skipping any log that doesn't match this wrapper.

The problem: if you're using ethers.js (via EthersProvider), that address never gets set in the first place — see packages/sdk/src/ethers/ethers-provider.ts, around lines 121-124. The code builds the log object but leaves out address:
```
logs: receipt.logs.map((log) => ({
  topics: log.topics.filter((t): t is Hex => t !== null),
  data: log.data as Hex,
})),
```
Since this line treats a missing address as "let it through," ethers.js users don't actually get this protection — any wrapper's log will pass the check. It only really works for viem/wagmi users, since those already include the address.

Fix: add address: log.address as Hex, to that mapping so ethers.js users get the same protection. Might also be worth adding a test that uses the real EthersProvider (not just the test mock), since that's how this slipped through.

eitjuh (changes requested):

packages/sdk/src/token/wrapped-token.ts
zmalatrax: fixed in [4aa7d87](https://github.com/zama-ai/sdk/pull/645/commits/4aa7d87fcab9bfa0784dc84651537e3db76bc4c1)

eitjuh (dismissed):
lgtm

eitjuh (dismissed):

eitjuh (approved):

PR #653, opened 2026-08-26
feat: v2 wildcard permits

eitjuh:
# Summary

Adds V2 (unified) decryption permit support to `@zama-fhe/sdk`: wildcard/permissive permits, requested by passing the new `WILDCARD_PERMIT` sentinel to `grantPermit`. One permit signature then covers every contract instead of an explicit list — an opt-in alternative to the existing bounded V1 model, never inferred from a large contract list.

- `Permission` is now a discriminated union (`PermissionV1 | PermissionV2`); coverage/widening/pruning/revocation logic (`uncoveredContracts`, `findPermitToWiden`, `pruneUnusable`, `withoutPermitsTouching`, `resolvePermit`) is wildcard-aware.
- `CredentialService.grantPermit` / `Permits.grantPermit` / `grantDelegationPermit` / `useGrantPermit` accept `WILDCARD_PERMIT` alongside the existing `Address[]` form.
- Wired `protocolConfigContractAddress` into `FheChain` — required for V2 signing to ever succeed; previously hardcoded to `undefined` for every chain, so V2 signing could never work regardless of a network's actual protocol version.
- New `UnifiedPermitNotSupportedError` (code `UNIFIED_PERMIT_NOT_SUPPORTED`), thrown before any wallet prompt when a wildcard/V2 permit is requested against a chain still on protocol <0.14, naming the required version explicitly.
- Fixed a storage-migration bug found during review: permits persisted by pre-this-change SDK versions (no top-level `version` field) would otherwise fail the new schema and get silently wiped from storage on next load. Added a backward-compatible backfill (`backfillLegacyPermissionVersion`) so existing users' stored permits keep working.
- V1 (existing) behavior verified unchanged — including a byte-for-byte call-ordering fix caught during review — and confirmed live against Sepolia (currently on protocol <0.14): `grantPermit([contract])` signs and persists correctly; `grantPermit(WILDCARD_PERMIT)` fails immediately with the new typed error and no wallet prompt.

Implements [SDK-304](https://linear.app/zama/issue/SDK-304/feat-add-v2wildcard-permit-types-and-signing), the first sub-issue of [SDK-292](https://linear.app/zama/issue/SDK-292/feat-add-support-for-unified-decryption).

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [x] Docs

## Validation

- `pnpm typecheck` — clean across all 7 packages
- `pnpm lint` (oxlint + ast-grep) — clean
- `pnpm test:run` — 169 files / 1894 tests passing, including new coverage for wildcard grant/reuse/revocation, the pre-v0.14 error path, and a dedicated storage-migration regression test
- `pnpm docs:check-links` — 110 pages, clean
- `pnpm api-report` / `api-report:check` — regenerated and reviewed (`sdk.api.md`, `sdk-query.api.md`, `sdk-viem.api.md`, `sdk-ethers.api.md`, `sdk-node.api.md`); scanned every generated report directly for embedded `ae-forgotten-export` / `ae-undocumented` warnings — none remain
- Manual live-network check against Sepolia (real relayer + RPC, no mocks): `grantPermit([contract])` signs and persists correctly (confirmed via a fresh `hasPermit` read-back); `grantPermit(WILDCARD_PERMIT)` fails in ~40ms with `UnifiedPermitNotSupportedError`, before any wallet interaction — confirming Sepolia's current pre-v0.14 status is handled cleanly end-to-end, not just in mocked unit tests

## Related

Closes [SDK-304](https://linear.app/zama/issue/SDK-304/feat-add-v2wildcard-permit-types-and-signing) (sub-issue of [SDK-292](https://linear.app/zama/issue/SDK-292/feat-add-support-for-unified-decryption)).

## Demo (if possible)

N/A — backend/SDK API change, no UI. Usage:

```ts
import { WILDCARD_PERMIT } from "@zama-fhe/sdk";

// One signature, covers every contract (present and future):
await sdk.permits.grantPermit(WILDCARD_PERMIT);

// vs. today's bounded, explicit form — unchanged:
await sdk.permits.grantPermit([tokenA, tokenB]);
```


zmalatrax:
need to update the fhevm/sdk to the 0.14 rc, it was reverted on `prerelease` head, and `alpha` is still on 0.13.2 as well

docs/gitbook/src/concepts/permit-model.md
```diff
@@ -79,11 +79,29 @@ This means users see fewer wallet popups over time. As they interact with more c
 Batch all contract addresses you expect to need into a single `permits.grantPermit()` call to minimize wallet popups. Each uncovered chunk of up to 10 contracts triggers one signature prompt.
 {% endhint %}
 
+## Wildcard (V2) permits
+
+Everything above describes the V1 permit model: a bounded, explicit list of up to 10 contract addresses per permit. `@zama-fhe/sdk` also supports V2 (unified) permits, which add one capability on top of that model: a **wildcard permit**, requested by passing the exported `WILDCARD_PERMIT` sentinel instead of a contract list —
```
zmalatrax, line 84: I think we should frame it as V2 permits first, then a subsection on the wildcard feature they enable
Because V2 permits have a slightly different shape, and indeed not necessarily wildcard permits (if you specify contracts in the `allowedContracts` field it's not a wildcard anymore)

Also, at some point V1 permit will be deprecated, once available on testnet/mainnet with V2 permits we want to enforce V2 permits for newly created ones, while pre-existing V1 permits are still usable

We even might want to rewrite the permit model page to replace V1 permit section with the V2 one 🤔

docs/gitbook/src/concepts/permit-model.md
```diff
@@ -79,11 +79,29 @@ This means users see fewer wallet popups over time. As they interact with more c
...
+
+await sdk.permits.grantPermit(WILDCARD_PERMIT);
+```
+
+A wildcard permit covers every contract the signer is allowed to decrypt from, present and future, with a single wallet signature — including contracts the app doesn't know about yet at grant time. It's the same additive/reuse model as a normal permit: granting it again, or requesting decryption for any contract while a valid wildcard permit is in scope, never re-prompts.
+
+A few things a wildcard permit does **not** change:
+
+- **On-chain delegation is unaffected.** A wildcard permit only removes the need to enumerate contracts in the _permit_ — it has no bearing on `FHE.delegateUserDecryption()` grants, which remain required per-contract (or via a separate wildcard delegation on the ACL contract) for delegated decryption to work at all. Permit coverage and ACL delegation are two independent mechanisms.
+- **It's opt-in, not inferred.** The SDK never signs a wildcard permit unless the caller explicitly asks for one — passing a large contract list to `grantPermit()` still signs a normal, bounded V1 permit for exactly those contracts.
```
zmalatrax, line 97: Here it should sign a V2 permit, which has the same upper bound when not a wildcard permit
https://github.com/zama-ai/fhevm/blob/6b38823e2043e7ae54c323093899e8f6ed1b47c6/gateway-contracts/contracts/Decryption.sol#L670-L673

But maybe you intented to rewrite it through [SDK-305](https://linear.app/zama/issue/SDK-305/feat-wire-v1v2-permit-branching-into-decryption) ?

packages/sdk/src/chains/configs.ts
```diff
@@ -1,5 +1,11 @@
 import type { FheChain } from "./types";
 
+// None of the built-in configs below set `protocolConfigContractAddress` yet —
```
zmalatrax: ProtocolConfig addresses are already available on both testnet and mainnet
mainnet: https://github.com/zama-ai/protocol-registry/blob/7aca6bc7588ec984cd4648436c7b4c5d126e26c5/mainnet.json#L619-L628
testnet (Polygon Amoy): https://github.com/zama-ai/protocol-registry/blob/7aca6bc7588ec984cd4648436c7b4c5d126e26c5/testnet.json#L129-L134
testnet (Ethereum Sepolia): https://github.com/zama-ai/protocol-registry/blob/7aca6bc7588ec984cd4648436c7b4c5d126e26c5/testnet.json#L731-L735

packages/sdk/src/chains/to-fhevm-chain.ts
```diff
@@ -18,7 +18,9 @@ export function toFhevmChain(chain: FheChain): FhevmChain {
         acl: { address: chain.aclContractAddress },
         inputVerifier: { address: chain.inputVerifierContractAddress },
         kmsVerifier: { address: chain.kmsContractAddress },
-        protocolConfig: undefined,
+        protocolConfig: chain.protocolConfigContractAddress
+          ? { address: chain.protocolConfigContractAddress }
+          : undefined,
```
zmalatrax, line 23: Are there still cases where ProtocolConfig address would not be available? 🤔 
Might be possible to remove the `undefined` type here

packages/sdk/src/credentials/types.ts
```diff
@@ -67,27 +67,55 @@ export interface SerializedPermit {
...
+/**
+ * A signed EIP-712 permit binding a signer (and optional delegator) to a set of
+ * contract addresses for a bounded time window. Wraps the serialized `@fhevm/sdk`
+ * permit ({@link SerializedPermit}) alongside the scope/coverage metadata the
+ * permission store indexes on.
+ *
+ * Discriminated by `version`: {@link PermissionV1} or {@link PermissionV2}.
+ * Mirrors {@link PermissionSchema}; kept in sync by a `.test-d.ts` guard.
+ */
+export type Permission = PermissionV1 | PermissionV2;
```
zmalatrax, line 116: `serializedPermit` already has a `version` field, I think we can compound on it as it comes from the fhevm/sdk permit definition

And instead of duplicating all fields, we could make a base interface and extend V1 and V2 with `durationDays` and `durationsSeconds`, sth like
```ts
interface PermissionBase { keypairPublicKey: Hex; ... }
export interface PermissionV1 extends PermissionBase { durationDays: number }
export interface PermissionV2 extends PermissionBase { durationSeconds: number }
export type Permission = PermissionV1 | PermissionV2;
```

packages/sdk/src/credentials/permission-store.ts
```diff
@@ -15,6 +15,22 @@ interface PermissionStoreConfig {
...
+/**
+ * A permit persisted by an SDK version before V2 (unified) permits existed has
+ * no top-level `version` field — the discriminant didn't exist yet, and every
+ * such stored permit is a V1 permit. Backfills `version: 1` on read so that
+ * upgrading to an SDK version whose `PermissionSchema` requires that field
+ * doesn't treat old, still-valid stored permits as corrupt and wipe them via
+ * {@link PermissionStore.clearScope}. Mirrors `@fhevm/sdk`'s own
+ * `parseSignedDecryptionPermit`: "if no version, interpret as permit v1".
+ */
+function backfillLegacyPermissionVersion(raw: unknown): unknown {
```
zmalatrax, line 27: this would become unnecessary if applying https://github.com/zama-ai/sdk/pull/653#discussion_r3872126330

packages/sdk/src/namespaces/permits.ts
```diff
@@ -61,11 +62,18 @@ export class Permits {
...
+   * wildcard permit decrypts anything the signer owns until it expires, so
+   * prefer a specific contract list unless the caller genuinely needs
+   * unbounded coverage.
+   *
+   * @param contracts - Contract addresses to authorize, or {@link WILDCARD_PERMIT}.
    * @throws if `transportKeyPairDerivationSecret` is configured and the keypair fails to wrap or
    *   unwrap. {@link KeyWrappingError}
    */
-  async grantPermit(contracts: Address[]): Promise<void> {
+  async grantPermit(contracts: Address[] | WildcardPermit): Promise<void> {
```
zmalatrax, line 76: what about a separate method for wildcard permits? `sdk.permits.grantWildcardPermit()` and `sdk.permits.grantWildcardDelegationPermit`

packages/sdk/src/namespaces/permits.ts
```diff
@@ -61,11 +62,18 @@ export class Permits {
...
     if (contracts.length === 0) {
```
zmalatrax: Should check if `contracts` is an array as well, no? Even though it works as `const WILDCARD_PERMIT = "wildcard"`

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -197,7 +231,9 @@ export class CredentialService {
     const requestedSet = new Set(requested);
     return {
       keypair,
-      permissions: permissions.filter((p) => p.contractAddresses.some((a) => requestedSet.has(a))),
+      permissions: permissions.filter(
+        (p) => isWildcardPermission(p) || p.contractAddresses.some((a) => requestedSet.has(a)),
```
zmalatrax: The "is already covered by permit" check is inlined in different shapes across multiple places, it might be worth refactoring into a shared generic helper or sth 🤔 

same line in `withoutPermitsTouching`:
https://github.com/zama-ai/sdk/blob/d298e6b06bd541d17ce1b4254ed47e2328695d61/packages/sdk/src/credentials/permissions.ts#L68

similar in `uncoveredContracts`: https://github.com/zama-ai/sdk/blob/d298e6b06bd541d17ce1b4254ed47e2328695d61/packages/sdk/src/credentials/permissions.ts#L23

`findPermissionFor`: https://github.com/zama-ai/sdk/blob/d298e6b06bd541d17ce1b4254ed47e2328695d61/packages/sdk/src/credentials/decrypt-permit.ts#L54

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -583,46 +620,78 @@ export class CredentialService {
...
+                ...permitInput,
+                delegatorAddress: scope.delegatorAddress,
+              })
+            : await relayer.signUnifiedDecryptionPermit(permitInput)
+          : isDelegated
+            ? await relayer.signDecryptionPermit({
+                ...permitInput,
+                delegatorAddress: scope.delegatorAddress,
+              })
+            : await relayer.signDecryptionPermit(permitInput);
```
zmalatrax: Both signUnifiedDecryptionPermit and signDecryptionPermit `parameters` have the same type signature, I think it could be simplified

```ts
const sign = version === 2 ? relayer.signUnifiedDecryptionPermit : relayer.signDecryptionPermit;
const signedPermit = await sign(isDelegated ? {...permitInput, delegatorAddress: scope.delegatorAddress} : permitInput);
```

ref:
[signUnifiedDecryptionPermit](https://github.com/zama-ai/fhevm/blob/v0.14.1-0/sdk/js-sdk/src/core/actions/base/signUnifiedDecryptionPermit.ts)
[signDecryptionPermit](https://github.com/zama-ai/fhevm/blob/v0.14.1-0/sdk/js-sdk/src/core/actions/base/signDecryptionPermit.ts)
[signLegacyDecryptionPermit](https://github.com/zama-ai/fhevm/blob/v0.14.1-0/sdk/js-sdk/src/core/actions/base/signLegacyDecryptionPermit.ts)

docs/gitbook/src/concepts/permit-model.md
eitjuh: Makes sense, but the intent is to cover that with SDK-305

docs/gitbook/src/concepts/permit-model.md
eitjuh: Makes sense, but the intent is to cover that with SDK-305

packages/sdk/src/chains/configs.ts
eitjuh: Fixed

packages/sdk/src/credentials/types.ts
eitjuh: Implemented that base interface like suggested. I've kept the version as it's own field instead of reusing serializedPermit, because TS can only tell v1 and v2 apart when the tag is directly on the object, not buried on a deeper level. Moving it means changing all the code which changes the permit version which I think is not worth it, current is more simple.

packages/sdk/src/namespaces/permits.ts
eitjuh: Hmmm, I like the simplicity of changing it based on the input, but if you feel very strongly about it I can change it

packages/sdk/src/namespaces/permits.ts
eitjuh: Fixed

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/chains/to-fhevm-chain.ts
eitjuh: I couldn't find all of them in protocol-registry - for example bscTestnet, happy to remove undefined once we populated them all

packages/sdk/src/credentials/permission-store.ts
eitjuh: I suggest this still needs to stay. I think keeping the separate version field is better (see other thread), mainly because TypeScript can only tell V1 and V2 apart automatically if the tag sits directly on the object, not buried inside serializedPermit. Since I suggest keeping that field, old saved permits don't have it yet, so we add it in when we load them, otherwise they'd break. Let me know if you don't like it

packages/sdk/src/credentials/credential-service.ts

zmalatrax (approved):
lgtm

zmalatrax (changes requested):
api-report check failed

zmalatrax (approved):

PR #683, opened 2026-08-28
feat(sdk): offload encryption to a web worker in the web() transport

zmalatrax:
Closes SDK-308

### DIFF
--- a/packages/sdk/src/relayer/fhevm-relayer.ts
+++ b/packages/sdk/src/relayer/fhevm-relayer.ts
@@ -3,6 +3,7 @@ import {
   createFhevmBaseClient,
   createFhevmDecryptClient,
   createFhevmEncryptClient,
+  hasFhevmRuntimeConfig,
 } from "@fhevm/sdk/viem";
 import {
   createFhevmCleartextBaseClient,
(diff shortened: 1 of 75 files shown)


packages/sdk/src/worker/encrypt-worker-client.ts
```diff
@@ -0,0 +1,664 @@
...
+    callOptions: CallOptions | undefined,
+  ): Promise<T> {
+    const { signal, onProgress, ...options } = callOptions ?? {};
+    // Checked before anything is spent: dispatching an already-aborted call
+    // would run the whole proof anyway, since the worker's event loop is blocked
+    // by the WASM work and the abort message only dequeues once it is done.
+    if (signal?.aborted) {
+      throw signal.reason ?? new DOMException("The operation was aborted.", "AbortError");
+    }
+    await this.init();
```
eitjuh, line 283: Init can take a while: spawn, download the encryption key, and initialize. The abort eventlistener is only added on line 299. If a caller aborts before the worker is initialized, we don't notice, and encryption runs anyway. I think we should check if there is an abort after init as well perhaps?

packages/sdk/src/relayer/fhevm-relayer.ts
```diff
@@ -89,19 +111,95 @@ export class FhevmRelayer implements RelayerSDK {
...
+    inline: {
+      prefetchKey: EncryptWorkerClientConfig["prefetchKey"];
+      createInlineClient: () => FhevmEncryptBackend;
+    },
+  ): FhevmEncryptBackend {
+    const offload = config.offloadEncrypt ?? false;
+    const strict = offload === true;
+    const workerAvailable =
+      typeof Worker === "function" &&
+      (config.offloadWorker !== undefined || BUNDLED_ENCRYPT_WORKER);
```
eitjuh, line 165: BUNDLED_ENCRYPT_WORKER is always false?

eitjuh (changes requested):

packages/sdk/src/relayer/fhevm-relayer.ts
zmalatrax: It's always `true` for ESM builds, while overriden to `false` for CJS builds (only added workers for the web target -> ESM builds)
https://github.com/zama-ai/sdk/blob/c047567cb1f549073bd9197a4ad865481fdafacb/packages/sdk/src/worker/spawn.ts#L3-L4
https://github.com/zama-ai/sdk/blob/c047567cb1f549073bd9197a4ad865481fdafacb/packages/sdk/rolldown.config.ts#L25

packages/sdk/src/worker/encrypt-worker-client.ts
zmalatrax: Indeed, I moved it before the init for early call abortion, but it could still happen during initialization, best fit would be to check before/after I suppose, will refactor

zmalatrax: fixed in [60a8d58](https://github.com/zama-ai/sdk/pull/683/commits/60a8d5812d6f9497b964a32c4c2ff72ed476931f)

eitjuh (approved):

PR #684, opened 2026-08-31
feat: wire v1v2 permit branching into decryption

eitjuh:
# Summary

`grantPermit` now probes the connected relayer (`canUseUnifiedDecryptionPermit`) and signs a V2 (unified) permit when that route is available, V1 otherwise — including when widening an existing permit. Decrypt dispatch is unchanged: `@fhevm/sdk` already routes on the permit's own version.

A stored V2/wildcard permit hitting a relayer that has not deployed `/v3/user-decrypt` now surfaces `UnifiedDecryptionUnsupportedError` (distinct from the existing signing-time `UnifiedPermitNotSupportedError`). V2 permit persistence keeps `delegatorAddress` as unsigned post-sign metadata so delegated V2 decrypts resolve `ownerAddress` to the delegator.

`@fhevm/sdk` in `package.json` / `pnpm-lock.yaml` is **not** in this push — bump that separately once the published version includes the V2 `delegatorAddress` parse/serialize round-trip.

## Scope

- [x] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [ ] Docs

## Validation

- `pnpm exec vitest run src/credentials/__tests__/credential-service.test.ts src/services/__tests__/decryption-service.test.ts src/errors/__tests__/decrypt.test.ts src/errors/__tests__/errors.test.ts` (226 tests, including typecheck)

## Related

- [SDK-305](https://linear.app/zama/issue/SDK-305/feat-wire-v1v2-permit-branching-into-decryption) (this work)
- Parent: [SDK-292](https://linear.app/zama/issue/SDK-292/feat-add-support-for-unified-decryption)
- Depends on: [`zama-ai/fhevm#3657`](https://github.com/zama-ai/fhevm/pull/3657) (V2 `delegatorAddress` through parse/serialize) — bump `@fhevm/sdk` once that ships

## Demo (if possible)

N/A — no UI changes.


packages/sdk/src/credentials/credential-service.ts
```diff
@@ -181,8 +222,8 @@ export class CredentialService {
     const signerAddress = checksum(account.address);
 
     if (contracts === WILDCARD_PERMIT) {
-      const keypair = await this.#vault.getOrCreate(signerAddress);
       const scope = this.#permissionScope(signerAddress, delegator);
+      const keypair = await this.#vault.getOrCreate(signerAddress);
```
zmalatrax, line 203: it should be gated by unified decryption check as well, as wildcard permits are only available for unified decryption

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -155,6 +163,39 @@ export class CredentialService {
...
+   */
+  async #canUseUnifiedPermit(): Promise<boolean> {
+    if (!this.#router.chain.protocolConfigContractAddress) {
+      return false;
+    }
+    const chainId = this.#router.chain.id;
+    const cached = this.#unifiedPermitByChain.get(chainId);
+    if (cached !== undefined && nowSeconds() < cached.expiresAt) {
+      return cached.supported;
+    }
```
zmalatrax: Looking back at `fhevm/sdk@0.14.x`, the fhevm/sdk actually probe the relayer "only once per client", so with that I think our cache is unnecessary, wdyt?
cf. https://github.com/zama-ai/fhevm/blob/v0.14.1-1/sdk/js-sdk/src/core/actions/base/canUseUnifiedDecryptionPermit.ts#L22
https://github.com/zama-ai/fhevm/blob/v0.14.1-1/sdk/js-sdk/src/core/modules/relayer/module/fetchFeatures.ts#L60-L70
https://github.com/zama-ai/fhevm/blob/v0.14.1-1/sdk/js-sdk/src/core/relayerFeatures/ensureRelayerFeatures-p.ts#L65-L74

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -155,6 +163,39 @@ export class CredentialService {
...
+   * {@link UNIFIED_PERMIT_CACHE_TTL_SECONDS}. A thrown probe is not stored,
+   * so the next grant can retry — and must not block issuance (caller signs
+   * V1 for that attempt). Never used for `WILDCARD_PERMIT` (always V2).
+   *
+   * V2 signing also requires the chain's `protocolConfigContractAddress` —
+   * without it, signing fails regardless of what the relayer reports, so we
+   * short-circuit here instead of probing the relayer only to fail at
+   * signing time.
+   */
+  async #canUseUnifiedPermit(): Promise<boolean> {
```
zmalatrax: what about making a #permitVersion that returns either `1` or `2` instead of calling this method in a ternary operator each time?

sth like

```ts
async #permitVersion(): Promise<1 | 2> {
  if (!this.#router.chain.protocolConfigContractAddress) return 1;
  return (await this.#router.relayer.canUseUnifiedDecryptionPermit()) ? 2 : 1;
}
```

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -208,17 +249,19 @@ export class CredentialService {
...
-        const widened = await this.#signPermit({ version: 1, chunk: widenedSet, keypair, scope });
+        const widened = await this.#signPermit({ version, chunk: widenedSet, keypair, scope });
         await swallow(
           "replace permit",
           () => this.#store.replace(scope, candidate.serializedPermit.signature, widened),
           this.#logger,
         );
         permissions[permissions.indexOf(candidate)] = widened;
       } else {
+        const version = (await this.#canUseUnifiedPermit()) ? 2 : 1;
```
zmalatrax: it's call in both if and else branches, it could be moved up as a single call

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts
eitjuh: Makes sense! Fixed

packages/sdk/src/credentials/credential-service.ts
eitjuh: Added a check

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -155,6 +155,31 @@ export class CredentialService {
...
+
+  /**
+   * A thrown probe is treated as unsupported for this call only, never
+   * cached — `@fhevm/sdk` already resolves and pins `canUseUnifiedDecryptionPermit`
+   * once per client, so caching it again here would just duplicate that.
+   */
+  async #permitVersion(): Promise<1 | 2> {
+    if (!this.#chainSupportsUnifiedSigning()) {
+      return 1;
+    }
```
zmalatrax: I don't think it's needed to separate ProtocolConfig check from permitVersion,

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -181,13 +206,21 @@ export class CredentialService {
...
     if (contracts === WILDCARD_PERMIT) {
-      const keypair = await this.#vault.getOrCreate(signerAddress);
       const scope = this.#permissionScope(signerAddress, delegator);
+      const keypair = await this.#vault.getOrCreate(signerAddress);
       const permissions = await this.#store.listUsableAndPrune(scope, keypair.publicKey);
       const existing = permissions.find(isWildcardPermission);
       if (existing) {
         return { keypair, permissions: [existing] };
       }
+      if (!this.#chainSupportsUnifiedSigning()) {
```
zmalatrax: ProtocolConfig already exists in 0.13, it's not a 0.14-specific contract so this check alone does not gate unified decryption, it should check both ProtocolConfig and probe the relayer (hence https://github.com/zama-ai/sdk/pull/684/changes#r3933051632)

I would call `#permitVersion` instead and throw if it's 1
```ts
if ((await this.#permitVersion()) === 1) {
  throw new UnifiedPermitNotSupportedError("...");
}
```

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts

zmalatrax (approved):

PR #701, opened 2026-09-01
fix(sdk): structured events + metadata for wallet signing/permit fai…

eitjuh:
# Summary

Wallet-signing failures during permit granting (`grantPermit`, `grantDelegationPermit`, `registerPermit`) were invisible to SDK observability — `CredentialService`/`Permits` never emitted any `ZamaSDKEvent`, so an integrator's `onEvent`/Sentry hook only ever saw them if they happened to catch the thrown error at the call site. The thrown errors themselves also carried little structured detail: the operation name was baked into free-text message prefixes, and the wallet/provider's RPC error code and error-class name were buried in an opaque `cause`.

This PR:

- Adds a `ZamaSDKEvents.PermitError` event (`PermitErrorEvent`, `PermitOperation`), emitted on every `grantPermit` / `grantDelegationPermit` / `registerPermit` failure — mirroring the existing `TransactionError` event for write transactions. This covers failures at both layers: the `Permits` namespace's own pre-flight guards (no signer configured, chain mismatch) and everything inside `CredentialService` (validation, permit-store reads/writes, wallet signing/verification) — via a `#failPermit` choke point in each class, so every failure path emits exactly once.
- Adds `operation`, `rpcCode`, and `walletErrorName` as structured fields on `SigningFailedError` / `SigningRejectedError`, extracted from the wallet/provider's raw error via two new chain-walking helpers in `utils/error.ts` (`extractRpcErrorCode`, `extractWalletErrorName`).
- Changes `wrapSigningError` to accept a structured `{ operation, description? }` context (instead of a free-text string), to return the classified error rather than throw it directly — mirroring `wrapDecryptError`/`wrapEncryptError` — and to pass an already-typed `ZamaError` through unchanged, so callers can emit the observability event before throwing without duplicating that check themselves.

**Deliberate exclusion:** credential/vault resolution (`vault.getOrCreate` / `vault.readStored`, and anything derived from them, e.g. `TransportKeyPairChangedError`, `KeyWrappingError`) stays outside the emit boundary on purpose — a pre-existing test (`zama-sdk-credentials-lifecycle.test.ts`) asserts those failures must never reach the public `onEvent` stream, since they can be adjacent to `transportKeyPairDerivationSecret` internals. That test passes unmodified.

Purely additive: no public method signatures changed, and existing thrown-error behavior for `grantPermit`/`grantDelegationPermit`/`registerPermit` is unchanged for callers that don't subscribe to `onEvent`.

## Scope

- [x] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [ ] Docs

## Validation

- `pnpm --filter @zama-fhe/sdk typecheck` (`tsc --noEmit`) — clean.
- `pnpm --filter @zama-fhe/sdk lint` (`oxlint`) — clean.
- `pnpm exec oxfmt --check` on all changed files — clean.
- `pnpm vitest run --project sdk` — 1436/1436 tests passing, including:
  - The original 4 tests asserting `PermitError` fires with the correct `operation`/`error` on `grantPermit`/`grantDelegationPermit`/`registerPermit` failures, and does **not** fire on success.
  - New tests covering the widened coverage: `grantPermit`/`grantDelegationPermit` emitting on a pre-signing failure (no signer configured) and on chain mismatch (via the public `sdk.permits.*` API), `registerPermit` emitting on chain-mismatch/expired-permit, and — critically — `registerPermit` **not** emitting for the vault-adjacent `TransportKeyPairChangedError`, confirming the credential-resolution exclusion holds.
  - The pre-existing `zama-sdk-credentials-lifecycle.test.ts` invariant test (unmodified), confirming `KeyWrappingError` from vault resolution still never reaches `onEvent`.
- `pnpm api-report` — regenerated, no public API diff (all new surface is `@internal`/private).
- `pnpm docs:check-links` / `pnpm llm:build` — clean; changelog and error-reference docs updated for the new fields/event.
- Extended `wrapSigningError`'s existing unit tests (`errors/__tests__/errors.test.ts`) to cover the new `rpcCode`/`walletErrorName` extraction and the `ZamaError` passthrough.

## Related

Closes [SDK-320](https://linear.app/zama/issue/SDK-320/featsdk-structured-events-metadata-for-wallet-signingpermit-failures).

Prompted by a production incident where a Super App user's `grantPermit` intermittently failed (`SigningFailedError` / `InvalidParamsRpcError`, "chainId should be same as current chainId") across several wallet connectors, eventually root-caused to a BigInt-serialization issue in one browser — a trace that took most of a day partly because the SDK gave no structured signal to work with (see SDK-320 for full context).

## Demo (if possible)

N/A — SDK-internal error/event model change, no UI surface.


packages/sdk/etc/sdk-query.api.md
```diff
@@ -1576,8 +1576,10 @@ export class ZamaSDK {
     terminate(): void;
 }
 
+// Warning: (ae-forgotten-export) The symbol "PermitErrorEvent" needs to be exported by the entry point index.d.ts
```
zmalatrax: looks like a missing a re-export; Same for underlying PermitOperation

Either add them to the list, or maybe export all the types from the file:
`export type * from "../events/sdk-events"` instead of explicit type list

packages/sdk/src/utils/error.ts
```diff
@@ -397,3 +397,27 @@ export function extractHttpStatus(error: unknown): number | undefined {
...
+/**
+ * A JSON-RPC / EIP-1193 numeric error code (e.g. `4001` for user rejection, or a
+ * wallet/provider code like `-32603`) found anywhere in a signing failure's cause
+ * chain. Gives a signing error a structured, groupable field instead of leaving
+ * the code buried in free-text `message` or an opaque nested `cause`.
+ */
+export function extractRpcErrorCode(error: unknown): number | undefined {
+  const node = findInErrorChain(error, (n) => typeof n.code === "number");
+  return node ? (node.code as number) : undefined;
+}
```
zmalatrax, line 415: We might want to dig further for the rpc error codes:
Viem has a map of known error codes, then map every other code into UnknownRpcError which is given a viem-specific error code of `-1` cf. https://github.com/wevm/viem/blob/3db6cdf15969e6b2ff9a2899e2ba65b6f90460f2/src/errors/rpc.ts#L58-L60

Current `extractRpcErrorCode` would then provide `-1` error code, swallowing the potential error code of the unknown rpc error (from viem pov)

This would only happen for error viem cannot classify though 🤔 (plain Object/Error thrown by a wallet? might be a rare edge case)

```suggestion
export function extractRpcErrorCode(error: unknown): number | undefined {
  const node = findInErrorChain(error, (n) => typeof n.code === "number" && n.code !== -1);
  return node ? (node.code as number) : undefined;
}
```

packages/sdk/etc/sdk-query.api.md
eitjuh: Fixed

packages/sdk/src/utils/error.ts
eitjuh: Fixed

packages/sdk/src/errors/signing.ts
```diff
@@ -1,43 +1,108 @@
...
+ * (see {@link extractRpcErrorCode}/{@link extractWalletErrorName}) so an
+ * integrator can fingerprint/alert on structured fields instead of parsing
+ * the message text.
  */
-export function wrapSigningError(error: unknown, context: string): never {
+export function wrapSigningError(
+  error: unknown,
+  context: { operation: string; description?: string },
+): ZamaError {
+  const { operation, description } = context;
```
zmalatrax: We should add the passthrough for already `ZamaError` errors as in `wrapEncryptError`, avoiding to have this branching in other methods

```ts
if (error instanceof ZamaError) {
  return error;
}
const { operation, description } = context;
```

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -349,10 +358,19 @@ export class CredentialService {
...
       });
     } catch (error) {
-      if (error instanceof ZamaError) {
-        throw error;
-      }
-      throw wrapSigningError(error, "registerPermit: signature verification failed");
+      const failure =
+        error instanceof ZamaError
+          ? error
+          : wrapSigningError(error, {
```
zmalatrax: https://github.com/zama-ai/sdk/pull/701/changes#r3915316337

packages/sdk/src/errors/index.ts
```diff
@@ -1,6 +1,6 @@
 export { ZamaError, ZamaErrorCode, isRetryable, retryAfterSeconds } from "./base";
 export { matchZamaError, type ErrorForCode } from "./match";
-export { SigningRejectedError, SigningFailedError } from "./signing";
+export { SigningRejectedError, SigningFailedError, type SigningErrorMetadata } from "./signing";
```
zmalatrax, line 3: is it required to be exported?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -620,16 +638,21 @@ export class CredentialService {
...
-        await this.#vault.evict(scope.signerAddress);
+        failure = error;
+      } else {
+        // A key pair the relayer can't re-derive (post KMS/TKMS rotation) is
+        // unusable: evict it so the next grantPermit regenerates a valid one, then
+        // surface the typed InvalidTransportKeyPairError via wrapSigningError.
+        if (error instanceof Error && isInvalidTransportKeyPairMessage(error.message)) {
+          await this.#vault.evict(scope.signerAddress);
+        }
+        failure = wrapSigningError(error, { operation, description: "credential signing failed" });
```
zmalatrax: the uppercase of the description was dropped

packages/sdk/src/errors/signing.ts
eitjuh: Fixed

packages/sdk/src/errors/index.ts
eitjuh: I temporarily removed the export and rebuilt: `api-extractor` immediately flagged `ae-forgotten-export` on `SigningErrorMetadata` in `sdk.api.md`. `SigningRejectedError`/`SigningFailedError`'s constructors both take `ErrorOptions & SigningErrorMetadata`, so the type is part of their exported constructor signature, the `.d.ts` needs a name for it.

packages/sdk/src/credentials/credential-service.ts
eitjuh: this is used like this, so then lowercase makes sense, no? `${operation}: ${description}: ...`

packages/sdk/src/credentials/credential-service.ts
eitjuh: fixed

pnpm-workspace.yaml
```diff
@@ -12,6 +12,13 @@ allowBuilds:
   sharp: false
   utf-8-validate: false
 
+# eslint-plugin-compat requires caniuse-lite at runtime but does not declare it.
```
zmalatrax, line 15: corresponding ticket on eslint-plugin-compat: https://github.com/amilajack/eslint-plugin-compat/issues/692

worth tracking in linear to remove this hack once fixed upstream

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -311,14 +357,22 @@ export class CredentialService {
...
-      throw new PreparedPermitExpiredError(
-        `registerPermit: the prepared permit's validity window (starting ${startTimestamp}, ` +
-          `${durationDays}d) has already elapsed — call preparePermit again.`,
+      throw this.#failPermit(
+        "registerPermit",
+        new PreparedPermitExpiredError(
+          `registerPermit: the prepared permit's validity window (starting ${startTimestamp}, ` +
+            `${durationDays}d) has already elapsed — call preparePermit again.`,
+        ),
+        "validation failed",
```
zmalatrax: Why do we wrap those 2 errors in `#failPermit`?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -348,35 +404,32 @@ export class CredentialService {
...
-        `registerPermit: expected a V1 signed permit, got version ${signedPermit.version}.`,
-      );
-    }
 
-    const serializedPermit = SerializedPermitSchema.parse(
-      await relayer.serializeSignedDecryptionPermit({ signedPermit }),
-    );
+      const serializedPermit = SerializedPermitSchema.parse(
+        await relayer.serializeSignedDecryptionPermit({ signedPermit }),
+      );
```
zmalatrax: Noticed it uses raw schema parsing while there is an helper function `parseSchema(schema, value)`
It would help in error handling as it maps the zod error to a `ZamaError`, properly handled by #failPermit after

Similar lines in #signPermit https://github.com/zama-ai/sdk/blob/6f786a2a420cbe5a5af8473bd308fe257b6841a9/packages/sdk/src/credentials/credential-service.ts#L691-L693

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -348,35 +404,32 @@ export class CredentialService {
...
+        serializedPermit,
+        startTimestamp: Number(signedPermit.eip712.message.startTimestamp),
+        durationDays: Number(signedPermit.eip712.message.durationDays),
+      };
+      const scope: PermissionScope = {
+        signerAddress,
+        chainId: activeChainId,
+        delegatorAddress: checksum(signedPermit.encryptedDataOwnerAddress),
+      };
+      await this.#store.replace(scope, serializedPermit.signature, permission);
```
zmalatrax: I think we need to `swallow` this store.replace, as done on the other storage write

https://github.com/zama-ai/sdk/blob/6f786a2a420cbe5a5af8473bd308fe257b6841a9/packages/sdk/src/credentials/credential-service.ts#L255-L259

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -157,17 +183,32 @@ export class CredentialService {
...
     const keypair = await this.#vault.getOrCreate(signerAddress);
     if (requested.length === 0) {
       return { keypair, permissions: [] };
     }
 
     const scope = this.#permissionScope(signerAddress, delegator);
-    const permissions = await this.#store.listUsableAndPrune(scope, keypair.publicKey);
+    let permissions: Permission[];
+    try {
+      permissions = await this.#store.listUsableAndPrune(scope, keypair.publicKey);
```
zmalatrax, line 217: Same as in https://github.com/zama-ai/sdk/pull/701/changes#r3924922480

Maybe better not to wrap?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -157,17 +183,32 @@ export class CredentialService {
...
+      delegator !== undefined ? "grantDelegationPermit" : "grantPermit";
+    let signerAddress: ChecksummedAddress;
+    try {
+      const signer = this.#requireSigner("grantPermit");
+      const account = signer.requireWalletAccount("grantPermit");
+      signerAddress = checksum(account.address);
+    } catch (error) {
+      throw this.#failPermit(operation, error, "grant failed");
+    }
     const requested = normalizeAddresses(contracts);
```
zmalatrax: `normalizeAddresses` calls `checksum` on contracts, so it should be in the try block, similar to `signerAddress = checksum(account.address);`, no?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -157,17 +183,32 @@ export class CredentialService {
...
+
+    // Credential/vault resolution — deliberately not wrapped; those failures
+    // must never reach onEvent.
     const keypair = await this.#vault.getOrCreate(signerAddress);
     if (requested.length === 0) {
       return { keypair, permissions: [] };
     }
 
     const scope = this.#permissionScope(signerAddress, delegator);
```
zmalatrax: same as https://github.com/zama-ai/sdk/pull/701/changes#r3925027421, it calls `checksum`

packages/sdk/src/errors/signing.ts
```diff
@@ -1,43 +1,107 @@
...
+ * return it rather than throw (mirroring {@link wrapDecryptError}/
+ * {@link wrapEncryptError}), so a caller can emit an observability event
+ * carrying the classified error before throwing it themselves. Enriches the
+ * result with `rpcCode`/`walletErrorName` extracted from `error`'s cause
+ * chain, and passes an already-typed `ZamaError` through unchanged.
  */
-export function wrapSigningError(error: unknown, context: string): never {
+export function wrapSigningError(
+  error: unknown,
+  context: { operation: string; description?: string },
```
zmalatrax: I'm wondering if the `description` field actually provides value here, it's for non-`ZamaError` only, and it's a bit redundant with the `operation` and the `originalMsg`, wdyt?

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts

packages/sdk/src/credentials/credential-service.ts

pnpm-workspace.yaml
eitjuh: Created: https://linear.app/zama/issue/SDK-332

packages/sdk/src/errors/signing.ts
eitjuh: Yeah agree, removed it 👍

packages/sdk/src/credentials/credential-service.ts
eitjuh: Routing these through #failPermit doesn't change the error itself, it just makes sure PermitError gets emitted on the SDK's onEvent stream. That's important because onEvent is the one place someone can observe every permit failure without depending on each call site's own error handling

packages/sdk/src/credentials/credential-service.ts
eitjuh: same as the other one:
> Routing these through #failPermit doesn't change the error itself, it just makes sure PermitError gets emitted on the SDK's onEvent stream. That's important because onEvent is the one place someone can observe every permit failure without depending on each call site's own error handling

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -335,11 +399,10 @@ export class CredentialService {
           "preparePermit again to rebind the signature request to the current key pair.",
       );
     }
-
     const transportKeyPair = await relayer.parseTransportKeyPair(keypair);
-    let signedPermit: SignedDecryptionPermit;
+
     try {
-      signedPermit = await relayer.parseSignedDecryptionPermit({
+      const signedPermit = await relayer.parseSignedDecryptionPermit({
```
zmalatrax: parseTransportKeyPair should be in the try block, no?

packages/sdk/src/credentials/credential-service.ts
zmalatrax: Okok, so we'd also want to have events emitted on validation errors? We might need a follow up on the other namespaces if we want so then

eitjuh: I believe so - I think right now there is only one namespace where it's missing - let's fix in a follow up PR

packages/sdk/src/credentials/credential-service.ts

zmalatrax (approved):
lgtm

PR #703, opened 2026-09-03
feat: expose wildcard delegation

eitjuh:
# Summary

Exposes wildcard delegation: a reserved ACL sentinel address that lets a delegator grant decryption rights across every confidential contract they own — current and future — in one call, instead of enumerating each contract. The ACL contract already supported this; the SDK just didn't surface it as a discoverable, typed constant.

- Exports `WILDCARD_CONTRACT` from `@zama-fhe/sdk`'s public surface (`packages/sdk/src/contracts/constants.ts`), sourced to match the value the underlying FHE SDK computes internally. Also top-level-exports `MAX_UINT64`/`MAX_UINT48`, closing a related gap where example apps hand-rolled the permanent-delegation sentinel instead of importing it.
- Documents the wildcard/per-contract redundancy caveat on `Delegations.delegateDecryption` (no batch-delegation API exists to detect the mix automatically, so it's called out in the JSDoc and the guide instead).
- Adds a "Delegate for all contracts (wildcard)" toggle to `DelegateDecryptionCard.tsx` in all 7 example apps and to the shared `delegation-panel.tsx` test component, replacing the hand-rolled sentinel constants those files used to carry.
- Adds a wildcard-delegation section to the delegated-decryption guide, including how it relates to (and composes with) `WILDCARD_PERMIT` at the permit layer.
- Adds unit tests (ABI-encoded call construction, service-level passthrough, react-hook passthrough) and two new Playwright scenarios covering the wildcard grant and decrypting a contract never named in the grant.

Purely additive — no existing exports or signatures changed.

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`
- [x] `@zama-fhe/test-components`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [x] Examples (`examples/`)
- [ ] CI/workflows
- [x] Docs

## Validation

- `pnpm typecheck` — clean across all 7 workspace packages.
- `pnpm test:run` — 1934/1934 unit tests pass (includes the new delegation/ACL/hook tests).
- `pnpm lint` (oxlint + ast-grep) — clean.
- `pnpm format:check` — clean.
- `pnpm api-report` / `pnpm api-report:check` — regenerated and verified; the diff to `packages/sdk/etc/sdk.api.md` is purely additive (`WILDCARD_CONTRACT`, `MAX_UINT64`, `MAX_UINT48`).
- `node scripts/docs/check-links.mjs` — all internal doc links/anchors resolve.
- `pnpm e2e:test` — **not completed**. The new Playwright scenarios are written (`test/playwright/tests/delegation.spec.ts`), but the local anvil devnet failed to deploy (`CleartextFHEVMExecutor` exceeds the EIP-170 size limit under the current `contracts/lib/fhevm` submodule state) — a pre-existing environment issue unrelated to this change. Needs a run in an environment where the e2e suite deploys cleanly.

## Related

- Closes [SDK-302](https://linear.app/zama/issue/SDK-302/feat-expose-rfc-017-wildcard-delegation)
- Plan: `engineering_plan-wildcard-delegation.md`

## Demo (if possible)

N/A — this is primarily an SDK/library change. The example-app addition is a checkbox toggle on the existing delegation form; it wasn't exercised in a running browser this session (typecheck and unit tests confirm it compiles and wires to `useDelegateDecryption` correctly, but no screenshot was captured).

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/docs/gitbook/src/guides/delegated-decryption.md
+++ b/docs/gitbook/src/guides/delegated-decryption.md
@@ -238,7 +238,7 @@ try {
 
 ### 6. Delegate for all contracts with the wildcard address (optional)
 
-Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress` to cover every confidential contract the delegator owns — current and future. This is a plain ACL delegation to a reserved sentinel address: the ACL honors it automatically during decryption, and `getStatus`/`getExpiry` account for it too, so checking delegation status for any specific contract still reflects a wildcard-only grant. `WILDCARD_CONTRACT` is only valid as `contractAddress` — passing it as `delegateAddress` throws `DelegationDelegateCannotBeWildcardError`.
+Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress`. The delegation then applies to any confidential contract, including ones deployed later. It does not widen access: the delegate can only read what the delegator can already read on each contract. `WILDCARD_CONTRACT` is only valid as `contractAddress`. `getStatus` and `getExpiry` for a specific contract also reflect a wildcard-only grant, so existing status checks keep working.
 
 {% tabs %}
 {% tab title="Core SDK" %}
--- a/docs/gitbook/src/guides/delegated-decryption.md
+++ b/docs/gitbook/src/guides/delegated-decryption.md
@@ -266,8 +266,8 @@ await delegate({ delegateAddress: "0xDelegate" });
 {% endtab %}
 {% endtabs %}
 
-{% hint style="warning" %}
-Once a delegate holds a wildcard grant, per-contract delegations to that same delegate are redundant — the wildcard already covers them. The SDK does not track prior grants across calls to detect the overlap, so avoid combining them yourself.
+{% hint style="info" %}
+Wildcard and per-contract delegations to the same delegate can coexist — `ACL.sol` honors both. If a per-contract grant would be redundant because the delegate already holds an active wildcard grant, `delegateDecryption` logs a warning; it still submits the transaction.
 {% endhint %}
 
 {% hint style="info" %}
--- a/docs/gitbook/src/guides/delegated-decryption.md
+++ b/docs/gitbook/src/guides/delegated-decryption.md
@@ -238,7 +238,7 @@ try {
 
 ### 6. Delegate for all contracts with the wildcard address (optional)
 
-Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress` to cover every confidential contract the delegator owns — current and future. This is a plain ACL delegation to a reserved sentinel address; `ACL.sol` recognizes it and honors it for any contract checked via `isHandleDelegatedForUserDecryption`, not just the literal sentinel.
+Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress` to cover every confidential contract the delegator owns — current and future. This is a plain ACL delegation to a reserved sentinel address: the ACL honors it automatically during decryption, and `getStatus`/`getExpiry` account for it too, so checking delegation status for any specific contract still reflects a wildcard-only grant. `WILDCARD_CONTRACT` is only valid as `contractAddress` — passing it as `delegateAddress` throws `DelegationDelegateCannotBeWildcardError`.
 
 {% tabs %}
 {% tab title="Core SDK" %}
--- a/docs/gitbook/src/guides/delegated-decryption.md
+++ b/docs/gitbook/src/guides/delegated-decryption.md
@@ -236,7 +236,45 @@ try {
 {% endtab %}
 {% endtabs %}
 
-### 6. Revoke delegation (optional)
+### 6. Delegate for all contracts with the wildcard address (optional)
+
+Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress` to cover every confidential contract the delegator owns — current and future. This is a plain ACL delegation to a reserved sentinel address; `ACL.sol` recognizes it and honors it for any contract checked via `isHandleDelegatedForUserDecryption`, not just the literal sentinel.
+
+{% tabs %}
+{% tab title="Core SDK" %}
+
+```ts
+import { WILDCARD_CONTRACT } from "@zama-fhe/sdk";
+
+await sdk.delegations.delegateDecryption({
+  contractAddress: WILDCARD_CONTRACT,
+  delegateAddress: "0xDelegate",
+});
+```
+
+{% endtab %}
+{% tab title="React SDK" %}
+
+```tsx
+import { WILDCARD_CONTRACT } from "@zama-fhe/sdk";
+import { useDelegateDecryption } from "@zama-fhe/react-sdk";
+
+const { mutateAsync: delegate } = useDelegateDecryption(WILDCARD_CONTRACT);
+await delegate({ delegateAddress: "0xDelegate" });
+```
+
+{% endtab %}
+{% endtabs %}
+
+{% hint style="warning" %}
+Once a delegate holds a wildcard grant, per-contract delegations to that same delegate are redundant — the wildcard already covers them. The SDK does not track prior grants across calls to detect the overlap, so avoid combining them yourself.
+{% endhint %}
+
+{% hint style="info" %}
+**A second, related "wildcard" exists at the permit layer.** This section covers wildcard _delegation_ — an on-chain ACL grant from the delegator that lets a delegate skip naming each contract. `@zama-fhe/sdk` also has a wildcard _permit_ (`WILDCARD_PERMIT`), which lets the delegate skip naming each contract on their own signed permit instead. The two compose: decrypting on someone's behalf needs an active delegation covering the contract (specific or wildcard, this section) _and_ a permit covering it, held by the delegate (auto-signed per contract on each delegated decrypt, or granted up front as `WILDCARD_PERMIT` via `sdk.permits.grantDelegationPermit`). Granting one does not grant the other. See [Wildcard (V2) permits](../concepts/permit-model.md#wildcard-v2-permits) for the permit side.
+{% endhint %}
+
+### 7. Revoke delegation (optional)
 
 {% tabs %}
 {% tab title="Core SDK" %}
@@ -262,7 +300,7 @@ await revoke({ delegateAddress: "0xDelegate" });
 {% endtab %}
 {% endtabs %}
 
-### 7. Handle errors (optional)
+### 8. Handle errors (optional)
 
 Delegation operations can throw several error types. The most common:
 
--- a/examples/example-bnb/src/components/DecryptAsCard.tsx
+++ b/examples/example-bnb/src/components/DecryptAsCard.tsx
@@ -4,7 +4,11 @@ import { useActionState } from "react";
 import { getAddress, formatUnits, isAddress } from "viem";
 import { useDelegationStatus, useDecryptBalanceAs } from "@zama-fhe/react-sdk";
 import type { Address, TokenWrapperPairWithMetadata } from "@zama-fhe/sdk";
-import { DelegationNotFoundError, DelegationExpiredError, MAX_UINT64 } from "@zama-fhe/sdk";
+import { DelegationNotFoundError, DelegationExpiredError } from "@zama-fhe/sdk";
+
+// Mirrors the SDK's MAX_UINT64 export — not yet in this app's pinned SDK
+// version. Switch to importing it once the pin is bumped.
+const MAX_UINT64 = 2n ** 64n - 1n;
 
 interface DecryptAsCardProps {
   token: TokenWrapperPairWithMetadata;
--- a/examples/example-bnb/src/components/DecryptAsCard.tsx
+++ b/examples/example-bnb/src/components/DecryptAsCard.tsx
@@ -6,9 +6,11 @@ import { useDelegationStatus, useDecryptBalanceAs } from "@zama-fhe/react-sdk";
 import type { Address, TokenWrapperPairWithMetadata } from "@zama-fhe/sdk";
 import { DelegationNotFoundError, DelegationExpiredError } from "@zama-fhe/sdk";
 
-// Mirrors the SDK's MAX_UINT64 export — not yet in this app's pinned SDK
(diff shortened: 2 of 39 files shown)


packages/sdk/src/namespaces/delegations.ts
```diff
@@ -46,7 +46,12 @@ export class Delegations {
    * window with a bounded internal retry, so you can decrypt right after this
    * resolves.
    *
-   * @param contractAddress - The confidential contract address to delegate on.
+   * @param contractAddress - The confidential contract address to delegate on. Pass
+   *   {@link WILDCARD_CONTRACT} instead of a specific address to delegate across every
+   *   confidential contract the delegator owns — current and future — rather than
+   *   enumerating each one. Granting the wildcard makes any per-contract delegation to
+   *   the same delegate redundant; the SDK does not track prior grants across calls to
+   *   detect that mix, so avoid combining them yourself.
```
zmalatrax: Not sure about this part, one can definitely have a mix of wildcard and per-contract delegations
cf. NatSpec from ACL contract: https://github.com/zama-ai/fhevm/blob/e7e7fecf10ea7a59c68603376caac2a308e18cca/host-contracts/contracts/ACL.sol#L320-L322

packages/sdk/src/namespaces/delegations.ts
```diff
@@ -46,7 +46,12 @@ export class Delegations {
    * window with a bounded internal retry, so you can decrypt right after this
    * resolves.
    *
-   * @param contractAddress - The confidential contract address to delegate on.
+   * @param contractAddress - The confidential contract address to delegate on. Pass
+   *   {@link WILDCARD_CONTRACT} instead of a specific address to delegate across every
+   *   confidential contract the delegator owns — current and future — rather than
```
zmalatrax: wdyt of
```ts
   * @param contractAddress - The confidential contract address to delegate on, or
   *   {@link WILDCARD_CONTRACT} to cover every contract with a single delegation.
```

packages/sdk/src/index.ts
```diff
@@ -65,6 +65,9 @@ export {
   ERC7984_WRAPPER_INTERFACE_ID,
 } from "./contracts";
 
+// Delegation/ACL sentinels
+export { MAX_UINT48, MAX_UINT64, WILDCARD_CONTRACT } from "./contracts";
```
zmalatrax: I don't think we need to export `MAX_UINT48`

test/playwright/tests/delegation.spec.ts
```diff
@@ -143,6 +144,54 @@ test("should delegate decryption to another account", async ({
...
+    abi: aclDelegateAbi,
+    functionName: "delegateForUserDecryption",
+    args: [account.address, WILDCARD_CONTRACT, 2n ** 64n - 1n],
+  });
+  await viemClient.waitForTransactionReceipt({ hash });
+
+  await page.goto(`/delegation?token=${contracts.cUSDT}&delegator=${account1.address}`);
+  await page.getByTestId("decrypt-delegate-button").click();
+
+  await expect(page.getByTestId("delegated-balance")).toContainText("0");
```
zmalatrax: This test actually short-circuits and does nothing, `decryptBalanceAs` early returns: https://github.com/zama-ai/sdk/blob/1d247d88e896982d0b7eb5e8d43beab0b6e28e7a/packages/sdk/src/token/token.ts#L224-L227
You'd need to fund the account v1 with cUSDT, and I'm not sure it'll work anyway if we don't update forge-fhevm to 0.14 as well

test/playwright/tests/delegation.spec.ts
```diff
@@ -143,6 +144,54 @@ test("should delegate decryption to another account", async ({
   await expect(page.getByTestId("delegate-success")).toContainText("Tx: 0x");
 });
 
+test("should delegate for all contracts via the wildcard checkbox", async ({
```
zmalatrax, line 147: >Wildcard grants are not honored by the SDK's delegated-decrypt preflight, so a wildcard-only delegation cannot be used end to end.
>
>Before any delegated decrypt, `DecryptionService` runs `#assertAllDelegationsActive`, which goes through `DelegationService.assertDelegationActive` -> `getStatus` -> `getDelegationExpiry`. That reads `getUserDecryptionDelegationExpirationDate(delegator, delegate, tokenAddress)`. On the ACL that getter is a raw per-contract storage read: a wildcard grant lives under the sentinel key, so for the token it returns 0 and we throw `DelegationNotFoundError` before ever reaching the relayer. Only `isHandleDelegatedForUserDecryption` ORs the wildcard row in ([ACL.sol#L527-L540](https://github.com/zama-ai/fhevm/blob/v0.14.0-0/host-contracts/contracts/ACL.sol#L527-L540)), and `@fhevm/sdk` uses that one, so the protocol would accept the decrypt. The rejection is ours.
>
>With account 1 funded, the new e2e test fails on exactly this:
>
>```bash
>Error: No active delegation from 0x7099...79C8 to 0xf39F...2266 for 0xDc64...F6C9
>```
>
>The same read backs `isDelegated`, `getStatus` and `useDelegationStatus`, so every status surface reports "no delegation" after a wildcard grant, including the example `DecryptAsCard` right after ticking the new checkbox.
>
>Fix in one place: in `DelegationService.getStatus`, read the token row and the `WILDCARD_CONTRACT` row in parallel (skip the second when the requested contract is already the wildcard) and report whichever is active. All callers above become wildcard-correct with no other changes. Since that preflight then already knows the wildcard expiry, the SDK-302 "warn when mixing wildcard and per-contract grants" item becomes a `logger.warn` in `delegateDecryption` rather than a docs caveat.

examples/example-bnb/src/components/DecryptAsCard.tsx
zmalatrax, line 1: Better to update the examples in a following PR imo, once exports are available as an alpha release
Also the examples are missing status/revoke for wildcard delegation

We can keep test-nextjs and test-vite though, they're relying on the workspace version, no need of a release for them

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -236,7 +236,45 @@ try {
...
+
+const { mutateAsync: delegate } = useDelegateDecryption(WILDCARD_CONTRACT);
+await delegate({ delegateAddress: "0xDelegate" });
+```
+
+{% endtab %}
+{% endtabs %}
+
+{% hint style="warning" %}
+Once a delegate holds a wildcard grant, per-contract delegations to that same delegate are redundant — the wildcard already covers them. The SDK does not track prior grants across calls to detect the overlap, so avoid combining them yourself.
```
zmalatrax: https://github.com/zama-ai/sdk/pull/703/changes#r3932432305

packages/sdk/src/namespaces/delegations.ts
eitjuh: Removed this

packages/sdk/src/namespaces/delegations.ts
eitjuh: I like it

packages/sdk/src/index.ts
eitjuh: Correct, it wasn't used anywhere anymore so I removed it 👍

test/playwright/tests/delegation.spec.ts
eitjuh: Fixed

test/playwright/tests/delegation.spec.ts
eitjuh: Fixed

docs/gitbook/src/guides/delegated-decryption.md
eitjuh: Fixed

packages/sdk/src/contracts/constants.ts
```diff
@@ -1,9 +1,15 @@
...
- * `setOperator`'s `until` argument is a `uint48` timestamp; this fits in a
- * JS number (< 2^53) and mirrors {@link MAX_UINT64}'s "permanent" role.
+ * Reserved wildcard sentinel address. Pass this as `contractAddress` to
+ * {@link Delegations.delegateDecryption} to delegate decryption rights across
+ * every confidential contract the delegator owns — current and future —
+ * instead of enumerating each contract individually. `ACL.sol` recognizes
+ * this address and honors it for any `(handle, contractAddress)` pair
+ * checked via `isHandleDelegatedForUserDecryption`, not just the literal
+ * sentinel itself.
  */
```
zmalatrax: this comment was not updated

examples/example-bnb/src/components/DecryptAsCard.tsx
eitjuh: Added https://linear.app/zama/issue/SDK-333/add-wildcard-delegation-to-example-apps-post-alpha

packages/sdk/src/contracts/constants.ts
eitjuh: Missed this one, thanks

packages/sdk/src/services/delegation-service.ts
zmalatrax, line 108: missed that in first review, there are missing check regarding wildcard permits:
delegate != WILDCARD_CONTRACT
There are done in fhevm/sdk: https://github.com/zama-ai/fhevm/blob/51cae14cb69cccf7e1be34baf1cd513c24ee3586/sdk/js-sdk/src/core/host-contracts/checkDelegation.ts#L50-L55

because they're part of the ACL contract: https://github.com/zama-ai/fhevm/blob/51cae14cb69cccf7e1be34baf1cd513c24ee3586/host-contracts/contracts/ACL.sol#L332-L334

We could add a new typed error mirroring the ACL one, `DelegationDelegateCannotBeWildcardError`

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -236,7 +236,45 @@ try {
 {% endtab %}
 {% endtabs %}
 
-### 6. Revoke delegation (optional)
+### 6. Delegate for all contracts with the wildcard address (optional)
+
+Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress` to cover every confidential contract the delegator owns — current and future. This is a plain ACL delegation to a reserved sentinel address; `ACL.sol` recognizes it and honors it for any contract checked via `isHandleDelegatedForUserDecryption`, not just the literal sentinel.
```
zmalatrax: this one as well, sry for the delay between both with constants.ts x) (https://github.com/zama-ai/sdk/pull/703/changes#r3934287438)

packages/sdk/src/services/delegation-service.ts
eitjuh: Good find, pushing a fix

docs/gitbook/src/guides/delegated-decryption.md
eitjuh: Fixed

packages/sdk/src/services/delegation-service.ts
eitjuh: Added a check for this

docs/gitbook/src/guides/delegated-decryption.md
```diff
@@ -236,7 +236,45 @@ try {
 {% endtab %}
 {% endtabs %}
 
-### 6. Revoke delegation (optional)
+### 6. Delegate for all contracts with the wildcard address (optional)
+
+Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress` to cover every confidential contract the delegator owns — current and future. This is a plain ACL delegation to a reserved sentinel address: the ACL honors it automatically during decryption, and `getStatus`/`getExpiry` account for it too, so checking delegation status for any specific contract still reflects a wildcard-only grant. `WILDCARD_CONTRACT` is only valid as `contractAddress` — passing it as `delegateAddress` throws `DelegationDelegateCannotBeWildcardError`.
```
zmalatrax: I wouldn't name the error here (`DelegationDelegateCannotBeWildcardError`), I'd rather list it under "### 8. Handle errors (optional)" with the other ones

```suggestion
Instead of granting delegation one contract at a time, pass `WILDCARD_CONTRACT` as `contractAddress`. The delegation then applies to any confidential contract, including ones deployed later. It does not widen access: the delegate can only read what the delegator can already read on each contract. `WILDCARD_CONTRACT` is only valid as `contractAddress`. `getStatus` and `getExpiry` for a specific contract also reflect a wildcard-only grant, so existing status checks keep working.
```

Also:
>The new error is missing from the three docs that enumerate delegation errors. Every sibling appears in guides/handle-errors.md line 46, reference/sdk/errors.md lines 38, 109, 651 and 659, and reference/sdk/delegation.md line 66. `DelegationDelegateCannotBeWildcardError` is only mentioned here

packages/sdk/src/contracts/constants.ts
```diff
@@ -1,9 +1,19 @@
...
+ * {@link Delegations.delegateDecryption} to delegate decryption rights across
+ * every confidential contract the delegator owns — current and future —
+ * instead of enumerating each contract individually.
+ *
+ * `ACL.sol`'s `isHandleDelegatedForUserDecryption` honors this address for
+ * any `(handle, contractAddress)` pair, not just the literal sentinel — but
+ * its raw per-contract expiry getter doesn't, so querying a specific contract
+ * directly misses a wildcard-only grant. {@link Delegations.getStatus} and
+ * {@link Delegations.getExpiry} fall back to the wildcard row themselves, so
+ * status reads stay correct either way.
```
zmalatrax: I was thinking of something lighter, in the same fashion as in https://github.com/zama-ai/sdk/pull/703#discussion_r3932447019

e.g.
```suggestion
/**
 * `contractAddress` covering every confidential contract with a single
 * delegation.
*/
```

docs/gitbook/src/guides/delegated-decryption.md
eitjuh: Fixed

packages/sdk/src/contracts/constants.ts
eitjuh: Fixed

zmalatrax (approved):
lgtm, ty!

PR #705, opened 2026-09-03
feat: unified-decryption-for-offline-permit-flow

eitjuh:
# Summary

Adds the V2 (unified) permit branch to the offline permit flow, closing [SDK-306](https://linear.app/zama/issue/SDK-306/feat-unified-decryption-for-offline-permit-flow) (sub-issue of SDK-292). Previously `CredentialService.preparePermit`/`registerPermit` only ever built/accepted V1 permits, even though the online `grantPermit` flow already branches between V1 and V2.

- `preparePermit` now prefers V2 (unified) permits whenever the connected chain supports them — mirroring `grantPermit` — and accepts `contracts: WILDCARD_PERMIT` to request a wildcard-scoped permit offline, the same way the online flow does.
- `registerPermit` accepts both V1 and V2 signed permits instead of hard-rejecting anything but V1, and correctly reads V2's `allowedContracts`/`durationSeconds` message fields and its top-level (post-sign) `delegatorAddress`.
- `PreparedPermit` is now a discriminated union (`PreparedPermitV1 | PreparedPermitV2`); `PreparePermitRequest.contracts` widens from `readonly Address[]` to `readonly Address[] | WildcardPermit`. Both changes are additive.
- Added `RelayerSDK.createUnsignedUnifiedDecryptionPermitEip712` (and its `FhevmRelayer`/mock implementations) as the signer-less counterpart to the existing V1 builder — confirmed by reading the installed `@fhevm/sdk@0.14.1-1` package directly, since the upstream naming wasn't documented anywhere else.
- Updated `docs/gitbook/src/concepts/permit-model.md` and `docs/gitbook/src/guides/offline.md` to cover the V2/wildcard offline path, and fixed a stale doc claim that a plain contract list to `grantPermit()` always signs a V1 permit (it already auto-upgrades to V2 when supported).

## Scope

- [x] `@zama-fhe/sdk`
- [ ] `@zama-fhe/react-sdk`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [x] Docs

## Validation

- `pnpm typecheck` — clean across all packages
- `pnpm lint` (oxlint + ast-grep) — clean
- `pnpm test:run` — 1950/1950 tests passing (added 12 new tests covering V2/wildcard offline prepare+register, delegation, expiry, tampered payloads, and unsupported-chain errors, plus a `PreparedPermit` type-drift guard)
- `pnpm docs:check-links` — all 111 pages resolve
- `pnpm --filter @zama-fhe/sdk build` — succeeds
- `pnpm api-report:check` — `etc/sdk.api.md` and `etc/sdk-query.api.md` regenerated, no `ae-forgotten-export` warnings

## Related

- Closes [SDK-306](https://linear.app/zama/issue/SDK-306/feat-unified-decryption-for-offline-permit-flow), sub-issue of [SDK-292](https://linear.app/zama/issue/SDK-292/feat-add-support-for-unified-decryption)
- Builds on the V1/V2 permit branching already landed earlier in this stack (`grantPermit`'s version selection, wildcard delegation)

## Demo (if possible)

N/A — SDK/API surface change, no UI.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


packages/sdk/src/credentials/credential-service.ts
```diff
@@ -320,19 +332,87 @@ export class CredentialService {
...
-        }),
-      ),
-    );
 
-    return { version: 1, eip712, signerAddress };
+    // A wildcard request always attempts V2 without a chain-support precheck;
+    // an unsupported chain surfaces as UnifiedPermitNotSupportedError instead.
+    // A specific-contract request checks chain support first and only
+    // upgrades to V2 once confirmed.
+    const version = isWildcard || (await this.#permitVersion()) === 2 ? 2 : 1;
```
zmalatrax: shouldn't it match the behavior in the "online path"?
It feels like a preparePermit(WILDCARD_PERMIT) can bypass the relayer probe "is unified decryption available?"
https://github.com/zama-ai/sdk/blob/ff26b20fb24191ede92a8c6499be4904682604d9/packages/sdk/src/credentials/credential-service.ts#L203-L217

zmalatrax: And if kept as is, it can be simplified to:
```ts
const version = isWildcard ? 2 : await this.#permitVersion()
```

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -415,24 +504,29 @@ export class CredentialService {
...
+            durationSeconds: Number(signedPermit.eip712.message.durationSeconds),
+          }
+        : {
+            version: 1,
+            keypairPublicKey: keypair.publicKey,
+            contractAddresses: normalizeAddresses(signedPermit.eip712.message.contractAddresses),
+            serializedPermit,
+            startTimestamp: Number(signedPermit.eip712.message.startTimestamp),
+            durationDays: Number(signedPermit.eip712.message.durationDays),
+          };
```
zmalatrax: There is a similar ternary op for the "online path", wdyt of refactoring this into a shared helper?
With a base from shared fields and then a small ternary to return the base + version-specific field (version, durationSeconds/durationDays..)
https://github.com/zama-ai/sdk/blob/ff26b20fb24191ede92a8c6499be4904682604d9/packages/sdk/src/credentials/credential-service.ts#L772-L788

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -320,19 +332,87 @@ export class CredentialService {
...
+      if (error instanceof Error && isUnsupportedUnifiedPermitMessage(error.message)) {
+        throw new UnifiedPermitNotSupportedError(
+          "preparePermit: V2 (unified) decryption permits — including wildcard permits — " +
+            "require protocol v0.14 or later on this network. The connected chain hasn't " +
+            "upgraded yet (or its ProtocolConfig contract address isn't configured on this " +
+            "FheChain). Prepare a V1 permit with an explicit contract list instead, or retry " +
+            "once the network upgrades.",
+          { cause: error },
+        );
+      }
```
zmalatrax: Similar to https://github.com/zama-ai/sdk/pull/705/changes#r3949106553
I'm wondering if it's worth refactoring into an helper in errors/credential.ts, with an `op: "grantPermit" | "preparePermit"` arg

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -320,19 +332,87 @@ export class CredentialService {
...
-    );
 
-    return { version: 1, eip712, signerAddress };
+    // A wildcard request always attempts V2 without a chain-support precheck;
+    // an unsupported chain surfaces as UnifiedPermitNotSupportedError instead.
+    // A specific-contract request checks chain support first and only
+    // upgrades to V2 once confirmed.
+    const version = isWildcard || (await this.#permitVersion()) === 2 ? 2 : 1;
+
+    const eip712 = await this.#createUnsignedPermitEip712({
```
zmalatrax: this could be worth being inlined, wdyt?

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -375,11 +455,16 @@ export class CredentialService {
...
+        ? Number(message.durationSeconds)
+        : Number(message.durationDays) * SECONDS_PER_DAY;
+    if (nowSeconds() >= startTimestamp + durationSeconds) {
+      const windowDescription =
+        parsed.version === 2 ? `${durationSeconds}s` : `${Number(message.durationDays)}d`;
       throw new PreparedPermitExpiredError(
         `registerPermit: the prepared permit's validity window (starting ${startTimestamp}, ` +
-          `${durationDays}d) has already elapsed — call preparePermit again.`,
+          `${windowDescription}) has already elapsed — call preparePermit again.`,
       );
```
zmalatrax: We could always print `durationSeconds` here instead of re-branching just for durationDays 🤔

packages/sdk/src/credentials/credential-service.ts
eitjuh: Fixed — `preparePermit` now runs the same chain-support precheck as `grantPermit`'s WILDCARD_PERMIT branch before ever building the typed data, so a wildcard request can no longer bypass the "is unified decryption available?" probe. Added a regression test asserting `canUseUnifiedDecryptionPermit` is called and `createUnsignedUnifiedDecryptionPermitEip712` is never reached when unsupported.

eitjuh: Addressed as part of the fix above — the precheck now runs unconditionally, so the version resolves to `isWildcard ? 2 : chainVersion` with no redundant ternary.

packages/sdk/src/credentials/credential-service.ts
eitjuh: Done — extracted a shared `buildPermission(base, versionFields)` helper in `permissions.ts` (base fields + a small discriminated-union ternary) and used it in both `registerPermit` and `#signPermit`.

packages/sdk/src/credentials/credential-service.ts
eitjuh: Done — extracted `unifiedPermitNotSupportedError(op, error)` into `errors/credential.ts` with an `op: "grantPermit" | "preparePermit"` arg, used by both `preparePermit`'s and `#signPermit`'s catch blocks.

packages/sdk/src/credentials/credential-service.ts
eitjuh: Inlined — now that the error-wrapping moved into the shared `errors/credential.ts` helper, the remaining body was small enough to fold directly into `preparePermit` (no more separate private method / options-bag).

packages/sdk/src/credentials/credential-service.ts
eitjuh: Done — always prints durationSeconds now, the durationDays re-branch is gone.

zmalatrax (approved):

PR #727, opened 2026-09-08
feat: extend KMS recovery & invalidation to V2 permits

eitjuh:
# Summary

Extends the SDK's existing KMS-recovery and revocation machinery to V2 (unified) and wildcard permits, and adds a new on-chain revocation mechanism:

- **On-chain decryption-signature invalidation.** New `sdk.permits.invalidateDecryptionSignatures(timestamp?)` (and `useInvalidateDecryptionSignatures`) calls `ACL.invalidateDecryptionSignaturesBefore`, so the KMS Connector itself starts rejecting any decryption request signed before the cutoff — the only mechanism that reaches an already-exfiltrated permit, not just this SDK's local cache. Use it on suspected signing-key compromise or a multisig (ERC-1271/Safe) owner rotation. On success, this signer's locally-cached permits for the current chain are cleared automatically. `IACL` doesn't declare the underlying contract function yet upstream ([zama-ai/fhevm#3800](https://github.com/zama-ai/fhevm/pull/3800), open); `contracts/src/IACLExtended.sol` is a local stopgap until that lands — see `UPSTREAM-REQUEST-fhevm-invalidateDecryptionSignaturesBefore.md` for the exact removal steps once it merges.
- **Wildcard-aware KMS-context recovery.** Fixed a bug where a wildcard (permissive) permit, if evicted during the SDK's automatic revoked-KMS-context recovery, was silently re-granted scoped to only the contracts the failing decrypt call happened to request — instead of restoring full wildcard coverage.
- **ERC-1271 (smart-contract-wallet) signatures on V2 permits.** Confirmed/tested that V2 permits' variable-length signature field round-trips a smart-wallet signature (e.g. a Safe multisig) the same way as a fixed-length EOA signature, through `grantPermit`, the offline `preparePermit`/`registerPermit` flow, and a rejected `isValidSignature` check.

Docs updated: permit model, security model, `ZamaSDK` reference, contract builders, a new `useInvalidateDecryptionSignatures` reference page, and errors reference.

## Scope

- [x] `@zama-fhe/sdk`
- [x] `@zama-fhe/react-sdk`
- [x] `@zama-fhe/test-components`
- [ ] `@zama-fhe/test-nextjs`
- [ ] `@zama-fhe/test-vite`
- [ ] Examples (`examples/`)
- [ ] CI/workflows
- [x] Docs

## Validation

- `pnpm typecheck`
- `pnpm lint`
- `pnpm test:run` (1976 tests)
- `pnpm test:integration`
- `pnpm api-report:check`
- `pnpm abi:check`
- `pnpm docs:check-links`

Not run: Playwright E2E / a live-KMS-Connector check for signature invalidation — the KMS Connector's rejection behavior for an invalidated signature isn't implemented/confirmed anywhere yet ([fhevm#3800](https://github.com/zama-ai/fhevm/pull/3800) only adds the `IACL` interface declaration, nothing connector-side), so there's no real target to test against yet.

## Related

- Closes [SDK-307](https://linear.app/zama/issue/SDK-307/feat-extend-kms-recovery-and-invalidation-to-v2-permits)
- Depends on (not merged): [zama-ai/fhevm#3800](https://github.com/zama-ai/fhevm/pull/3800) — see `UPSTREAM-REQUEST-fhevm-invalidateDecryptionSignaturesBefore.md` for the stopgap and its removal steps

## Demo (if possible)

No visual/UX redesign — the only UI change is a new "Invalidate All Signatures" button in the internal `test/test-components` permits panel, used for integration testing. No screenshot/GIF.


packages/sdk/src/credentials/credential-service.ts
```diff
@@ -615,6 +627,53 @@ export class CredentialService {
...
+   */
+  async invalidateDecryptionSignatures(timestamp?: Date): Promise<TransactionResult> {
+    const signer = this.#requireSigner("invalidateDecryptionSignatures");
+    const account = signer.requireWalletAccount("invalidateDecryptionSignatures");
+    const signerAddress = checksum(account.address);
+    const resolvedTimestamp = timestamp ? BigInt(Math.floor(timestamp.getTime() / 1000)) : 0n;
+    const acl = this.#router.relayer.chain.aclContractAddress;
+
+    const result = await submitTransaction({
+      operation: "invalidateDecryptionSignatures",
```
zmalatrax: > This tx call skips the ACL revert mapping every other ACL write gets, so `InvalidationTimestampTooLow` / `InvalidationTimestampInTheFuture` surface as a generic `TransactionRevertedError`:
> https://github.com/zama-ai/sdk/blob/4f07a22339d20eccadd1406901680e9d245fd7ed/packages/sdk/src/services/delegation-service.ts#L291-L325
>
> `#submitAclTransaction` is private on `DelegationService`, so two ways to get there:
>
> - Move the write into `DelegationService` (`invalidateDecryptionSignaturesBefore(signer, timestamp?)`), keep only the local clear in `CredentialService`, and let `Permits` orchestrate guard + write + clear, same shape as `Delegations`. This also drops the `provider` / `emitEvent` fields the PR adds to `CredentialService` (signing + storage until now) and the double account derivation (`requireAlignedWalletAccount` in `Permits`, then `requireWalletAccount` in the service). The service stops being delegation-only at that point, a follow-up rename to `acl-service` might fit best
> - Or lift `#submitAclTransaction` into a shared `utils/submit-acl-transaction.ts` and call it from `CredentialService`. Smaller move, but the extra plumbing stays in `CredentialService`.
> 
> Either way the revert map needs the two selectors, with the error classes next to the permit errors rather than in `errors/delegation.ts`, and `AclTransactionOperation` needs `"invalidateDecryptionSignatures"`:
> https://github.com/zama-ai/sdk/blob/4f07a22339d20eccadd1406901680e9d245fd7ed/packages/sdk/src/errors/acl-revert.ts#L31

I think that I prefer re-using the `DelegationService` approach, it reduces the wallet/provider plumbing duplication

packages/sdk/src/credentials/credential-service.ts
```diff
@@ -898,12 +957,17 @@ export class CredentialService {
     }
 
     const survivors = stored.filter((p) => !stale.has(p.serializedPermit.signature));
+    // A wildcard permit's `contractAddresses` is `[]` (a permissive scope), so
+    // flattening it into `prior` below would silently lose that coverage —
+    // check for it directly instead and re-request WILDCARD_PERMIT explicitly.
+    const hadWildcard = stored.some(isWildcardPermission);
     const prior = stored.flatMap((p) => p.contractAddresses);
     const recovery = this.#clearAndRegrantScope(
```
zmalatrax: we could merge `hadWildcard` and `prior` into a single param, and move the ternary op from the `#clearAndRegrantScope` helper here, e.g.
```ts
const regrant: WildcardPermit | ChecksummedAddress[] = stored.some(isWildcardPermission)
  ? WILDCARD_PERMIT
  : sortedUnion(prior, normalizeAddresses(contracts));
```

packages/sdk/src/credentials/permission-store.ts
```diff
@@ -161,6 +161,34 @@ export class PermissionStore {
...
   }
 
+  /**
+   * Delete every permission for the given signer on one chain, including every
+   * delegator scope on that chain. Other chains are left intact.
+   *
+   * On-chain signature invalidation is per host-chain ACL (`msg.sender` on that
+   * chain only), so locally evicting other chains would drop still-valid permits.
+   */
+  async clearAllForSignerOnChain(
```
zmalatrax, line 173: Isn't it a duplication of
https://github.com/zama-ai/sdk/blob/4f07a22339d20eccadd1406901680e9d245fd7ed/packages/sdk/src/credentials/permission-store.ts#L153-L162
https://github.com/zama-ai/sdk/blob/4f07a22339d20eccadd1406901680e9d245fd7ed/packages/sdk/src/credentials/permission-store.ts#L215-L228
and https://github.com/zama-ai/sdk/blob/4f07a22339d20eccadd1406901680e9d245fd7ed/packages/sdk/src/credentials/storage-keys.ts#L23-L25

I think it can be refactored to use a shared private helper

packages/react-sdk/src/permits/use-invalidate-decryption-signatures.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @example
+ * ```tsx
+ * const invalidate = useInvalidateDecryptionSignatures();
+ * invalidate.mutate({});
+ * ```
+ */
+export function useInvalidateDecryptionSignatures(
+  options?: UseMutationOptions<TransactionResult, Error, InvalidateDecryptionSignaturesParams>,
+) {
```
zmalatrax, line 30: nit but it leaks empty `mutate({})` in api docs, a similar issue happened with `useRevokePermits`, fixed with `| void`, cf. https://github.com/zama-ai/sdk/blob/4f07a22339d20eccadd1406901680e9d245fd7ed/packages/react-sdk/src/permits/use-revoke-permits.ts#L25

packages/sdk/src/credentials/credential-service.ts
eitjuh: I agree, fixed it

packages/sdk/src/credentials/permission-store.ts
eitjuh: Yes, fixed

packages/sdk/src/credentials/credential-service.ts

packages/react-sdk/src/permits/use-invalidate-decryption-signatures.ts
eitjuh: Fixed

packages/sdk/src/namespaces/permits.ts
```diff
@@ -327,4 +338,44 @@ export class Permits {
...
+   */
+  async invalidateDecryptionSignatures(timestamp?: Date): Promise<TransactionResult> {
+    const service = this.#requireCredentialService("invalidateDecryptionSignatures");
+    const signer = this.#requireSigner("invalidateDecryptionSignatures");
+    const account = await requireAlignedWalletAccount(
+      "invalidateDecryptionSignatures",
+      this.#signer,
+      this.#provider,
+    );
+    const signerAddress: ChecksummedAddress = checksum(account.address);
```
zmalatrax: type annotation can be removed

eitjuh: Fixed

zmalatrax (approved):
lgtm

PR #730, opened 2026-09-10
fix: relayers chains validation

zmalatrax:
Closes SDK-336

### DIFF
--- a/packages/sdk/src/config/__tests__/types.test-d.ts
+++ b/packages/sdk/src/config/__tests__/types.test-d.ts
@@ -1,12 +1,7 @@
 import { assertType, describe, expectTypeOf, test } from "vitest";
 import { mainnet, sepolia } from "../../chains";
-import type { AtLeastOneChain, FheChain } from "../../chains";
-import { createConfig } from "../create";
-import { createConfig as createViemConfig } from "../../viem/config";
-import { createConfig as createEthersConfig } from "../../ethers/config";
-import type { GenericProvider } from "../../types";
-import type { ZamaConfigBase, RelayerConfig, RelayersFor } from "../types";
-import type { EIP1193Provider, PublicClient, WalletClient } from "viem";
+import type { FheChain } from "../../chains";
+import type { ZamaConfigBase, RelayerConfig } from "../types";
 import type { ZamaConfigViem } from "../../viem/types";
 import type { ZamaConfigEthers } from "../../ethers/types";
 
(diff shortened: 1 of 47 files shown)


packages/sdk/src/config/__tests__/types.test-d.ts
```diff
@@ -65,3 +70,81 @@ describe("ZamaConfigBase (mapped relayers)", () => {
...
+        // @ts-expect-error: sepolia has a relayer but no chains entry
+        relayers: { [mainnet.id]: relayer, [sepolia.id]: relayer },
+        provider,
+      }),
+    );
+  });
+
+  test("rejects an orphaned relayer key in a map held in a variable", () => {
+    const relayers = { [mainnet.id]: relayer, [sepolia.id]: relayer };
+    // @ts-expect-error: sepolia has a relayer but no chains entry
```
eitjuh: This makes me think this PR shouldn't be `fix:` but `feat!:`.

```ts
const relayers = { [mainnet.id]: relayer, [sepolia.id]: relayer };
createConfig({ chains: [mainnet], relayers, provider });
```

This currently compiles fine, but with this PR, it doesn't compile fine anymore?

packages/sdk/src/config/resolve.ts
```diff
@@ -33,47 +34,37 @@ export interface ResolvedChainRelayer {
...
-  const result = new Map<number, ResolvedChainRelayer>();
 
-  for (const id of chainMap.keys()) {
-    const chain = chainMap.get(id);
-    const relayerConfig = relayerMap.get(String(id));
+  const orphaned = Object.keys(relayers)
+    .map(Number)
+    .filter((id) => !ids.includes(id));
+  if (orphaned.length > 0) {
+    logger.warn(
```
eitjuh, line 52: Should this be `logger.warnAlways` instead of `logger.warn`? Seems like we have it exactly for that reason?

packages/sdk/src/config/__tests__/types.test-d.ts
zmalatrax: Indeed, but at the same time this example would have simply thrown at runtime before, so it does not break any running program (was already broken at runtime)

However, considering that we now simply warn log at runtime, I'm wondering if we couldn't just drop the typecheck hardening (commit 2), it kinda contradicts the "it's fine, just dead relayer"

packages/sdk/src/config/resolve.ts
zmalatrax: I hesitated between both, I ended up choosing `logger.warn` rather than `logger.warnAlways` because it didn't felt like a fatal error that needs to be always logged; especially if the typecheck throws about it, it would have usually been fixed before reaching runtime imo (unless ts error silenced/ignored)

I'm fine refactoring to warnAlways if you have a strong opinion on this, and I'd lean towards it if we remove the typecheck (cf. the other comment https://github.com/zama-ai/sdk/pull/730#discussion_r3987349740)

packages/sdk/src/config/__tests__/types.test-d.ts
ankurdotb: Isn't this restoring expected behaviour rather than a change in behaviour? The key question for whether to include `!` is whether this warrants bumping version to v4.x vs staying within v3.x family (I think it's the latter)

zmalatrax: Agreeing with you Ankur, we should stay on v3.x.x here imo

eitjuh: Also agree 👍

eitjuh (dismissed):

packages/sdk/src/config/__tests__/types.test-d.ts
zmalatrax: reverted the typecheck in https://github.com/zama-ai/sdk/pull/730/commits/ef12cb36fc8b9b0955b1a693f0a21cfa5791bc8a

eitjuh (approved):

PR #732, opened 2026-09-10
feat(sidecar): expose SDK decryption with signer and storage channels

zmalatrax:
Closes SDK-322
Closes SDK-323
References SDK-327


packages/sdk/src/errors/revive.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/** Restores callback errors for SDK recovery; other codes produce a base error without subclass-specific fields. */
+export function reviveZamaError(
+  code: ZamaErrorCode,
+  message: string,
+  options: { retryable: boolean; retryAfter?: number },
+): ZamaError {
+  const Constructor = callbackErrors[code];
+  const error = Constructor ? new Constructor(message) : new ZamaError(code, message, options);
+  return Object.assign(error, options);
```
eitjuh: `Object.assign(error, options)` puts the wire's `retryable` on top of whatever the error class set for itself. Checked against the built SDK:

```js
const e = reviveZamaError(ZamaErrorCode.SigningRejected, "user rejected", { retryable: true });
e.retryable; // true — a normally-constructed SigningRejectedError is false
isRetryable(e); // true
```

`ZamaError.retryable` is `readonly` and its JSDoc says only `RelayerRequestFailedError` overrides it per-instance. `isRetryable`'s JSDoc goes further: "every `ZamaError` subclass declares its own retryability, so a consumer never has to hardcode a set of retryable codes." That's no longer true.

In the sidecar this value comes from the client app over the callback channel, so a buggy client can hand back a wallet rejection tagged retryable and anything trusting `isRetryable()` will treat "the user said no" as "try again."

Could we let the constructor decide, and only assign `retryAfter`?

```ts
const error = Constructor ? new Constructor(message) : new ZamaError(code, message, options);
return options.retryAfter === undefined
  ? error
  : Object.assign(error, { retryAfter: options.retryAfter });
```

Minor related thing: as written it also sets `retryAfter: undefined` on every revived error that didn't have one.

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+  rpc WarmTransportKeyPair(OperationRequest) returns (Empty);
+  rpc WarmTransportKeyPairScope(ScopeRequest) returns (Empty);
+  rpc RevokeTransportKeyPair(ScopeRequest) returns (Empty);
+}
+message Empty {}
+message GetInfoRequest {}
+message GetInfoResponse { string sdk_version = 1; }
+message WalletAccount { bytes address = 1; uint64 chain_id = 2; }
+message CreateContextRequest {
+  string config_json = 1;
```
eitjuh: Not a big fan of having json in proto files. Would be nice to get validation and generated types for this.
Could we do something like:

```proto
message ChainConfig {
  uint64 id = 1;
  optional string rpc_url = 2;
  optional ChainAuth auth = 3;
  optional uint64 gateway_chain_id = 4;
  optional string relayer_url = 5;
  optional bytes acl_contract_address = 6;
}
message ContextConfig {
  repeated ChainConfig chains = 1;
  optional uint64 chain_id = 2;
  optional uint64 permit_ttl = 3;
  optional uint64 transport_key_pair_ttl = 4;
  optional string transport_key_pair_scope = 5;
  optional uint64 registry_ttl = 6;
}
```

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
+syntax = "proto3";
+package zama.sdk.v1alpha1;
+option go_package = "github.com/zama-ai/sdk/clients/go/gen/zama/sdk/v1alpha1;sdkv1alpha1";
+service SidecarService {
```
eitjuh, line 7: No comments in this file. They'd become the doc comments in the generated Go, Rust and TS.
Can we at least comment the RPCs? Even if it just shows what SDK method it matches with exactly like here, developers would already have a lot more context: https://github.com/zama-ai/sdk/pull/732/changes#diff-ae1706c4818260d41095a2f58c105647c1b1debaece2b318c36e14c9a7f5d8fdR27

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+  rpc GetInfo(GetInfoRequest) returns (GetInfoResponse);
+  rpc CreateContext(CreateContextRequest) returns (CreateContextResponse);
+  rpc CloseContext(ContextRequest) returns (Empty);
```
eitjuh: I don't like this shared `Empty` too much, because at some moment we might want to add a field to it.
Can each RPC have its own response message instead? We can keep it empty today, but then we can always add fields to it if we ever wish to?

```suggestion
  rpc CloseContext(ContextRequest) returns (CloseContextResponse);
```

with (so similar to Empty, but have each rpc have it's own)
```
message CloseContextResponse {}
```

Same for the other rpc's that return `Empty`

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+  rpc HasDelegationPermit(DelegationContractsRequest) returns (HasPermitResponse);
+  rpc RevokePermits(RevokePermitsRequest) returns (Empty);
+  rpc ClearPermits(OperationRequest) returns (Empty);
+  rpc WarmTransportKeyPair(OperationRequest) returns (Empty);
+  rpc WarmTransportKeyPairScope(ScopeRequest) returns (Empty);
+  rpc RevokeTransportKeyPair(ScopeRequest) returns (Empty);
+}
+message Empty {}
+message GetInfoRequest {}
+message GetInfoResponse { string sdk_version = 1; }
```
eitjuh: should this have protocol_version too? (I am not sure myself, maybe not)

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+  WalletAccount account = 3;
+  StorageBinding storage = 4;
+  StorageBinding permit_storage = 5;
+}
+message CreateContextResponse { string context_id = 1; }
+message ContextRequest { string context_id = 1; }
+message UpdateAccountRequest { string context_id = 1; WalletAccount account = 2; }
+message Operation { string context_id = 1; string operation_id = 2; }
+message OperationRequest { Operation operation = 1; }
+message EncryptedInput { bytes encrypted_value = 1; bytes contract_address = 2; }
```
eitjuh: I think we can add additional buf validation and we get it for free in the other clients if we wish

```proto
bytes contract_address = 2 [(buf.validate.field).bytes.len = 20];
```

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+    string string_value = 3;
+    bool undefined_value = 4;
+    double number_value = 5;
+  }
+}
+message ClearEntry { bytes encrypted_value = 1; ClearValue value = 2; }
+message DecryptValuesRequest { Operation operation = 1; repeated EncryptedInput inputs = 2; optional double timeout_ms = 3; }
+message DecryptValuesResponse { repeated ClearEntry values = 1; }
+message DelegatedDecryptValuesRequest {
+  Operation operation = 1; repeated EncryptedInput inputs = 2;
```
eitjuh: Any reason for this weird formatting? (other lines too)

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+message UpdateAccountRequest { string context_id = 1; WalletAccount account = 2; }
+message Operation { string context_id = 1; string operation_id = 2; }
+message OperationRequest { Operation operation = 1; }
+message EncryptedInput { bytes encrypted_value = 1; bytes contract_address = 2; }
+message ClearValue {
+  oneof value {
+    string bigint_value = 1;
+    bool bool_value = 2;
+    string string_value = 3;
+    bool undefined_value = 4;
```
eitjuh:
```suggestion
    Empty undefined_value = 4;
```

small nit, but the value is already clear from `undefined_value` so I think this is more clear

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+    double number_value = 5;
+  }
+}
+message ClearEntry { bytes encrypted_value = 1; ClearValue value = 2; }
+message DecryptValuesRequest { Operation operation = 1; repeated EncryptedInput inputs = 2; optional double timeout_ms = 3; }
```
eitjuh: why `double`? Seems wrong to me? Others in this file also seem to have `double`, for example `max_concurrency`, which also doesn't seem to make sense to me

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+}
+message DecryptPublicValuesRequest { Operation operation = 1; repeated bytes encrypted_values = 2; optional double timeout_ms = 3; }
+message DecryptPublicValuesResponse { repeated ClearEntry values = 1; bytes abi_encoded_clear_values = 2; bytes decryption_proof = 3; }
+message DelegatedBatchDecryptValuesRequest {
+  Operation operation = 1; repeated EncryptedInput inputs = 2;
+  bytes delegator_address = 3; optional bytes account_address = 4;
+  optional double max_concurrency = 5; optional bool wait_for_propagation = 6;
+}
+message SdkError { string code = 1; string message = 2; bool retryable = 3; optional double retry_after_seconds = 4; }
+message BatchItem { bytes encrypted_value = 1; bytes contract_address = 2; ClearValue value = 3; SdkError error = 4; }
```
eitjuh: `value` and `error` are independent fields, so "both set" and "neither set" both decode fine and mean nothing. We have some validation in TS for this, but I think we should enforce this through the proto file through something like this:
```suggestion
message BatchItem {
  bytes encrypted_value = 1;
  bytes contract_address = 2;
  oneof result {
    ClearValue value = 3;
    SdkError error = 4;
  }
}
```

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+message PreparePermitRequest { Operation operation = 1; bytes signer_address = 2; repeated bytes contract_addresses = 3; optional bytes delegator_address = 4; optional double duration_days = 5; }
+message PreparePermitResponse { string prepared_permit_json = 1; }
+message RegisterPermitRequest { Operation operation = 1; string prepared_permit_json = 2; bytes signature = 3; }
+message ContractsRequest { Operation operation = 1; repeated bytes contract_addresses = 2; }
+message DelegationContractsRequest { Operation operation = 1; bytes delegator_address = 2; repeated bytes contract_addresses = 3; }
+message HasPermitResponse { bool has_permit = 1; }
+message ContractList { repeated bytes addresses = 1; }
+message RevokePermitsRequest { Operation operation = 1; ContractList contracts = 2; }
+message ScopeRequest { Operation operation = 1; string scope_id = 2; }
+message SignerReply { string operation_id = 1; string action_id = 2; bytes signature = 3; SdkError error = 4; }
```
eitjuh: same as https://github.com/zama-ai/sdk/pull/732/changes#r4015436054

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+message DelegatedBatchDecryptValuesRequest {
+  Operation operation = 1; repeated EncryptedInput inputs = 2;
+  bytes delegator_address = 3; optional bytes account_address = 4;
+  optional double max_concurrency = 5; optional bool wait_for_propagation = 6;
+}
+message SdkError { string code = 1; string message = 2; bool retryable = 3; optional double retry_after_seconds = 4; }
+message BatchItem { bytes encrypted_value = 1; bytes contract_address = 2; ClearValue value = 3; SdkError error = 4; }
+message DelegatedBatchDecryptValuesResponse { repeated BatchItem items = 1; }
+message PreparePermitRequest { Operation operation = 1; bytes signer_address = 2; repeated bytes contract_addresses = 3; optional bytes delegator_address = 4; optional double duration_days = 5; }
+message PreparePermitResponse { string prepared_permit_json = 1; }
```
eitjuh: The client only needs one thing out of this — the EIP-712 payload to sign, nested at `.eip712`. It then hands the whole string straight back to `RegisterPermit` untouched.

So today Go and Rust have to parse a blob they otherwise never look inside, just to reach one field.

Can we return the two pieces separately instead?

```proto
message PreparePermitResponse {
  bytes prepared_permit = 1;   // opaque — hand back to RegisterPermit as-is
  string typed_data_json = 2;  // the payload to sign
}
```

`typed_data_json` is the same shape we already send on `SignerAction` (line 81), so clients reuse the signing path they've already got. `RegisterPermitRequest` would take `bytes prepared_permit` to match.

Leaving the typed data itself as JSON is right, by the way — EIP-712 `domain` and `message` are open by design. Just worth a comment saying that's deliberate.

proto/zama/sdk/v1alpha1/sidecar.proto
```diff
@@ -0,0 +1,92 @@
...
+message SignerClientMessage { oneof message { ContextRequest attach = 1; SignerReply reply = 2; } }
+message SignerAction { string operation_id = 1; string action_id = 2; WalletAccount account = 3; string typed_data_json = 4; }
+message SignerReplyError { string operation_id = 1; string action_id = 2; SdkError error = 3; }
+message SignerActionCancelled { string operation_id = 1; string action_id = 2; }
+message SignerServerMessage { oneof message { Empty attached = 1; SignerAction action = 2; SignerReplyError reply_error = 3; SignerActionCancelled cancelled = 4; } }
+
+message StorageBinding { oneof backend { Empty memory = 1; string persistent = 2; string application = 3; } }
+enum StorageMethod { STORAGE_METHOD_UNSPECIFIED = 0; STORAGE_METHOD_GET = 1; STORAGE_METHOD_SET = 2; STORAGE_METHOD_DELETE = 3; }
+message StorageAction { string request_id = 1; string backend_id = 2; StorageMethod method = 3; string key = 4; bytes value = 5; }
+message StorageReply { string request_id = 1; optional bytes value = 2; SdkError error = 3; }
```
eitjuh:
```suggestion
message StorageReply {
  string request_id = 1;
  oneof result {
    bytes value = 2;
    Empty not_found = 3;
    SdkError error = 4;
  }
}
```

packages/sdk/src/errors/revive.ts
```diff
@@ -0,0 +1,19 @@
+import { ZamaError, ZamaErrorCode } from "./base";
+import { InvalidTransportKeyPairError } from "./credential";
+import { fatalBatchErrors } from "./fatal-batch";
+
+const callbackErrors: Partial<Record<ZamaErrorCode, new (message: string) => ZamaError>> = {
+  ...fatalBatchErrors,
+  [ZamaErrorCode.InvalidTransportKeyPair]: InvalidTransportKeyPairError,
+};
```
eitjuh: This covers just 7 errors, I guess that's intentional?

packages/sdk/src/index.ts
```diff
@@ -105,6 +106,7 @@ export {
   PreparedPermitExpiredError,
   RelayerRequestFailedError,
   retryAfterSeconds,
+  reviveZamaError,
```
eitjuh: This line makes `reviveZamaError` public: everyone who installs the SDK gets it, and we can't remove it later without a breaking release. But nothing inside `packages/sdk` calls it, the only caller is `sdk-sidecar`...

I think the function should exist, but can we export it from an internal entry point — `@zama-fhe/sdk/internal`, instead of the root? The sidecar imports the built package, so it does have to be exported somewhere, just not where app developers see it. If we'd rather keep it public, it needs JSDoc and a docs page, which I'd like to avoid?

packages/sdk-sidecar/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vitest/config";
+import { fileURLToPath } from "node:url";
+
+export default defineConfig({
+  root: fileURLToPath(new URL(".", import.meta.url)),
+  test: { name: "sdk-sidecar", include: ["test/**/*.test.ts"], environment: "node" },
+});
```
eitjuh, line 10: This doesn't pick up the `sharedResolve` aliases from the root config, so it resolves `@zama-fhe/sdk` to `dist` while every other project uses `src`. Ignore if in the stacked PRs

proto/zama/sdk/v1alpha1/sidecar.proto
zmalatrax: no, didn't enforce proper formatting

proto/zama/sdk/v1alpha1/sidecar.proto
zmalatrax: mmmh, it moved to `double` because of JS leniant `number`, but actual fields could be typed to integer.
Maybe `retry_after_seconds` could benefit from fractions but I'll move it to integer as well for now, unless someone really wants to have <1s retry-after config it's not an issue imho

proto/zama/sdk/v1alpha1/sidecar.proto
zmalatrax: indeed, iirc `oneof` still allow for "neither set" so we'll still keep a vit on ts validation, but def an improvement

proto/zama/sdk/v1alpha1/sidecar.proto
zmalatrax: mmmh not convinced either, I'd say that protocol version is already conveyed by the sdk_version, especially if we align the sidecar version with it (zama-fhe/sdk@3.6.0 <> "sidecar" v3.6.0)

packages/sdk/src/errors/revive.ts
zmalatrax: yes it covers the "fatal errors" as declared in the ts sdk, to preserve corresponding behavior as it could fail on a callback (so that it throws on the sidecar as well), as they block retryability, and some are part of recovery

But `InvalidTransportKeyPairError` is useless, it can be removed from this list

packages/sdk/src/index.ts
zmalatrax: Indeed, I'm ok with the `@zama-fhe/sdk/internal` export proposal

proto/zama/sdk/v1alpha1/sidecar.proto
zmalatrax: iiuc for annotations to be processed and have an actual validator check at runtime it requires something like https://github.com/bufbuild/protovalidate-es . Currently we only have ts-proto, I'm not sure if both are compatible. Adding custom annotations and enforcing them would be a great addition, but requires a bit more investigation first. I'd rather post-pone adding them

packages/sdk-sidecar/test/package-exports.test.mjs
```diff
@@ -0,0 +1,21 @@
+import assert from "node:assert/strict";
```
eitjuh, line 1: this test never runs?
 `vitest.config.ts:9` includes `test/**/*.test.ts`, it will never cover`.mjs`?

eitjuh: Never mind, I see it's covered by https://github.com/zama-ai/sdk/pull/734

eitjuh (approved):

packages/sdk-sidecar/src/encoding.ts
```diff
@@ -0,0 +1,87 @@
...
+export function bytes(value: `0x${string}`): Buffer {
+  return Buffer.from(hexToBytes(value));
+}
+export function clearValue(value: SdkClearValue | number): ClearValue {
+  switch (typeof value) {
+    case "bigint":
+      return { value: { $case: "bigintValue", bigintValue: value.toString() } };
+    case "number":
+      return {
+        value: { $case: "numberValue", numberValue: unsignedInteger(value, "Clear value") },
```
eitjuh, line 44: Can a clear value be a number over 4294967295? If so this throws?

PR #733, opened 2026-09-10
feat(sidecar): add Rust and Go clients with managed SDK contexts

zmalatrax:
References SDK-324
References SDK-325
References SDK-327


clients/rust/src/signer.rs
```diff
@@ -0,0 +1,204 @@
...
+            return Ok(());
+        }
+        let request = SigningRequest {
+            operation_id: action.operation_id,
+            action_id: action.action_id,
+            account: action
+                .account
+                .context("missing signer account")?
+                .try_into()?,
+            typed_data: serde_json::from_str(&action.typed_data_json)?,
```
eitjuh: Because of the `?`, when a client sends a broken request (`typed_data_json: "blablabla"`), it doesn't just fail the request, but it closes the connection. I think in Go we are parsing the error in the reply for the one request and keep the channel open, so I think we should do the same here

clients/rust/src/alloy.rs
```diff
@@ -0,0 +1,47 @@
...
+        })?;
+        let signature = self.0.sign_hash(&hash).await.map_err(|_| {
+            signing_error("Wallet signing failed. Check the wallet connection and approval.")
+        })?;
+        Ok(signature.as_bytes().to_vec())
+    }
+}
+fn signing_error(message: &str) -> SdkError {
+    SdkError {
+        code: "SIGNING_FAILED".into(),
```
eitjuh: Every error here becomes `SIGNING_FAILED`, also when a user rejects a transaction.
For Go, it seems like we have `ErrSigningRejected` (`errors.go:75`), can we do the same for Rust so apps can distinguish between "signing rejected" and truly "failed"?

zmalatrax: indeed, ty! Reworked the error handling to classify known errors (notably added `CHAIN_MISMATCH`), and preserve `SdkError` ones

eitjuh (approved):

PR #734, opened 2026-09-10
feat(sidecar): add typed Docker examples and native CI

zmalatrax:
Closes SDK-324
Closes SDK-325
References SDK-327


.github/workflows/sidecar.yml
```diff
@@ -0,0 +1,120 @@
...
+          go-version-file: clients/go/go.mod
+          cache-dependency-path: clients/go/go.sum
+      - name: Install repository toolchain
+        run: rustup show
+      - name: Regenerate TypeScript and Go protobuf bindings
+        run: |
+          pnpm exec buf format --diff --exit-code
+          pnpm sidecar:generate
+          sh clients/go/generate.sh
+          git diff --exit-code -- packages/sdk-sidecar/src/generated clients/go/gen
```
eitjuh, line 118: This catches "someone changed the proto and forgot to regenerate", but not "someone changed the proto in a way that breaks existing clients".

`buf.yaml:14` already sets up `breaking: use: [FILE]`, but nothing ever runs it

Can we add a `buf breaking` step here? Non-blocking is fine while we're on `v1alpha1`, but we have to prep for getting it to stable

eitjuh (approved):

PR #744, opened 2026-09-11
docs(sidecar): add Rust quick start guide

eitjuh:
## Summary
- Adds a Rust getting-started guide mirroring the TS SDK's quick-start.md, covering starting the sidecar, adding the client as a git dependency, and connecting/decrypting a value.
- Not wired into SUMMARY.md nav yet since the sidecar is still a prototype — link it once the crate/image are published.

## Test plan
- [ ] Proofread against the current sidecar API (builder.rs / decryption.rs / types.rs) for drift
- [ ] Confirm pinned Alloy versions in the snippet still match clients/rust/Cargo.toml

Note: pre-commit typecheck was skipped for this docs-only commit — `sidecar-demo` currently fails `pnpm typecheck` in `@zama-fhe/sdk-sidecar` due to unrelated, pre-existing API drift against `@zama-fhe/sdk` (see `handlers.ts`, `callback-errors.ts`, `sdk.ts:45`).

### DIFF
--- a/docs/gitbook/src/tutorials/quick-start-rust.md
+++ b/docs/gitbook/src/tutorials/quick-start-rust.md
@@ -1,0 +1,139 @@
+---
+title: Quick start (Rust)
+description: Decrypt a confidential value from a native Rust application, through the SDK sidecar.
+---
+
+# Quick start (Rust)
+
+{% hint style="warning" %}
+**Prototype status.** The Rust client and the sidecar it talks to are an early prototype, not a released product. There is no published crate on crates.io and no published Docker image yet — you build both from a checkout of the [`zama-ai/sdk`](https://github.com/zama-ai/sdk) repository. The API described here can change. See the [sidecar trust boundary](../../../../packages/sdk-sidecar/SECURITY.md) before using it with anything beyond a test wallet.
+{% endhint %}
+
+Rust doesn't run the SDK directly. Instead, a small server called the **sidecar** runs `@zama-fhe/sdk` inside a Node.js process, and your Rust application talks to it over a private local socket. Your wallet's private key stays in your Rust application; encrypted values and decryption happen inside the sidecar. This guide gets you from a running sidecar to a decrypted confidential balance.
+
+## Prerequisites
+
+- Docker and Docker Compose
+- Rust 1.98+ and Cargo
+- A Linux host, or Docker Desktop for local development only — Docker Desktop on macOS cannot share a Unix socket between a container and a native host process, so on macOS the Rust side also needs to run in a container
+- A Sepolia RPC URL and a **test wallet** with a small amount of Sepolia ETH (never use a wallet holding real funds)
+
+## 1. Start the sidecar
+
+Clone the repository and start the sidecar with Docker Compose. This is the only way to run it today — there's no standalone image to pull yet.
+
+```sh
+git clone https://github.com/zama-ai/sdk.git
+cd sdk
+pnpm install
+cp -n .env.sidecar.example .env.sidecar.local
+chmod 600 .env.sidecar.local
+```
+
+Fill in `.env.sidecar.local` with your Sepolia RPC URL, wallet address and test wallet private key. Leave `RELAYER_API_KEY` empty to use the public Sepolia relayer.
+
+```sh
+export SIDECAR_UID="$(id -u)"
+export SIDECAR_GID="$(id -g)"
+
+docker compose --env-file .env.sidecar.local \
+  -f packages/sdk-sidecar/compose.yaml up --wait sidecar
+```
+
+Check that it started:
+
+```sh
+docker compose -f packages/sdk-sidecar/compose.yaml ps
+```
+
+{% hint style="info" %}
+The sidecar has no process-wide chain or wallet. Every Rust `Client` you connect creates its own SDK context with its own chain configuration and signer, so one running sidecar can serve multiple applications.
+{% endhint %}
+
+## 2. Add the client to your project
+
+The Rust client isn't published, so pull it in as a Git dependency and pin it to match the versions the sidecar was generated against:
+
+```toml
+[dependencies]
+zama-sdk-sidecar = { git = "https://github.com/zama-ai/sdk.git", features = ["alloy"] }
+tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+alloy-signer-local = { version = "=1.8.3", default-features = false, features = ["zeroize"] }
+```
+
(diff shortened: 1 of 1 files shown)


docs/gitbook/src/tutorials/quick-start-rust.md
ankurdotb, line 1: Where do you see this page being shown on Gitbook?

- One option is to include all contents in current quick start page, but that might be lot.
- The other option (see other comment) is to have a separate sidebar section for this method, and reference it from the [existing quick start](https://docs.zama.org/protocol/sdk/getting-started/quick-start) with a callout on top for other languages? Would also reference this in wallet/exchange guide
- Do we want to be publicly describing this as a "Sidecar" in the docs/tabs, or something else?
- I would say this feature is in Beta status in the callout

**General note**: Use relative links not absolute, so that it shows up properly on Gitbook AND on Github, including on different branches

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
+---
+title: Quick start (Rust)
+description: Decrypt a confidential value from a native Rust application, through the SDK sidecar.
+---
+
+# Quick start (Rust)
+
+{% hint style="warning" %}
+**Prototype status.** The Rust client and the sidecar it talks to are an early prototype, not a released product. There is no published crate on crates.io and no published Docker image yet — you build both from a checkout of the [`zama-ai/sdk`](https://github.com/zama-ai/sdk) repository. The API described here can change. See the [sidecar trust boundary](../../../../packages/sdk-sidecar/SECURITY.md) before using it with anything beyond a test wallet.
```
ankurdotb, line 9: Good note/callout but one we can solve next week but actually wiring up those in CI. Do you want to keep it Github only for now?

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
...
+{% endhint %}
+
+Rust doesn't run the SDK directly. Instead, a small server called the **sidecar** runs `@zama-fhe/sdk` inside a Node.js process, and your Rust application talks to it over a private local socket. Your wallet's private key stays in your Rust application; encrypted values and decryption happen inside the sidecar. This guide gets you from a running sidecar to a decrypted confidential balance.
+
+## Prerequisites
+
+- Docker and Docker Compose
+- Rust 1.98+ and Cargo
+- A Linux host, or Docker Desktop for local development only — Docker Desktop on macOS cannot share a Unix socket between a container and a native host process, so on macOS the Rust side also needs to run in a container
+- A Sepolia RPC URL and a **test wallet** with a small amount of Sepolia ETH (never use a wallet holding real funds)
```
ankurdotb, line 19: Seems a bit too AI. We don't make this callout on normal SDK quickstart setup, and implies Sepolia only. Should probably make more generic?

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
...
+
+```sh
+git clone https://github.com/zama-ai/sdk.git
+cd sdk
+pnpm install
+cp -n .env.sidecar.example .env.sidecar.local
+chmod 600 .env.sidecar.local
+```
+
+Fill in `.env.sidecar.local` with your Sepolia RPC URL, wallet address and test wallet private key. Leave `RELAYER_API_KEY` empty to use the public Sepolia relayer.
```
ankurdotb, line 33: Isn't there a default? Seems like a configuration option

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
...
+    // A test wallet — never a wallet holding real funds.
+    let signer: PrivateKeySigner = "0xYOUR_TEST_WALLET_PRIVATE_KEY".parse()?;
+    let account = WalletAccount {
+        address: signer.address(),
+        chain_id: 11_155_111, // Sepolia
+    };
+
+    let config = SdkConfig::from_chains(
+        account.chain_id,
+        vec![ChainConfig::new(account.chain_id, "https://your-sepolia-rpc.example")],
```
ankurdotb, line 88: Shouldn't this be reading from the env file provided during compose up?

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
...
+        _ => anyhow::bail!("unexpected decrypted value type"),
+    }
+    Ok(())
+}
+```
+
+The first call to `decrypt_values` for a given account prompts an EIP-712 wallet signature over the sidecar's signer channel — the SDK caches the resulting permit in whichever storage backend you configured, so later calls don't re-prompt.
+
+{% hint style="info" %}
+**Storage.** The example above uses in-memory storage, which is lost when your process exits. To keep credentials across restarts, either point the sidecar at its own SQLite volume (`PersistentStorage("your-app")`) or supply your own backend by implementing `NativeStorage` — for example, to store the opaque credential bytes in your own database. See [Choose credential storage](https://github.com/zama-ai/sdk/blob/main/packages/sdk-sidecar/README.md#choose-credential-storage) for the options.
```
ankurdotb, line 130: See previous comment on relative vs absolute links

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
...
+{% endhint %}
+
+## Next steps
+
+- [Full sidecar README](https://github.com/zama-ai/sdk/blob/main/packages/sdk-sidecar/README.md) — storage choices, transport limits, reconnection, and the development/test workflow
```
ankurdotb, line 135: Think of Gitbook as the primary destination for these docs. I would not refer to this as a "README", as we don't call it this on the other docs pages.

- I think the right place to fit this is either all in one page under "Guides" section. Or do you think a "Support in other languages" section in the sidebar?

docs/gitbook/src/tutorials/quick-start-rust.md
```diff
@@ -0,0 +1,139 @@
...
+- [Sidecar trust boundary](https://github.com/zama-ai/sdk/blob/main/packages/sdk-sidecar/SECURITY.md) — what the sidecar can see, and what it can't
+- [Wire contract](https://github.com/zama-ai/sdk/blob/main/proto/README.md) — every supported RPC and the SDK method it maps to
```
ankurdotb, line 137: Interesting one on whether to show in Gitbook or not, it's quite niche so makes sense but still make link relative?

ankurdotb (changes requested):
Lots of comments are related to how the docs show up. Note that Gitbook creates preview on PR branches, so if you login you can take a look (lmk if you need access)

PR #746, opened 2026-09-11
docs(changelog): refresh beta raw-material scaffold

ankurdotb:
## Summary

Re-runs `pnpm docs:changelog`, which hadn't been re-run since the branch-model restructure (`prerelease` → `main`/`beta`/`alpha`) landed. The changelog page's plain-language sections were already current, but the "raw material" scratch block (regenerated each run, dropped on promotion) was stale/missing, so it's refreshed here along with the LLM corpus (`llms-full.txt`).

Also confirmed `pnpm docs:check-target beta` passes — the committed doc URLs already correctly target the `beta` GitBook space, so no retargeting was needed there.

## Verification

- `pnpm docs:check-target beta` → passes
- `pnpm docs:check-links` → 112 pages, all internal links/anchors resolve
- `pnpm llm:check` → passes clean post-commit

## Note on the live "beta" GitBook page appearing blank

Repo-side, everything checks out: the beta changelog page has real drafted content, and doc URLs correctly target `beta`. That points the likely cause elsewhere — GitBook's own space→branch git-sync setting for the Beta space may still be pointed at the old `prerelease` branch (removed in the rename to the three-branch model), which is a GitBook admin-panel setting outside this repo and outside what I can check or fix from here. Worth having someone with GitBook admin access confirm/update the synced branch for that space.

Separately, per the SDK Release Process doc: the `sdk-changelog` skill itself (lives in `zama-marketplace`, not this repo) still has old `alpha`-cycle terminology in its own `SKILL.md`/templates from before the branch rename — that's already being tracked as a separate fix in that repo.


ankurdotb:
Closing per repo owner — will be handled separately.

PR #780, opened 2026-09-16
feat(sidecar): expose credentials and configuration options

zmalatrax:
Closes SDK-362


packages/sdk-sidecar/src/config-options.ts
```diff
@@ -0,0 +1,176 @@
...
+  return Object.hasOwn(supported, value);
+}
+
+function supportedValue<Value extends string>(
+  value: string,
+  values: Readonly<Record<Value, true>>,
+  name: string,
+): Value {
+  if (!isSupported(value, values)) {
+    throw new ConfigurationError(`Unsupported ${name}.`);
```
eitjuh: Maybe 
 ````suggestion
     throw new ConfigurationError(
       `Unsupported ${name} "${value}". Supported values: ${Object.keys(values).join(", ")}.`,
     );
 ````

packages/sdk-sidecar/src/sdk.ts
```diff
@@ -1,23 +1,35 @@
...
+} from "@zama-fhe/sdk";
+import { createHttpProvider } from "./provider.js";
 import type { StorageManager } from "./storage-manager.js";
 import { parseContextConfig } from "./sdk-config.js";
 import type { ContextFactory } from "./runtime.js";
 
+const noop = () => {};
+const stderrLogger: GenericLogger = {
+  error: (message) => process.stderr.write(`${message}\n`),
+  warn: (message) => process.stderr.write(`${message}\n`),
```
eitjuh: Should this also have a `data` prop (for example from `keypair-vault.ts`)
( `this.#logger.error("transport key pair unwrap failed unexpectedly", { key, error });` )

clients/rust/examples/balance/support.rs
```diff
@@ -52,32 +51,102 @@ impl Settings {
...
-        .confidentialBalanceOf(settings.account.address)
-        .call()
-        .await?;
-    Ok((name, encrypted))
+
+fn optional_bool(values: &HashMap<String, String>, key: &str) -> Result<Option<bool>> {
+    values
+        .get(key)
+        .filter(|value| !value.is_empty())
+        .map(|value| value.parse().map_err(|_| anyhow::anyhow!("invalid {key}")))
```
eitjuh: nit for folks writing `SDK_SINGLE_THREAD=1` in their env
 ````suggestion
         .map(|value| match value.to_ascii_lowercase().as_str() {
             "true" | "1" = Ok(true),
             "false" | "0" = Ok(false),
             _ = Err(anyhow::anyhow!("invalid {key}")),
         })
 ````

clients/go/config_options.go
```diff
@@ -0,0 +1,214 @@
...
+
+func (v PinnedModuleVersions) wire() *pb.ModuleVersions {
+	return &pb.ModuleVersions{Selection: &pb.ModuleVersions_Pinned{Pinned: &pb.PinnedModuleVersions{
+		Tfhe: v.TFHE, Kms: v.KMS, CheckCompatibility: v.CheckCompatibility,
+	}}}
+}
+
+// ProcessRuntime is applied once per sidecar process; the first SDK config wins.
+type ProcessRuntime struct {
+	WasmAssetLoadMode *string
```
eitjuh: We give relayers typed constants (`RelayerNode` / `RelayerCleartext`), but the WASM load mode and the TFHE/KMS versions are plain strings. The sidecar has a fixed list of allowed values for all of these, so today users only find out they typed it wrong after a round trip. Can we add constants for these too, the same way we did for the relayer type?

clients/go/config_options.go
```diff
@@ -0,0 +1,214 @@
...
+type AutoModuleVersions struct{}
+
+func (AutoModuleVersions) wire() *pb.ModuleVersions {
+	return &pb.ModuleVersions{Selection: &pb.ModuleVersions_Auto{Auto: &pb.Empty{}}}
+}
+
+type PinnedModuleVersions struct {
+	TFHE               *string
+	KMS                *string
+	CheckCompatibility *string
```
eitjuh: Same as https://github.com/zama-ai/sdk/pull/780/changes#r4029308846

clients/rust/src/config_options.rs
```diff
@@ -0,0 +1,240 @@
...
+            selection: Some(selection),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct PinnedModuleVersions {
+    pub tfhe: Option<String>,
+    pub kms: Option<String>,
+    pub check_compatibility: Option<String>,
```
eitjuh: Same as https://github.com/zama-ai/sdk/pull/780/changes#r4029308846

clients/rust/src/config_options.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+/// Applied once per sidecar process; the first SDK config wins.
+#[derive(Clone, Debug, Default)]
+pub struct ProcessRuntime {
+    pub wasm_asset_load_mode: Option<String>,
```
eitjuh: Same as https://github.com/zama-ai/sdk/pull/780/changes#r4029308846

clients/rust/src/config.rs
```diff
@@ -79,6 +88,8 @@ impl SdkConfig {
         Self {
             chain_id,
             chains,
+            process_runtime: None,
+            relayers: Default::default(),
```
eitjuh: Small nit: `relayers: Default::default()` is just `None` here, and the line above it writes `None` directly. Can we use `None` so both read the same?

```suggestion
            relayers: None,
```

clients/go/config_options.go
```diff
@@ -0,0 +1,246 @@
...
+
+type AutoModuleVersions struct{}
+
+func (AutoModuleVersions) wire() *pb.ModuleVersions {
+	return &pb.ModuleVersions{Selection: &pb.ModuleVersions_Auto{Auto: &pb.Empty{}}}
+}
+
+type TfheVersion string
+
+type KmsVersion string
```
eitjuh, line 19:
```suggestion
type TfheVersion string

const (
	TfheVersionV1_5_3 TfheVersion = "1.5.3"
	TfheVersionV1_6_2 TfheVersion = "1.6.2"
)

type KmsVersion string

const (
	KmsVersionV0_13_10   KmsVersion = "0.13.10"
	KmsVersionV0_13_20_0 KmsVersion = "0.13.20-0"
)
```

eitjuh (approved):
have a small nit, feel free to fix it or not

clients/rust/src/config_options.rs
```diff
@@ -0,0 +1,324 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub struct ProviderOptions {
+    pub headers: Option<BTreeMap<String, String>>,
+    pub timeout: Option<u32>,
+    pub retry_count: Option<u32>,
+    pub retry_delay: Option<u32>,
+    pub batch: Option<ProviderBatch>,
+    pub polling_interval: Option<u32>,
+}
```
eitjuh, line 176: nit for redacting the header values too - if you do println!("{config:?}"), it can print the RPC API key header

clients/go/config_options.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+type ProviderOptions struct {
+	Headers         map[string]string
+	Timeout         *uint32
+	RetryCount      *uint32
+	RetryDelay      *uint32
+	Batch           ProviderBatch
+	PollingInterval *uint32
+}
```
eitjuh, line 112: same as https://github.com/zama-ai/sdk/pull/780/changes#r4035285535

clients/rust/src/config_options.rs
eitjuh: Never mind, I thought the relayer API key would be in it, but that's already in ChainAuth and transport key in DerivationSecret

clients/go/config_options.go
eitjuh: Never mind, I thought the relayer API key would be in it, but that's already in ChainAuth and transport key in DerivationSecret

PR #784, opened 2026-09-17
feat(sidecar): add encryption api

zmalatrax:
Closes SDK-363


clients/go/encryption.go
```diff
@@ -0,0 +1,93 @@
...
+
+func integerEncryptWire(kind string, value *big.Int) (*pb.EncryptInput, error) {
+	if value == nil {
+		return nil, errors.New("encryption integer is nil")
+	}
+	return &pb.EncryptInput{Type: kind, Value: &pb.EncryptInput_BigintValue{BigintValue: value.String()}}, nil
+}
+func (v IntegerInput) encryptWire() (*pb.EncryptInput, error) {
+	return integerEncryptWire(string(v.Type), v.Value)
+}
```
eitjuh:
```suggestion
func (v IntegerInput) encryptWire() (*pb.EncryptInput, error) {
	switch v.Type {
	case EUint8, EUint16, EUint32, EUint64, EUint128, EUint256:
		return integerEncryptWire(string(v.Type), v.Value)
	default:
		return nil, errors.New("unsupported encryption integer type: " + string(v.Type))
	}
}
```

packages/sdk-sidecar/src/encryption.ts
```diff
@@ -0,0 +1,71 @@
...
+        return { type, value: integer(value.bigintValue) as 0n | 1n };
+      }
+      break;
+    case "eaddress":
+      if (value?.$case === "addressValue") {
+        return { type, value: address(value.addressValue) };
+      }
+      break;
+  }
+  throw invalidArgument("Encryption input type and value encoding do not match.");
```
eitjuh:
```suggestion
  throw invalidArgument(
    `Encryption input type and value encoding do not match: received type ${JSON.stringify(type)} with ${value?.$case ?? "no"} value.`,
  );
```

packages/sdk-sidecar/src/encryption.ts
```diff
@@ -0,0 +1,71 @@
...
+  EncryptResponse,
+  EncryptInput as WireInput,
+} from "./generated/zama/sdk/v1alpha1/sidecar.js";
+import type { ContextSdk } from "./runtime.js";
+import { address, bytes, unsignedInteger } from "./encoding.js";
+import { invalidArgument } from "./errors.js";
+
+function integer(value: string): bigint {
+  if (!/^(0|-[1-9][0-9]*|[1-9][0-9]*)$/.test(value)) {
+    throw invalidArgument("Encryption integers must use canonical decimal encoding.");
```
eitjuh:
```suggestion
    throw invalidArgument(
      `Encryption integers must use canonical decimal encoding: received ${JSON.stringify(value)}.`,
    );
```

clients/go/encryption.go
zmalatrax: good catch ty! I'm refactoring the proto message to use `oneof` rather than string, mismatch won't be an issue anymore

packages/sdk-sidecar/src/encryption.ts
zmalatrax: https://github.com/zama-ai/sdk/pull/784#discussion_r4036387209

clients/go/encryption.go
```diff
@@ -0,0 +1,122 @@
...
+
+func Eaddress(value common.Address) EncryptInput {
+	return EncryptInput{wire: &pb.EncryptInput{Value: &pb.EncryptInput_Eaddress{Eaddress: value.Bytes()}}}
+}
+
+func (s *SDKContext) Encrypt(ctx context.Context, params EncryptParams, options EncryptOptions) (*EncryptResult, error) {
+	values := make([]*pb.EncryptInput, len(params.Values))
+	for i, value := range params.Values {
+		if value.wire == nil {
+			return nil, fmt.Errorf("encryption input %d: missing value", i)
```
eitjuh, line 99: nit:

`Euint64(nil)` returns a zero `EncryptInput`, which is the same value you get from a plain `EncryptInput{}`. So `Encrypt` can't tell a nil integer from an input the caller forgot to build, and reports both as `encryption input 1: missing value` — `encryption_test.go:149-151` pins all three causes to that one string.

It's easy to hit, because `SetString` returns nil when the parse fails:

```go
amount, ok := new(big.Int).SetString(userInput, 10) // ok is false, amount is nil
sidecar.Euint64(amount)                             // silently an empty input
```

The caller then goes looking for an input they forgot rather than a bad amount string. Can we keep the intended type on the struct so the message can say `encryption input 1: euint64 value is nil`? Rust can't reach this state at all, so Go is the only client where the message has to explain it.

eitjuh (approved):

eitjuh (approved):

PR #785, opened 2026-09-17
feat(sidecar): expose offline transaction preparation

zmalatrax:
Closes SDK-365


packages/sdk-sidecar/src/offline.ts
```diff
@@ -0,0 +1,166 @@
+import { bytesToHex } from "viem";
+import type { PrepareOptions, PrepareTransactionRequest } from "@zama-fhe/sdk";
+import type * as rpc from "./generated/zama/sdk/v1alpha1/sidecar.js";
+import type { ContextSdk } from "./runtime.js";
+import { address, bytes, safeInteger } from "./encoding.js";
+import { invalidArgument } from "./errors.js";
+
+function decimal(value: string): bigint {
+  if (!/^(0|-?[1-9][0-9]*)$/.test(value)) {
+    throw invalidArgument("Bigint must be a canonical decimal string.");
```
eitjuh: we only use this for unsigned, so technically can do this (for now?)
```suggestion
  if (!/^(0|[1-9][0-9]*)$/.test(value)) {
    throw invalidArgument("Bigint must be a canonical unsigned decimal string.");
```

zmalatrax: `decimal` is actually a duplicate of `integer()`, I'm removing it to use `integer` instead, will push soon https://github.com/zama-ai/sdk/blob/953d037677b0a2026d7654501f9a3455135138c6/packages/sdk-sidecar/src/encoding.ts#L59

clients/rust/src/offline.rs
```diff
@@ -0,0 +1,222 @@
...
+
+#[derive(Clone, Debug)]
+pub struct PrepareFees {
+    pub max_fee_per_gas: BigInt,
+    pub max_priority_fee_per_gas: BigInt,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PreparedTransaction {
+    pub kind: String,
```
eitjuh: can we do this the same as in go instead of just a String

https://github.com/zama-ai/sdk/blob/f8fdadb4ef95c4b3ad8f51da4418ad93208c211a/clients/go/offline.go#L33-L45

clients/rust/examples/balance/offline.rs
```diff
@@ -0,0 +1,105 @@
...
+                transaction: Transaction::SetOperator {
+                    token,
+                    operator: Address::repeat_byte(0x11),
+                    until: 1,
+                },
+            },
+            None,
+        )
+        .await?;
+    ensure!(prepared.kind == "SetOperator", "unexpected prepared kind");
```
eitjuh: same as https://github.com/zama-ai/sdk/pull/785#discussion_r4038237574

clients/go/examples/balance/offline.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	signed, err := signPreparedTransaction(prepared, privateKey)
+	if err != nil {
+		return err
+	}
+	encoded, err := signed.MarshalBinary()
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Offline %s: locally signed transaction %s\nSigned bytes: 0x%x\nCaller owns broadcasting; this demo does not broadcast.\n", prepared.Kind, signed.Hash(), encoded)
```
eitjuh: :meat-proxy: Every other step prints byte-identical text in both examples — `User Address:`, `Token:`, `Encrypted balance:`, `Decrypted balance:`, `Encrypted input %d:`, `Input proof:` — which is what lets the README carry one shared output block. The offline step is the first one where they diverge: Go prints `Offline SetOperator: locally signed transaction …` then `Signed bytes:`, Rust prints `Prepared operator revocation and signed locally:` then `Signed transaction hash:`. That is why the README had to drop "Both examples print:" and describe this step in prose instead. Can we align the three lines so the block covers the whole run again?

https://github.com/zama-ai/sdk/blob/f8fdadb4ef95c4b3ad8f51da4418ad93208c211a/clients/rust/examples/balance/offline.rs#L35-L40

clients/rust/src/offline.rs
zmalatrax: fixed in [95c2ffb](https://github.com/zama-ai/sdk/pull/785/commits/95c2ffb3ec90e05c6366ba17ca901d7ccc526104)

packages/sdk-sidecar/src/offline.ts
zmalatrax: fixed in [0e43436](https://github.com/zama-ai/sdk/pull/785/commits/0e4343619d139831c41d245f3af9b692b7ea21ef)

clients/go/examples/balance/offline.go
zmalatrax: fixed in [2ff28c3](https://github.com/zama-ai/sdk/pull/785/commits/2ff28c3c3c5b625694376ce7de72e46912684038)

eitjuh (approved):

PR #786, opened 2026-09-17
feat(sidecar): add native transaction signing

zmalatrax:
Closes SDK-364


packages/sdk-sidecar/src/remote-signer.ts
```diff
@@ -8,41 +8,108 @@ import {
...
+    }
+    return false;
+  }
+  abortReason(operationId: string): Error {
+    return this.#hasPendingWrite(operationId)
+      ? transactionOutcomeUnknown(
+          "Operation cancelled after a transaction was requested; the wallet may have broadcast it.",
+        )
+      : cancelled();
+  }
```
eitjuh, line 132: > `abortReason` only looks at writes still waiting for the wallet. Once the hash reply is in and the SDK is sitting in `waitForTransactionReceipt`, the pending entry is gone, so a cancel, deadline, account update or context close in that window reports plain `CANCELLED`. That is the longest window in the whole flow, and `CANCELLED` is exactly what a native caller will retry, which is the double-submit the docs are trying to prevent. TRANSACTIONS.md line 52 promises `TRANSACTION_OUTCOME_UNKNOWN` for "write already dispatched".
>
> Tested it: `Token.setOperator` through `SidecarRuntime.execute`, replied with a 32-byte hash, left the receipt mock hanging, then aborted the caller signal:
>
> ```
> AssertionError: expected { code: 'CANCELLED', …(3) } to match object { code: 'TRANSACTION_OUTCOME_UNKNOWN' }
> ```
>
> Can we remember which operations already got a hash back (and drop them when the operation settles) so `abortReason` keeps returning `TRANSACTION_OUTCOME_UNKNOWN` there? Putting the hash in that message would help too, since the SDK-operation caller never sees it otherwise.

zmalatrax: fixed in [0b1b58f](https://github.com/zama-ai/sdk/pull/786/commits/0b1b58f3301a68d617377daaa77bcb2c082b5d4a)

eitjuh (approved):

PR #787, opened 2026-09-18
feat(sidecar): expose delegation api

zmalatrax:
Closes SDK-366

In the examples, when using the delegation api to grant/revoke delegation, the sdk events/responses are not forwarded to the client, it works but the native client is slightly blind when calling those functions, this would be solved with SDK-370


packages/sdk-sidecar/DELEGATIONS.md
```diff
@@ -0,0 +1,111 @@
...
+| -------------------------------------- | ------------------------------------- |
+| `AlreadyDelegatedOrRevokedInSameBlock` | `DELEGATION_COOLDOWN`                 |
+| `NotDelegatedYet`                      | `DELEGATION_NOT_FOUND`                |
+| `SenderCannotBeDelegate`               | `DELEGATION_SELF_NOT_ALLOWED`         |
+| `DelegateCannotBeContractAddress`      | `DELEGATION_DELEGATE_EQUALS_CONTRACT` |
+| `ExpirationDateAlreadySetToSameValue`  | `DELEGATION_EXPIRY_UNCHANGED`         |
+| `EnforcedPause`                        | `ACL_PAUSED`                          |
+| `SenderCannotBeContractAddress`        | `DELEGATION_CONTRACT_IS_SELF`         |
+
+The SDK ACL ABI declares functions only, so ACL reverts decode to their 4-byte selector; the SDK maps named ACL errors when its ABI carries the error entries.
```
eitjuh, line 73: I don't understand this table if we are always showing TRANSACTION_REVERTED. When a revert happens, we get back something like `unrecognized error 0x<selector>`, for which we don't have any mapping? The SDK's ACL ABI lists functions only, no errors, so the revert bytes cannot be decoded to a name?

clients/rust/src/alloy_transactions.rs
```diff
@@ -111,6 +118,23 @@ where
...
+    let data = payload.data.as_ref().and_then(|raw| {
+        let text = serde_json::from_str::<String>(raw.get()).ok()?;
+        let hex = text.strip_prefix("0x")?;
+        alloy_primitives::hex::decode(hex).ok()
+    });
+    let is_revert_message = payload
+        .message
+        .to_lowercase()
+        .starts_with("execution reverted");
+    (payload.code == 3 || is_revert_message || data.is_some()).then(|| data.unwrap_or_default())
```
eitjuh, line 136: Same node error, different answer in Go and Rust.

Take a node reply with code `-32000`, message `insufficient funds`, and a `data` field of `"0x"`. Go says this is not a revert and returns `SIGNING_FAILED`. Rust says it is a revert and returns `TRANSACTION_REVERTED`. The only reason is the `|| data.is_some()` on this line. Go has no such check (`ethereum_signer.go:241-258`), and the comment above this Rust function only mentions code 3 and the message prefix.

Can we drop the `data.is_some()` so both adapters follow the same rule?

```suggestion
   (payload.code == 3 || is_revert_message).then(|| data.unwrap_or_default())
```

packages/sdk-sidecar/DELEGATIONS.md
zmalatrax: removed it

eitjuh (approved):

